Issue #8, opened 2020-10-22
Replace the double-concatenation in the BaseCollateFunction


philippmwirth:
Currently, the BaseCollateFunction concatenates the input images as follows:
```python
# three concatenations
b0 = torch.cat([transform(img) for img in imgs], 0)
b1 = torch.cat([transform(img) for img in imgs], 0)
return torch.cat([b0, b1], 0)
```
This is inefficient because new memory has to be allocated for every concatenation operation (of which there are three).
An alternative with a single concatenation would be:
```python
# single concatenation
b = [transform(imgs[i % bsz]) for i in range(2*bsz)]
return torch.cat(b, 0)
```

A short experiment for `bsz=512` and `image_height=128` shows the potential speed-up:
```bash
Required time (so far): 0.1442425012588501s
Required time (new): 0.09266984462738038s  
```

Issue #7, opened 2020-10-20
Add prebuilt collate functions for SimCLR, MoCo, and SwAV


philippmwirth:
This will save the user a line of code when trying to reproduce results and can make our example code leaner.

I suggest something like the following lines in `lightly.data.collate.py` 
```python

class SimCLRCollateFunction(ImageCollateFunction):
    """Description...

    """

    def __init__(self, input_size=32):
        super(SimCLRCollateFunction, self).__init__(
            input_size=input_size,
            # put all the SimCLR settings here
        )
```

PR #14, opened 2020-10-26
Add Memory Bank and MoCo

philippmwirth:
# Add Memory Bank and MoCo

I have added a memory bank parent class for the NTXentLoss. This way the loss can be used with and without memory bank. Furthermore, I have implemented MoCo, addressed issues #7 and #8, and updated a few docstrings.

## Changes
- Added a memory bank module as parent class for the NTXentLoss function. The NTXentLoss will query the memory bank for negative samples. If the size of the memory bank is zero, the NTXentLoss will continue as before. Otherwise, it will use the negative samples from the memory bank to calculate the loss.
- Added an example to the NTXentLoss docstring.
- Added `memory_bank_size: 0` argument to the config file. This should not be changed at the moment as SimCLR tends to diverge.
- Added default collate functions for `SimCLR` and `MoCo` for user friendliness.
- Addressed issue #8 for efficiency reasons.
- Added the possibility to use `ImageCollateFunction` without Gaussian blur.
- Embeddings can now be trained with a learning rate scheduler. 
- Added an implementation of the MoCo architecture.

## Why?
- The way the NTXentLoss with memory bank is implemented allows a user to choose any type of embedding model and transformation and use them together with the NTXentLoss irrespective of whether a memory bank is required or not. 
For example, a user could try any of the following combinations:
   - SimCLR & no memory bank -> like the paper
   - SimCLR & memory bank -> diverges, unfortunately
   - MoCo & no memory bank -> ???
   - MoCo & memory bank -> like the paper
- This flexible design should be continued for all loss functions which are added in the future.

### DIFF
--- a/docs/source/lightly.loss.rst
+++ b/docs/source/lightly.loss.rst
@@ -7,3 +7,8 @@ lightly.loss
 ---------------
 .. autoclass:: lightly.loss.ntx_ent_loss.NTXentLoss
    :members:
+
+.memory_bank
+-------------
+.. autoclass:: lightly.loss.memory_bank.MemoryBankModule
+   :members:
--- a/docs/source/lightly.models.rst
+++ b/docs/source/lightly.models.rst
@@ -13,6 +13,11 @@ lightly.models
 .. automodule:: lightly.models.simclr
    :members:
 
+.moco
+---------------
+.. automodule:: lightly.models.moco
+   :members:
+
 .zoo
 ---------------
 .. automodule:: lightly.models.zoo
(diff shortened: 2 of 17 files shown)


IgorSusmelj (approved):

IgorSusmelj:
MOCO and memory bank seems not to work properly for me:
Blue is without memory bank. Orange with memory bank. Trained for 100 epochs on CIFAR10.

image

IgorSusmelj:
Memory bank should not update when `requires_grad=False`

PR #41, opened 2020-11-19
Add Video Dataset

philippmwirth:
# Add Video Dataset
Added support to train on, embed, upload, and download video files without extracting all frames beforehand.
To do so, store the video files in an input directory, e.g. like this:

```bash
dataset/
+-- my_video.mp4
+-- other_video.avi
+-- last_video.mov
```

and point the lightly command to the input directory:
```bash
lightly-magic input_dir=dataset/
```
The framework will automatically detect the different video files and will process individual frames as input images.

# Changes
- In `lightly.api.routes.users.service.py` the URL was faulty. Therefore, the function always entered the `except` clause. Fixed the problem and now return the status code of the response as well for better debugging.
- In `lightly.api.upload.py`, refactored the `upload_images_from_folder` function to work with `VideoDatasets` and implemented the feedback from the code review.
- In `lightly.api.utils.py`, implemented the feedback from the code review.
- In `lightly.cli.download_cli.py`, replaced the copying by the new `dataset.dump` function.
- In `lightly.data._helpers.py`: Try to import `VideoDataset` and set a flag if not possible (because of additional requirements). If the input directory contains videos, try to load a `VideoDataset`. Raise exception if not possible. 
Moved the image datasets to `_image.py`.
- `lightly.data._image.py`, now holds the `DatasetFolder` (for images).
- `lightly.data._video.py` contains the implementation of the `VideoDataset`.
- In `lightly.data.dataset.py`, changed the implementation of the `LightlyDataset` to support the `VideoDataset` and added the `dump` function.
- Changed the developer dependencies.
- Added video dependencies.
- Added tests and changed the tox to test videos separately.

### DIFF
--- a/lightly/data/_helpers.py
+++ b/lightly/data/_helpers.py
@@ -11,8 +11,9 @@ from lightly.data._image import DatasetFolder
 try:
     from lightly.data._video import VideoDataset
     VIDEO_DATASET_AVAILABLE = True
-except Exception:
+except Exception as e:
     VIDEO_DATASET_AVAILABLE = False
+    VIDEO_DATASET_ERRORMSG = e
 
 
 IMG_EXTENSIONS = ('.jpg', '.jpeg', '.png', '.ppm', '.bmp',
@@ -71,7 +72,8 @@ def _load_dataset_from_folder(root: str, transform):
         raise ValueError(f'The input directory {root} contains videos '
                          'but the VideoDataset is not available. \n'
                          'Make sure you have installed the right '
-                         'dependencies.')
+                         'dependencies. The error from the imported '
+                         f'module was: {VIDEO_DATASET_ERRORMSG}')
 
     if contains_videos:
         # root contains videos -> create a video dataset
--- a/lightly/data/_helpers.py
+++ b/lightly/data/_helpers.py
@@ -15,17 +15,20 @@ except Exception:
     VIDEO_DATASET_AVAILABLE = False
 
 
-from lightly.data._image_loaders import default_loader
-
 IMG_EXTENSIONS = ('.jpg', '.jpeg', '.png', '.ppm', '.bmp',
                   '.pgm', '.tif', '.tiff', '.webp')
 
 VIDEO_EXTENSIONS = ('.mp4', '.mov', '.avi')
 
 
-def _contains_videos(root, extensions):
-    """
+def _contains_videos(root: str, extensions: tuple):
+    """Checks whether directory contains video files.
 
+    Args:
+        root: Root directory path.
+
+    Returns:
+        True if root contains subdirectories else false.
     """
     list_dir = os.listdir(root)
     is_video = \
@@ -33,14 +36,14 @@ def _contains_videos(root, extensions):
     return any(is_video)
 
 
-def _contains_subdirs(root):
-    """Check whether directory contains subdirectories
+def _contains_subdirs(root: str):
+    """Checks whether directory contains subdirectories.
 
     Args:
-        root: (str) Root directory path
+        root: Root directory path.
 
     Returns:
-        True if root contains subdirectories else false
+        True if root contains subdirectories else false.
 
     """
     list_dir = os.listdir(root)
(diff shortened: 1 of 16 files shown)


lightly/data/_helpers.py
```diff
@@ -134,10 +60,21 @@ def _load_dataset_from_folder(root, transform):
         Dataset consisting of images in the root directory
     """
 
-    if _contains_subdirs(root):
-        dataset = datasets.ImageFolder(root, transform=transform)
+    contains_videos = _contains_videos(root, VIDEO_EXTENSIONS)
+    if contains_videos and not VIDEO_DATASET_AVAILABLE:
+        raise ValueError(f'The input directory {root} contains videos '
```
shruti-shyam, line 71: the error message could be more detailed (it can say it failed because torch version is old or some other package is not installed)

lightly/data/_helpers.py
```diff
@@ -134,10 +60,21 @@ def _load_dataset_from_folder(root, transform):
...
+                         'but the VideoDataset is not available.')
+
+    if contains_videos:
+        dataset = VideoDataset(root,
+                               extensions=VIDEO_EXTENSIONS,
+                               transform=transform)
+    elif _contains_subdirs(root):
+        dataset = datasets.ImageFolder(root,
```
shruti-shyam, line 83: are we assuming that we won't have videos inside the subdirs?

lightly/data/_video.py
```diff
@@ -0,0 +1,137 @@
...
+        instances.append(path)
+
+    # get timestamps
+    timestamps, fpss = [], []
+    for instance in instances:
+        ts, fps = read_video_timestamps(instance, pts_unit=pts_unit)
+        timestamps.append(ts)
+        fpss.append(fps)
+
+    # get offsets
```
shruti-shyam: not clear to me what offsets are and why we need them, maybe add a small comment to explain

lightly/data/_video.py
```diff
@@ -0,0 +1,137 @@
...
+        if index < 0 or index >= self.__len__():
+            raise IndexError(f'Index {index} is out of bounds for VideoDataset'
+                             f' of size {self.__len__()}.')
+
+        # find video of the frame
+        i = 0
+        while i < len(self.offsets) - 1:
+            if self.offsets[i] >= index:
+                break
+            i = i + 1
```
shruti-shyam: could be more clear

lightly/data/_video.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # find and return the frame as PIL image
+        target = i
+        sample = self.loader(self.videos[i],
+                             self.video_timestamps[i][index - self.offsets[i]])
```
shruti-shyam, line 180: add description in the docstring on what is target, offsets

lightly/data/_image.py
```diff
@@ -0,0 +1,121 @@
...
+# All Rights Reserved
+
+import os
+import torchvision.datasets as datasets
+
+from lightly.data._image_loaders import default_loader
+
+
+def _make_dataset(directory, extensions=None, is_valid_file=None):
+    """Returns a list of all image files with targets in the directory.
```
shruti-shyam, line 13: instead of calling it target, we should call it meta_label, i think

tests/data/test_VideoDataset.py
```diff
@@ -0,0 +1,71 @@
...
+    import cv2
+    VIDEO_DATASET_AVAILABLE = True
+except Exception:
+    VIDEO_DATASET_AVAILABLE = False
+
+class TestVideoDataset(unittest.TestCase):
+
+    def ensure_dir(self, path_to_folder: str):
+        if not os.path.exists(path_to_folder):
+            os.makedirs(path_to_folder)
```
shruti-shyam, line 20: can replace with os.makedirs(path_to_folder, exists_ok=True)

tests/data/test_LightlyDataset.py
```diff
@@ -68,7 +95,15 @@ def test_create_lightly_dataset_from_folder(self):
         self.assertEqual(len(dataset), n_tot_files)
         self.assertListEqual(sorted(fnames), sorted(filenames))
 
+        out_dir = tempfile.mkdtemp()
+        dataset.dump(out_dir)
+        self.assertEqual(
+            sum(len(os.listdir(os.path.join(out_dir, subdir))) for subdir in os.listdir(out_dir)),
```
shruti-shyam, line 101: I prefer to use os.walk() instead if os.listdir()

tests/data/test_LightlyDataset.py
```diff
@@ -120,3 +157,29 @@ def test_not_existing_folder_dataset(self):
             LightlyDataset(
                 from_folder='/a-random-hopefully-non/existing-path-to-nowhere/'
             )
+
+    def test_video_dataset(self):
+
+        if not VIDEO_DATASET_AVAILABLE:
+            tmp_dir = tempfile.mkdtemp()
+            self.ensure_dir(tmp_dir)
```
shruti-shyam, line 165: can be omitted

tests/data/test_LightlyDataset.py
```diff
@@ -120,3 +157,29 @@ def test_not_existing_folder_dataset(self):
...
+            # simulate a video
+            path = os.path.join(tmp_dir, 'my_file.png')
+            dataset = torchvision.datasets.FakeData(size=1, image_size=(3, 32, 32))
+            image, _ = dataset[0]
+            image.save(path)
+            os.rename(path, os.path.join(tmp_dir, 'my_file.avi'))
```
shruti-shyam, line 171: not sure if this is the best way to create a video

shruti-shyam (approved):

PR #47, opened 2020-11-25
Fix PL warning

philippmwirth:
# Fix PL warning and add overview

- Fixed the following pytorch-lightning warning:
   ```bash
   Passing a ModelCheckpoint instance to Trainer(checkpoint_callbacks=...) is deprecated since v1.1 
   and will no longer be supported in v1.3
   ```
   The fix checks the local pytorch-lightning version and acts accordingly. Let's remove the fix once we increment the version requirements.
- Checkpoint callback is now customizable:
   ```
   lightly-train input_dir=my_dataset/ save_last=True save_top_k=3
   ```
- Added an overview of the submodules to `lightly/__init__.py`

### DIFF
--- a/lightly/cli/config/config.yaml
+++ b/lightly/cli/config/config.yaml
@@ -74,6 +74,12 @@ trainer:
   max_epochs: 100             # Number of epochs to train for.
   precision: 32               # If set to 16, will use half-precision.
 
+# checkpoint_callback namespace: Modify the checkpoint callback
+checkpoint_callback:
+  save_last: True             # Whether to save the checkpoint from the last epoch
+  save_top_k: 1               # Save the top k checkpoints
+  dirpath:                    # Where to store the checkpoints
+
 # seed
 seed: 1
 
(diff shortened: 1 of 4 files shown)


lightly/cli/config/config.yaml
```diff
@@ -74,6 +74,12 @@ trainer:
   max_epochs: 100             # Number of epochs to train for.
   precision: 32               # If set to 16, will use half-precision.
 
+# checkpoint_callback namespace: Modify the checkpoint callback
+checkpoint_callback:
+  save_last: True             # Whether to save the checkpoint from the last epoch
+  save_top_k: 1               # Save the top k checkpoints
+  dirpath:                    # Where to store the checkpoints
```
IgorSusmelj, line 81: Would add hint what happens if no dirpath is set (where is the checkpoint stored by default?)

IgorSusmelj (changes requested):

IgorSusmelj (approved):

PR #61, opened 2020-12-05
Implement SimSiam

### LINKED ISSUE #54: Implement SimSiam Representation Learning
As proposed by users u/lfotofilter and u/AiDreamer on Reddit.

Add an implementation of the SimSiam Representation Learning framework.
- [Paper](https://arxiv.org/pdf/2011.10566.pdf)

busycalibrating:
WIP implementation for the SimSiam model. This implementation allows you to use any backbone that you'd like. The model trains, but the results need to be validated. I also refactored `import torchvision.transforms as transforms` to `import torchvision.transforms as T` because there were variables with the same name that could cause bugs. Let me know of whatever feedback you may have!

This would close #54. 

TODO:

- [ ] Validate performance/results
- [x] Implement unit tests
- [x] ~~Create a class to prebuild resnet based backbones (like the SimCLR implementation)~~ 
- [x] Tidy up docstrings

### DIFF
--- a/lightly/data/collate.py
+++ b/lightly/data/collate.py
@@ -9,7 +9,7 @@ import torch.nn as nn
 from typing import List
 
 import torchvision
-import torchvision.transforms as transforms
+import torchvision.transforms as T
 from lightly.transforms import GaussianBlur
 from lightly.transforms import RandomRotate
 
@@ -147,31 +147,31 @@ class ImageCollateFunction(BaseCollateFunction):
         else:
             input_size_ = input_size
 
-        color_jitter = transforms.ColorJitter(
+        color_jitter = T.ColorJitter(
             cj_bright, cj_contrast, cj_sat, cj_hue
         )
 
-        transform = [transforms.RandomResizedCrop(size=input_size,
-                                          scale=(min_scale, 1.0)),
+        transform = [T.RandomResizedCrop(size=input_size,
+                                         scale=(min_scale, 1.0)),
              RandomRotate(prob=rr_prob),
-             transforms.RandomHorizontalFlip(p=hf_prob),
-             transforms.RandomVerticalFlip(p=vf_prob),
-             transforms.RandomApply([color_jitter], p=cj_prob),
-             transforms.RandomGrayscale(p=random_gray_scale),
+             T.RandomHorizontalFlip(p=hf_prob),
+             T.RandomVerticalFlip(p=vf_prob),
+             T.RandomApply([color_jitter], p=cj_prob),
+             T.RandomGrayscale(p=random_gray_scale),
              GaussianBlur(
                  kernel_size=kernel_size * input_size_,
                  prob=gaussian_blur),
-             transforms.ToTensor()
+             T.ToTensor()
         ]
 
         if normalize:
             transform += [
-             transforms.Normalize(
+             T.Normalize(
                 mean=normalize['mean'],
                 std=normalize['std'])
              ]
            
-        transform = transforms.Compose(transform)
+        transform = T.Compose(transform)
 
         super(ImageCollateFunction, self).__init__(transform)
 
@@ -317,3 +317,12 @@ class MoCoCollateFunction(ImageCollateFunction):
             rr_prob=rr_prob,
             normalize=imagenet_normalize,
         )
+
+
+class SimSiamCollateFunction(SimCLRCollateFunction):
+    """SimSiam uses the same augmentations as SimCLR. Perhaps the
+    SimCLR augmentations should be abstracted into a more generic
+    base class, as those seem to be a common launching point for
+    most self-supervised methods?
+    """
+    pass
--- a/lightly/data/collate.py
+++ b/lightly/data/collate.py
@@ -317,12 +317,3 @@ class MoCoCollateFunction(ImageCollateFunction):
             rr_prob=rr_prob,
             normalize=imagenet_normalize,
         )
-
-
-class SimSiamCollateFunction(SimCLRCollateFunction):
-    """SimSiam uses the same augmentations as SimCLR. Perhaps the
-    SimCLR augmentations should be abstracted into a more generic
-    base class, as those seem to be a common launching point for
-    most self-supervised methods?
-    """
-    pass
--- a/lightly/data/collate.py
+++ b/lightly/data/collate.py
@@ -243,7 +243,7 @@ class SimCLRCollateFunction(ImageCollateFunction):
             vf_prob=vf_prob,
             hf_prob=hf_prob,
             rr_prob=rr_prob,
-            normalize=imagenet_normalize,
+            normalize=normalize,
         )
 
 
--- a/lightly/loss/sym_neg_cos_sim_loss.py
+++ b/lightly/loss/sym_neg_cos_sim_loss.py
@@ -1,3 +1,8 @@
+""" Symmetrized Negative Cosine Similarity Loss Functions """
+
+# Copyright (c) 2020. Lightly AG and its affiliates.
+# All Rights Reserved
+
 import torch
 
 class SymNegCosineSimilarityLoss(torch.nn.Module):
@@ -13,9 +18,7 @@ class SymNegCosineSimilarityLoss(torch.nn.Module):
         >>> t1 = transforms(images)
         >>>
         >>> # feed through SimSiam model
-        >>> # returns z1, z2, p1, p2 concat in one tensor
-        >>> batch = torch.cat((t0, t1), dim=0)
-        >>> output = model(batch)  
+        >>> output = model(t0, t1)  
         >>>
         >>> # calculate loss
         >>> loss = loss_fn(output)
@@ -23,19 +26,6 @@ class SymNegCosineSimilarityLoss(torch.nn.Module):
     """
 
     def _neg_cosine_simililarity(self, x, y):
-        """Computes the negative cosine similarity between the predictions x 
-        and projections y.
-
-        Args:
-            x:
-                Predictions with shape: n x d.
-            y: 
-                Projections with shape: n x d. Gradients wll not flow through this
-                variable.
-
-        Returns:
-            [type]: [description]
-        """
         v = - torch.nn.functional.cosine_similarity(x, y.detach(), dim=-1).mean()
         return v
 
@@ -48,10 +38,14 @@ class SymNegCosineSimilarityLoss(torch.nn.Module):
                 out0:
                     Output projections of the first set of transformed images.
                     Expects the tuple to be of the form (z0, p0), where z0 is
-                    the output of the backbone and projection mlp, and p0 is 
+                    the output of the backbone and projection mlp, and p0 is the
+                    output of the prediction head.
                 out1:
                     Output projections of the second set of transformed images.
-
+                    Expects the tuple to be of the form (z1, p1), where z1 is
+                    the output of the backbone and projection mlp, and p1 is the
+                    output of the prediction head.
+ 
             Returns:
                 Contrastive Cross Entropy Loss value.
 
--- a/lightly/loss/sym_neg_cos_sim_loss.py
+++ b/lightly/loss/sym_neg_cos_sim_loss.py
@@ -1,7 +1,7 @@
 import torch
 import numpy as np
 
-class SymmetrizedLoss(torch.nn.Module):
+class SymNegCosineSimilarityLoss(torch.nn.Module):
     """Implementation of the Symmetrized Loss.
     
     Examples:
@@ -24,7 +24,7 @@ class SymmetrizedLoss(torch.nn.Module):
     """
 
     def __init__(self):
-        super(SymmetrizedLoss, self).__init__()
+        super(SymNegCosineSimilarityLoss, self).__init__()
 
     def _neg_cosine_simililarity(self, x, y):
         """Computes the negative cosine similarity between the predictions x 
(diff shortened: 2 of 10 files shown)


lightly/data/collate.py
```diff
@@ -304,3 +304,12 @@ def __init__(self,
             rr_prob=rr_prob,
             normalize=imagenet_normalize,
         )
+
+
+class SimSiamCollateFunction(SimCLRCollateFunction):
```
philippmwirth, line 309: I wouldn't introduce an extra class, if the settings are the exact same as for SimCLR. For example, we didn't write a new class for the MoCo v2 augmentations because they are exactly the same as SimCLR.

busycalibrating: Agreed - would prefer to keep these augmentations with the simclr name, or pull them into a more generic/high level class of base augmentations?

philippmwirth: I'd say keep them with the SimCLR name

lightly/loss/symmetrized_loss.py
```diff
@@ -0,0 +1,85 @@
...
+        if output.shape[0] % 4:
+            raise ValueError('Expected output of shape 4*bsz x dim but got '
+                             f'shape {output.shape[0]} x {output.shape[1]}.')
+
+        # device = output.device
+        batch_size, dim = output.shape
+        batch_size = batch_size // 4
+
+        # normalize the output to length 1
+        output = torch.nn.functional.normalize(output, dim=1)
```
IgorSusmelj, line 72: I think we don't need normalization since we use cosine similarity and not L2 distance.

lightly/models/simsiam.py
```diff
@@ -0,0 +1,159 @@
...
+    l1 = nn.Sequential(nn.Linear(in_dims, h_dims),
+                       nn.BatchNorm1d(h_dims),
+                       nn.ReLU(inplace=True))
+
+    l2 = nn.Sequential(nn.Linear(h_dims, h_dims),
+                       nn.BatchNorm1d(h_dims),
+                       nn.ReLU(inplace=True))
+
+    l3 = nn.Sequential(nn.Linear(h_dims, out_dims),
+                       nn.BatchNorm1d(out_dims))
```
IgorSusmelj, line 70: I'm not sure whether we should add here or on line 33 (before linear layer) another ReLU activation. Currently, we have

- linear
- batch_norm
- linear

two linear layers without non-linearity between seem odd to me. 

I would add ReLU at line 33. That way we don't have a sparse output after the projection head.

lightly/loss/symmetrized_loss.py
busycalibrating: Whoops yes you're right - there's some dead code I need to tidy up

lightly/models/simsiam.py
busycalibrating: That's true it is not typical. However this would match the implementation described in [the paper (pg 3)](https://arxiv.org/pdf/2011.10566.pdf): 

> _Projection MLP_. The projection MLP _(f)_ has BN applied to each fully-connected (fc) layer, including its output fc. Its **output fc has no ReLU**. The hidden fc is 2048 d. This MLP has 3 layers. 
>
> _Prediction MLP._ The prediction MLP _(h)_ has BN applied to its hidden fc layers.   Its output fc does not have BN (ablation in Sec. 4.4) or ReLU. This MLP has 2 layers. The dimension of _h_’s input and output (_z_ and _p_) is d=2048, and _h_’s hidden layer’s dimension is 512, making _h_ a bottleneck structure (ablation in supplement).

This could be worth experimenting with to see the difference in performance?

lightly/data/collate.py
laurenmoos: +1 for this advice, almost did this exact same thing (extend SimCLR for BYOL

lightly/loss/symmetrized_loss.py
```diff
@@ -0,0 +1,85 @@
+import torch
+import numpy as np
+
+class SymmetrizedLoss(torch.nn.Module):
+    """Implementation of the Symmetrized Loss.
```
philippmwirth, line 5: I'd suggest to give this loss a more descriptive name (for example `SymmetricNegCosineSimilarityLoss` or `NegCosineSimilarityLoss`).

lightly/models/simsiam.py
IgorSusmelj: Got this reply from Kaiming:
Dear Igor,

 

Thank you for your interest!

 

The predictor head by default is: Linear->BN->ReLU->Linear. Also note the dimension is 2048->512->2048. And you are right, there is no ReLU between the output of projection head and the input of prediction head. This is because this embedding (output of projection head) is expected to be normalized to the unit sphere, so ReLU is not desired.

 

Regards,

Kaiming

lightly/models/simsiam.py
```diff
@@ -0,0 +1,159 @@
...
+            TODO
+        pred_hidden_dim:
+            TODO
+        out_dim:
+            Dimension of the output (after the projection head).
+
+    """
+
+    def __init__(self,
+                 backbone: nn.Module,
```
philippmwirth, line 96: Please also rename `backbone` to `features` for consistency

busycalibrating: If you're refactoring the models would you be open to renaming the other names to backbone? I think it's more informative than features in this context

busycalibrating: On board with keeping it consistent though obviously

philippmwirth: Yes, we also like `backbone` better. In that case you can keep it the way it is and we will change the other names.

lightly/models/simsiam.py
```diff
@@ -0,0 +1,159 @@
...
+        # device = output.device
+        batch_size = len(x)
+        batch_size = batch_size // 2
+
+        x1, x2 = x[:batch_size], x[batch_size:]
+
+        emb1, emb2 = self.backbone(x1), self.backbone(x2)
+        emb1, emb2 = emb1.squeeze(), emb2.squeeze()
+        z1, z2 = self.projection_mlp(emb1), self.projection_mlp(emb2)
+        p1, p2 = self.prediction_mlp(z1), self.prediction_mlp(z2)
```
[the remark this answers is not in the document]
busycalibrating: I think what he meant is that based on the design architecture, the output of the embedding is already scattered on the unit hypersphere [(paper pg 3-4)](https://arxiv.org/pdf/2011.10566.pdf):

> The blue curve in Figure 2 (middle) shows that with stop-gradient, the std value is near 1/√d. This indicates that the outputs do not collapse, and they are scattered on the unit hypersphere.

Would you agree with my interpretation?

busycalibrating:
Currently waiting for #66 to go through and then I'll adjust my implementation based on that.

lightly/models/simsiam.py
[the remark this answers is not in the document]
philippmwirth: The description of the plot you are referencing (Figure 2) is:
> Middleplot:  the per-channel std of the **l<sub>2</sub>-normalized output**, plotted as the averaged std over all channels.

So my interpretation is that the **l<sub>2</sub>-normalized** output of the embeddings is scattered on the unit hypersphere (as opposed to collapsing and basically pointing in a single direction). Therefore I would agree with @IgorSusmelj  here. Although it's not clear to me from the paper that the outputs need to be normalized.

lightly/models/simsiam.py
```diff
@@ -0,0 +1,159 @@
...
+        self.proj_hidden_dim = proj_hidden_dim
+        self.pred_hidden_dim = pred_hidden_dim
+        self.out_dim = out_dim
+
+        self.projection_mlp = \
+            _projection_mlp(num_ftrs, proj_hidden_dim, out_dim)
+
+        self.prediction_mlp = \
+            _prediction_mlp(num_ftrs, pred_hidden_dim, out_dim)
+
```
philippmwirth, line 115: This should be 
```
self.prediction_mlp = \
    _prediction_mlp(out_dim, pred_hidden_dim, out_dim)
```
Because then I can change the `num_ftrs` without breaking the forward pass, right?

lightly/models/simsiam.py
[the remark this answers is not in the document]
IgorSusmelj: Maybe just try one implementation. Adding normalization should be rather easy with one line of code. We can add tensorboard logging of the std using pytorch-lightning `self.log(..)` and verify whether the results match with the paper.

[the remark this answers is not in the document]
philippmwirth: +1 on this, just leave it like it is for the moment.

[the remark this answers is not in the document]
busycalibrating: Sounds good!

lightly/models/simsiam.py
busycalibrating: If anything, would it not be

```Python
self.prediction_mlp = _prediction_mlp(num_ftrs, pred_hidden_dim, num_ftrs)
```

Not totally sure what you mean by changing `num_ftrs` without breaking the forward pass?

philippmwirth: Currently, if `num_ftrs != out_dim`, the forward pass breaks because the output of the projection head has a different shape than the input of the prediction head. Therefore, my suggestion was to change the input dimension of the prediction head to the output dimension of the projection head.

busycalibrating: Ah gotcha, yes you're right. I still mix up prediction and projection in my head 😀

philippmwirth:
Btw for the validation of the model we can run experiments when the PR is ready in case you don't have the compute :) 

I already did a few quick tests on cifar10 and the linear evaluation accuracy is around 63% after 50 epochs which roughtly corresponds to what we can see in [Figure D.1](https://arxiv.org/pdf/2011.10566.pdf). So it seems we are on the right path.

lightly/loss/sym_neg_cos_sim_loss.py
```diff
@@ -0,0 +1,68 @@
...
+
+        Args:
+            x:
+                Predictions with shape: n x d.
+            y: 
+                Projections with shape: n x d. Gradients wll not flow through this
+                variable.
+
+        Returns:
+            [type]: [description]
```
busycalibrating, line 37: Fix this

lightly/loss/sym_neg_cos_sim_loss.py
```diff
@@ -0,0 +1,68 @@
...
+                out1:
+                    Output projections of the second set of transformed images.
+
+            Returns:
+                Contrastive Cross Entropy Loss value.
+
+            Raises:
+                ValueError if shape of output is not multiple of batch_size.
+        """
+        z0, p0 = out0
```
busycalibrating, line 61: Should we do a check for whether the forward pass from the `SimSiam` model was in `return_features` mode?

lightly/models/simsiam.py
```diff
@@ -0,0 +1,169 @@
...
+        self.projection_mlp = \
+            _projection_mlp(num_ftrs, proj_hidden_dim, out_dim, num_mlp_layers)
+
+        self.prediction_mlp = \
+            _prediction_mlp(out_dim, pred_hidden_dim, out_dim)
+        
+    def forward(self, 
+                x0: torch.Tensor, 
+                x1: torch.Tensor,
```
busycalibrating, line 130: @philippmwirth I didn't add support for `x1=None` as per the latest simclr/moco models forward pass. Is this something that you want across the board?

busycalibrating:
I rebased my branch and updated my implementation to account for the API change. I think this is ready for the CIFAR-10 validation, and I just need to implement some tests. Should be good to go soon!

lightly/models/simsiam.py
IgorSusmelj: It would be great to have a more or less standardized interface across the different models.

busycalibrating:
I added some basic tests (mimicking what was implemented for SimCLR), but I think these tests are pretty basic and at some point should test for numeric results with known inputs. But that should probably be its own issue. 

This should be now ready for proper review and validation

lightly/loss/sym_neg_cos_sim_loss.py
philippmwirth: No, I think it's better to be clear about what inputs the function expects and leave it to the user to pass the right arguments.

lightly/loss/sym_neg_cos_sim_loss.py
```diff
@@ -0,0 +1,62 @@
...
+        >>> # generate two random transforms of images
+        >>> t0 = transforms(images)
+        >>> t1 = transforms(images)
+        >>>
+        >>> # feed through SimSiam model
+        >>> output = model(t0, t1)  
+        >>>
+        >>> # calculate loss
+        >>> loss = loss_fn(output)
+
```
philippmwirth, line 25: The example is not up-to-date, it should be:
```python
>>> out0, out1 = model(t0, t1)
>>> loss = loss_fn(out0, out1)
```

and you're missing the brackets for the constructor:
```python
>>> loss_fn = SymmetrizedLoss()
```

busycalibrating: Good catch!

IgorSusmelj:
I just run a test modifying the MoCo tutorial to use SimSiam (last commit: 826524b2e293b63774a434936adf3ed93d90bd9e). Run with the same settings (100 epochs, resnet-18, etc.)
Final accuracy is 77.9% on the test set

image
image
image

philippmwirth (approved):

Issue #60, opened 2020-12-05
Update ResNet to have optional `num_classes`


busycalibrating:
Currently, the ResNet implementation always adds a classification head but for almost all of the self supervised approaches, you only care about the actual backbone (and so you strip away this linear layer).

I'd propose to make the `num_classes` argument optional, and if `None`, omit the final linear layer so you only get the actual backbone. I can open a pull request for this if you're open to this.

Also, It looks like the library's implementation is not totally consistent with the default PyTorch ones (some missing max pools, different kernel sizes). I think a lot of people are likely to use the default PyTorch models for a lot of things, which would make the ResNets trained using Lightly incompatible. Would there be an interest in switching to the default PyTorch implementations?

philippmwirth:
Yes, making `num_classes` optional is a good idea. Feel free to open the PR.

Regarding the ResNet implementation: For the future it would be good to change to the default PyTorch ResNet. However, we need to be careful to not break backwards compatability since we have certain pre-trained models and users may also have "old" checkpoints lying around. I'd suggest you open a new issue for this particular problem and we will discuss possible solutions there.

Issue #57, opened 2020-12-02
Refactor models


philippmwirth:
The SimCLR and MoCo architectures are currently implemented as standalone architectures.
Before adding new architectures like BYOL or SimSiam we should probably make use of the existing code, e.g.
- Divide architectures into more general frameworks (in our case: with momentum updates vs. without momentum updates).
- Have SimCLR, MoCo, BYOL, SimSiam, etc inherit from the respective generalized architectures.

busycalibrating:
+1

Do you have an idea for base classes/interfaces that you'd like to enforce, and an idea of how long it will be before things are refactored into this format?

philippmwirth:
At the moment, I'm thinking of something like this (names are provisionally):

```
ResNetEncoder
+-- SingleResNetEncoder
    +-- ResNetSimCLR
    +-- ResNetSimSiam
+-- DualResNetEncoder
    +-- ResNetMoCo
    +-- ResNetBYOL
```

Unfortunately, I can't give you an exact timeline but since we have rather quick development cycles I would say in the next weeks.

busycalibrating:
Do you tie each self-supervised model to a specific backbone? I'm fairly new to this subject so I don't know the details of all the different types of approaches, but it was my understanding that they were _fairly_ model agnostic?

For example this, [BYOL implementation](https://github.com/lucidrains/byol-pytorch) allowed you to hook up your own model into the BYOL architecture.

philippmwirth:
Good point, currently the layer of abstraction is implemented in `lightly.embedding`, so if someone wants to use a custom model that person would have to write its own implementation of the self-supervised model. This refactoring would be a great time to change this and accept different backbones. I have changed the example names in the comment above.

PR #64, opened 2020-12-05
Added BYOL loss to losses and simple MLP to models

laurenmoos:
**this does NOT have** 
1. model refactoring (as I will iteratively follow your changes once you cement the parent classes)
2. momentum (it is passed as parameter but the logic isn't implemented yet)

**this does have:**
1. models (simple MLP)  under models
2. loss (forward pass) under loss
3. I have way too many classes in the loss file but perhaps this will give you/us inspiration for some core abstractions that would be helpful (wrapper class for models to access hidden states/dual-model & consistency loss abstractions & augmentation abstractions)- BYOL is simple logic but somewhat complicated framework so needs a lot of OOP

**my next steps would be the following** 
1. adding a benchmark/testing (I think it would be cool to demonstrate use of batch norm  influencing functionality!)  and training loop
2. adding momentum implementation
3. waiting for your commit to make architecture homogeneous with the higher level abstractions you propose (probably should have abstractions of Online/Target model as well as the low level model architectures

### DIFF
--- a/lightly/loss/byol_loss.py
+++ b/lightly/loss/byol_loss.py
@@ -1,0 +1,189 @@
+import singleton as singleton
+import torch
+from torch import nn
+
+import torch.nn.functional as F
+
+from models.mlp import MLP
+
+
+class ExperimentParams:
+    """
+    Encapsulate hyper-parameters for performing and logging experiments.
+    """
+
+    def __init(self, use_momentum, beta, augment_functions, net, projection_size, projection_hidden_size, layer):
+        # TODO: implement momentum
+        self.use_momentum = use_momentum
+        self.beta = beta
+        self.augment_functions = augment_functions
+        self.net = net
+        self.projection_size = projection_size
+        self.projection_hidden_size = projection_hidden_size
+        self.layer = layer
+
+    def __str__(self):
+        return "Running experiment with the following hyper-parameters.\n" \
+               "Momentum: {}\nBeta: {}\nAugment Function: {}\nNet: {}\nProjection Size: {}\n" \
+               "Projection Hidden Size: {}\n" "Layer: {}".format(self.use_momentum, self.beta, self.augment_functions,
+                                                                 self.net, self.projection_size,
+                                                                 self.projection_hidden_size, self.layer)
+
+
+class NetWrapper(nn.Module):
+    """
+    Utilities for accessing hidden layers.
+    """
+    def __init__(self, net, projection_size, projection_hidden_size, layer=-2):
+        super().__init__()
+        self.net = net
+        self.layer = layer
+
+        self.projector = None
+        self.projection_size = projection_size
+        self.projection_hidden_size = projection_hidden_size
+
+        self.hidden = None
+        self.hook_registered = False
+
+    def _find_layer(self):
+        if type(self.layer) == str:
+            modules = dict([*self.net.named_modules()])
+            return modules.get(self.layer, None)
+        elif type(self.layer) == int:
+            children = [*self.net.children()]
+            return children[self.layer]
+        return None
+
+    def _hook(self, _, __, output):
+        self.hidden = self.flatten(output)
+
+    def _register_hook(self):
+        layer = self._find_layer()
+        assert layer, f'hidden layer ({self.layer}) not found'
+        layer.register_forward_hook(self._hook)
+        self.hook_registered = True
+
+    @singleton('projector')
+    def get_projector(self, hidden):
+        _, dim = hidden.shape
+        projector = MLP(dim, self.projection_size, self.projection_hidden_size)
+        return projector.to(hidden)
+
+    def get_representation(self, x):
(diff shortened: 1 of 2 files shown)


lightly/loss/byol_loss.py
```diff
@@ -0,0 +1,189 @@
+import singleton as singleton
+import torch
+from torch import nn
+
+import torch.nn.functional as F
+
+from models.mlp import MLP
+
+
```
philippmwirth, line 9: Please remove `ExperimentParams` from `byol_loss.py`.

If there is the wish for better experiment tracking, open an issue about it and then we can discuss the next steps.

lightly/loss/byol_loss.py
```diff
@@ -0,0 +1,189 @@
...
+        self.layer = layer
+
+    def __str__(self):
+        return "Running experiment with the following hyper-parameters.\n" \
+               "Momentum: {}\nBeta: {}\nAugment Function: {}\nNet: {}\nProjection Size: {}\n" \
+               "Projection Hidden Size: {}\n" "Layer: {}".format(self.use_momentum, self.beta, self.augment_functions,
+                                                                 self.net, self.projection_size,
+                                                                 self.projection_hidden_size, self.layer)
+
+
```
philippmwirth, line 32: This is a cool idea! We still have to find the best way to get the activations from intermediate levels of the backbone net. However, I don't think `byol_loss.py` is the right place for such a feature.

Please remove `NetWrapper` from this file and feel free to make a suggestion in #60 on how to solve this issue.

lightly/loss/byol_loss.py
```diff
@@ -0,0 +1,189 @@
...
+    @staticmethod
+    def default(val, def_val):
+        return val if val else def_val
+
+    @staticmethod
+    def flatten(t):
+        return t.reshape(t.shape[0], -1)
+
+
```
philippmwirth, line 94: The classes `OnlineModel` and `TargetModel` will receive an abstract parent class with #57 (I will make a PR soon).

IMO it would be best to reuse the parent class as soon as it's ready.

lightly/loss/byol_loss.py
```diff
@@ -0,0 +1,189 @@
...
+            old_weight, up_weight = ma_params.data, current_params.data
+            ma_params.data = self._update_average(old_weight, up_weight)
+
+    def _update_average(self, old, new):
+        if self.ema:
+            return old * self.beta + (1 - self.beta) * new
+        else:
+            return new
+
+
```
philippmwirth, line 129: The implementation of the loss should follow the style of the current implementations. I.e. the loss takes as input the outputs of the model. See:
- [Lightly at a Glance: Training](https://docs.lightly.ai/getting_started/lightly_at_a_glance.html)
- [NTXentLoss implementation](https://github.com/lightly-ai/lightly/blob/develop/lightly/loss/ntx_ent_loss.py)
- [Negative Cosine Similarity Loss by busycalibrating](https://github.com/lightly-ai/lightly/pull/61/files)

This allows for a more flexible setup where a single loss function can be combined with different architectures.

lightly/loss/byol_loss.py
```diff
@@ -0,0 +1,189 @@
...
+            target_encoder = self.target.encoder
+            target_proj_one = target_encoder(image_one).detach()
+            target_proj_two = target_encoder(image_two).detach()
+
+        loss_one = self._loss_fn(online_pred_one, target_proj_two.detach())
+        loss_two = self._loss_fn(online_pred_two, target_proj_one.detach())
+
+        loss = loss_one + loss_two
+        return loss.mean()
+
```
philippmwirth, line 177: Please refrain from using static methods and use module level functions instead:

```python
def _neg_cosine_loss(x, y):
   """Calculates the asymmetric negative cosine loss.

   """
   pass

class BYOLLoss(nn.Module):
...

def forward(self, x):
   """Uses neg_cosine_loss to calculate the symmetric negative cosine loss.

   """
   pass
```

Source: [Google Python Styleguide](https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators)

lightly/loss/byol_loss.py
```diff
@@ -0,0 +1,189 @@
...
+    def _update_average(self, old, new):
+        if self.ema:
+            return old * self.beta + (1 - self.beta) * new
+        else:
+            return new
+
+
+class BYOLLoss:
+    """
```
philippmwirth, line 131: For the docstrings format, please check out the [Google Python Styleguide](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) to see how we do it :)

lightly/models/mlp.py
```diff
@@ -0,0 +1,15 @@
+from torch import nn
+
+
```
philippmwirth, line 3: #57 will also introduce utility classes for the projection heads and predictor heads.

PR #66, opened 2020-12-07
Refactor models

philippmwirth:
# Refactor Models

Addressing #57

Refactored SimCLR and MoCo such that they take a backbone as input and are independent of ResNet. Both SimCLR and MoCo are implementations of a PyTorch `nn.Module`. MoCo additionally makes use of the `_MomentumEncoderMixin` for the implementation of the momentum encoder. This mixin can be reused for other models (e.g. BYOL). Renamed the `features` to more descriptive `backbone`.

Made changes to the command-line tool and to the tests so that they are compatible with the changes.

To Do:
- [x] Clean up code
- [x] Clean up comments and docstrings
- [x] Update documentation and tutorials


lightly/cli/_helpers.py
```diff
@@ -65,3 +65,65 @@ def load_state_dict_from_url(url, map_location=None):
...
+        # found running mean or running var with different shapes
+        state_dict[key] = state.repeat(num_splits)
+
+    return state_dict
+
+
+def _filter_state_dict(state_dict):
+    """Prevents unexpected key error when loading PyTorch-Lightning checkpoints.
+
+    Removes the "model." prefix from all keys in the state dictionary.
```
laurenmoos, line 99: instead of making this a doc maybe make the offset a context that denotes that it is removing this prefix?

lightly/cli/_helpers.py
```diff
@@ -65,3 +65,65 @@ def load_state_dict_from_url(url, map_location=None):
...
+    """Loads the model weights from the state dictionary.
+
+    """
+    # step 1: filter state dict
+    if apply_filter:
+        state_dict = _filter_state_dict(state_dict)
+ 
+    # step 2: expand batchnorm weights
+    state_dict = \
+        _expand_batchnorm_weights(model.state_dict(), state_dict, 0)
```
laurenmoos, line 126: this feels like an initialization, would there be a way of using a closure or lightweight encapsulation to make this easier to read? Also if the function signature doesn't have 'splits' in it should probably pass 0 as a named variable at least

laurenmoos: not to nitpick too much about the function name but 'maybe' might be useful prefix if the transformation isn't guaranteed to occur (which appears it is not to be)

lightly/models/_momentum.py
```diff
@@ -0,0 +1,97 @@
...
+# Copyright (c) 2020. Lightly AG and its affiliates.
+# All Rights Reserved
+
+import copy
+
+import torch
+import torch.nn as nn
+
+
+def _deactivate_requires_grad(params):
```
laurenmoos, line 12: this seems like a more general function than just for momentum? Should this go in a more general utilities file?

lightly/models/moco.py
```diff
@@ -58,87 +40,28 @@ class ResNetMoCo(nn.Module, _StateDictLoaderMixin):
...
-        self.key_features, self.key_projection_head = \
-            _get_features_and_projections(
-                ResNetGenerator(name=name, width=width, num_splits=num_splits),
-                self.num_ftrs,
-                self.out_dim,
-                num_splits
-            )
-
-        # set key-encoder weights to query-encoder weights
-        for param_k in self.key_features.parameters():
```
laurenmoos, line 92: this is an example where you could use a more generic function that wraps a loop that deactivates gradient for parameters

lightly/models/moco.py
```diff
@@ -58,87 +40,28 @@ class ResNetMoCo(nn.Module, _StateDictLoaderMixin):
...
-            apply_filter:
-                If True, removes the `model.` prefix from keys in the state_dict.
+        self.backbone = backbone
+        self.projection_head = _get_moco_projection_head(num_ftrs, out_dim)
+        self.momentum_features = None
+        self.momentum_projection_head = None
 
-        """
-        self._custom_load_from_state_dict(state_dict, strict, apply_filter)
-        self._momentum_update_key_encoder(0.)
```
laurenmoos, line 114: I think it is pretty much impossible to eliminate magic numbers in ML engineering but at least using a named parameter when you pass a number (if not a constant) would be helpful re. readability (obviously there is a limit to this but in general it is a best practice)

philippmwirth: True, the part you commented on was actually removed from the code but it happens again further down...

lightly/models/_momentum.py
philippmwirth: I agree that the function can be more general but at the moment there are no other applications for it than in momentum and hence I would leave it here. We can always move it to a more general file if necessary.

docs/source/tutorials_source/package/tutorial_moco_memory_bank.py
```diff
@@ -191,19 +193,32 @@
...
     def __init__(self):
         super().__init__()
+        
+        # create a ResNet backbone and remove the classification head
+        resnet = lightly.models.ResNetGenerator('resnet-18', 1, num_splits=8)
+        backbone = nn.Sequential(
+            *list(resnet.children())[:-1],
+            nn.AdaptiveAvgPool2d(1),
+        )
+
```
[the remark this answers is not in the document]
philippmwirth: We will keep the `ResNetGenerator` for now to uphold backwards compatability but with the new implementation of `MoCo` and `SimCLR` (which are now independent from ResNet), it will be much easier to swap to the torchvision models. For now, I think it would be best if the new models don't depend on a specific model.

docs/source/getting_started/lightly_at_a_glance.rst
```diff
@@ -64,11 +64,20 @@ with the normalized temperature-scaled cross entropy loss and simple stochastic
...
     import lightly.models as models
     import lightly.loss as loss
 
-    # build a resnet-34 with 32 embedding neurons
-    model = models.ResNetSimCLR(name='resnet-34', num_ftrs=32)
+    # use a resnet backbone
+    resnet = torchvision.models.resnet.resnet18()
+    resnet = nn.Sequential(
+        *list(resnet.children())[:-1],
+        nn.AdaptiveAvgPool2d(1),
```
busycalibrating, line 76: I think this is redundant - the last 2 layers of ResNet18 are:

```
AdaptiveAvgPool2d-67            [-1, 512, 1, 1]               
Linear-68                       [-1, 1000]
```

So this would just be stripping the linear layer and adding another Adaptive Avg Pool?

philippmwirth: You are right, the Adaptive Avg Pool is redundant. Thanks!

IgorSusmelj (approved):

PR #74, opened 2020-12-13
Documentation update, Add SimCLR example

### LINKED ISSUE #73: remove checkpointing from embedding module
With the current setup the embedding module `SelfSupervisedEmbedding` automatically creates checkpoints when trained. This also causes problems if checkpoints already exist, basically making it hard to use on its own. I would propose to remove the checkpoint_callback code https://github.com/lightly-ai/lightly/blob/d1b471174f8a37c88432d1592292b3e057940329/lightly/embedding/_base.py#L46 
from the `embedding` module.

I think that should be a rather small change but it could help with usability.

IgorSusmelj:
### Changes

A few minor corrections in the documentation of the pip package:

 - `data/_video.py` docstring now states that we use specific timestamps and not random ones
 - `data/_video.py` unused `pts_unit` has been removed
 - `embedding/embedding.py` docstring is now more clear about the return value (it's a tuple!)
 - `docker/examples/datasets_in_the_wild.rst` contains now a section for comparing storage usage when working with videos directly vs extracting all frames beforehand
 - `tutorials_source/package/tutorial_moco_memory_bank.py` uses now CosineAnnealingLR scheduler and batch size 512
- `tutorials_source/package/tutorial_simclr_clothing.py `: New tutorial on how to use SimCLR and the embedding module

### Todo:

- [x] I had to add a fix to manually remove the created checkpoints. This can be removed after fix #73


docs/source/docker/examples/datasets_in_the_wild.rst
```diff
@@ -72,6 +72,36 @@ extract the frames with the following command:
     ffmpeg -i raw/video.mp4 -filter:v "fps=5" frames_ffmpeg/%d.png
 
 
+Extracting the frames without introducing compression artifacts is using lots of 
+storage. In this example, we have a small video of 6.4 MBytes. Once extracted 
+the .png frames together with the video consume 553.4 MBytes. This is a
```
philippmwirth, line 77: Comma after extracted, i.e. "Once extracted, ..."

docs/source/tutorials_source/package/tutorial_moco_memory_bank.py
```diff
@@ -324,6 +333,7 @@ def configure_optimizers(self):
 
 # %%
 # Train the Classifier
+model.eval()
 classifier = Classifier(model.resnet_moco)
 trainer = pl.Trainer(max_epochs=max_epochs, gpus=gpus,
                      progress_bar_refresh_rate=100)
```
philippmwirth, line 339: Did you verify that the results work after changing the learning rate, batch size etc?

lightly/data/_video.py
```diff
@@ -11,22 +11,22 @@
...
+        A PIL image
 
     """
     # random access read from video (slow)
     frame, _, _ = read_video(path,
                              start_pts=timestamp,
-                             end_pts=timestamp,
-                             pts_unit=pts_unit)
+                             end_pts=timestamp)
     # read_video returns tensor of shape 1 x W x H x C
```
philippmwirth, line 30: Why did you remove the `pts_unit`? This can lead to errors because we read the timestamps in seconds.

IgorSusmelj: Good catch. I saw that the pts_unit isn't set for the private method, so I thought I would remove it. I forgot to switch to pts_unit='sec'.

docs/source/tutorials_source/package/tutorial_moco_memory_bank.py
IgorSusmelj: Yes, accuracy is higher, model trains faster due to larger batch size. However, I also added cosine LR and therefore increased the default number of epochs. I will update the exact numbers in the docs. So far I got around 80% after 200 epochs.

philippmwirth (approved):

PR #79, opened 2020-12-17
[WIP] Add new VideoLoader class for faster video reading.

IgorSusmelj:
### Changes

- Use new `video_reader` backend from torchvision for faster video loading

### Todo

- [x] Add tests
- [x] Remove old `_video_loader` 
- [ ] Use new loader for faster embedding of videos

### DIFF
--- a/lightly/data/_video.py
+++ b/lightly/data/_video.py
@@ -4,6 +4,7 @@
 # All Rights Reserved
 
 import os
+from typing import List
 from PIL import Image
 
 import torchvision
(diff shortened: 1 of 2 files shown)


IgorSusmelj:
@philippmwirth, the new video interface is much faster when reading sequential frames. That might not work when using multiple workers. 

How should we change the embedding part of the code? Just switch to 0 or 1 worker in case we can use the new backend?

philippmwirth:
> 
> 
> @philippmwirth, the new video interface is much faster when reading sequential frames. That might not work when using multiple workers.
> 
> How should we change the embedding part of the code? Just switch to 0 or 1 worker in case we can use the new backend?

I would leave it to the user and document it clearly. IMO it's not a good idea to enforce a certain number of workers.

lightly/data/_video.py
```diff
@@ -47,6 +48,117 @@ def _video_loader(path, timestamp, pts_unit='sec'):
...
+
+        else: # fallback on pyav
+            if timestamp == None:
+                timestamp = self.timestamps[self.current_timestamp_idx]
+            frame, _, _ = io.read_video(self.path,
+                                        start_pts=timestamp,
+                                        end_pts=timestamp,
+                                        pts_unit=self.pts_unit)    
+            self.last_timestamp_idx = self.timestamps.index(timestamp)    
+
```
philippmwirth, line 145: Nice idea! I have some feedback:
- Be consistent in the way you check for `None` (I would always go with `if x is None` or `if x is not None`).
- A few comments would make reading this part a lot easier.

IgorSusmelj:
We 

> > @philippmwirth, the new video interface is much faster when reading sequential frames. That might not work when using multiple workers.
> > How should we change the embedding part of the code? Just switch to 0 or 1 worker in case we can use the new backend?
> 
> I would leave it to the user and document it clearly. IMO it's not a good idea to enforce a certain number of workers.

We can improve the embedding part in another PR. This one would already fix multiple issues we faced when working with videos.

philippmwirth (approved):

Issue #84, opened 2020-12-21
Error while using VideoDataset without extension specified


IgorSusmelj:
### Issue
When using the VideoDataset I end up with an exception if the extension argument is not set. I can reproduce the error on different lightly versions as well as video files.

![image](https://user-images.githubusercontent.com/1522766/102757196-3f761200-4371-11eb-85e6-657fdf29845e.png)

OS: CentOS
Python 3.7
Lightly1.07

PR #86, opened 2020-12-22
Restructure Dataset

philippmwirth:
# Restructure Dataset and Collate Function

To make it easier to understand the code and the package, we have decided to change the collate functions in such a way that they return a tuple of transformed images instead of a single batch with twice the batch size, i.e. the new collate function returns:
```python
batch = collate_fn(...)
(transformed_0, transformed_1), labels, filenames = batch
```
All of the models and loss functions were also adjusted so that the new workflow reads:
```python
# schematic workflow for training

for e in range(1, epochs+1):
   
    for (x0, x1), _, _ in dataloader:
        ...
        
        # forward pass of the transformed images
        y0, y1 = model(x0, x1)

        # calculate loss from the projections of the transformed images
        loss = loss_fn(y0, y1)

        ...

```

Furthermore, I picked up some issues along the way, including: 
- #84 Error while using VideoDataset without extension specified 
- Remove dead code (e.g. the VAEEmbedding)
- Clean up autocomplete

## TODO
- [x] Update docstrings
- [x] Benchmark the models to make sure they still work
- [x] Update the documentation


IgorSusmelj (approved):

IgorSusmelj:
Don't forget to update the code for the tutorials. Or do you want me to have a look at them when the PR is ready?

philippmwirth:
> 
> 
> Don't forget to update the code for the tutorials. Or do you want me to have a look at them when the PR is ready?

I will change the code for the tutorials and then you can take a look at them in the PR.

docs/source/tutorials_source/package/tutorial_simclr_clothing.py
```diff
@@ -167,14 +167,14 @@
...
-                                  max_epochs=max_epochs)
+encoder.train_embedding(gpus=gpus, 
+                        progress_bar_refresh_rate=100,
+                        max_epochs=max_epochs)
 
 # %%
 # Now, let's make sure we move the trained model to the gpu if we have one
 device = 'cuda' if gpus==1 else 'cpu'
-encoder.to(device)
+encoder = encoder.to(device)
```
IgorSusmelj, line 177: I think this is not needed

philippmwirth: yep

IgorSusmelj (approved):

PR #91, opened 2021-01-08
Add simsiam tutorial

philippmwirth:
# SimSiam Tutorial: Sentinel-2 Dataset Analysis

Made a tutorial using the recently added SimSiam model. The tutorial highlights how to train and embed images with the SimSiam model in *old-school PyTorch* style. 

Furthermore, I fixed a typo in the SimCLR tutorial and added an alternative way to download the docker image when the initial command has failed.


docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,447 @@
...
+# %%
+# Setup data augmentations and loaders
+# ------------------------------------
+# Since we're working on satellite images, it makes sence to use horizonal and
+# vertical flips as well as random rotation transformations. We apply weak color 
+# jitter to learn an invariance of the model with respect to slight changes in
+# the color of the water.
+#
+
+# the collate function defines the transformations for self-supervised learning
```
MalteEbner, line 90: The term 'data augmentation' is never used here, even though it is the most common word for such operations (not 'transformations' or even 'collate functions'. This confused me.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,447 @@
...
+
+# create a torchvision transformation for embedding the dataset after training
+test_transforms = torchvision.transforms.Compose([
+    torchvision.transforms.Resize((input_size, input_size)),
+    torchvision.transforms.ToTensor(),
+    torchvision.transforms.Normalize(
+        mean=lightly.data.collate.imagenet_normalize['mean'],
+        std=lightly.data.collate.imagenet_normalize['std'],
+    )
+])
```
MalteEbner, line 113: No reasoning is given for these transformations, it might be good to explain them.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,447 @@
...
+
+# create a lightly dataset for training (no transforms!)
+dataset_train_simsiam = lightly.data.LightlyDataset(
+    input_dir=path_to_data
+)
```
MalteEbner, line 118: It might be good to explain why no transforms are made here, even though the data augmentation transforms are needed only for training but not for testing. In general, it took me quite a while to understand which of the two transformations defined before is needed for training / testing and why this is so. 
Perhaps it is easier to understand if the order of the 6 blocks is changed: First the collate function (data augmentation function) is defined, then the train dataset, then the train data loader. After that, the three blocks for embedding images are explained: the resizing transform, the dataset definition and the data loader definition.
It should also be made clear why the resizing transform is added to the dataset, but the data augmentation transform to the data loader.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,447 @@
...
+    momentum=0.9,
+    weight_decay=5e-4
+)
+
+
+# %% 
+# Use a GPU if available
+
+
+device = 'cuda' if torch.cuda.is_available() else 'cpu'
```
MalteEbner, line 188: Perhaps put this block in line 208 just before the 'model.to(device)' as they belong together.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,447 @@
...
+        # calculate the per-dimension standard deviation of the outputs
+        # we can use this later to check whether the embeddings are collapsing
+        output, _ = y0
+        output = output.detach()
+        output = torch.nn.functional.normalize(output, dim=1)
+        
+        output_std = torch.std(output, 0)
+        output_std = output_std.mean()
+
+        # use running averages to track the loss and standard deviation
```
MalteEbner, line 239: 'running averages' -> 'moving averages'

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,447 @@
...
+import umap
+from sklearn import random_projection
+
+# %%
+# Then, we transform the embeddings using UMAP and rescale them to fit in the 
+# [0, 1] square.
+#
+
+
+# use a random Gaussian projection to get 2d embeddings
```
MalteEbner, line 310: You could explain why we need 2d-embeddings here and thus the projection.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,447 @@
...
+]
+
+
+def get_image_as_np_array(filename: str):
+    """Returns an image as a numpy array.
+    """
+    img = Image.open(filename)
+    return np.asarray(img)
+
+
```
MalteEbner, line 397: Some comments are very obvious and essentially say the same as the line of code above or below in the same words. Such comments do not add any value to the reader, but are just another thing to read. This applies to following comments: 

>  392: """Returns an image as a numpy array."""
> 402: " get shape of the image "
> 429: "add the subplot"
> 436: "plot the image"
> 441: "disable the axis"

I would rather have fewer and larger comments explaining the idea/summary of what is done than a comment for every line of code.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,447 @@
...
+    """Plots the example image and its eight nearest neighbors.
+    """
+    n_subplots = 9
+    # initialize empty figure
+    fig = plt.figure()
+    fig.suptitle(f"Nearest Neighbor Plot {i + 1}")
+    #
+    example_idx = filenames.index(example_image)
+    # get distances to the cluster center
+    distances = embeddings - embeddings[example_idx]
```
MalteEbner, line 422: Why not simply use np.pairwise_distances() ?

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,447 @@
...
+    distances = np.power(distances, 2).sum(-1).squeeze()
+    # sort indices by distance to the center
+    argsort = np.argsort(distances)
```
MalteEbner, line 425: 'argsort' is quite a bad name, because it is also a function. Furthermore doing a complete sort is not efficient, replace it with `indices_closest_images = np.argpartition(distances,n_subplots-1)[:n_subplots]`

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,447 @@
...
+    # show images
+    plot_indices = argsort[:n_subplots]
+    for plot_offset in range(n_subplots):
```
MalteEbner, line 428: Why not `for plot_offset, image_id in enumerate(closest_indices):` ? It would also replace line 432

docs/source/tutorials_source/package/tutorial_simclr_clothing.py
```diff
@@ -133,7 +133,9 @@
 # %%
 # Create the SimCLR model
 # -----------------------
```
MalteEbner, line 135: That does not belong in this commit

docs/source/docker/getting_started/setup.rst
```diff
@@ -48,6 +48,12 @@ We assume *container-credentials.json* is in your current directory.
 
     cat container-credentials.json | docker login -u json_key --password-stdin https://eu.gcr.io
 
+If the above command does not work, try the following:
```
MalteEbner, line 51: That does not belong in this commit

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
philippmwirth: I am not aware of such a Numpy function, do you mean [this one](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html)? In any case I think the code here is fine.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
philippmwirth: The advantage of the complete sort is that the first image is still the example image whereas with argpartition the order of the indices is undefined. Furthermore, this line is by far no bottleneck - I will continue using the full sort but change the name.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
MalteEbner: You are right, it is from sklearn. I just thought about it, as it is used for the python implementation of coreset.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
MalteEbner: If it is not the bottleneck, then the performance is indeed irrelevant. I didn't think about the order within the 9 nearest images being relevant, sorry.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,443 @@
...
+dataset_train_simsiam = lightly.data.LightlyDataset(
+    input_dir=path_to_data
+)
+
+# create a dataloader for training
+dataloader_train_simsiam = torch.utils.data.DataLoader(
+    dataset_train_simsiam,
+    batch_size=batch_size,
+    shuffle=True,
+    collate_fn=collate_fn,
```
IgorSusmelj, line 117: I would add `drop_last=True` for training of the models. Besides performance boost (because the input tensor shape is always the same) it could also help with stability during training.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,443 @@
...
+# vectors. If it is close to one divided by the square root of the output 
+# dimension, everything is fine (you can read
+# up on this idea `here <https://arxiv.org/abs/2011.10566>`_).
+
+device = 'cuda' if torch.cuda.is_available() else 'cpu'
+model.to(device)
+
+avg_loss = 0.
+avg_output_std = 0.
+for e in range(1, epochs + 1):
```
IgorSusmelj, line 213: Would just do `for epoch in range(epochs):`. Maybe I missed it, but I don't see any problem having the first epoch be 0.

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
```diff
@@ -0,0 +1,443 @@
...
+        # backpropagation
+        loss = criterion(y0, y1)
+        loss.backward()
+
+        optimizer.step()
+        optimizer.zero_grad()
+
+        # calculate the per-dimension standard deviation of the outputs
+        # we can use this later to check whether the embeddings are collapsing
+        output, _ = y0
```
IgorSusmelj, line 233: I miss somehow the information that the SimSiam forward pass returns a tuple of (prediction, projection). Maybe adding a comment or renaming the variables would help?

docs/source/tutorials_source/package/tutorial_simsiam_esa.py
philippmwirth: Ok, I don't mind adding `drop_last=True`. Not sure it has such a big impact though.

IgorSusmelj (approved):

Issue #83, opened 2020-12-19
Improve Documentation


IgorSusmelj:
Let's add some high-level information about the structure of the package. And also explain how the different modules are connected. Maybe even make an illustration?

PR #110, opened 2021-01-27
Add lightly overview to docs

philippmwirth:
# Overview of lightly in docs

#83 

I added a diagram to show the different components of lightly and how they interact. This should help users and possible contributors to understand the framework better.

Below you can see the diagram I added:
lightly_overview

And here's a pdf of the start page of the docs: [lightly_overview.pdf](https://github.com/lightly-ai/lightly/files/5881745/lightly_overview.pdf)

@IgorSusmelj should I also add the diagram to the README? What do you think?


IgorSusmelj (approved):

IgorSusmelj:
It looks great. Maybe add it to the "terminology" section in the readme?

MalteEbner:
The grey box of the diagram shows that the user can choose custom input data, backbone and optimiser, but not a custom dataset, collate function, loss or model. 

The lightly docs, specifically the [options for the CLI](https://docs.lightly.ai/lightly.cli.html#ref-cli-config-default) list the input_data, model, loss criterion, collate function, dataloader and trainer to be customisable. This contradicts the border of the grey box and might be confusing to our users.

My proposition would be to put only the input_data and output (trained_model and embeddings) outside of the grey box, as lightly is providing everything else including the backbone and optimiser.

MalteEbner:
I like the first page of the documentation starting with the diagram. What I like especially is that the terms the diagram lists in bold are explained below and the tutorials are linked directly to these terms. This provides a very connected and fluent introduction into lightly.

What I would put directly under the diagram is a very short code snippet showing how easy it is to use lightly to get the two outputs from the input data. It could look like this:

```
import lightly

embeddings, pre_trained_backbone, filenames = lightly.train_model_and_embed_images(input_dir='my/cute/cats/dataset')
```

philippmwirth:
> 
> 
> The grey box of the diagram shows that the user can choose custom input data, backbone and optimiser, but not a custom dataset, collate function, loss or model.
> 
> The lightly docs, specifically the [options for the CLI](https://docs.lightly.ai/lightly.cli.html#ref-cli-config-default), list the input_data, model, loss criterion, collate function, dataloader and trainer to be customisable. This contradicts the border of the grey box and might be confusing to our users.
> 
> My proposition would be to put only the input_data and output (trained_model and embeddings) outside of the grey box, as lightly is providing everything else including the backbone and optimiser.

But this is only true for the command-line tool and for the one-line-code (the one you made the example with). What I want to highlight with this diagram is the flexibility of lightly: You can choose the backbone, you can choose the optimizer, etc. If I move the backbone inside it could be misleading because we only provide resnet backbones.

philippmwirth:
> 
> 
> I like the first page of the documentation starting with the diagram. What I like especially is that the terms the diagram lists in bold are explained below and the tutorials are linked directly to these terms. This provides a very connected and fluent introduction into lightly.
> 
> What I would put directly under the diagram is a very short code snippet showing how easy it is to use lightly to get the two outputs from the input data. It could look like this:
> 
> ```
> import lightly
> 
> embeddings, pre_trained_backbone, filenames = lightly.train_model_and_embed_images(input_dir='my/cute/cats/dataset')
> ```

That's a great idea! Maybe I will use the code for the self-supervised embedding though as it shows directly how the different components I showed are connected, i.e.

```python
encoder = lightly.embedding.SelfSupervisedEmbedding(model, loss, optimizer, dataloader)
encoder.train(gpus=1, max_epochs=10)
```

WDYT?

philippmwirth:
I didn't change the diagram for now but I added it to the README and also added a tiny code example for the self-supervised embedding to the index. [Here is how it looks now.](https://github.com/lightly-ai/lightly/files/5888591/lightly_overview.pdf)

IgorSusmelj (approved):

IgorSusmelj:
Looks great!

Issue #111, opened 2021-01-27
Add support for new tag representation (binary mask)


IgorSusmelj:
The new Lightly platform will use a new tag representation. The transmitted data will be encoded as (16bit) hex strings.

The goal is to provide a simple helper class to work with this new format and switch between hex representation and binary representation.

Some functionalities which need to be implemented:

- convert a hex string from the API to a binary mask
```
'ab3f' --> '1010101100111111'
```
- get the set indices from a binary mask
- basic arithmetics (&, ~) to set individual bits or invert the bitmask

PR #113, opened 2021-01-28
Add bitmask representation

philippmwirth:
# Add bitmask representation

#111, *this PR is to the `active_learning` branch*

## Description
I added a helper class to load and manipulate tags in the new format. It supports:
- initialization from binary and hexstrings
- conversion to binary and hexstrings
- union and intersection
- setting/clearing/getting bits by index
- function to get the list of indices of nonzero bits
- tests

**Currently missing:** A proper way to calculate the inverse of a tag.

### DIFF
--- a/lightly/api/bitmask.py
+++ b/lightly/api/bitmask.py
@@ -109,6 +109,14 @@ class BitMask:
         >>> mask = BitMask.from_hex('0x6')
         >>> mask = Bitmask.from_bin('0b0110')
 
+        >>> # for a dataset with 10 images, assume the following tag
+        >>> # 0001011001 where the 1st, 4th, 5th and 7th image are selected
+        >>> # this tag would be stored as 0x59.
+        >>> hexstring = 0x59                    # what you receive from the api
+        >>> mask = BitMask.from_hex(hexstring)  # create a bitmask from it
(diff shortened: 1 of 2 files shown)


lightly/api/bitmask.py
```diff
@@ -0,0 +1,216 @@
...
+
+    Attributes:
+        x:
+            An integer representation of the binary mask.
+
+    Examples:
+        >>> # the following are equivalent
+        >>> mask = BitMask(6)
+        >>> mask = BitMask.from_hex('0x6')
+        >>> mask = Bitmask.from_bin('0b0110')
```
IgorSusmelj, line 110: Please add an example covering the data as it would come from the API:`"A8B42"` or so) `0x..` notation is Python internal once it's a hex string. But input to the BitMask is a normal string. 
Also, add an example to get the string (same format as API).

philippmwirth: I requested a change on your PR to use the `0x` notation and @japrescott agreed with me. Let's agree first and then make changes.

IgorSusmelj: Now, you can add an example :)

philippmwirth:
@IgorSusmelj how's this? :)

IgorSusmelj (approved):

PR #115, opened 2021-01-29
Tutorial on custom augmentations

philippmwirth:
# Tutorial on custom augmentations

The idea is to show how to easily use custom augmentations with lightly. The augmentations are based on [this paper](https://arxiv.org/pdf/2101.04909.pdf).
The [dataset](https://www.kaggle.com/c/vinbigdata-chest-xray-abnormalities-detection/data) is also a collection of chest X-rays.

The general structure of the tutorial is as follows:
1. Write custom augmentations and show what they do on an example image.
2. Train MoCo with the custom augmentations
3. Show that the distribution of the labels of the nearest neighbors of an image similar to the labels of the image. (The dataset is meant for object detection and this is just one way to show that the embeddings make sense)

Unfortunately, the last step doesn't work very well with the current training settings:
- ResNet18
- Batch size 64
- Num splits 0 (because the batch size is tiny)
- Input size 224

Therefore, it would be great if @IgorSusmelj could try to run the tutorial with an increased batch size (and num_splits > 0) and maybe also a better model. Let me know how it goes! Otherwise we'll have to find other plots to show at the bottom.

PS here's an example report: [custom_augmentations_tutorial.pdf](https://github.com/lightly-ai/lightly/files/5894145/custom_augmentations_tutorial.pdf)

### DIFF
--- a/docs/source/tutorials_source/package/tutorial_custom_augmentations.py
+++ b/docs/source/tutorials_source/package/tutorial_custom_augmentations.py
@@ -1,0 +1,244 @@
+"""
+.. _lightly-custom-augmentation-5:
+
+Tutorial 5: Custom Augmentations
+==============================================
+
+"""
+
+# %%
+# Imports
+# -------
+#
+# Import the Python frameworks we need for this tutorial.
+import os
+import glob
+import torch
+import torch.nn as nn
+import torchvision
+import pytorch_lightning as pl
+import lightly
+import matplotlib.pyplot as plt
+from sklearn.neighbors import NearestNeighbors
+from sklearn.preprocessing import normalize
+from PIL import Image, ImageOps
+import numpy as np
+import pandas
+
+# %%
+# Configuration
+# -------------
+# TODO
+#
+num_workers = 8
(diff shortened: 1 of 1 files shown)


docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,244 @@
...
+max_epochs = 100
+num_ftrs = 500
+
+# %%
+# Let's set the seed for our experiments
+pl.seed_everything(seed)
+
+# %%
+# Make sure `path_to_data` points to the downloaded x-ray dataset.
+# You can download the dataset `from kaggle <https://www.kaggle.com/c/vinbigdata-chest-xray-abnormalities-detection/overview>`_.
```
IgorSusmelj, line 46: Images in this dataset are in DICOM format which is great. I would also use DICOM in this tutorial since it's a very common format for medical imaging. To convert from DICOM to NumPy/torch we can use this library: https://github.com/pydicom/pydicom

PR #121, opened 2021-02-03
Tests: include data/_utils.py:check_images()

MalteEbner:


tests/data/test_LightlyDataset.py
```diff
@@ -135,6 +136,26 @@ def test_create_lightly_dataset_from_folder_nosubdir(self):
         for i in range(n_tot):
             sample, target, fname = dataset[i]
 
+    def test_check_images(self):
```
IgorSusmelj, line 139: This test does not cover corrupt images. How about creating a few files with random strings in them and store them as `.jpg` to test whether the methods also detects the corrupt images?

IgorSusmelj (approved):

PR #123, opened 2021-02-11
109 Tutorial custom augmentations pw

philippmwirth:
# Tutorial on custom augmentations

Added a tutorial on self-supervised learning on X-ray images using custom augmentations.

The tutorial can be seen [here](https://docs.lightly.ai/tutorials/package/tutorial_custom_augmentations.html).

### DIFF
--- a/docs/source/tutorials_source/package/tutorial_custom_augmentations.py
+++ b/docs/source/tutorials_source/package/tutorial_custom_augmentations.py
@@ -1,0 +1,244 @@
+"""
+.. _lightly-custom-augmentation-5:
+
+Tutorial 5: Custom Augmentations
+==============================================
+
+"""
+
+# %%
+# Imports
+# -------
+#
+# Import the Python frameworks we need for this tutorial.
+import os
+import glob
+import torch
+import torch.nn as nn
+import torchvision
+import pytorch_lightning as pl
+import lightly
+import matplotlib.pyplot as plt
+from sklearn.neighbors import NearestNeighbors
+from sklearn.preprocessing import normalize
+from PIL import Image, ImageOps
+import numpy as np
+import pandas
+
+# %%
+# Configuration
+# -------------
+# TODO
+#
+num_workers = 8
+batch_size = 64
+num_splits = 0
+seed = 1
+max_epochs = 100
+num_ftrs = 500
+
+# %%
+# Let's set the seed for our experiments
+pl.seed_everything(seed)
+
+# %%
+# Make sure `path_to_data` points to the downloaded x-ray dataset.
+# You can download the dataset `from kaggle <https://www.kaggle.com/c/vinbigdata-chest-xray-abnormalities-detection/overview>`_.
+
+# %%
+
+path_to_data = '/datasets/vinbigdata/train'
+
+
+# %%
+# TODO
+
+class HistogramEqualize(object):
+    """
+
+    """
+
+    def __call__(self, sample: Image):
+        """
+
+        """
+        return ImageOps.equalize(sample)
+
+class GaussianNoise(object):
+    """
+
+    """
+
+    def __call__(self, sample: torch.Tensor):
+        """
+
+        """
+        mu = sample.mean()
+        snr = np.random.randint(low=4, high=8)
+        sigma = mu / snr
+        noise = torch.normal(torch.zeros(sample.shape), sigma)
+        return sample + noise
+
+# %%
+# Setup data augmentations and loaders
+# ------------------------------------
+#
+# TODO
+transform = torchvision.transforms.Compose([
+    torchvision.transforms.RandomResizedCrop(size=224, scale=(0.2, 1.0)),
+    torchvision.transforms.RandomHorizontalFlip(p=0.5),
+    torchvision.transforms.RandomVerticalFlip(p=0.5),
+    torchvision.transforms.GaussianBlur(21),
+    HistogramEqualize(),
+    torchvision.transforms.ToTensor(),
+    GaussianNoise(),
+])
+
+# %%
+#
+
+example_image_name = os.path.join(path_to_data, 'ffeffc54594debf3716d6fcd2402a99f.jpg')
+example_image = Image.open(example_image_name)
+
+augmented_image_1 = transform(example_image).numpy()
+augmented_image_2 = transform(example_image).numpy()
+
+fig, axs = plt.subplots(1, 3)
+
+axs[0].imshow(np.asarray(example_image))
+axs[0].set_axis_off()
+axs[0].set_title('Original Image')
+
+axs[1].imshow(augmented_image_1.squeeze())
+axs[1].set_axis_off()
+
+axs[2].imshow(augmented_image_2.squeeze())
+axs[2].set_axis_off()
+
+
+# %%
+#
+
+collate_fn = lightly.data.BaseCollateFunction(transform)
+
+
+dataset_train_simclr = lightly.data.LightlyDataset(
+    input_dir=path_to_data
+)
+
+dataloader_train_simclr = torch.utils.data.DataLoader(
+    dataset_train_simclr,
+    batch_size=batch_size,
+    shuffle=True,
+    collate_fn=collate_fn,
+    drop_last=True,
+    num_workers=num_workers
+)
+
+# %%
+# Create the SimCLR model
+# -----------------------
+# Create a ResNet backbone and remove the classification head
+
+# TODO
+resnet = lightly.models.ResNetGenerator('resnet-18', num_splits=num_splits)
+last_conv_channels = list(resnet.children())[-1].in_features
+backbone = nn.Sequential(
+    *list(resnet.children())[:-1],
+    nn.Conv2d(last_conv_channels, num_ftrs, 1),
+    nn.AdaptiveAvgPool2d(1)
+)
+
+# create the MoCo model using the newly created backbone
+model = lightly.models.MoCo(backbone, num_ftrs=num_ftrs, m=0.99)
+
+# TODO
+criterion = lightly.loss.NTXentLoss(
+    temperature=0.1,
+    memory_bank_size=4096,
+)
+
+optimizer = torch.optim.SGD(model.parameters(), lr=0.1, momentum=0.9, weight_decay=1e-4)
+encoder = lightly.embedding.SelfSupervisedEmbedding(
+    model,
+    criterion,
+    optimizer,
+    dataloader_train_simclr
+)
+
+gpus = 1 if torch.cuda.is_available() else 0
+encoder.train_embedding(gpus=gpus, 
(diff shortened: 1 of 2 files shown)


docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+
+The original dataset this tutorial is based on can be found `on Kaggle <https://www.kaggle.com/c/vinbigdata-chest-xray-abnormalities-detection/overview>`_.
+These images are in the DICOM format. For simplicity and efficiency reasons, 
+we randomly selected ~4000 images from the above dataset, resized them such that the
+maximum of the width and height of each image is no larger than 512, and converted
+them to the 16-bit TIFF format. To do so, we used ImageMagick:
+
+.. code::
+
+    mogrify -path path/to/new/dataset -resize 512x512 -format tiff "*.dicom"
```
IgorSusmelj, line 26: Did you check whether the tiff image has 16-bit?

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+does not make sense.
+
+We will show how to address these problems and how to train a ResNet-18 with MoCo
+on a set of 16-bit X-ray images in TIFF format.
+
+The original dataset this tutorial is based on can be found `on Kaggle <https://www.kaggle.com/c/vinbigdata-chest-xray-abnormalities-detection/overview>`_.
+These images are in the DICOM format. For simplicity and efficiency reasons, 
+we randomly selected ~4000 images from the above dataset, resized them such that the
+maximum of the width and height of each image is no larger than 512, and converted
+them to the 16-bit TIFF format. To do so, we used ImageMagick:
```
IgorSusmelj, line 22: Maybe add a hint on how to install or that it's already installed on most Linux systems.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+        sigma = mu / snr
+        noise = torch.normal(torch.zeros(sample.shape), sigma)
+        return sample + noise
+
+
+# %%
+# Now that we have implemented our custom augmentations, we can combine them
+# with available augmentations from the torchvision library. Make sure, that
+# the first augmentation is the histogram normalization, and that the Gaussian
+# noise is applied after converting the image to a tensor.
```
IgorSusmelj, line 134: Mention that we follow the augmentations from the Facebook paper.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+
+axs[1].imshow(augmented_image_1)
+axs[1].set_axis_off()
+
+axs[2].imshow(augmented_image_2)
+axs[2].set_axis_off()
+
+# %%
+# Finally, in order to use the augmentation pipeline we defined for self-supervised
+# learning, we need to create a Lightly collate function like so:
```
IgorSusmelj, line 182: Let's use `lightly` for anything related to the pip package and `Lightly` if we talk about the company :)

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+
+def tiff_loader(f):
+    """Loads a 16-bit tiff image and returns it as a numpy array.
+
+    """
+    with open(f, 'rb') as f:
+        image = Image.open(f)
+        return np.array(image)
+
+# create the dataset and overwrite the image loader
```
IgorSusmelj, line 204: I would like to get a bit more context here. E.g. mention that lightly follows the torchvision dataset and therefore has a `.loader` property which is called to load an image.
Can you point to the source code to show how the `loader` is used?

The reason is, this is very new. Not everyone looks at the torchvision source code. And even less look at lightly. So we need to explain why there is a loader and why we overwrite it.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+dataset_train = lightly.data.LightlyDataset(input_dir=path_to_data)
+dataset_train.dataset.loader = tiff_loader
```
IgorSusmelj, line 206: Overwriting is btw. a very elegant solution. Just a few lines of code. I haven't thought about it :)

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+# %%
+# Setup criterion and optimizer
+# -----------------------------
+# For the criterion, we use the NTXentLoss which should always be used with MoCo.
+# MoCo also requires a memory bank - we set its size to 4096.
+#
+# The choice of the optimizer is left to the user. Here, we go with simple stochastic
+# gradient descent with momentum.
+criterion = lightly.loss.NTXentLoss(
+    temperature=0.1,
```
IgorSusmelj, line 246: Is there are the reason we temperature of 0.1? I would either mention why we use it or switch to default or use the one from the paper.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+    memory_bank_size=4096,
```
IgorSusmelj, line 247: Same here, mention why we use 4096 or use MoCo default which is 16384

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+)
+
+optimizer = torch.optim.SGD(model.parameters(), lr=0.1, momentum=0.9, weight_decay=1e-4)
```
IgorSusmelj, line 250: Same here, why do we use this lr? momentum is usually 0.9 (that's fine for me). But also weight_decay. Number from paper or "this just worked best for us".

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+#
+# We choose three example images. For each example image, we find 50 nearest neighbors.
+# Then, we plot the critical findings in the example image (dark blue) and the distribution
+# of the critical findings in the nearest neighbor images (light blue) as bar plots.
+
+
+# transform the original annotations to multiclass labels
+fnames = [fname.split('.')[0] for fname in fnames]
+
+df = pandas.read_csv('/datasets/vinbigdata/train.csv')
```
IgorSusmelj, line 326: This block is a bit too magic. Maybe show the first 3-4 rows of the csv file (including heading) in the comments/ text. So someone just reading/ skimming through the tutorial would get an idea why we have the next few lines?

IgorSusmelj: Maybe also mention what the output is. Show an example of what is in `multilabels` after the "magic" :)

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+
+def plot_knn_multilabels(embeddings, multilabels, samples_idx, n_neighbors=50):
+    """Plots multiple rows of random images with their nearest neighbors
+    """
+    # lets look at the nearest neighbors for some samples
+    # we use the sklearn library
+    nbrs = NearestNeighbors(n_neighbors=n_neighbors).fit(embeddings)
+    _, indices = nbrs.kneighbors(embeddings)
+
+    # TODO
```
IgorSusmelj, line 348: remove todo

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+        bars1 = multilabels[idx]
+        bars2 = np.mean(multilabels[indices[idx]], axis=0)
+
+        plt.bar(r1, bars1, color='steelblue', edgecolor='black', width=bar_width)
+        plt.bar(r2, bars2, color='lightsteelblue', edgecolor='black', width=bar_width)
+        plt.xticks(0.5 * (r1 + r2), classes, rotation=90)
+        plt.tight_layout()
+
+
+plot_knn_multilabels(embeddings, multilabels, [4111, 3340, 1796])
```
IgorSusmelj, line 366: Add more information to the plot. Why do we have this plot (like you did in the presentation).

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -0,0 +1,366 @@
...
+for filename, label in zip(df.image_id, df.class_name):
+    try:
+        i = fnames.index(filename.split('.')[0])
+        j = classes.index(label)
+        multilabels[i, j] = 1.
+    except:
+        pass
+
+
+def plot_knn_multilabels(embeddings, multilabels, samples_idx, n_neighbors=50):
```
IgorSusmelj, line 340: With only 4'000 samples kNN with k=50 might be quite high, no? Since you mentioned that you want to use the plot to see similar cases. I'm not sure how balanced the dataset is. Or did you already do some experiments with lower numbers (e.g. 10 or 20)?

IgorSusmelj (changes requested):
Would explain some parts a bit better

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
philippmwirth: yes

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
philippmwirth: ahh I always forget...

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
philippmwirth: tbh, it's because that worked well on cifar10... I will look into it

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
philippmwirth: I think I'll just add an explanation, the CSV file is not so informative.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
philippmwirth: We could try lower numbers. IMO the larger the number the more impressive the result ;)

IgorSusmelj (approved):

PR #125, opened 2021-02-11
Openapi generated client

MalteEbner:
PR consist of following steps / commits:

1. copied templates to generate client -> all changed files in the folder `openapi_client_generator/python/`
2. wrote script to generated the client at  `openapi_client_generator/generate_openapi_client.sh`-> should be tested
3. generated the client using the script -> all changed files in the folder `lightly/openapi_generated/`
4. changed the .mustache template to allow required and nullable values to be None (e.g. the previousTagId). First the model.mustache template was changed, then the script war rerun, which caused the generated code to change.


lightly/openapi_generated/requirements.txt
```diff
@@ -0,0 +1,5 @@
+certifi >= 14.05.14
+six >= 1.10
+python_dateutil >= 2.5.3
+setuptools >= 21.0.0
+urllib3 >= 1.15.1
```
philippmwirth, line 5: This introduces new requirements, they should be specified in the PIP package requirements.

lightly/openapi_generated/swagger_client/api/annotations_api.py
```diff
@@ -0,0 +1,232 @@
...
+            body=body_params,
+            post_params=form_params,
+            files=local_var_files,
+            response_type='list[AnnotationData]',  # noqa: E501
+            auth_settings=auth_settings,
+            async_req=params.get('async_req'),
+            _return_http_data_only=params.get('_return_http_data_only'),
+            _preload_content=params.get('_preload_content', True),
+            _request_timeout=params.get('_request_timeout'),
+            collection_formats=collection_formats)
```
philippmwirth, line 232: What do we need the annotations api for?

lightly/openapi_generated/swagger_client/api/internal_api.py
```diff
@@ -0,0 +1,133 @@
...
+            body=body_params,
+            post_params=form_params,
+            files=local_var_files,
+            response_type=None,  # noqa: E501
+            auth_settings=auth_settings,
+            async_req=params.get('async_req'),
+            _return_http_data_only=params.get('_return_http_data_only'),
+            _preload_content=params.get('_preload_content', True),
+            _request_timeout=params.get('_request_timeout'),
+            collection_formats=collection_formats)
```
philippmwirth, line 133: I'm pretty sure that this code should not be public. Can you remove all the generated API code which is not being used?

lightly/openapi_generated/swagger_client/models/annotation_state.py
```diff
@@ -0,0 +1,105 @@
...
+            return False
+
+        return self.to_dict() == other.to_dict()
+
+    def __ne__(self, other):
+        """Returns true if both objects are not equal"""
+        if not isinstance(other, AnnotationState):
+            return True
+
+        return self.to_dict() != other.to_dict()
```
philippmwirth, line 105: All the annotation data / requests are not needed by the PIP package, or am I wrong?

lightly/openapi_generated/swagger_client/models/create_cf_bucket_activity_request.py
```diff
@@ -0,0 +1,151 @@
...
+            return False
+
+        return self.to_dict() == other.to_dict()
+
+    def __ne__(self, other):
+        """Returns true if both objects are not equal"""
+        if not isinstance(other, CreateCFBucketActivityRequest):
+            return True
+
+        return self.to_dict() != other.to_dict()
```
philippmwirth, line 151: This is not needed in the PIP package

openapi_client_generator/python/model.mustache
```diff
@@ -99,8 +99,10 @@ class {{classname}}(object):
         :type: {{datatype}}
         """
 {{#required}}
+{{^nullable}}
         if self._configuration.client_side_validation and {{name}} is None:
             raise ValueError("Invalid value for `{{name}}`, must not be `None`")  # noqa: E501
+{{/nullable}}
 {{/required}}
 {{#isEnum}}
 {{#isContainer}}
```
philippmwirth, line 108: Ok do we have to change this everytime we generate the API code? In that case I would make an additional change in the `generate_openapi_client.sh` script which applies this as a patch.

philippmwirth (changes requested):
Please correct me if I'm wrong.

In my understanding, the following points hold:
- We have a part of [this](https://github.com/swagger-api/swagger-codegen/tree/master/modules/swagger-codegen/src/main/resources/python) repo in our repo
- Everytime we build the api from scratch, we also need to make changes to `tag_data.py` and `model.mustache`

If so, please make changes such that the necessary code from the [other repo](https://github.com/swagger-api/swagger-codegen/tree/master/modules/swagger-codegen/src/main/resources/python) is not part of our repository. Instead, give instructions on how to get the required code or clone it automatically in the generator script. Also, can we apply the necessary changes to the two files as a patch or similar?

Let me know if anything is unclear.

MalteEbner:
@philippmwirth:
Do I understand you correctly, that you basically wish the following changes?

1. If one generates a new python client from a new api.yml, no manual changes, to anything (neither .mustache files, nor generated files) should be necessary. This point is already fulfilled:
The change with the /nullable to the model.mustach-template only has the be made once (in this PR), it basically creates our own python code generator. If a new api is written, only the script has to be rerun, no manual changes of any code have to be performed.
2. We do not want that many copied files as template. Ideally, we could use the generator without customers templates. To achieve this, I create the following PR for swagger-codegen: https://github.com/swagger-api/swagger-codegen/pull/10906
3. The lightly pip package should not contain any code needed for generating the client. We could solve this by putting the files in the folder `openapi_client_generator` into the lightly-core package. What do you think about that?
4. There should be as little files and code in the repo as possible. Thus all generated files not needed should be deleted.
5. The new requirements added should be specified in the PIP package requirements.

philippmwirth:
> 
> 
> @philippmwirth:
> Do I understand you correctly, that you basically wish the following changes?
> 
>     1. If one generates a new python client from a new api.yml, no manual changes, to anything (neither .mustache files, nor generated files) should be necessary. This point is already fulfilled:
>        The change with the /nullable to the model.mustach-template only has the be made once (in this PR), it basically creates our own python code generator. If a new api is written, only the script has to be rerun, no manual changes of any code have to be performed.
> 
>     2. We do not want that many copied files as template. Ideally, we could use the generator without customers templates. To achieve this, I create the following PR for swagger-codegen: [swagger-api/swagger-codegen#10906](https://github.com/swagger-api/swagger-codegen/pull/10906)
> 
>     3. The lightly pip package should not contain any code needed for generating the client. We could solve this by putting the files in the folder `openapi_client_generator` into the lightly-core package. What do you think about that?
> 
>     4. There should be as little files and code in the repo as possible. Thus all generated files not needed should be deleted.
> 
>     5. The new requirements added should be specified in the PIP package requirements.



> 
> 
> @philippmwirth:
> Do I understand you correctly, that you basically wish the following changes?
> 
>     1. If one generates a new python client from a new api.yml, no manual changes, to anything (neither .mustache files, nor generated files) should be necessary. This point is already fulfilled:
>        The change with the /nullable to the model.mustach-template only has the be made once (in this PR), it basically creates our own python code generator. If a new api is written, only the script has to be rerun, no manual changes of any code have to be performed.
> 
>     2. We do not want that many copied files as template. Ideally, we could use the generator without customers templates. To achieve this, I create the following PR for swagger-codegen: [swagger-api/swagger-codegen#10906](https://github.com/swagger-api/swagger-codegen/pull/10906)
> 
>     3. The lightly pip package should not contain any code needed for generating the client. We could solve this by putting the files in the folder `openapi_client_generator` into the lightly-core package. What do you think about that?
> 
>     4. There should be as little files and code in the repo as possible. Thus all generated files not needed should be deleted.
> 
>     5. The new requirements added should be specified in the PIP package requirements.

Yes, this is pretty accurate. Only one thing: I don't think it would be such a good idea to put the generator in the lightly-core repo. Can we not simply clone the code and apply the changes with a patch? My guess is we could write a script which retrieves the code, applies the patch, and then generates the openapi code.

PR #131, opened 2021-02-16
Active learning sampling without scores (step 2)

MalteEbner:
- added bitmask files, copied directly from https://github.com/lightly-ai/lightly/pull/113/files:
- added sampling workflow including sampler_config and tests 
- created active learning agent including tests and scorer
- Bugfix: added  generated `__init__.py`
- Created` test_UNMOCKED_complety.py`, please run this script to test the new workflow end2end on the server: Create a new dataset, set the environment variable TOKEN, set the dataset id and the path to the dataset and then run the script.

Known issues:
- return value of `ActiveLearningAgent` is list of filenames - what is required?
- `LightlyDataset` should have some functionality to provide files exactly in the order specified by a list of filenames -> new issue?
- internal methods starting with `_` are not necessarily docstring-documented.


lightly/active_learning/agents/agent.py
```diff
@@ -0,0 +1,55 @@
...
+from lightly.api.bitmask import BitMask
+
+
+class ActiveLearningAgent:
+    """A basic class providing an active learning policy
+
+    Args:
+        api_workflow_client: The client to connect to the api
+        preselected_tag_id: The tag defining the already chosen samples (e.g. already labelled ones), default: None
+    """
```
philippmwirth, line 15: Classes don't have args but attributes (see [3.8.4](https://google.github.io/styleguide/pyguide.html)). Also, don't forget the newline after the colon and add punctuation
```python
Attributes:
    api_workflow_client:
        The client to connect to the api.
```

lightly/active_learning/config/sampler_config.py
```diff
@@ -0,0 +1,51 @@
...
+
+from lightly.openapi_generated.swagger_client.models.sampling_method import SamplingMethod
+from lightly.openapi_generated.swagger_client.models.sampling_config import SamplingConfig
+from lightly.openapi_generated.swagger_client.models.sampling_create_request import SamplingCreateRequest
+from lightly.openapi_generated.swagger_client.models.sampling_config_stopping_condition import \
+    SamplingConfigStoppingCondition
+
+
+class SamplerConfig:
+    """The configuration parameters of a sampler
```
philippmwirth, line 11: Why are the pre-selected and query tags not part of the config?

lightly/active_learning/config/sampler_config.py
```diff
@@ -0,0 +1,51 @@
...
+        self.method = method
+        self.batch_size = batch_size
+        self.min_distance = min_distance
+        if name is None:
+            date_time = datetime.now().strftime("%m/%d/%Y, %H:%M:%S")
+            name = f"{self.method}_{self.batch_size}_{self.min_distance}_{date_time}"
+        self.name = name
+
+    def _get_as_api_sampling_create_request(self, preselected_tag_id: str = None,
+                                            query_tag_id: str = None) -> SamplingCreateRequest:
```
philippmwirth, line 31: `_get_as_api_sampling_create_request` -> `to_api_sampling_create_request`?

lightly/active_learning/agents/agent.py
```diff
@@ -0,0 +1,55 @@
...
+        new_tag_data = self.api_workflow_client.sampling(
+            sampler_config=sampler_config, al_scores=scores,
+            preselected_tag_id=self.preselected_tag_id, query_tag_id=query_tag_id)
+
+        # set the newly chosen tag as the new preselected_tag_id
+        self.preselected_tag_id = new_tag_data.id
+
+        # extract the chosen samples
+        chosen_samples_ids = BitMask.from_hex(new_tag_data.bit_mask_data).to_indices()
+        chosen_filenames = [self.api_workflow_client.filenames_on_server[i] for i in chosen_samples_ids]
```
philippmwirth, line 53: nice! although slow for large tags :/

lightly/api/api_workflow_sampling.py
```diff
@@ -0,0 +1,62 @@
...
+        payload = sampler_config._get_as_api_sampling_create_request(
+            preselected_tag_id=preselected_tag_id, query_tag_id=query_tag_id)
+        payload.row_count = 15  # TODO: remove after solving https://github.com/lightly-ai/lightly-core/issues/150
+        response = self.samplings_api.trigger_sampling_by_id(payload, self.dataset_id, self.embedding_id)
+        job_id = response.job_id
+
+        # poll the job status till the job is finished
+        exception_counter = 0  # TODO; remove after solving https://github.com/lightly-ai/lightly-core/issues/150
+        while True:
+            try:
```
philippmwirth, line 43: I guess `while True` is a no-no... Do you think we can find a save alternative?

tests/api_workflow/test_UNMOCKED_complete.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # perform_a_sampling
+    print("Starting performing a sampling")
+    agent = ActiveLearningAgent(api_workflow_client)
+    sampler_config = SamplerConfig(batch_size=8)
+    chosen_samples_ids, chosen_filenames = agent.sample(sampler_config=sampler_config)
+    print(f'chosen_filenames: {chosen_filenames}')
+    print("Finished the sampling")
+
+
```
philippmwirth, line 54: Can you add a second test here where we sample using `preselectedTag` and `queryTag`?

philippmwirth (changes requested):
The general idea and layout makes sense and looks good to me. I have added comments about the docstring format and also a few questions.

One thing that bothers me is that the user has to directly pass the `queryTagId` and `preselectedTagId`. Would it not be more sensible to use the tag name for this? Otherwise it's very "unfriendly" for the user. Additionally, we need to implement an easy way for the user to do active learning (this should be step 2): It should be very low effort to sample an initial pool of samples from the initial tag and then subsequently extend it by more sampling iterations (meaning we need to set the pre-selected tag to whatever was sampled and the query tag to whatever was not sampled).

lightly/api/api_workflow_sampling.py
```diff
@@ -0,0 +1,93 @@
...
+            ApiException, ValueError
+        """
+
+        # upload the active learning scores to the api
+        if al_scores is not None:
+            raise NotImplementedError  # TODO: fill out in later branches
+
+        # trigger the sampling
+        payload = self._create_sampling_create_request(sampler_config,preselected_tag_id,query_tag_id)
+        payload.row_count = 15  # TODO: remove after solving https://github.com/lightly-ai/lightly-core/issues/150
```
philippmwirth, line 42: So now we can only use it on this specific test case? Can we not determine the row count by the initial-tag size? Would make it much much easier to test

lightly/api/bitmask.py
```diff
@@ -0,0 +1,197 @@
...
+        # if the number is odd, there is a nonzero bit at offset
+        if x % 2 > 0:
+            nonzero_bit_indices.append(offset)
+        # increment the offset and divide the number x by two (rounding down)
+        offset += 1
+        x = x // 2
+    return nonzero_bit_indices
+
+
+def _invert(x: int) -> int:
```
philippmwirth, line 48: Let's create an issue to implement this, should be a quick one.

lightly/active_learning/config/sampler_config.py
```diff
@@ -0,0 +1,32 @@
...
+            The method to use for sampling, e.g. RANDOM.
+        batch_size:
+            The maximum number of samples to be chosen by the sampler.
+        min_distance:
+            The minimum distance of samples in the chosen set, one of stopping conditions.
+        name:
+            The name of this sampling, defaults to a name consisting of all other attributes and the datetime
+
+    """
+    def __init__(self, method: SamplingMethod = SamplingMethod.RANDOM, batch_size: int = 32, min_distance: float = -1,
```
philippmwirth, line 22: Maybe switch default to coreset?

lightly/active_learning/agents/agent.py
```diff
@@ -0,0 +1,79 @@
...
+        api_workflow_client:
+            The client to connect to the api.
+        preselected_tag_id:
+            The id of the tag containing the already labeled samples, default: None == no labeled samples yet.
+        query_tag_id:
+            The id of the tag defining where to sample from, default: None resolves to initial_tag
+
+    """
+
+    def __init__(self, api_workflow_client: ApiWorkflowClient, query_tag_id: str = None, preselected_tag_id: str = None):
```
philippmwirth, line 22: Are we using the `tag_id` until tag-names are unique? Or what's the reasoning here?

philippmwirth (commented):
Looks good to me! Maybe have @IgorSusmelj look at the non-generated files and then we can merge :)

lightly/active_learning/agents/agent.py
MalteEbner: I solved that, now the tag_names are used

lightly/api/bitmask.py
MalteEbner: See https://github.com/lightly-ai/lightly/pull/131/commits/48dfca68d534e09dc09f0c4f01e0b9b4cbd32bac

lightly/active_learning/agents/agent.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def __init__(self, api_workflow_client: ApiWorkflowClient, query_tag_name: str = None, preselected_tag_name: str = None):
+
+        self.api_workflow_client = api_workflow_client
+        if query_tag_name is not None or preselected_tag_name is not None:
+            tag_name_id_dict = dict([tag.name, tag.id] for tag in self.api_workflow_client._get_all_tags())
+            if query_tag_name is not None:
+                self.query_tag_id = tag_name_id_dict[query_tag_name]
+            if preselected_tag_name is not None:
+                self.preselected_tag_id_tag_id = tag_name_id_dict[preselected_tag_name]
```
philippmwirth, line 30: typo: `self.preselected_tag_id_tag_id`

lightly/openapi_generated/swagger_client/models/jobs_data.py
```diff
@@ -0,0 +1,258 @@
...
+
+    Lightly.ai enables you to do self-supervised learning in an easy and intuitive way. The lightly.ai OpenAPI spec defines how one can interact with our REST API to unleash the full potential of lightly.ai  # noqa: E501
+
+    OpenAPI spec version: 1.0.0
+    Contact: support@lightly.ai
+    Generated by: https://github.com/swagger-api/swagger-codegen.git
+"""
+
+
+import pprint
```
IgorSusmelj, line 14: is this a new dependency?

lightly/openapi_generated/swagger_client/models/jobs_data.py
```diff
@@ -0,0 +1,258 @@
...
+import re  # noqa: F401
+
+import six
```
IgorSusmelj, line 17: and this one too?

tests/api_workflow/test_UNMOCKED_complete.py
```diff
@@ -0,0 +1,69 @@
...
+from lightly.api.bitmask import BitMask
+from lightly.data import LightlyDataset
+from lightly.api.upload import upload_dataset
+
+from lightly.utils import save_embeddings
+
+from lightly.active_learning.config.sampler_config import SamplerConfig
+
+
+def t_est_unmocked_complete_workflow(path_to_dataset: str, token: str, dataset_id: str):
```
IgorSusmelj, line 18: type in method name `t_est` --> `test`

IgorSusmelj (commented):
Looks good to me. Can you also let us know about any changes in the requirements.txt? I wasn't aware that we have now more dependencies. We should have a look at this.

philippmwirth (changes requested):
Found one typo but other than that it looks good to me, let me know when you fixed it then I  will run it on my side

lightly/openapi_generated/swagger_client/models/jobs_data.py
MalteEbner: print is part of the standard library, see https://docs.python.org/3/library/pprint.html

tests/api_workflow/test_UNMOCKED_complete.py
MalteEbner: This is not a typo, but I did that to prevent unittest from finding and running that function in CI.

lightly/openapi_generated/swagger_client/models/jobs_data.py
MalteEbner: Yes, the additional dependencies from the generated files are already in the branch develop_active_learning, see https://github.com/lightly-ai/lightly/blob/90741ff6e723633c64eeba069c31dc54229c55ee/requirements/base.txt#L9. All the dependencies are quite small.

philippmwirth (approved):

PR #136, opened 2021-02-18
168 active learning pip package endpoints me

MalteEbner:
closes: https://github.com/lightly-ai/lightly-core/issues/168, https://github.com/lightly-ai/lightly/pull/136
- added generate files for the endpoints
- added a scorer for classification with tests
- added support for sampling with active learning scores

HOW TO TEST:
- run the script `test_al_step_3.py`


lightly/active_learning/scorers/classification.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def entropy(probs: np.ndarray, axis: int = 1):
+    zeros = np.zeros_like(probs)
+    log_probs = np.log2(probs, out=zeros, where=probs > 0)
+    entropies = -1 * np.sum(probs * log_probs, axis=axis)
+    return entropies
+
+
+class ScorerClassification(Scorer):
```
IgorSusmelj, line 15: Some docstrings here would be helpful. Is this a scorer for "image classification"? Will there be another one for object detection?

lightly/active_learning/scorers/classification.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, model_output: np.ndarray):
```
IgorSusmelj, line 16: Please describe the expected shape from `model_output` in the docstring. Are we talking about something like this?
`NxC` meaning we have N samples and C classes. 
Do we care about the range of the values? Should they be in the range of [0,1] or can they be logits?

IgorSusmelj (changes requested):
I would like documentation on some of the functions where it's not clear whether there are any expectations regarding input/output.

lightly/active_learning/scorers/classification.py
```diff
@@ -0,0 +1,31 @@
+from typing import *
+
+import numpy as np
+
+from lightly.active_learning.scorers.scorer import Scorer
+
+
+def entropy(probs: np.ndarray, axis: int = 1):
```
philippmwirth, line 8: Docstrings, always ;)

philippmwirth: Also, make it private by prefixing with _

lightly/active_learning/scorers/classification.py
philippmwirth: In the long run we should probably support [0, 1] and logits but let's choose one to start with

lightly/api/api_workflow_sampling.py
```diff
@@ -37,7 +40,12 @@ def sampling(self, sampler_config: SamplerConfig, al_scores: Dict[str, List[int]
 
         # upload the active learning scores to the api
         if al_scores is not None:
-            raise NotImplementedError  # TODO: fill out in later branches
+            if preselected_tag_id is None:
+                raise ValueError
+            for score_type, score_values in al_scores.items():
```
philippmwirth, line 45: I'd suggest to make a check in the pip package whether the number of scores coincides with the size of the tag they're being uploaded to. @MalteEbner @IgorSusmelj thoughts on that?

tests/UNMOCKED_end2end_tests/test_al_step3.py
```diff
@@ -0,0 +1,78 @@
...
+from lightly.openapi_generated.swagger_client import SamplingMethod
+
+from lightly.utils import save_embeddings
+
+from lightly.active_learning.config.sampler_config import SamplerConfig
+
+
+def t_est_unmocked_complete_workflow(path_to_dataset: str, token: str, dataset_id: str):
+    # define the api_client and api_workflow
+    api_workflow_client = ApiWorkflowClient(host="https://api-dev.lightly.ai", token=token, dataset_id=dataset_id)
```
philippmwirth, line 22: Can you switch to getting the host from the env variable? Also in the other unmocked script because I don't feel so comfortable exposing our developer api to everyone.

tests/active_learning/test_ScorerClassification.py
```diff
@@ -0,0 +1,24 @@
...
+        predictions_normalized = predictions / np.sum(predictions, axis=1)[:, np.newaxis]
+        model_output = predictions_normalized
+        scorer = ScorerClassification(model_output)
+        scores = scorer._calculate_scores()
+        scores_prediction_entropy = scores["prediction-entropy"]
+        scores_prediction_margin = scores["prediction-margin"]
+
+        assert scores_prediction_entropy.shape == (n_samples, )
+        assert scores_prediction_margin.shape == (n_samples, )
+        assert all(scores_prediction_entropy > 0)
```
philippmwirth, line 22: This assertion will fail for the following example input:
```
[
   [1.0, 0.0, 0.0, 0.0, 0.0]
]
```
because the logarithm retains the original value for all values <= 0.

philippmwirth (changes requested):
Please add docstrings to all the functions. If you think they are self-explanatory you can simply add a single sentence like so:
```Python
def my_fun():
    """This is my function.
    
    """
```

Otherwise it's looking good! Great work @MalteEbner!

tests/UNMOCKED_end2end_tests/test_al_step3.py
MalteEbner: I would not call it "exposing" the api, as it will be accessible from everywhere even if we do not give the easily-to-guess URL in the pip package. I someone really wants to get to it, trying out "api-dev" is not really harder than searching our pip-package for some URLs. The URL is protected by a token anyways. In my eyes removing the url is more like preventing burglary on your house by removing your address from the telephone book.
If you really want me to, I could change that, but I do not see it as a security measure.

philippmwirth: I don't see it as a security threat, more as a bad practice but focus on the important stuff first

lightly/api/api_workflow_sampling.py
MalteEbner: I did that in the ActiveLearningAgent.query()

lightly/active_learning/scorers/classification.py
MalteEbner: Yes, there will be another one for object detection in the future, but first we need to define the formats of the prediction of an  object detection model  which will be supported.

lightly/active_learning/scorers/classification.py
MalteEbner: @philippmwirth I do not know how to calculate the prediction entropy and prediction margin with logits, do you know any literature on that?

tests/UNMOCKED_end2end_tests/test_al_step3.py
MalteEbner: I would keep it there as it means one less error which could be made while testing.

lightly/active_learning/scorers/classification.py
philippmwirth: Check out the documentation on the cross entropy loss [here](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html). But again, it's ok to add this later as long as we specify the input format clearly.

lightly/api/api_workflow_sampling.py
philippmwirth: Didn't see that, thanks for the hint

lightly/active_learning/scorers/classification.py
MalteEbner: I created an issue for it here: https://github.com/lightly-ai/lightly/issues/137

lightly/active_learning/scorers/classification.py
MalteEbner: So the cross entropy loss out of logits is basically computed by first applying a softmax, as it computes the `exp(x_i)/sum(exp(x))`. The softmax gives the probability distribution as we want it. Thus our users would need to run the logits through a softmax first before setting the model_output in the scorer.

philippmwirth: Yes, let's do it this way for the moment. The advantage from calculating the entropy from logits is improved numerical stability

philippmwirth (approved):

IgorSusmelj (approved):

PR #152, opened 2021-02-24
138 create active learning run with a real dataset and kNN classifier

### LINKED ISSUE #138: [Active Learning] Create minimal active learning run with a real dataset and kNN classifier
Create a script to run an active learning episode. The input is not a dataset, but an embedding.csv with the soft labels being the true labels. As an example, use the clothing dataset: https://github.com/alexeygrigorev/clothing-dataset-small and embed its train and test set using the CLI.

The episode itself consists of following steps:

1. Upload either the true images, thumbnails of it or fake data.
2. Upload the `embedding.csv`
3. Perform a sampling and get the chosen filenames back.
4. Subset the dataset to the labeled_set provided by the filenames
5. use a kNN. e.g. from https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html, fit it on the labeled_set (5.1), evaluate on a separate test set (5.2), and predict_proba on the unlabeled_set (5.3) and 
6. save the predictions on the unlabeled_set in the al_scorer
Continue the loop starting from step 3.

MalteEbner:
closes #138

### DIFF
--- a/lightly/active_learning/agents/agent.py
+++ b/lightly/active_learning/agents/agent.py
@@ -86,6 +86,13 @@ class ActiveLearningAgent:
             the filenames of the samples in the new labeled_set
 
         """
+        # check input
+        if sampler_config.n_samples < len(self.labeled_set):
+            print("ERROR: The number of samples which should be sampled according to the config"
+                  " (including the current labeled set)"
+                  "is smaller than the number of samples in the current labeled set.")
+            return self.labeled_set
+
         # calculate scores
         if al_scorer is not None:
             no_unlabeled_samples = len(self.unlabeled_set)
--- a/lightly/active_learning/agents/agent.py
+++ b/lightly/active_learning/agents/agent.py
@@ -89,7 +89,7 @@ class ActiveLearningAgent:
         """
         # check input
         if sampler_config.n_samples < len(self.labeled_set):
-            warnings.warng("ActiveLearningAgent.query: The number of samples which should be sampled "
+            warnings.warn("ActiveLearningAgent.query: The number of samples which should be sampled "
                            "including the current labeled set "
                            "(sampler_config.n_samples) "
                             "is smaller than the number of samples in the current labeled set.")
--- a/lightly/active_learning/agents/agent.py
+++ b/lightly/active_learning/agents/agent.py
@@ -1,4 +1,5 @@
 from typing import *
+import warnings
 
 from lightly.active_learning.config.sampler_config import SamplerConfig
 from lightly.active_learning.scorers.scorer import Scorer
@@ -88,9 +89,10 @@ class ActiveLearningAgent:
         """
         # check input
         if sampler_config.n_samples < len(self.labeled_set):
-            print("ERROR: The number of samples which should be sampled according to the config"
-                  " (including the current labeled set)"
-                  "is smaller than the number of samples in the current labeled set.")
+            warnings.warng("ActiveLearningAgent.query: The number of samples which should be sampled "
+                           "including the current labeled set "
+                           "(sampler_config.n_samples) "
+                            "is smaller than the number of samples in the current labeled set.")
             return self.labeled_set
 
         # calculate scores
--- a/lightly/active_learning/config/sampler_config.py
+++ b/lightly/active_learning/config/sampler_config.py
@@ -11,22 +11,22 @@ class SamplerConfig:
     Attributes:
         method:
             The method to use for sampling, e.g. RANDOM.
-        batch_size:
-            The maximum number of samples to be chosen by the sampler.
+        n_samples:
+            The maximum number of samples to be chosen by the sampler including the samples in the preselected tag.
         min_distance:
             The minimum distance of samples in the chosen set, one of stopping conditions.
         name:
             The name of this sampling, defaults to a name consisting of all other attributes and the datetime
 
     """
-    def __init__(self, method: SamplingMethod = SamplingMethod.CORESET, batch_size: int = 32, min_distance: float = -1,
+    def __init__(self, method: SamplingMethod = SamplingMethod.CORESET, n_samples: int = 32, min_distance: float = -1,
                  name: str = None):
 
         self.method = method
-        self.batch_size = batch_size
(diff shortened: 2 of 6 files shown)


tests/UNMOCKED_end2end_tests/real_active_learning_run.py
```diff
@@ -0,0 +1,121 @@
...
+        if iteration != len(ratios)-1:  # skip the following for the last iteration
+            # 5.3 predict on the unlabeled set
+            predictions = classifier.predict_proba(training_set.get_features(agent.unlabeled_set))
+
+            # 6. Save the predictions in a scorer
+            al_scorer = ScorerClassification(model_output=predictions)
+
+
+if __name__ == "__main__":
+    path_to_dataset = "/Users/malteebnerlightly/Documents/datasets/clothing-dataset-small-master/train"
```
IgorSusmelj, line 111: pack this into an ENV variable or switch to argparse for this

tests/UNMOCKED_end2end_tests/real_active_learning_run.py
```diff
@@ -0,0 +1,121 @@
...
+    token = os.getenv("TOKEN")
+    dataset_id = "603606ed2176760032373065"
```
IgorSusmelj, line 113: pack this into an ENV variable or switch to argparse for this

tests/UNMOCKED_end2end_tests/real_active_learning_run.py
```diff
@@ -0,0 +1,121 @@
...
+    path_to_train_embeddings_csv = "/Users/malteebnerlightly/Documents/datasets/clothing-dataset-small-master/train/lightly_outputs/2021-02-23/23-38-25/embeddings.csv"
```
IgorSusmelj, line 114: here as well

tests/UNMOCKED_end2end_tests/real_active_learning_run.py
```diff
@@ -0,0 +1,121 @@
...
+    path_to_test_embeddings_csv = "/Users/malteebnerlightly/Documents/datasets/clothing-dataset-small-master/test/lightly_outputs/2021-02-23/23-41-09/embeddings.csv"
+    for i in range(1):
+        print(f"ITERATION {i}:")
+        real_active_learning_run(path_to_dataset, token, dataset_id,
+                                 path_to_train_embeddings_csv=path_to_train_embeddings_csv,
+                                 path_to_test_embeddings_csv=path_to_test_embeddings_csv)
+        print("")
```
IgorSusmelj, line 121: Since the whole script is a "test" you should also mention what the expected output looks like. Print "success" if it runs through or compare values or something that tells the user "it worked, test passed"

tests/UNMOCKED_end2end_tests/real_active_learning_run.py
```diff
@@ -0,0 +1,121 @@
...
+
+if __name__ == "__main__":
+    path_to_dataset = "/Users/malteebnerlightly/Documents/datasets/clothing-dataset-small-master/train"
+    token = os.getenv("TOKEN")
+    dataset_id = "603606ed2176760032373065"
+    path_to_train_embeddings_csv = "/Users/malteebnerlightly/Documents/datasets/clothing-dataset-small-master/train/lightly_outputs/2021-02-23/23-38-25/embeddings.csv"
+    path_to_test_embeddings_csv = "/Users/malteebnerlightly/Documents/datasets/clothing-dataset-small-master/test/lightly_outputs/2021-02-23/23-41-09/embeddings.csv"
+    for i in range(1):
+        print(f"ITERATION {i}:")
+        real_active_learning_run(path_to_dataset, token, dataset_id,
```
IgorSusmelj, line 118: Food for thought:
How about testing both options (CORAL and COREST) after each other? So we have 2 end2end tests in one file?

lightly/active_learning/config/sampler_config.py
```diff
@@ -11,22 +11,22 @@ class SamplerConfig:
     Attributes:
         method:
             The method to use for sampling, e.g. RANDOM.
-        batch_size:
-            The maximum number of samples to be chosen by the sampler.
+        n_samples:
```
philippmwirth, line 14: Why did you switch from `batch_size` to `n_samples`?

lightly/active_learning/agents/agent.py
```diff
@@ -86,6 +86,13 @@ def query(self, sampler_config: SamplerConfig, al_scorer: Scorer = None) -> List
             the filenames of the samples in the new labeled_set
 
         """
+        # check input
+        if sampler_config.n_samples < len(self.labeled_set):
+            print("ERROR: The number of samples which should be sampled according to the config"
```
philippmwirth, line 91: prints "error" but continues? I'd say use warnings for this

lightly/active_learning/config/sampler_config.py
MalteEbner: Before doing that, I had the following wokflow:
- I want to have n_samples and M of them are in the preselected set.
- First I have to calculate M and substract it to calculate batch_size = n_samples-M
- Inside the sampling itself, the M has to be added again, thus the preselected tag is loaded from the API to calculate M and then again n_samples = batch_size + M is set.

Instead of substracting M first, only to add it afterwards and peforming API request for each of this, it is much easier to have the sampler_config to store the n_samples instead of the batch_size.

philippmwirth: But is it not more user-friendly if we use the batch size?

MalteEbner: A user could say:
- I am planning my AL run to have 5%, then 10%, then 20% of the data labeled -> n_samples is more user friendly
- I found that my model is not good enough, I want to sample batch_size additional samples -> batch_size is more user friendly.

As the ActiveLearningAgent.query() returns the full labeled set with len(labeled_set)==n_samples, having the config with n_samples fits better to the overall workflow.

philippmwirth (approved):

PR #157, opened 2021-02-25
Active learning tutorial

### LINKED ISSUE #146: [Active learning] Tutorial
# Tutorial for Active Learning

## Description
Add a simple tutorial on active learning to the [docs](https://docs.lightly.ai/). 

## Tasks

- [x] Add a tutorial about active learning for classification to the docs

This tuturial includes the following steps:
1. Define the prerequisites: a dataset as images on the disk and the corresponding labels. The validation set also has to be defined (or split in a later step). 
2. Create embeddings for the dataset. Decision: Using Code - another 100 lines? Using CLI?
3. Create a new dataset in the Webapp and copy the token and dataset_id.
4. Upload the `embedding.csv`
5. Define a classifier on your dataset with training, predict_probabilities and evaluation methods. It must support defining the labeled set as a list of filenames or an external function performing this split must be created.
6. (Optional, only needed if no validation set is defined): Split the dataset using the random sampler into two parts with a predefined ratio. One is used as the training set/query tag, the other one as the validation set. 
7. Perform an initial sampling (e.g. with coreset or random) on the training set and get the filenames of the labeled set back. 
8. Train the classifier on the labeled set.
9. Evaluate the classifier on the validation set and store the result somewhere.
10. Predict the probabilities of the classes with the classifer on the unlabeled set. 
11. Save the predictions on the unlabeled_set in the al_scorer
12. Peform a CORAL sampling with the al_scorer (internally uploads the scores) and get the filenames of the labeled set back.
13. Continue the loop starting from step 8 for a few iterations.
14. Plot the performance of the agent (saved in step 9) over multiple active learning steps.

## Acceptance Criteria

- [x] Tutorial works
- [x] Tutorial introduces all concepts which are new in lightly AL.

MalteEbner:
close #146

### DIFF
--- a/docs/source/tutorials_source/package/tutorial_active_learning.py
+++ b/docs/source/tutorials_source/package/tutorial_active_learning.py
@@ -34,15 +34,23 @@ Prerequisites:
 
 # %%
 # Creation of the dataset on the lightly platform with embeddings
-# To perform samplings, we need to
-# A. Train a model on the dataset, e.g. using the CLI https://docs.lightly.ai/getting_started/command_line_tool.html#train-a-model-using-the-cli
-# B. Create embeddings for the dataset, e.g. using the CLI https://docs.lightly.ai/getting_started/command_line_tool.html#create-embeddings-using-the-cli
+# To perform samplings, we need to perform several steps, ideally with the CLI.
+# More documentation on each step is found at https://docs.lightly.ai/getting_started/command_line_tool.html#
+# A. Train a model on the dataset, e.g. with
+#    lightly-train input_dir="path/to/clothing-dataset-small"
+# B. Create embeddings for the dataset, e.g. with
+#    lightly-embed input_dir="path/to/clothing-dataset-small/train" checkpoint=mycheckpoint.ckpt
 #    Save the path to the embeddings.csv, you will need it later.
 #    for uploading the embeddings and for defining the dataset for the classifier
 # C. Create a new dataset on the lightly platform as described in https://docs.lightly.ai/getting_started/platform.html#.
 #    Save the token and dataset id, you will need them later to upload the images and embeddings
 #    and to run the active learning samplers.
-# D. Upload the images to the platform, e.g. using the CLI https://docs.lightly.ai/getting_started/command_line_tool.html#upload-data-using-the-cli
+# D. Upload the images to the platform, e.g. with
+#    lightly-upload input_dir="path/to/clothing-dataset-small/train" \
+#    embedding="path/to/clothing-dataset-small/train/.../embeddings.csv" \
+#    token="yourToken" dataset_id="yourDatasetId"
+
+
 
 
 # %%
@@ -64,10 +72,9 @@ from lightly.openapi_generated.swagger_client import SamplingMethod
 
 # %%
 # Definition of parameters
-path_to_embeddings_csv = "path/to/embeddings.csv"
-host = "https://api.lightly.ai"  # the URL of the web platform
-YOUR_TOKEN = "ABCDEF"  # your token of the web platform
-YOUR_DATASET_ID = "ASDFASDF"  # the id of your dataset on the web platform
+path_to_embeddings_csv = "path/to/clothing-dataset-small/train/.../embeddings.csv"
+YOUR_TOKEN = "yourToken"  # your token of the web platform
+YOUR_DATASET_ID = "yourDatasetId"  # the id of your dataset on the web platform
 
 
 # %%
@@ -113,7 +120,7 @@ class CSVEmbeddingDataset:
 
 # %%
 # Upload the embeddings to the lightly web platform
-api_workflow_client = ApiWorkflowClient(host=host, token=YOUR_TOKEN, dataset_id=YOUR_DATASET_ID)
+api_workflow_client = ApiWorkflowClient(token=YOUR_TOKEN, dataset_id=YOUR_DATASET_ID)
 api_workflow_client.upload_embeddings(name="embedding-1", path_to_embeddings_csv=path_to_embeddings_csv)
 
 # %%
--- a/docs/source/tutorials_source/package/tutorial_active_learning.py
+++ b/docs/source/tutorials_source/package/tutorial_active_learning.py
@@ -112,15 +112,15 @@ class CSVEmbeddingDataset:
 
 
 # %%
-# Upload the embeddings and dataset
+# Upload the embeddings to the lightly web platform
 api_workflow_client = ApiWorkflowClient(host=host, token=YOUR_TOKEN, dataset_id=YOUR_DATASET_ID)
 api_workflow_client.upload_embeddings(name="embedding-1", path_to_embeddings_csv=path_to_embeddings_csv)
 
 # %%
-# Definition of dataset for classifer,the active learning agent and the classifier
+# Definition of the dataset for the classifer,the classifier and the active learning agent
 dataset = CSVEmbeddingDataset(path_to_embeddings_csv=path_to_embeddings_csv)
-agent = ActiveLearningAgent(api_workflow_client=api_workflow_client)
 classifier = KNeighborsClassifier(n_neighbors=20, weights='distance')
+agent = ActiveLearningAgent(api_workflow_client=api_workflow_client)
 
 # %%
 # 1. Choose an initial subset of your dataset.
--- a/docs/source/tutorials_source/package/tutorial_active_learning.py
+++ b/docs/source/tutorials_source/package/tutorial_active_learning.py
@@ -1,0 +1,177 @@
+"""
+
+Tutorial 6: Active Learning
+==============================================
+
+In this tutorial, we will run an active learning loop using both the lightly package and the platform.
+An active learning loop is a sequence of multiple samplings each choosing only a subset
+of all samples in the dataset. This workflow has the following structure:
+
+1. Choose an initial subset of your dataset e.g. using one of our samplers like the coreset sampler.
+    Split your dataset accordingly into a labeled set and unlabeled set.
+Next, a loop with following steps occurs:
+2. Train a classifier on the labeled set.
+3. Use the classifier to predict on the unlabeled set.
+4. Calculate active learning scores from the prediction.
+5. Use an active learning agent to choose the next samples to be labeled based on the active learning scores.
+6. Update the labeled set to include the newly chosen samples and remove them from the unlabeled set.
+
+In this tutorial we use a k-nearest-neighbor classifier that predicts the class of a sample
+based on the class of the k nearest samples in the labeled set.
+We use the euclidean distance between a sample's embeddings as the distance metric.
+The advantage of such a classifier compared to CNNs is that it is very fast and easily implemented.
+
+To make the definition of the dataset for training the classifier easy,
+we recommend using a dataset where the samples are grouped into folder according to their class.
+We use the clothing-dataset-small. You can download it using
+`git clone https://github.com/alexeygrigorev/clothing-dataset-small.git`
+
+Prerequisites:
+- Make sure you are familiar with the command line tools described at https://docs.lightly.ai/getting_started/command_line_tool.html#
+
+"""
+
+
+# %%
+# Creation of the dataset on the lightly platform with embeddings
+# To perform samplings, we need to
+# A. Train a model on the dataset, e.g. using the CLI https://docs.lightly.ai/getting_started/command_line_tool.html#train-a-model-using-the-cli
+# B. Create embeddings for the dataset, e.g. using the CLI https://docs.lightly.ai/getting_started/command_line_tool.html#create-embeddings-using-the-cli
+#    Save the path to the embeddings.csv, you will need it later.
+#    for uploading the embeddings and for defining the dataset for the classifier
+# C. Create a new dataset on the lightly platform as described in https://docs.lightly.ai/getting_started/platform.html#.
+#    Save the token and dataset id, you will need them later to upload the images and embeddings
+#    and to run the active learning samplers.
+# D. Upload the images to the platform, e.g. using the CLI https://docs.lightly.ai/getting_started/command_line_tool.html#upload-data-using-the-cli
+
+
+# %%
+# Imports
+# -------
+#
+# Import the Python frameworks we need for this tutorial.
+
+import csv
+from typing import List
+import numpy as np
+from sklearn.neighbors import KNeighborsClassifier
+
+from lightly.active_learning.agents.agent import ActiveLearningAgent
+from lightly.active_learning.config.sampler_config import SamplerConfig
+from lightly.active_learning.scorers.classification import ScorerClassification
+from lightly.api.api_workflow_client import ApiWorkflowClient
+from lightly.openapi_generated.swagger_client import SamplingMethod
+
+# %%
+# Definition of parameters
+path_to_embeddings_csv = "path/to/embeddings.csv"
+host = "https://api.lightly.ai"  # the URL of the web platform
+token = "ABCDEF"  # the token of the web platform
+dataset_id = "ASDFASDF"  # the id of your dataset on the web platform
+n_samples_to_sample = [100, 200]  # We first want to 100 samples to be labeled and
+# then another 100 to have 200 samples in total
+methods_to_sample = [SamplingMethod.CORESET, SamplingMethod.CORAL]  # We perform the initial sampling with CORESET
+# and then use CORAL for the following sampling with active learning scores
+
+
+# %%
+# Definition of a dataset for the classifier based on the embeddings.csv
+# The kNN-classifier needs the embeddings as features for its classification. Thus we define a class to create
+# such a dataset out of the embeddings.csv. It also allows to choose only a subset of all samples
+# dependent on the filenames given.
+
+class CSVEmbeddingDataset:
+    def __init__(self, path_to_embeddings_csv: str):
+        with open(path_to_embeddings_csv, 'r') as f:
+            data = csv.reader(f)
+
+            rows = list(data)
+            header_row = rows[0]
+            rows_without_header = rows[1:]
+
+            index_filenames = header_row.index('filenames')
+            filenames = [row[index_filenames] for row in rows_without_header]
+
+            index_labels = header_row.index('labels')
+            labels = [row[index_labels] for row in rows_without_header]
+
+            embeddings = rows_without_header
+            indexes_to_delete = sorted([index_filenames, index_labels], reverse=True)
+            for embedding_row in embeddings:
+                for index_to_delete in indexes_to_delete:
+                    del embedding_row[index_to_delete]
+
+        self.dataset = dict([(filename, (np.array(embedding_row, dtype=float), int(label)))
+                             for filename, embedding_row, label in zip(filenames, embeddings, labels)])
+
+    def get_features(self, filenames: List[str]) -> np.ndarray:
+        features_array = np.array([self.dataset[filename][0] for filename in filenames])
+        return features_array
+
+    def get_labels(self, filenames: List[str]) -> np.ndarray:
+        labels = np.array([self.dataset[filename][1] for filename in filenames])
+        return labels
+
+
+# %%
+# Definition of dataset, active learning agent and classifier
+dataset = CSVEmbeddingDataset(path_to_embeddings_csv=path_to_embeddings_csv)
+agent = ActiveLearningAgent(ApiWorkflowClient(host=host, token=token, dataset_id=dataset_id))
+classifier = KNeighborsClassifier(n_neighbors=20, weights='distance')
+
+# %%
+# Upload of the embeddings
+agent.api_workflow_client.upload_embeddings(name="embedding-1", path_to_embeddings_csv=path_to_embeddings_csv)
+
+# %%
+# 1. Choose an initial subset of your dataset.
+print("Starting the initial sampling")
+sampler_config = SamplerConfig(name="initial-sampling", n_samples=n_samples_to_sample[0], method=methods_to_sample[0])
+agent.query(sampler_config=sampler_config)
+print(f"There are {len(agent.labeled_set)} samples in the labeled set.")
+
+# %%
+# 2. Train a classifier on the labeled set.
+labeled_set_features = dataset.get_features(agent.labeled_set)
+labeled_set_labels = dataset.get_labels(agent.labeled_set)
+classifier.fit(X=labeled_set_features, y=labeled_set_labels)
+
+# %%
+# 3. Use the classifier to predict on the unlabeled set.
+unlabeled_set_features = dataset.get_features(agent.unlabeled_set)
+predictions = classifier.predict_proba(X=unlabeled_set_features)
+
+# %%
+# 4. Calculate active learning scores from the prediction.
+active_learning_scorer = ScorerClassification(model_output=predictions)
+
+# %%
+# 5. Use an active learning agent to choose the next samples to be labeled based on the active learning scores.
+sampler_config = SamplerConfig(name="2nd-sampling", n_samples=n_samples_to_sample[1], method=methods_to_sample[1])
+agent.query(sampler_config=sampler_config, al_scorer=active_learning_scorer)
+print(f"There are {len(agent.labeled_set)} samples in the labeled set.")
+
+# %%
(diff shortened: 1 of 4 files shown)


docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,177 @@
...
+Prerequisites:
+- Make sure you are familiar with the command line tools described at https://docs.lightly.ai/getting_started/command_line_tool.html#
+
+"""
+
+
+# %%
+# Creation of the dataset on the lightly platform with embeddings
+# To perform samplings, we need to
+# A. Train a model on the dataset, e.g. using the CLI https://docs.lightly.ai/getting_started/command_line_tool.html#train-a-model-using-the-cli
```
IgorSusmelj, line 38: I would stick with the same enumeration pattern. Maybe use numbers everywhere?
Or is there a reason you switched to characters?

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,177 @@
...
+from lightly.active_learning.agents.agent import ActiveLearningAgent
+from lightly.active_learning.config.sampler_config import SamplerConfig
+from lightly.active_learning.scorers.classification import ScorerClassification
+from lightly.api.api_workflow_client import ApiWorkflowClient
+from lightly.openapi_generated.swagger_client import SamplingMethod
+
+# %%
+# Definition of parameters
+path_to_embeddings_csv = "path/to/embeddings.csv"
+host = "https://api.lightly.ai"  # the URL of the web platform
```
IgorSusmelj, line 68: Would remove from the tutorial. This should anyway be the default host

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,177 @@
...
+token = "ABCDEF"  # the token of the web platform
```
IgorSusmelj, line 69: Call this `YOUR_TOKEN`

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,177 @@
...
+dataset_id = "ASDFASDF"  # the id of your dataset on the web platform
```
IgorSusmelj, line 70: Call this `YOUR_DATASET_ID`

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,177 @@
...
+n_samples_to_sample = [100, 200]  # We first want to 100 samples to be labeled and
+# then another 100 to have 200 samples in total
+methods_to_sample = [SamplingMethod.CORESET, SamplingMethod.CORAL]  # We perform the initial sampling with CORESET
```
IgorSusmelj, line 73: would not create a list but specify the method used directly when using it

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,177 @@
...
+classifier = KNeighborsClassifier(n_neighbors=20, weights='distance')
+
+# %%
+# Upload of the embeddings
+agent.api_workflow_client.upload_embeddings(name="embedding-1", path_to_embeddings_csv=path_to_embeddings_csv)
+
+# %%
+# 1. Choose an initial subset of your dataset.
+print("Starting the initial sampling")
+sampler_config = SamplerConfig(name="initial-sampling", n_samples=n_samples_to_sample[0], method=methods_to_sample[0])
```
IgorSusmelj, line 129: use `SamplingMethod.CORESET` here directly

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,177 @@
...
+            index_labels = header_row.index('labels')
+            labels = [row[index_labels] for row in rows_without_header]
+
+            embeddings = rows_without_header
+            indexes_to_delete = sorted([index_filenames, index_labels], reverse=True)
+            for embedding_row in embeddings:
+                for index_to_delete in indexes_to_delete:
+                    del embedding_row[index_to_delete]
+
+        self.dataset = dict([(filename, (np.array(embedding_row, dtype=float), int(label)))
```
IgorSusmelj, line 104: could you add in comments below the rough structure of what we have now in `self.dataset`. This would help a lot for the understanding.

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,177 @@
...
+# Creation of the dataset on the lightly platform with embeddings
+# To perform samplings, we need to
+# A. Train a model on the dataset, e.g. using the CLI https://docs.lightly.ai/getting_started/command_line_tool.html#train-a-model-using-the-cli
+# B. Create embeddings for the dataset, e.g. using the CLI https://docs.lightly.ai/getting_started/command_line_tool.html#create-embeddings-using-the-cli
+#    Save the path to the embeddings.csv, you will need it later.
+#    for uploading the embeddings and for defining the dataset for the classifier
+# C. Create a new dataset on the lightly platform as described in https://docs.lightly.ai/getting_started/platform.html#.
+#    Save the token and dataset id, you will need them later to upload the images and embeddings
+#    and to run the active learning samplers.
+# D. Upload the images to the platform, e.g. using the CLI https://docs.lightly.ai/getting_started/command_line_tool.html#upload-data-using-the-cli
```
IgorSusmelj, line 45: Please show the CLI commands you used for A, B, D

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def get_labels(self, filenames: List[str]) -> np.ndarray:
+        labels = np.array([self.dataset[filename][1] for filename in filenames])
+        return labels
+
+
+# %%
+# Definition of dataset, active learning agent and classifier
+dataset = CSVEmbeddingDataset(path_to_embeddings_csv=path_to_embeddings_csv)
+agent = ActiveLearningAgent(ApiWorkflowClient(host=host, token=token, dataset_id=dataset_id))
```
IgorSusmelj, line 119: Just a general note. Why can't the `ActiveLearningAgent` create the `ApiWorkflowClient`? The `ApiWorkflowClient` si something the users should not have to deal with. He only should care about `token` and `dataset_id`.

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,177 @@
...
+
+# %%
+# Upload of the embeddings
+agent.api_workflow_client.upload_embeddings(name="embedding-1", path_to_embeddings_csv=path_to_embeddings_csv)
+
+# %%
+# 1. Choose an initial subset of your dataset.
+print("Starting the initial sampling")
+sampler_config = SamplerConfig(name="initial-sampling", n_samples=n_samples_to_sample[0], method=methods_to_sample[0])
+agent.query(sampler_config=sampler_config)
```
IgorSusmelj, line 130: Same quesiton here as above. What's the reason I have to create the sampler config separately? Why not just call 
`agent.query(name="initial-sampling", n_samples=n_samples_to_sample[0], method=methods_to_sample[0])`

IgorSusmelj (changes requested):
Only a few minor changes. Regarding the interface, we can have a chat or discuss in slack.

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def get_labels(self, filenames: List[str]) -> np.ndarray:
+        labels = np.array([self.dataset[filename][1] for filename in filenames])
+        return labels
+
+
+# %%
+# Upload the embeddings to the lightly web platform
+api_workflow_client = ApiWorkflowClient(token=YOUR_TOKEN, dataset_id=YOUR_DATASET_ID)
+api_workflow_client.upload_embeddings(name="embedding-1", path_to_embeddings_csv=path_to_embeddings_csv)
```
philippmwirth, line 124: This would still be replaced by the CLI now right? 

FIY because maybe you didn't know. You can embed, and upload in a single command like this:
```bash
lightly-magic input_dir=path/to/input/dir trainer.max_epochs=0 dataset_id=XYZ token=123
```

docs/source/tutorials_source/package/tutorial_active_learning.py
MalteEbner: I would keep the enumeration pattern, as the enumeration in the code (1-6) refers to the enumeration in the beginning (1-6) , but not the one in the CLI steps (A-C)

docs/source/tutorials_source/package/tutorial_active_learning.py
MalteEbner: As discussed, we keep it like this.

docs/source/tutorials_source/package/tutorial_active_learning.py
MalteEbner: As discussed, we keep it like this.

docs/source/tutorials_source/package/tutorial_active_learning.py
MalteEbner: No, it cannot, as the upload_embeddings sets the internal api_workflow_client.embedding_id.

philippmwirth: Hm ok that's not nice...

There's an endpoint at ` /v1/datasets/{datasetId}/embeddings` to get all embeddings by dataset id. Can we use this one instead and have the user pass the name of the embedding? 

@IgorSusmelj thoughts?

IgorSusmelj: I would make the following:
- user can also not provide an embedding_id --> we just take the latest one in the db

Cases, where I have multiple embeddings and want to take a specific one, are rare for now.

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,185 @@
...
+
+"""
+
+
+# %%
+# Creation of the dataset on the lightly platform with embeddings
+# To perform samplings, we need to perform several steps, ideally with the CLI.
+# More documentation on each step is found at https://docs.lightly.ai/getting_started/command_line_tool.html#
+# A. Train a model on the dataset, e.g. with
+#    lightly-train input_dir="path/to/clothing-dataset-small"
```
IgorSusmelj, line 40: use code blocks (rst syntax) for this. You can have a look at the other tutorials.

docs/source/tutorials_source/package/tutorial_active_learning.py
```diff
@@ -0,0 +1,185 @@
+"""
+
+Tutorial 6: Active Learning
```
IgorSusmelj, line 3: Rename to tutorial 2 and put into the platform folder

docs/source/tutorials_source/package/tutorial_active_learning.py
philippmwirth: So, just that I understand correctly, the user cannot even pass an embedding name? The program just picks one?

IgorSusmelj (changes requested):
Move the tutorial to the platform folder. I still need to run the tutorial locally on my machine. Will do this tonight or latest tomorrow.

docs/source/tutorials_source/package/tutorial_active_learning.py
MalteEbner: The user could use 
```python
api_workflow_client.upload_embeddings(name="existing-embedding-name", path_to_embedding_csv="") 
```
to set the `api_workflow_client.embedding_id` for an embedding specified by the name.
I would propose  to have a separate method for that, but in another issue.

MalteEbner:
@IgorSusmelj To test the script, I made following code changes:
image

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -0,0 +1,190 @@
...
+
+"""
+
+
+# %%
+# Creation of the dataset on the lightly platform with embeddings
+# To perform samplings, we need to perform several steps, ideally with the CLI.
+# More documentation on each step is found at https://docs.lightly.ai/getting_started/command_line_tool.html#
+# A. Train a model on the dataset, e.g. with
+#    ```lightly-train input_dir="path/to/clothing-dataset-small"```
```
IgorSusmelj, line 46: Add `trainer.max_epochs=5` to the lightly command

IgorSusmelj: We don't want our users to train for hours :)

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -0,0 +1,190 @@
...
+        return features_array
+
+    def get_labels(self, filenames: List[str]) -> np.ndarray:
+        labels = np.array([self.dataset[filename][1] for filename in filenames])
+        return labels
+
+
+# %%
+# Upload the embeddings to the lightly web platform
+api_workflow_client = ApiWorkflowClient(token=YOUR_TOKEN, dataset_id=YOUR_DATASET_ID)
```
IgorSusmelj, line 128: I get an error here because of the missing host. I used a quick fix and just added the host argument. But we wanted to hide it!?
```
  File "tutorial_active_learning.py", line 128, in <module>
    api_workflow_client = ApiWorkflowClient(token=YOUR_TOKEN, dataset_id=YOUR_DATASET_ID)
TypeError: __init__() missing 1 required positional argument: 'host'
```

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -0,0 +1,190 @@
...
+classifier.fit(X=labeled_set_features, y=labeled_set_labels)
+
+# evaluate on unlabeled set
+unlabeled_set_features = dataset.get_features(agent.unlabeled_set)
+unlabeled_set_labels = dataset.get_labels(agent.unlabeled_set)
+accuracy_on_unlabeled_set = classifier.score(X=unlabeled_set_features, y=unlabeled_set_labels)
+print(f"accuracy on unlabeled set: {accuracy_on_unlabeled_set}")
+
+# evaluate on test set
+dataset_test = CSVEmbeddingDataset("path/to/clothing-dataset-small/test/.../embeddings.csv")
```
IgorSusmelj, line 187: Replace with variable URL we set in beginning

IgorSusmelj (changes requested):
There are some smaller bugs. let's fix them.

IgorSusmelj:
A problem I encountered during the run of the tutorial. This error happened after running the same code twice on the same dataset:
```
Aborting upload, embedding with name='embedding-1' already exists.
Starting the initial sampling
  File "tutorial_active_learning.py", line 142, in <module>
    agent.query(sampler_config=sampler_config)
  File "/opt/conda/envs/lightly/lib/python3.7/site-packages/lightly/active_learning/agents/agent.py", line 115, in query
    query_tag_id=self.query_tag_id)
  File "/opt/conda/envs/lightly/lib/python3.7/site-packages/lightly/api/api_workflow_sampling.py", line 72, in sampling
    raise ValueError(f"Sampling job with job_id {job_id} failed with error {job_status_data.error}")
ValueError: Sampling job with job_id 15 failed with error [object Object]
```

philippmwirth:
> 15

@IgorSusmelj  you can check the job status in the admin-dev app. The problem is that the tag you want to create already exists (from your previous run). @MalteEbner the best solution is probably to give more descriptive feedback to the user.

docs/source/tutorials_source/platform/tutorial_active_learning.py
IgorSusmelj, line 46: Also make it clear that the users should use `input_dir="path/to/clothing-dataset-small/train"`

IgorSusmelj:
> A problem I encountered during the run of the tutorial. This error happened after running the same code twice on the same dataset:
> 
> ```
> Aborting upload, embedding with name='embedding-1' already exists.
> Starting the initial sampling
> Traceback (most recent call last):
>   File "tutorial_active_learning.py", line 142, in <module>
>     agent.query(sampler_config=sampler_config)
>   File "/opt/conda/envs/lightly/lib/python3.7/site-packages/lightly/active_learning/agents/agent.py", line 115, in query
>     query_tag_id=self.query_tag_id)
>   File "/opt/conda/envs/lightly/lib/python3.7/site-packages/lightly/api/api_workflow_sampling.py", line 72, in sampling
>     raise ValueError(f"Sampling job with job_id {job_id} failed with error {job_status_data.error}")
> ValueError: Sampling job with job_id 15 failed with error [object Object]
> ```

f strings in python only work with ' not with " to my knowledge.

So these lines need to be updated: `raise ValueError(f"Sampling job with job_id {job_id} failed with error {job_status_data.error}")`

docs/source/tutorials_source/platform/tutorial_active_learning.py
MalteEbner: Please use the quick fix now, as you want to set it to "https://api-dev.lightly.ai" anyways. I will restructure the host and its usage in all files in a later issue.

MalteEbner:
> A problem I encountered during the run of the tutorial. This error happened after running the same code twice on the same dataset:
> 
> ```
> Aborting upload, embedding with name='embedding-1' already exists.
> Starting the initial sampling
> Traceback (most recent call last):
>   File "tutorial_active_learning.py", line 142, in <module>
>     agent.query(sampler_config=sampler_config)
>   File "/opt/conda/envs/lightly/lib/python3.7/site-packages/lightly/active_learning/agents/agent.py", line 115, in query
>     query_tag_id=self.query_tag_id)
>   File "/opt/conda/envs/lightly/lib/python3.7/site-packages/lightly/api/api_workflow_sampling.py", line 72, in sampling
>     raise ValueError(f"Sampling job with job_id {job_id} failed with error {job_status_data.error}")
> ValueError: Sampling job with job_id 15 failed with error [object Object]
> ```

The reason is that the same name for the new tag is set in both runs, thus the tag already exists. 
```python
sampler_config = SamplerConfig(name="initial-sampling", n_samples=100, method=SamplingMethod.CORESET)
```
You could either
- set the new name to something else e.g. initial-sampling-2 and 2nd-sampling-2
- not set any name at all (then the name ist set automatically to something containing the current date and time)

A better error message for this should be done in another issue.

IgorSusmelj:
> > A problem I encountered during the run of the tutorial. This error happened after running the same code twice on the same dataset:
> > ```
> > Aborting upload, embedding with name='embedding-1' already exists.
> > Starting the initial sampling
> > Traceback (most recent call last):
> >   File "tutorial_active_learning.py", line 142, in <module>
> >     agent.query(sampler_config=sampler_config)
> >   File "/opt/conda/envs/lightly/lib/python3.7/site-packages/lightly/active_learning/agents/agent.py", line 115, in query
> >     query_tag_id=self.query_tag_id)
> >   File "/opt/conda/envs/lightly/lib/python3.7/site-packages/lightly/api/api_workflow_sampling.py", line 72, in sampling
> >     raise ValueError(f"Sampling job with job_id {job_id} failed with error {job_status_data.error}")
> > ValueError: Sampling job with job_id 15 failed with error [object Object]
> > ```
> 
> The error is not cause by the f-string, I only used the f-string for the error message of another error.
> 
> The reason for that error (see it in the admin-dev) is that the same name for the new tag is set in both runs, thus the tag already exists.
> 
> ```python
> sampler_config = SamplerConfig(name="initial-sampling", n_samples=100, method=SamplingMethod.CORESET)
> ```
> 
> You could either
> 
> * set the new name to something else e.g. initial-sampling-2 and 2nd-sampling-2
> * not set any name at all (then the name ist set automatically to something containing the current date and time)
> 
> A better error message for this should be done in another issue.

Yes, I know that the error is not caused by the f-string. I simply wanted to mention that the error message is no descriptive because the f-strings don't work if you use `"` instead of `'` to the best of my knowledge.

MalteEbner:
> > > A problem I encountered during the run of the tutorial. This error happened after running the same code twice on the same dataset:
> > > ```
> > > Aborting upload, embedding with name='embedding-1' already exists.
> > > Starting the initial sampling
> > > Traceback (most recent call last):
> > >   File "tutorial_active_learning.py", line 142, in <module>
> > >     agent.query(sampler_config=sampler_config)
> > >   File "/opt/conda/envs/lightly/lib/python3.7/site-packages/lightly/active_learning/agents/agent.py", line 115, in query
> > >     query_tag_id=self.query_tag_id)
> > >   File "/opt/conda/envs/lightly/lib/python3.7/site-packages/lightly/api/api_workflow_sampling.py", line 72, in sampling
> > >     raise ValueError(f"Sampling job with job_id {job_id} failed with error {job_status_data.error}")
> > > ValueError: Sampling job with job_id 15 failed with error [object Object]
> > > ```
> > 
> > 
> > The error is not cause by the f-string, I only used the f-string for the error message of another error.
> > The reason for that error (see it in the admin-dev) is that the same name for the new tag is set in both runs, thus the tag already exists.
> > ```python
> > sampler_config = SamplerConfig(name="initial-sampling", n_samples=100, method=SamplingMethod.CORESET)
> > ```
> > 
> > 
> > You could either
> > 
> > * set the new name to something else e.g. initial-sampling-2 and 2nd-sampling-2
> > * not set any name at all (then the name ist set automatically to something containing the current date and time)
> > 
> > A better error message for this should be done in another issue.
> 
> Yes, I know that the error is not caused by the f-string. I simply wanted to mention that the error message is no descriptive because the f-strings don't work if you use `"` instead of `'` to the best of my knowledge.

Your example shows that the f-string works with "", just look at the last line. Your point that we need a better error message is very valid, I created an issue for it: #169

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+# %%
+# Creation of the dataset on the lightly platform with embeddings
+# To perform samplings, we need to perform several steps, ideally with the CLI.
+# More documentation on each step is found at https://docs.lightly.ai/getting_started/command_line_tool.html#
+# A. Train a model on the dataset, e.g. with
+#    ```lightly-train input_dir="path/to/clothing-dataset-small trainer.max_epochs=5"```
+# B. Create embeddings for the dataset, e.g. with
+#    ```lightly-embed input_dir="path/to/clothing-dataset-small/train" checkpoint=mycheckpoint.ckpt```
```
IgorSusmelj, line 48: Further down we now also use an embedding for the test set. However, we never tell the user to create one. If we use a test set we should mention here that the user should also create embeddings for it.

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -0,0 +1,191 @@
...
+
+from lightly.active_learning.agents.agent import ActiveLearningAgent
+from lightly.active_learning.config.sampler_config import SamplerConfig
+from lightly.active_learning.scorers.classification import ScorerClassification
+from lightly.api.api_workflow_client import ApiWorkflowClient
+from lightly.openapi_generated.swagger_client import SamplingMethod
+
+# %%
+# Definition of parameters
+path_to_embeddings_csv = "path/to/clothing-dataset-small/train/.../embeddings.csv"
```
IgorSusmelj, line 80: call this train embeddings: `path_to_train_embeddings_csv`

philippmwirth (approved):

philippmwirth:
One more problem with the tutorial. It should run through when we build all the docs and for this, we need to set the token and dataset_id. Additionally, the path to the csv file should be fixed to: `/datasets/clothing-dataset/embeddings.csv`.

Any good idea how we solve the token & dataset_id problem? @MalteEbner @IgorSusmelj

philippmwirth:
Another problem I encountered when building the docs, the lightly commands are poorly formatted.
grafik

philippmwirth:
@IgorSusmelj can you please review again? We're ready to merge.

Here's an example of the tutorial:
[al_tutorial_example.pdf](https://github.com/lightly-ai/lightly/files/6090793/al_tutorial_example.pdf)

IgorSusmelj (approved):

PR #172, opened 2021-03-03
Add resize argument to workflow upload

philippmwirth:
Fixed the `size` argument which got forgotten.

### DIFF
--- a/lightly/api/api_workflow_upload_dataset.py
+++ b/lightly/api/api_workflow_upload_dataset.py
@@ -4,8 +4,6 @@ import warnings
 from concurrent.futures.thread import ThreadPoolExecutor
 from typing import TYPE_CHECKING, Union
 
-import torchvision
-
 from lightly.openapi_generated.swagger_client.models.sample_create_request import SampleCreateRequest
 
 from lightly.api.utils import check_filename, check_image, get_thumbnail_from_img, PIL_to_bytes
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_upload_dataset.py
```diff
@@ -22,7 +24,7 @@
 class _UploadDatasetMixin:
 
     def upload_dataset(self: ApiWorkflowClient, input: Union[str, LightlyDataset], max_workers: int = 8,
-                       mode: str = 'thumbnails', verbose: bool = True):
+                       size: Union[int, tuple] = -1,  mode: str = 'thumbnails', verbose: bool = True):
```
MalteEbner, line 27: rather Union[int, Tuple[int, int]]] and Tuple should be imported from typing.
Using types directly as typehints (i.e. tuple instead of tuple) is only supported in python 3.9.

lightly/api/api_workflow_upload_dataset.py
```diff
@@ -35,6 +37,12 @@ def upload_dataset(self: ApiWorkflowClient, input: Union[str, LightlyDataset], m
             max_requests:
                 Maximum number of requests a single worker can do before he has
                 to wait for the others.
+            size:
```
MalteEbner, line 40: Using the term 'output' here is very misleading in my eyes, as no image is outputted. Instead, the images  are resized before they are uploaded, thus the variable is also more like size_image_to_upload.
You could also distinguish the three possible cases better e.g. with a newline:
- if size is a negative int: images are uploaded without resizing
- if size is a positive int: ....
- if size is a tuple of (height, width) ...

MalteEbner (changes requested):
We could also think about putting the resize logic into the _upload_cli.py instead of here and then simply pass the dataset with the resize transform. I will leave this up to you.

MalteEbner (approved):

PR #174, opened 2021-03-03
145 active learning docs pw

philippmwirth:
I added a section about active learning to the docs and to the starting page of the docs. I also fixed some of the docstrings in the active learning code - will probably have to go over this a second time.

Here is a print out of the docs for those who can't or don't have the time to build them:
[al_docs_v0.pdf](https://github.com/lightly-ai/lightly/files/6076372/al_docs_v0.pdf)

### DIFF
--- a/docs/source/getting_started/active_learning.rst
+++ b/docs/source/getting_started/active_learning.rst
@@ -6,15 +6,6 @@ Lightly enables active learning with only a few lines of additional code. Learn
 here, how to get the most out of your data by maximizing the available information
 in your annotated dataset.
 
-
-Introcution to Pool-based Active Learning
-------------------------------------------
-TODO
-
-
-Active Learning with Lightly
------------------------------
-
 Preparations
 ~~~~~~~~~~~~~~~~~
 Before you read on, make sure you have read the section on the :ref:`lightly-platform`. 
@@ -88,7 +79,11 @@ Next, you will need to initialize the `ApiWorkflowClient` and the `ActiveLearnin
 
 .. note::
 
-   Note about query_tag. TODO.
+   It may not always be a good idea to sample from the full dataset. For example,
+   it could be that a large portion of the images is blurry. In that case, it's 
+   possible to create a tag in the web-app which only contains the sharp images
+   and tell the `ActiveLearningAgent` to only sample from this tag. To do so, set
+   the `query_tag_name` argument in the constructor.
 
 Let's configure the sampling request and request an initial selection next:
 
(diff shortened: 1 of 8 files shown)


docs/source/getting_started/active_learning.rst
```diff
@@ -0,0 +1,164 @@
...
+
+Now you have everything to get the next batch of images. One important thing to mention
+here is that the argument `n_samples` always refers to the total size of the labeled set.
+
+.. code-block:: Python
+
+   # we want a total of 200 images after the first iteration
+   # this time, we use the CORAL sampler and provide a scorer to the query
+   config = SamplerConfig(n_samples=200, method=SamplingMethod.CORAL, name='al-iteration-1')
+   al_iteration_1 = al_agent.query(sampler_config, scorer)
```
MalteEbner, line 158: I would call the variable `al_iteration_1` rather `labeled_set_al_iteration_1`. Perhaps you could also print the first 3 elements of it to make clear that it returns the filenames.

docs/source/getting_started/active_learning.rst
```diff
@@ -0,0 +1,164 @@
...
+trained model is used to get predictions on the unlabeled data and the sampler then
+decides based on these predictions. To get a list of all filenames in the unlabeled set,
+you can simply call
+
+.. code-block:: Python
+
+   # get all filenames in the unlabeled set
+   unlabeled_set = al_agent.unlabeled_set
+
+Use this list to get predictions on the unlabeled images.
```
MalteEbner, line 135: I would say here, that the user has to use the own classifier to run the predictions. E.g. add

```python
your_classifier.train(al_agent.labeled_set)
predictions = your_classifer.predict_proba(al_agent.unlabeled_set)
```

lightly/active_learning/scorers/classification.py
```diff
@@ -27,18 +27,14 @@ def _entropy(probs: np.ndarray, axis: int = 1) -> np.ndarray:
 
 
 class ScorerClassification(Scorer):
-    """A class to compute active learning scores out of the model_output (i.e. the predictions of a model) for a classification task.
```
MalteEbner, line 30: I do not get the point of all the changes in the docs of this class. 
'Helper class' - There is no widespread definition of a helper class thus it does not provide any informations
'i.e. predictions ... classification task' - I think it should be specified that this class is only targeted at classification tasks.
change of description of model_output - in my eyes the previous doc is much more informative and the relevant information is easier to find. I also think that we can include some pseudocode here, as it helps to understand how to generate the desired format.

MalteEbner (commented):
In general: I like the docs a lot, a very clear introduction into the topic and our concepts/classes how to use it.
The changes I proposed are a matter of taste to a certain degree, thus feel free to decide not to follow them.

docs/source/getting_started/active_learning.rst
philippmwirth: Maybe I'll just go with `labeled_set_iteration_1`. What do you think?

docs/source/getting_started/active_learning.rst
philippmwirth: Hm I'm not sure this is so informative. I mean you cannot simply call `your_classifier.predict_proba` on a set of filenames, no? I agree, however, that this part might need clarification. Do you have any other ideas?

lightly/active_learning/scorers/classification.py
philippmwirth: I agree to a certain degree but there were a few problems with your docstrings:
- Line 30 is way too long - that's why I changed it. The "helper" is not necessary.
- The model_output part was formatted in a way the documentation could not display.

I will rework the part about the model_output to include all the information it had before and add an example.

docs/source/getting_started/active_learning.rst
MalteEbner:
```python
dataset_labeled = your_dataset.subset_from_filenames(agent.labeled_set)
your_classifier.train(dataset_labeled)
dataset_unlabeled = your_dataset.subset_from_filenames(agent.unlabeled)
predictions = your_classifer.predict_proba(dataset_unlabeled)
```

We could also add such a subset_from_filenames method to the LightlyDataset, we need it latest for the AL-tutorial with images.

lightly/active_learning/scorers/classification.py
MalteEbner: Ok, so the previous docs only looked well-formatted in python, but badly-formatted in the documentation. Yes, this is a very good reason to change the docs.

docs/source/getting_started/active_learning.rst
philippmwirth: That's a good idea. I'd propose something like this:

```python
dataset = LightlyDataset.subset_from_filenames(input_dir, agent.unlabeled_set)
dataloader = torch.utils.data.DataLoader(dataset, shuffle=False, batch_size=1)
```

I will create an issue for this. Let's not add it to the docs before we have it implemented.

MalteEbner (approved):

PR #176, opened 2021-03-04
api workflow client.upload dataset() check initial tag me

### LINKED ISSUE #167: ApiWorkflowClient.upload_dataset() has no check if initial tag already exists
Such a check would allow to warn the user why the upload fails before the upload is started. It would tell the user that no samples can be uploaded to this dataset anymore, as the initial tag was already created.

MalteEbner:
close #167

### DIFF
--- a/lightly/api/api_workflow_upload_dataset.py
+++ b/lightly/api/api_workflow_upload_dataset.py
@@ -46,7 +46,7 @@ class _UploadDatasetMixin:
         """
         no_tags_on_server = len(self._get_all_tags())
         if no_tags_on_server > 0:
-            warnings.warn(f"Dataset has already been uploaded to the platform. Skipping upload.")
+            warnings.warn(f"Dataset with id {self.dataset_id} has already been completely uploaded to the platform. Skipping upload.")
             return
 
         # Check input variable 'input'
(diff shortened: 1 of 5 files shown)


lightly/api/api_workflow_upload_dataset.py
```diff
@@ -40,11 +40,14 @@ def upload_dataset(self: ApiWorkflowClient, input: Union[str, LightlyDataset], m
...
         Raises:
-            ValueError if dataset is too large.
+            ValueError if dataset is too large or input has the wrong type
             RuntimeError if the connection to the server failed.
-            RuntimeError if dataset already has an initial tag.
 
         """
+        no_tags_on_server = len(self._get_all_tags())
+        if no_tags_on_server > 0:
+            warnings.warn(f"Dataset has already been uploaded to the platform. Skipping upload.")
```
philippmwirth, line 49: I'd maybe reformulate the warning. What about f"Dataset with id {dataset_id} has already been completed. Skipping upload."

philippmwirth: Also, not sure if we want to throw a warning or just print the message and return.

MalteEbner: I changed the warning text, but would keep the warning, as messages are easily overlook. And it is indeed a warning, as the user did not follow the workflow.

philippmwirth (approved):

PR #180, opened 2021-03-04
169 check existing tag name pw

### LINKED ISSUE #169: ApiWorkflowClient.sampling(): check if tag_name already exists
The name in the sampler_config is used as the name of the new tag created after the sampling. There must not be a tag with that name already in the database. Otherwise an error occurs in the queueworker after a sampling is performed. 

This error is shown in the queueworker as
```
Error: {"status":400,"statusText":"Bad Request","url":"https://api-dev.lightly.ai/v1/datasets/603e008523de290032d7b9c4/tags?token=347f1dfbc3879a142d536d0b","size":0,"body":{"code":"TAG_NAME_EXISTS","error":"TAG_NAME_EXISTS"},"headers":{"content-type":["application/json; charset=utf-8"],"access-control-allow-origin":["*"],"etag":["W/\"41-ZHM/mN4+VRsqQFOimQGdZ/6FSbE\""],"vary":["Accept-Encoding"],"x-cloud-trace-context":["086bb37ac3f1bd91fe40e19e6457f11a"],"date":["Tue, 02 Mar 2021 11:50:57 GMT"],"server":["Google Frontend"],"content-length":["65"],"connection":["close"]}}
```
## Solution on Pip side
Instead, it could be checked at the start of a sampling whether this tag already exists. There are the following possible options to handle such a case:
1. Throw an Exception.
2. Warn the user and skip the sampling.
3. Warn the user, and generate a new sampling name by appending a counter, similar to this code for filenames
```python
def uniquify(path):
    filename, extension = os.path.splitext(path)
    counter = 1

    while os.path.exists(path):
        path = filename + " (" + str(counter) + ")" + extension
        counter += 1

    return path
```
Disadvantages:
- race conditions: What happens if a tag is created on the server side during the sampling?
- before every sampling, all tags (and thus all bitmasks) have to be downloaded.

## Solution on api side:
If the tag_name already exists, add a counter to it, similar to solution 3 on the pip side.
Disadvantage: still chance of race conditions, but much less likely.

philippmwirth:
# Check existing tag name

Closes #169 

I added a check for the tag name before the sampling request is started.

### DIFF
--- a/lightly/api/api_workflow_sampling.py
+++ b/lightly/api/api_workflow_sampling.py
@@ -36,7 +36,13 @@ class _SamplingMixin:
 
         Raises:
             ApiException, ValueError
+
         """
+
+        # make sure the tag name does not exist yet
+        tags = self._get_all_tags()
+        if sampler_config.name in [tag.name for tag in tags]:
+            raise RuntimeError('There already exists a tag with tag_name {sampler_config.name}.')
 
         # upload the active learning scores to the api
         if al_scores is not None:
--- a/lightly/api/api_workflow_sampling.py
+++ b/lightly/api/api_workflow_sampling.py
@@ -35,7 +35,7 @@ class _SamplingMixin:
             The newly created tag of the sampling.
 
         Raises:
-            ApiException, ValueError
+            ApiException, ValueError, RuntimeError
 
         """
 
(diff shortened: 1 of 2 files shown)


MalteEbner (approved):
You could add a unittest that this RuntimeError is thrown, but this is optional.

PR #187, opened 2021-03-05
186 Fix job status polling

### LINKED ISSUE #186: ApiWorkflowClientSampling doesn't check for job status WAITING
# ApiWorkflowClientSampling doesn't check for job status WAITING

Like the title says, the client only checks if the job is running but it can be waiting as well.
In particular, we need to check for job status:
- UNKNOWN
- RUNNING
- WAITING

before completion and 
- FINISHED
- FAILED

upon completion

philippmwirth:
# Fix job status polling

Closes #186 

There was a problem with how the PIP package polls the job status as a job can be waiting and the pip package didn't cover that case which would lead to an error when trying to read the `tag_id` from the `job_status_data`. I fixed this by retrying if the job is either `WAITING`, `RUNNING`, or `UNKNOWN`.

### DIFF
--- a/lightly/api/api_workflow_sampling.py
+++ b/lightly/api/api_workflow_sampling.py
@@ -75,8 +75,13 @@ class _SamplingMixin:
         job_status_data = None
 
         wait_time_till_next_poll = getattr(self, "wait_time_till_next_poll", 1)
-        while job_status_data is None or job_status_data.status == JobState.RUNNING:
+        while job_status_data is None \
+            or job_status_data.status == JobState.RUNNING \
+                or job_status_data.status == JobState.WAITING \
+                    or job_status_data.status == JobState.UNKNOWN:
+            # sleep before polling again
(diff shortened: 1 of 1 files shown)


japrescott (approved):
lgtm!
Suggestion: Maybe you want to also alert the user that there was an error on the worker side by checking for `JobState.FAILED` and outputting the `error`

philippmwirth:
> 
> 
> lgtm!
> Suggestion: Maybe you want to also alert the user that there was an error on the worker side by checking for `JobState.FAILED` and outputting the `error`

See the lines above :)

japrescott:
> > lgtm!
> > Suggestion: Maybe you want to also alert the user that there was an error on the worker side by checking for `JobState.FAILED` and outputting the `error`
> 
> See the lines above :)

sorry about this. I hate it when github thinks hiding 6lines is a good idea ;)

PR #191, opened 2021-03-05
185 v2 unmocked tests have a single test doing all

MalteEbner:
- creates the `_DatasetsMixin` to create and delete datasets and set the `_dataset_id`
- adds the script `test_api.py` creating a dataset, uploading, sampling with all samplers, w/wo preselected and query tags and finally deleting the dataset. As the script only needs the token and path/to/dataset to run, it could potentially be added to CI.


lightly/api/api_workflow_datasets.py
```diff
@@ -0,0 +1,96 @@
...
+from typing import List
+
+from lightly.openapi_generated.swagger_client.models.create_entity_response import CreateEntityResponse
+from lightly.openapi_generated.swagger_client.models.dataset_create_request import DatasetCreateRequest
+from lightly.openapi_generated.swagger_client.models.dataset_data import DatasetData
+
+
+class _DatasetsMixin:
+
+    def set_dataset_id(self: ApiWorkflowClient, dataset_name: str):
```
philippmwirth, line 16: A better name would be `set_dataset_id_by_name`

lightly/api/api_workflow_datasets.py
```diff
@@ -0,0 +1,96 @@
...
+        '''Sets the dataset id given the name of the dataset
+
+        Args:
+            dataset_name:
+                The name of the dataset for which the dataset_id
+                should be set as attribute
+
+        Returns:
```
philippmwirth, line 24: Leave the Returns away if it doesn't return anything

lightly/api/api_workflow_datasets.py
```diff
@@ -0,0 +1,96 @@
...
+                The name of the dataset to be created.
+
+        '''
+        try:
+            self.set_dataset_id(dataset_name)
+        except ValueError:
+            self._create_dataset_without_check_existing(dataset_name=dataset_name)
+
+    def _create_dataset_without_check_existing(self: ApiWorkflowClient, dataset_name: str):
+        ''' Creates a dataset on the webplatform
```
philippmwirth, line 54: No space after the '''

lightly/api/api_workflow_datasets.py
```diff
@@ -0,0 +1,96 @@
...
+        Args:
+            dataset_name:
+                The name of the dataset to be created.
+
+        '''
+        body = DatasetCreateRequest(name=dataset_name)
+        response: CreateEntityResponse = self.datasets_api.create_dataset(body=body)
+        self._dataset_id = response.id
+
+    def create_new_dataset_with_counter(self, dataset_basename: str):
```
philippmwirth, line 66: This looks like a function which is only used for testing. If this is the case I'd prefer to move it to the test file instead of having it here.

lightly/api/api_workflow_datasets.py
```diff
@@ -0,0 +1,96 @@
...
+
+        counter = 1
+        dataset_name = f"{dataset_basename}_{counter}"
+        while dataset_name in current_datasets_names:
+            counter += 1
+            dataset_name = f"{dataset_basename}_{counter}"
+        self._create_dataset_without_check_existing(dataset_name=dataset_name)
+
+    def delete_dataset(self: ApiWorkflowClient):
+        ''' Deletes a dataset on the web platform
```
philippmwirth, line 92: No space after '''

lightly/api/api_workflow_datasets.py
```diff
@@ -0,0 +1,96 @@
...
+
+        '''
+        self.datasets_api.delete_dataset_by_id(dataset_id=self.dataset_id)
```
philippmwirth, line 95: This function is dangerous! It takes the default dataset_id and just deletes it? I see big problems here...

Please remove this one and replace it by 'delete_dataset_by_id'. This way the user has at least to provide the id to delete a dataset.

philippmwirth (changes requested):
Just a few remarks

lightly/api/api_workflow_datasets.py
MalteEbner: I could imagine that some users want to just create a dataset on the webapp to do some samplings and do not care about the name of it at all and do not want to need to open the webapp. Thus I would keep it there.

philippmwirth: I disagree, this is something any user can write with two lines of code and I don't see why we should clutter our code with it. If you insist on keeping it here then I'd suggest to rename it to `create_new_dataset_with_unique_name` (or similar).

MalteEbner: I disagree that every user can create it with two lines of code, as the function has 10 lines of code. 
I could imagine that the function is needed for a colab tutorial where the user will want to rerun the tutorial without needing to create a new dataset again. 
I agree that the name you proposed is better. 

Another idea I had would be to put the multiple `create_dataset` functions into one and and add a parameter to them describing how to handle it if the dataset already exists:
- `"raise Error"`: If a dataset with that name already exists, raise an Error.
- `"take existing"`: Takes the existing dataset by setting the self._dataset_id
- `"make name unique"`: Adds a counter to the dataset name till it is unique, then creates it.

philippmwirth (approved):

PR #196, opened 2021-03-08
test.yml: use several python vesions

MalteEbner:


philippmwirth:
@MalteEbner the only concern I have is that the test runs will now take forever :/

IgorSusmelj:
I don't see the point in testing all versions. Let's just test the oldest one we support?
I guess this is 3.6 or 3.7?

philippmwirth:
> 
> 
> I don't see the point in testing all versions. Let's just test the oldest one we support?
> I guess this is 3.6 or 3.7?

3.6

philippmwirth (approved):

PR #213, opened 2021-03-12
Fix imports in __init__ and add tests

philippmwirth:
# Fix import errors

Closes #213 

I fixed the import errors, e.g. 
```
import lightly
lightly.data.LightlyDataset
```
should work now. I also added tests for all import that I found important.

Bumped the version number so we can quickly release this hotfix.

Added SimSiam to the list of models in the description.


tests/imports/nested_imports.py
```diff
@@ -0,0 +1,88 @@
...
+import torch
+
+import lightly
+
+
+class TestNestedImports(unittest.TestCase):
+
+    def test_nested_imports(self):
+        # active learning (commented out don't work)
+        #lightly.active_learning.agents.agent.ActiveLearningAgent
```
IgorSusmelj, line 11: Could you please create a follow-up issue for how to improve the handling of the imports? And mention in the issue that some of the direct imports don't work at the moment (the ones commented out here)

IgorSusmelj (commented):
I won't be able to test this right now on my machine. Still running some stuff in the background using another lightly version. I can have a look in the afternoon. @MalteEbner, do you have time to test this earlier?

IgorSusmelj (approved):

tests/imports/nested_imports.py
philippmwirth: Yes

IgorSusmelj (approved):

PR #222, opened 2021-03-14
Add benchmarking module for kNN evaluation of models

### LINKED ISSUE #77: Add callback for knn-validation
Add hooks for validation at the end of each epoch (only informative if there are labels). Let's make sure it can be switched off and that the user can pass a validation set of his choice.

IgorSusmelj:
### Changes

closes #77 

This PR adds a benchmarking module that can be used to quickly evaluate new SSL models on simple classification tasks using datasets such as CIFAR10, STL-10, ImageNet or others.

- Adds a `BenchmarkModule`. Inherit from it and pass it a PyTorch DataLoader to build up a feature bank to benchmark models.
- Add Docstrings and update documentation
- Update benchmarking code used for CIFAR10 benchmarks using the new module

I didn't add any unit tests since it's not a crucial part of the framework. However, I tested the implementation using our benchmark and did run it for a few epochs on a GPU as well as on CPU only. Everything seems to work fine.

### DIFF
--- a/docs/source/getting_started/benchmarks/cifar10_benchmark.py
+++ b/docs/source/getting_started/benchmarks/cifar10_benchmark.py
@@ -246,26 +246,26 @@ gpu_memory_usage = []
 
 # loop through configurations and train models
 for batch_size in batch_sizes:
-    for Model in models:
+    for BenchmarkModel in models:
         runs = []
         for seed in range(n_runs):
             pl.seed_everything(seed)
             dataloader_train_ssl, dataloader_train_kNN, dataloader_test = get_data_loaders(batch_size)
-            model = Model(dataloader_train_kNN, classes)
+            benchmark_model = BenchmarkModel(dataloader_train_kNN, classes)
             trainer = pl.Trainer(max_epochs=max_epochs, gpus=gpus,
                                 progress_bar_refresh_rate=100,
                                 distributed_backend=distributed_backend)
             trainer.fit(
-                model,
+                benchmark_model,
                 train_dataloader=dataloader_train_ssl,
                 val_dataloaders=dataloader_test
             )
             gpu_memory_usage.append(torch.cuda.max_memory_allocated())
             torch.cuda.reset_peak_memory_stats()
-            runs.append(model.max_accuracy)
+            runs.append(benchmark_model.max_accuracy)
 
             # delete model and trainer + free up cuda memory
-            del model
+            del benchmark_model
             del trainer
             torch.cuda.empty_cache()
         bench_results.append(runs)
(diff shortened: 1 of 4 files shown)


docs/source/getting_started/benchmarks/cifar10_benchmark.py
```diff
@@ -338,7 +243,7 @@ def configure_optimizers(self):
         for seed in range(n_runs):
             pl.seed_everything(seed)
             dataloader_train_ssl, dataloader_train_kNN, dataloader_test = get_data_loaders(batch_size)
-            model = Model(dataloader_train_kNN)
+            model = Model(dataloader_train_kNN, classes)
```
philippmwirth, line 246: Didn't mention it the last time but you could name this `BenchmarkModel` to be a bit clearer.

lightly/utils/__init__.py
```diff
@@ -13,3 +13,5 @@
 from lightly.utils.io import load_embeddings
 from lightly.utils.io import load_embeddings_as_dict
 from lightly.utils.embeddings_2d import fit_pca
+from lightly.utils.benchmarking import BenchmarkModule
+from lightly.utils.benchmarking import knn_predict
```
philippmwirth, line 17: This is not necessary to do the benchmarks, right? It's already included in the `BenchmarkModule`. You just put it here if someone wants to use it on his own?

lightly/utils/benchmarking.py
```diff
@@ -0,0 +1,215 @@
...
+    def __init__(self,
+                 dataloader_kNN: DataLoader,
+                 num_classes: int,
+                 knn_k: int=200,
+                 knn_t: float=0.1):
+        super().__init__()
+        self.backbone = nn.Module()
+        self.max_accuracy = 0.0
+        self.dataloader_kNN = dataloader_kNN
+        #self.gpus = gpus
```
philippmwirth, line 159: Why is this commented out?

philippmwirth (approved):
Very nice! Food for thought: What if our `SelfSupervisedEmbedding` or our `BaseEmbedding` class would inherit from the benchmarking module instead of the lightning module? We could still make the benchmarking voluntary. What do you think?

lightly/utils/__init__.py
IgorSusmelj: yes. In general, I don't like that we hide many useful methods within lightly. Having as much as possible private does not help anyone using lightly to build a custom model.

lightly/utils/benchmarking.py
IgorSusmelj: I removed it

Issue #220, opened 2021-03-12
Github actions call production api from __init__.py


MalteEbner:
- see https://github.com/lightly-ai/lightly/blob/e6565c0cd58e17472841a16943c360266de212a2/lightly/__init__.py#L119

PR #224, opened 2021-03-15
github actions: set LIGHTLY_SERVER_LOCATION="localhost:-1"

MalteEbner:
closes: #220

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,7 @@ jobs:
       run: pip install -e '.[all]'
     - name: Run Pytest
       run: |
+        LIGHTLY_SERVER_LOCATION="localhost:-1"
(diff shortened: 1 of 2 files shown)


philippmwirth (approved):

.github/workflows/test.yml
```diff
@@ -27,6 +27,7 @@ jobs:
       run: pip install -e '.[all]'
     - name: Run Pytest
       run: |
+        LIGHTLY_SERVER_LOCATION="localhost:-1"
```
japrescott, line 30: what kind of evil python magic is this? ;) care to elaborate to a NAP?

philippmwirth: We're simply setting the environment variable to something which doesn't exist so the package is basically tested offline.

PR #226, opened 2021-03-15
lightly dataset.subset from filenames me

### LINKED ISSUE #175: Add subset from filenames functionality to LightlyDataset
The `ActiveLearningAgent` returns a list of filenames for which the user will need to get predictions. Since this can be difficult and it's important, that the order of the filenames stays the same, it would be a good option to add the following functionality to the `LightlyDataset`.

```python
dataset = LightlyDataset.subset_from_filenames(input_dir, agent.unlabeled_set)
dataloader = torch.utils.data.DataLoader(dataset, shuffle=False, batch_size=1)
```

## Question A: Which kind of base dataset are supported during the creation?

1. input_dir: `str`
2. dataset: `torchvision.datasets.ImageFolder`
3. dataset: `lightly.data._helpers.DatasetFolder`
4. dataset: `lightly.data._video.VideoDataset`
5. dataset: `generic torchvision dataset` - Problem: it does not necessarily contain filenames
6. dataset: `LightlyDataset`

## Question B: How is the `SubsetDataset` represented:

1. just like the base dataset - not compatible with Option A1
2. a `LightlyDataset`
3. a `LightlyDatasetSubset`

## Question C: How is the `SubsetDataset` created:

1. Add a method to the base dataset - only compatible with option A6
2. Add a constructor to its representation - only compatible with B2 and B3

MalteEbner:
close #175

### DIFF
--- a/lightly/data/lighty_subset.py
+++ b/lightly/data/lighty_subset.py
@@ -4,16 +4,16 @@ from lightly.data.dataset import LightlyDataset
 
 
 class LightlySubset(LightlyDataset):
-    def __init__(self, base_dataset: LightlyDataset, subset_filenames: List[str]):
+    def __init__(self, base_dataset: LightlyDataset, filenames_subset: List[str]):
         self.base_dataset = base_dataset
 
         dict_base_dataset_filename_index: Dict[str, int] = \
             dict([
-                (base_dataset.index_to_filename(index), index)
+                (base_dataset.index_to_filename(self.dataset, index), index)
                 for index in range(len(base_dataset))
             ])
         self.mapping_subset_index_baseset = \
-            [dict_base_dataset_filename_index[filename] for filename in subset_filenames]
+            [dict_base_dataset_filename_index[filename] for filename in filenames_subset]
 
     def __getitem__(self, index_subset: int) -> Tuple[object, object, str]:
         index_baseset = self.mapping_subset_index_baseset[index_subset]
@@ -22,6 +22,11 @@ class LightlySubset(LightlyDataset):
 
     def __len__(self) -> int:
         return len(self.mapping_subset_index_baseset)
+
+    def index_to_filename(self, dataset, index_subset: int):
+        index_baseset = self.mapping_subset_index_baseset[index_subset]
+        fname = self.base_dataset.index_to_filename(dataset, index_baseset)
+        return fname
 
     @property
     def input_dir(self):
--- a/lightly/data/lighty_subset.py
+++ b/lightly/data/lighty_subset.py
@@ -1,0 +1,32 @@
+from typing import List, Dict, Tuple
+
+from lightly.data.dataset import LightlyDataset
+
+
+class LightlySubset(LightlyDataset):
+    def __init__(self, base_dataset: LightlyDataset, subset_filenames: List[str]):
+        self.base_dataset = base_dataset
+
+        dict_base_dataset_filename_index: Dict[str, int] = \
+            dict([
+                (base_dataset.index_to_filename(index), index)
+                for index in range(len(base_dataset))
+            ])
+        self.mapping_subset_index_baseset = \
+            [dict_base_dataset_filename_index[filename] for filename in subset_filenames]
+
+    def __getitem__(self, index_subset: int) -> Tuple[object, object, str]:
+        index_baseset = self.mapping_subset_index_baseset[index_subset]
+        sample, target, fname = self.base_dataset.__getitem__(index_baseset)
+        return sample, target, fname
+
+    def __len__(self) -> int:
+        return len(self.mapping_subset_index_baseset)
+
+    @property
+    def input_dir(self):
+        return self.base_dataset.input_dir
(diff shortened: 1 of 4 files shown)


lightly/data/lighty_subset.py
```diff
@@ -0,0 +1,37 @@
...
+
+from lightly.data.dataset import LightlyDataset
+
+
+class LightlySubset(LightlyDataset):
+    def __init__(self, base_dataset: LightlyDataset, filenames_subset: List[str]):
+        self.base_dataset = base_dataset
+
+        dict_base_dataset_filename_index: Dict[str, int] = \
+            dict([
```
philippmwirth, line 11: I think we should refrain from using list comprehension when the expressions are more complex. Why not simply use a loop?

lightly/data/lighty_subset.py
```diff
@@ -0,0 +1,37 @@
...
+                (base_dataset.index_to_filename(self.dataset, index), index)
+                for index in range(len(base_dataset))
+            ])
+        self.mapping_subset_index_baseset = \
```
philippmwirth, line 15: I'd rename it to be more clear: `subset_index_to_baseset_index`, `subset_index_to_dataset_index`, `map_subset_index_to_baseset_index` are possible candidates.

lightly/data/lighty_subset.py
```diff
@@ -0,0 +1,37 @@
...
+            [dict_base_dataset_filename_index[filename] for filename in filenames_subset]
+
+    def __getitem__(self, index_subset: int) -> Tuple[object, object, str]:
+        index_baseset = self.mapping_subset_index_baseset[index_subset]
+        sample, target, fname = self.base_dataset.__getitem__(index_baseset)
```
philippmwirth, line 20: You could also leave away this extra line but that's just a minor thing.

lightly/data/lighty_subset.py
```diff
@@ -0,0 +1,37 @@
...
+        return sample, target, fname
+
+    def __len__(self) -> int:
+        return len(self.mapping_subset_index_baseset)
+
+    def index_to_filename(self, dataset, index_subset: int):
```
philippmwirth, line 26: Why not simply store the list of filenames which are passed in the beginning and the return the filename at index?

philippmwirth (commented):
Logic is straight-forward and looks good. I have some suggestions for better names, please take a look.

An additional problem I see is the fact that the `LightlySubset` is inheriting from and wrapping around `LightlyDataset` at the same time. This looks like it could spell trouble in the future -> very easy source for mistakes and bugs. Don't you think a PyTorch like implementation would be more straight-forward and make more sense although it provides a little less user comfort?

lightly/data/lighty_subset.py
MalteEbner: I assumed that the list comprehension would be faster. However, now that I profiled it, it isn't, thus I changed it to use a for loop.

lightly/data/lighty_subset.py
MalteEbner: I prefer naming the return values explicitly, as it makes it much easier and faster during coding to assign the return values of a call of this function to variables:
```python
for i in range(len(dataset)):
    sample, target, fname = dataset.__getitem__(i)
```
Without naming the variables explicitly, I would need to dig into the nested function `self.base_dataset.__getitem__(index_baseset)` and possibly even further to get what the return values are and stand for.

MalteEbner:
About the idea to use a pytorch subset instead of a LightlySubset being both a child class and a wrapper:

The current implementation is very close to the one of the [`pytorch Subset`](https://pytorch.org/docs/stable/_modules/torch/utils/data/dataset.html#Subset), which is also both a child and a wrapper. Thus the `pytorch Subset` has the same problems, e.g. its transform is not settable directly as `dataset.transform`, but only nested through `dataset.dataset.transform`. 
Thus using a pytorch Subset is also a potential source for trouble in the future. I would consider it as even more dangerous, as the wrappers for the properties are not made explicitly. 
Another disadvantage of the `pytorch Subset` is that it would not support the `get_filenames` and `dump` methods.

philippmwirth:
> 
> 
> About the idea to use a pytorch subset instead of a LightlySubset being both a child class and a wrapper:
> 
> The current implementation is very close to the one of the [`pytorch Subset`](https://pytorch.org/docs/stable/_modules/torch/utils/data/dataset.html#Subset), which is also both a child and a wrapper. Thus the `pytorch Subset` has the same problems, e.g. its transform is not settable directly as `dataset.transform`, but only nested through `dataset.dataset.transform`.
> Thus using a pytorch Subset is also a potential source for trouble in the future. I would consider it as even more dangerous, as the wrappers for the properties are not made explicitly.
> Another disadvantage of the `pytorch Subset` is that it would not support the `get_filenames` and `dump` methods.

Hm good point, I didn't know that.

philippmwirth (approved):

philippmwirth (approved):

PR #230, opened 2021-03-16
Barlow twins loss and architecture

AdrianArnaiz:
Implemented 2 files:

* `lightly/models/barlow_twins_arch.py`: architecture with flexible backbone and projection head.
* `lightly/loss/barlow_twins_loss.py`: loss from original paper.


lightly/loss/barlow_twins_loss.py
```diff
@@ -0,0 +1,47 @@
...
+        >>>
+        >>> # feed through SimSiam model
+        >>> out0, out1 = model(t0, t1)
+        >>>
+        >>> # calculate loss
+        >>> loss = loss_fn(out0, out1)
+
+    """
+
+    def __init__(self, device, lambda_param=5e-3):
```
philippmwirth, line 25: I would not require the device here but rather get the device during the forward pass as you can see in our [`NTXEntLoss` (this is a link)](https://github.com/lightly-ai/lightly/blob/develop/lightly/loss/ntx_ent_loss.py) on line 128.

In your `forward` implementation it could be something like:
```Python

device = z_a.device
```

lightly/loss/barlow_twins_loss.py
```diff
@@ -0,0 +1,47 @@
...
+        z_a_norm = (z_a - z_a.mean(0)) / z_a.std(0) # NxD
+        z_b_norm = (z_b - z_b.mean(0)) / z_b.std(0) # NxD
+
+        N = z_a.size(0)
+        D = z_a.size(1)
+
+        # cross-correlation matrix
+        c = torch.mm(z_a_norm.T, z_b_norm) / N # DxD
+        # loss
+        c_diff = (c - torch.eye(D, device=self.device)).pow(2) # DxD
```
philippmwirth, line 41: When you change the way you get the device, don't forget to change it here as well :)

philippmwirth (changes requested):
Looks really good! I'm requesting one tiny change with respect to how the `device` is retrieved in the loss function (see my comment). After that, we're good to go! 

Let me know if anything's unclear :)

lightly/loss/barlow_twins_loss.py
AdrianArnaiz: Perfect! I'll commit soon

lightly/loss/barlow_twins_loss.py
AdrianArnaiz: Thanks :)

lightly/loss/barlow_twins_loss.py
AdrianArnaiz: Done

philippmwirth (approved):

PR #252, opened 2021-03-19
Object detection scorer

IgorSusmelj:
### Changes
- Added the framework for computing scores for object detection predictions
- Add the scorer for "frequency-detection" which prefers images with many objects and diverse objects

### DIFF
--- a/lightly/active_learning/scorers/detection.py
+++ b/lightly/active_learning/scorers/detection.py
@@ -1,0 +1,59 @@
+from typing import *
+
+import numpy as np
+
+from lightly.active_learning.scorers.scorer import Scorer
+
+
+def _object_frequency(model_output: List, frequency_pentaly: float, min_score: float) -> np.ndarray:
+    """Computes the entropy of a probability matrix over one array
+    """
+    n_objs = []
+    for output in model_output:
+        val = 0
+        objs = {}
+        for label_prob in output['class_probabilities']:
+            label = np.argmax(label_prob)
+            if label in objs:
+                objs[label] += frequency_pentaly
+            else:
+                objs[label] = 1
+        for k, v in objs.items():
+            val += v
+        n_objs.append(val)
+    
+    scores = [np.interp(x, (x.min(), x.max()), (min_score, 1.0)) for x in n_objs]
+    return scores
+
+
+class ScorerObjectDetection(Scorer):
+    """Class to compute active learning scores from the model_output of a object detection task.
+
+    Attributes:
+        model_output:
+            List of length N containing the predictions of an object detection
+            model. For each sample we expect a dictionary containing `boxes`,
+            `object_probabilities` and `class_probabilities`.
+
+
+    """
+    def __init__(self, model_output: List, config: Dict):
+        self.model_output = model_output
+        self.config = config
+
+    def _calculate_scores(self) -> Dict[str, np.ndarray]:
+        scores = dict()
+        scores["object-frequency"] = _object_frequency
+        return scores
+
+    def _get_object_frequency(self):
+        scores = _object_frequency(self.model_output, 0.25, 0.9)
+        return scores
+
+    def _get_prediction_margin_score(self):
+        uncertainties = np.array([1 - max(class_probabilities) for class_probabilities in self.model_output])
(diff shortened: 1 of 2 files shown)


lightly/active_learning/scorers/detection.py
```diff
@@ -0,0 +1,76 @@
+from typing import *
+
+import numpy as np
+
+from lightly.active_learning.scorers.scorer import Scorer
+
+
+def _object_frequency(model_output: List[Dict],
+                      frequency_pentaly: float,
```
philippmwirth, line 9: "pentaly" ;)

philippmwirth (changes requested):
Looks good, I have some minor feedback.

One thing I'd also like to discuss: We're requesting a rather specific format, I know that for example torchvision FasterRCNN provides a different output. Maybe we should already anticipate this and think about which is the most general type of boxes we can get -> make this the parent class and then have the others inherit from that. What do you think?

IgorSusmelj:
> Looks good, I have some minor feedback.
> 
> One thing I'd also like to discuss: We're requesting a rather specific format, I know that for example torchvision FasterRCNN provides a different output. Maybe we should already anticipate this and think about which is the most general type of boxes we can get -> make this the parent class and then have the others inherit from that. What do you think?

I like the idea of being more flexible. But there are so many options in object detection. I feel the best way is to have a clear format we want but help the user get there. Maybe we could offer some helper functions to the user. E.g. to change bounding box representation (x1, y1, x2, y2) --> (x1, y1, w, h).
Do you think that would work?

lightly/active_learning/scorers/detection.py
```diff
@@ -0,0 +1,76 @@
...
+            model. For each sample we expect a dictionary containing `boxes`,
+            `object_probabilities` and `class_probabilities`.
+            `boxes` has shape [B, 4] for B found bounding boxes and the 4 values
+            x1, y1, x2, y2.
+            `object_probabilities` has shape [B] with a objectness value for each
+            bounding box.
+            `class_probabilities` has shape [B, C] for the B bounding boxes and
+            C classes. The class probabilities should all sum up to 1.
+        config:
+            A dictionary containing additional parameters for the scorers.
```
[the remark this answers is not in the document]
IgorSusmelj: I will add one

lightly/active_learning/scorers/detection.py
IgorSusmelj: arrgh, will fix this :)

philippmwirth:
> I like the idea of being more flexible. But there are so many options in object detection. I feel the best way is to have a clear format we want but help the user get there. Maybe we could offer some helper functions to the user. E.g. to change bounding box representation (x1, y1, x2, y2) --> (x1, y1, w, h).
> Do you think that would work?

I think there are many approaches we could take there:
- Bounding Box Class for each type of possible bbox and all provide the necessary functions for our scorer.
- Scorer Class for each type of possible bbox and all calculate the scores they can
- Helper functions which make it easy for the user to switch between formats

I would prefer the first one but maybe this should be covered in a different ticket.

IgorSusmelj:
> > I like the idea of being more flexible. But there are so many options in object detection. I feel the best way is to have a clear format we want but help the user get there. Maybe we could offer some helper functions to the user. E.g. to change bounding box representation (x1, y1, x2, y2) --> (x1, y1, w, h).
> > Do you think that would work?
> 
> I think there are many approaches we could take there:
> 
> * Bounding Box Class for each type of possible bbox and all provide the necessary functions for our scorer.
> * Scorer Class for each type of possible bbox and all calculate the scores they can
> * Helper functions which make it easy for the user to switch between formats
> 
> I would prefer the first one but maybe this should be covered in a different ticket.

Let's discuss this further in another issue. I'm also not sure where to put it. To `utils`? I would prefer 1 or 3. I could imagine that the 3rd option could be the easiest to implement. A few methods to transform the bounding boxes. I'm not sure a class would help.

IgorSusmelj:
I created an issue for the further discussion: https://github.com/lightly-ai/lightly/issues/253

philippmwirth (approved):

PR #256, opened 2021-03-22
137 detection active learning pw

### LINKED ISSUE #137: [Active Learning] Compute Active learning scores for object detection
Create a scorer for object detection that can compute scores like prediction entropy, ...

Input of predictions:
For each of the N samples, there are B bounding boxes giving the probability of one of the C classes.

- [x] Find a good way to represent the model predictions of an object detection model
- [x] Decide which kind of models to support (Yolo, SSD, ...), see also https://heartbeat.fritz.ai/a-2019-guide-to-object-detection-9509987954c3
- [x] Decide which kinds of scores to compute and find a good way to compute them, e.g. see https://heartbeat.fritz.ai/a-2019-guide-to-object-detection-9509987954c3 or https://arxiv.org/pdf/2004.04699.pdf
- [x] Implement it

philippmwirth:
# Active Learning Scores for Object Detection and Refactoring

Closes #137.

I refactored the input to the object detection scorer so that we can always expect the same input format. I introduced the following classes:
- `BoundingBox`: A helper to create sound bounding boxes
- `ObjectDetectionOutput`: A helper to unify the interface to the output of object detection models.

Furthermore, I added an additional scoring method which calculates the `prediction-margin` of the object detection output.

### DIFF
--- a/lightly/active_learning/scorers/detection.py
+++ b/lightly/active_learning/scorers/detection.py
@@ -3,12 +3,13 @@ from typing import *
 import numpy as np
 
 from lightly.active_learning.scorers.scorer import Scorer
+from lightly.active_learning.utils.object_detection_output import ObjectDetectionOutput
 
 
-def _object_frequency(model_output: List[Dict],
+def _object_frequency(model_output: List[ObjectDetectionOutput],
                       frequency_penalty: float,
                       min_score: float) -> np.ndarray:
-    """Scorer which prefers samples with many and diverse objects
+    """Score which prefers samples with many and diverse objects.
 
     Args:
         model_outputs:
@@ -22,13 +23,13 @@ def _object_frequency(model_output: List[Dict],
         
     Returns:
         Numpy array of length N with the computed scores
+
     """
     n_objs = []
     for output in model_output:
         val = 0
         objs = {}
-        for label_prob in output['class_probabilities']:
-            label = np.argmax(label_prob)
+        for label in output.labels:
             if label in objs:
                 objs[label] += frequency_penalty
             else:
@@ -43,20 +44,29 @@ def _object_frequency(model_output: List[Dict],
     return np.asarray(scores)
 
 
+def _prediction_margin(model_output: List[ObjectDetectionOutput]):
+    """TODO
+
+    """
+    scores = []
+    for output in model_output:
+        if len(output.scores) > 0:
+            # prediction margin is 1 - max(class probs), therefore the mean margin
+            # is mean(1 - max(class probs)) which is 1 - mean(max(class probs))
+            score = 1. - np.mean(output.scores)
+        else:
+            # set the score to 0 if there was no bounding box detected
+            score = 0.
+        scores.append(score)
+    return np.asarray(scores)
+
+
 class ScorerObjectDetection(Scorer):
-    """Class to compute active learning scores from the model_output of a object detection task.
+    """Class to compute active learning scores from the model_output of an object detection task.
 
     Attributes:
         model_output:
-            List of length N containing the predictions of an object detection
-            model. For each sample we expect a dictionary containing `boxes`,
-            `object_probabilities` and `class_probabilities`.
-            `boxes` has shape [B, 4] for B found bounding boxes and the 4 values
-            x1, y1, x2, y2.
-            `object_probabilities` has shape [B] with a objectness value for each
-            bounding box.
-            `class_probabilities` has shape [B, C] for the B bounding boxes and
-            C classes. The class probabilities should all sum up to 1.
+            TODO
         config:
             A dictionary containing additional parameters for the scorers.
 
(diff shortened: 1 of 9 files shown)


lightly/active_learning/utils/object_detection_output.py
```diff
@@ -0,0 +1,98 @@
+""" TODO """
```
MalteEbner, line 1: TODO

lightly/active_learning/utils/bounding_box.py
```diff
@@ -0,0 +1,88 @@
...
+                f'[{x0}, {y0}, {x1}, {y1}]'
+            )
+
+        self.x0 = x0
+        self.y0 = y0
+        self.x1 = x1
+        self.y1 = y1
+
+    @classmethod
+    def from_x_y_w_h(cls, x: float, y: float, w: float, h: float):
```
IgorSusmelj, line 60: why not using already integrated methods from torchvision? I would suggest maybe use an internal format close to theirs (x1, y1, x2, y2) and then use their methods for conversion or to get the area...
see: https://github.com/pytorch/vision/blob/master/torchvision/ops/boxes.py

lightly/active_learning/utils/object_detection_output.py
```diff
@@ -0,0 +1,98 @@
...
+        >>>
+        >>> # create detection output
+        >>> detection_output = ObjectDetectionOutput(boxes, scores, labels)
+
+    """
+
+    def __init__(self,
+                 boxes: List[BoundingBox],
+                 scores: List[float],
+                 labels: List[int]):
```
IgorSusmelj, line 43: Why use labels instead of the class probabilities? I thought we discussed that we use class probs, bounding boxes and object probs

MalteEbner (approved):
I like the structure and the tests a lot. Great work!

lightly/active_learning/utils/bounding_box.py
philippmwirth: Mainly because they are using tensors and I really don't see a reason to convert to tensors and back just to get the area. What would be the advantage? I agree that this makes sense for more elaborate functions (e.g. I would not write my own non-maximum suppression).

lightly/active_learning/utils/object_detection_output.py
philippmwirth: Yes, the intuition here was that you can always go from class probabilities to labels/scores but you can't go the other way. If we require class probabilities there are many people who can't use our scorer. Additionally, we don't have any scorers yet which require the class probability so I thought we could add it as an option later?

lightly/active_learning/scorers/detection.py
```diff
@@ -43,42 +44,72 @@ def _object_frequency(model_output: List[Dict],
...
 
     Examples:
-        >>> predictions = [{
-        >>>     'boxes': [[14, 16, 52, 85]],
-        >>>     'object_probabilities': [0.1024],
-        >>>     'class_probabilities': [[0.5, 0.41, 0.09]]
-        >>> }]
-        >>> scorer = ScorerObjectDetection(predictions)
+        >>> # from output with scores
+        >>> detection_outputs = ObjectDetectionOutput(boxes, scores, labels)
```
IgorSusmelj, line 82: I would expand the example here and start with boxes, scores, and class probabilities or labels. If I just quickly want to work with the scorer I first have to understand the ObjectDetectionOutput. To simplify I would give an example already here. E.g. from model output to scores. So I don't have to read up on the other parts of the framework all the time.

lightly/active_learning/utils/bounding_box.py
```diff
@@ -0,0 +1,88 @@
...
+    Examples:
+    >>> # simple case
+    >>> bbox = BoundingBox(0.1, 0.2, 0.3, 0.4)
+    >>>
+    >>> # same bounding box in x, y, w, h format
+    >>> bbox = BoundingBox.from_x_y_w_h(0.1, 0.2, 0.2, 0.2)
+    >>>
+    >>> # same bounding box with prior normalization
+    >>> W, H = 100, 100 # get image shape
+    >>> bbox = BoundingBox(10 / W, 20 / H, 30 / W, 40 / H)
```
IgorSusmelj, line 29: Not sure this example is adding any value

lightly/active_learning/utils/bounding_box.py
```diff
@@ -0,0 +1,88 @@
...
+            x0 coordinate (normalized to [0, 1])
+        y0:
+            y0 coordinate (normalized to [0, 1])
+        x1:
+            x1 coordinate (normalized to [0, 1])
+        y1:
+            y1 coordinate (normalized to [0, 1])
+
+    Examples:
+    >>> # simple case
```
IgorSusmelj, line 21: mention the bounding box format here

lightly/active_learning/utils/bounding_box.py
philippmwirth: I can remove it, thought it might help

lightly/active_learning/utils/bounding_box.py
IgorSusmelj: We could just use their conversion functions.

lightly/active_learning/utils/bounding_box.py
MalteEbner: I would say it does.

MalteEbner (approved):

lightly/active_learning/utils/bounding_box.py
MalteEbner: I would say it does help

IgorSusmelj (changes requested):
Would consider using torchvision for the conversion of box formats. Use class probs if we can and labels as a fallback.

lightly/active_learning/utils/bounding_box.py
IgorSusmelj: Then I would add some more explanation. Using words like "prior normalization" is complicated but the scaling of the box to 0-1 is more for a simple thing. So I would use simple language for both. Maybe add more info about the input bounding box dimensions?

lightly/active_learning/utils/object_detection_output.py
IgorSusmelj: I feel we set here a standard. Switching it later might be complicated. We can always simplify (e.g. replace class probs with argmax internally). But the other way around might be breaking, no?

lightly/active_learning/utils/bounding_box.py
IgorSusmelj: We can also look at this later since this is pure internal handling/conversion.

philippmwirth: Yes, let's discuss this later :)

IgorSusmelj (approved):

MalteEbner (approved):

PR #263, opened 2021-03-29
245 cli lightly version pw

### LINKED ISSUE #245: Add option to show which lightly version is used by CLI
# Add option to show which lightly version is used by CLI

It would be a good feature to show the version which is used by the CLI so as to help a user debug if something goes wrong.

philippmwirth:
# Add cli command to show installed version

Closes #245.

I added a new CLI command which shows which version of lightly is installed and used for the cli.
This can be very helpful when working with different environments.

Example:
```
philipp_lightly_ai@philipp-gpu:~/$ lightly-version
lightly version 1.1.3
```

### DIFF
--- a/docs/source/getting_started/command_line_tool.rst
+++ b/docs/source/getting_started/command_line_tool.rst
@@ -24,6 +24,22 @@ the CLI.
     </div>
 
 
+Check the installation of lightly
+-----------------------------------
+To see if the lightly command-line tool was installed correctly, you can run the
+following command which will print the installed lightly version:
+
+.. code-block:: bash
+
+    lightly-version
+
+If lightly was installed correctly, you should see something like this:
+
+.. code-block:: bash
+
+    lightly version 1.1.4
+
+
 Train a model using the CLI
 ---------------------------------------
 Training a model using default parameters can be done with just one command. Let's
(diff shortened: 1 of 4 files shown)


japrescott:
is there a special reason why we have so many different commands and not one `lightly` command? we have lightly-upload, lightly-embed, lighty-download, lightly-?! Is this the python way? or the pytorch way?

philippmwirth:
> 
> 
> is there a special reason why we have so many different commands and not one `lightly` command? we have lightly-upload, lightly-embed, lighty-download, lightly-?! Is this the python way? or the pytorch way?

It's just a design choice and it's related to how hydra works (which is what we use to parse the command line arguments) and how our CLI is structured.

One reason why we cannot do something like this:
```
lightly --what upload
```
is that hydra requires the following syntax:
```
lightly what=upload
```
which is IMO less intuitive than naming the commands exactly after what they are doing.

However, refactoring the CLI is in our pipeline anyways so feel free to start a discussion about setup/naming/etc :)

MalteEbner (commented):
Could you add a test of this command to 
https://github.com/lightly-ai/lightly/blob/5b733261bbd31543fc4229f4a6f0253b2301af1c/.github/workflows/test_setup.yml#L31-L37 ?

MalteEbner (approved):

PR #267, opened 2021-03-30
126 consistency regularization pw

philippmwirth:
# Consistency Regularization

Implementation of the [CO2 paper](https://arxiv.org/abs/2010.02217) for consistency regularization based on [this repo](https://github.com/philippmwirth/co2).

- Implemented the CO2 regularizer as a standalone feature (can be added to any existing loss)
- Added documentation
- Added tests

For benchmarks, check out my [other repo](https://github.com/philippmwirth/co2). Regularized models consistently outperform the default ones but only by a tiny margin.

### DIFF
--- a/lightly/loss/regularizer/co2.py
+++ b/lightly/loss/regularizer/co2.py
@@ -1,0 +1,131 @@
+""" CO2 Regularizer """
+
+# Copyright (c) 2020. Lightly AG and its affiliates.
+# All Rights Reserved
+
+import torch
+from lightly.loss.memory_bank import MemoryBankModule
+
+
+class CO2Regularizer(MemoryBankModule):
+    """Implementation of the CO2 regularizer [0] for self-supervised learning.
+
+    [0] CO2, 2021, https://arxiv.org/abs/2010.02217
+
+    Attributes:
+        alpha:
+            Weight of the regularization term.
+        t_consistency:
+            Temperature used during softmax calculations.
+        memory_bank_size:
+            Number of negative samples to store in the memory bank.
+            Use 0 to use the second batch for negative samples.
+
+    Examples:
+        >>> # initialize loss function for MoCo
+        >>> loss_fn = NTXentLoss(memory_bank_size=4096)
+        >>>
+        >>> # initialize CO2 regularizer
+        >>> co2 = CO2Regularizer(alpha=1.0, memory_bank_size=4096)
+        >>>
+        >>> # generate two random trasnforms of images
+        >>> t0 = transforms(images)
+        >>> t1 = transforms(images)
+        >>>
+        >>> # feed through the MoCo model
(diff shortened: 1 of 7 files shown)


lightly/loss/regularizer/co2.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def _get_pseudo_labels(self,
+                           out0: torch.Tensor,
+                           out1: torch.Tensor,
+                           negatives: torch.Tensor = None):
+        """Computes the soft pseudo labels across negative samples.
+
+        Args:
+            out0:
+                Output projections of the first set of transformed images.
```
MalteEbner, line 61: Could you elaborate a bit more on what the tensors are and on their size/shape?

To understand what they are, it would help if you could relate `out0`, `out1` and `negatives` to the three terms uses in the paper in illustration 1b: x_p, x_q, {x_1, x_2 ...x_k ... x_K}). Is the following assumption correct?
-  out0 = {x_p_1, x_p_2, ... x_p_N} = positive samples; with N=batch size,
-  out1 = {x_q_1, x_q_2, ... x_q_N} = query samples; with N=batch size
- negatives = {x_1, x_2 ...x_k ... x_K} = negative crops
- all three arguments: the tensor itself are of shape NxC respectively KxC with C being the embedding dimension.
Intermediate values: I cannot infer their shape from the code, what is it? I assume:
-l_pos: shape: Nx1
-l_neg: shape: NxK with negatives, Nx(N-1) with out1 as negatives
- logits: shape: Nx(1+K) with negatives, Nx(1 + (N-1)) with out1 as negatives
- return value: same shape of logits, but you should specify this shape more.

lightly/loss/regularizer/co2.py
```diff
@@ -0,0 +1,131 @@
...
+        # ask memory bank for negative samples and extend it with out1 if 
+        # out1 requires a gradient, otherwise keep the same vectors in the 
+        # memory bank (this allows for keeping the memory bank constant e.g.
+        # for evaluating the loss on the test set)
+        out1, negatives = \
+            super(CO2Regularizer, self).forward(out1, update=True)
+        
+        # get log probabilities
+        p = self._get_pseudo_labels(out0, out1, negatives)
+        q = self._get_pseudo_labels(out1, out0, negatives)
```
MalteEbner, line 128: `self._get_pseudo_labels()` is only used with negatives, thus we did you implement it also without negatives?

MalteEbner (approved):
I approve, but it would be great if you could add more documentation.

lightly/loss/regularizer/co2.py
philippmwirth: The negatives can be `None` if `memory_bank_size` is set to 0. I will extend the comment.

PR #269, opened 2021-03-30
225 replace put request with api client.request

### LINKED ISSUE #225: Replace put_request(url, body) to use api_client.requests("PUT", url, body)
Advantages:
- smaller codebase
- less error prone: use generated code for timeout, error handling, ...
- Allows to drop the response library, once the docker endpoints are also refactored to use the new API

MalteEbner:
close #225 

After trying for several hours to upload directly with the urllib, I decided not to try it anymore. Instead, I 
- restructured the code and inlined it into the only calling instance
- deprecated unused arguments
- added more precise docstrings

Unmocked tests are running through.


lightly/api/api_workflow_client.py
```diff
@@ -130,7 +132,47 @@ def filenames_on_server(self):
...
+            max_backoff:
+                Maximal backoff before retrying.
+            max_retries:
+                Maximum number of retries before timing out.
+
+        Returns:
+            The response of the put request, usually a 200 for the success case.
+
+        """
+
```
philippmwirth, line 154: Why not use [this](https://github.com/lightly-ai/lightly/blob/7e68ddf03b898a9e9c00549918366c3af71ab865/lightly/openapi_generated/swagger_client/api_client.py#L343) instead?

philippmwirth (changes requested):

lightly/api/api_workflow_client.py
MalteEbner: I tried to use it for several hours, but it does not have a `data` attribute and expects the data to be json-serialisable, as it is designed for REST apis requests.

philippmwirth: And what about [this](https://docs.python.org/3/library/urllib.request.html#urllib.request.Request)?

philippmwirth: But then your `signed_write_url` has expired. This is not a problem of urllib but the server who refuses to authorize your request.

MalteEbner: urllib2 is not part of the standard package, urllib3 does not have a `data` attribute neither. I looked at many SO-sites for how to send an image/file with one of the urllib libraries and did not find any working ones. Oftentimes, it was recommended to use the requests library instead, also see https://stackoverflow.com/questions/2018026/what-are-the-differences-between-the-urllib-urllib2-urllib3-and-requests-modul

PR #277, opened 2021-04-01
273 retry wrapper pw

### LINKED ISSUE #273: Wrap upload and download requests in retry wrapper
# Wrap upload and download requests in retry wrapper

Since we start bombing the api with requests when uploading/downloading, it can happen that our api returns a 502 which leads to an exception at the user side. We can wrap the critical api calls such that they are repeated upon failure until they exceed a maximum number of retries.

philippmwirth:
# retry wrapper

Closes #273.

I wrapped all function calls to the api which are used in `_upload_single_image` in retry wrappers so that a single failed request will not interrupt the upload process.

### DIFF
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -152,27 +152,11 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin, _SamplingMixin, _UploadDatasetMi
 
         """
 
-        counter = 0
-        backoff = 1. + random.random() * 0.1
-        success = False
-        while not success:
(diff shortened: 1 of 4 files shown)


japrescott (approved):
very nice <3

MalteEbner (changes requested):
In general: very good idea and implementation.

Please also replace the following code by a retry call, to use the code even more:
https://github.com/lightly-ai/lightly/blob/426dbda1e096bec4bae98b643ee0e8ea8b857f0c/lightly/api/api_workflow_client.py#L135

MalteEbner (approved):

Issue #250, opened 2021-03-18
Add Alignment and Uniformity Loss


philippmwirth:
# Add Alignment and Uniformity Loss
Following the discussion in #249 and as suggested by @EelcoHoogendoorn.

Implement the loss function described [here](https://arxiv.org/abs/2005.10242), which achieves competitive results with more cited ones (symmetric negative cosine similarity & contrastive loss) while providing better interpretability.

EelcoHoogendoorn:
Thanks for facilitating this; ive made a fork on my account and made a feature branch there branching from dev. Any further details on how youd like to work towards a PR?

philippmwirth:
Since there are no dependencies with this issue, you can simply create a PR once you feel comfortable with your implementation and then we'll take a look. It would be good to rebase your feature branch onto our develop before you open the PR. 

If you have any questions just let us know :)

PR #278, opened 2021-04-02
Hypersphere loss

EelcoHoogendoorn:
Pr for https://github.com/lightly-ai/lightly/issues/250

Played around a lot with all kinds of different variations to this type of loss function, but decided it was getting out of hand, so I just left this PR with the literal implementation as provided in the published paper.

I dont like the temperature parameter this method has, its similar to the one found in the crossentropy; nor do I like the weight between the two losses. In practice I find that the behavior of the loss is data dependent. I have observed funny behavior in the past some for types of data and architectures. Ideally, the average distance between negative samples tends to sqrt(2), which is its maximum on a hypersphere of any dimension. Often this is approximated, but this certainly isnt guaranteed without tweaking. However, as provided this loss does (EDIT: said 'not' here but that was a typo) seem quite robust to catastropic collapse.

Also, ive experimented with various ways to make the 'temperature' parameter trainable, or scaling with the data distribution. This is quite important to convergence and quality; but its a huge rabbit hole, so yeah ill just stick to copying the paper here.


lightly/loss/hypersphere_loss.py
```diff
@@ -0,0 +1,96 @@
...
+https://arxiv.org/pdf/2005.10242.pdf
+
+NOTE: hypersphere is perhaps bad naming as I am not sure it is the essence;
+alignment-and-uniformity loss perhaps? Does not sound as nice.
+"""
+
+import torch
+
+
+def normalize_l2(x, dim=-1):
```
IgorSusmelj, line 11: Can we use the PyTorch functional `normalize` for this?
(see https://pytorch.org/docs/1.1.0/nn.html#normalize)

```python
import torch.nn.functional as F
x_normalized = F.normalize(x)
```

lightly/loss/hypersphere_loss.py
```diff
@@ -0,0 +1,96 @@
...
+
+        Returns:
+            torch.Tensor, [], float
+                scalar loss value
+
+        """
+        # FIXME: this is necessary for the loss to function as advertized,
+        #  but not technically part of the loss i feel
+        #  as it needs to be applied during inference as well
+        #  leave that responsibility to the end user?
```
IgorSusmelj, line 85: I would leave it here since it's needed for the loss calculation. For inference, I would leave it to the user to do the normalization.

lightly/loss/hypersphere_loss.py
```diff
@@ -0,0 +1,96 @@
...
+        x = normalize_l2(z_a)
+        y = normalize_l2(z_b)
+
+        def lalign(x, y):
+            return (x - y).norm(dim=1).pow(self.alpha).mean()
+        def lunif(x):
+            sq_pdist = torch.pdist(x, p=2).pow(2)
+            # NOTE: add nan_to_num to support batch-size==1 case.
+            # not a very practically relevant case but the tests as copy-pasted seem to demand it
+            return sq_pdist.mul(-self.t).exp().mean().log().nan_to_num()
```
IgorSusmelj, line 95: You can drop the `nan_to_num` here. As you mentioned, the case of batch size 1 makes not much sense. There is no need to make the tests pass that case. 
`nan_to_num` would additionally require us to switch the minimum PyTorch version to 1.8.

IgorSusmelj (changes requested):
Great work! Thank you very much for the contribution. I tested the code and it works :)

I will do some benchmarking and a more in-depth analysis of the new loss later.

Would do some minor modifications to the code before merging:
- Use PyTorch built-in normalize function
- No need to support batch size 1 for the test case

lightly/loss/hypersphere_loss.py
EelcoHoogendoorn: Alright; will do. Wasn't keen on the overhead of another call in the graph for purely theoretical reasons either.

lightly/loss/hypersphere_loss.py
EelcoHoogendoorn: Alright; will make sure the docstring calls this out as explicitly as possible.

lightly/loss/hypersphere_loss.py
EelcoHoogendoorn: Seems like it. Implementing it myself is a reflex of numpy refusing to do so for the part 10 years ive been using it I suppose.

EelcoHoogendoorn:
Adressed the above comments; the only open question I have is about the naming of things; not sure that Hypersphere loss really captures the essence of it... but its not like I have better ideas.

IgorSusmelj (approved):

PR #283, opened 2021-04-12
Bugfix: active learning scores numpy float32 me

### LINKED ISSUE #259: ActiveLearningScores cannot be sent when of type numpy.float32
When the active learning scores are of type `numpy.float32`, they cannot be parsed to a swagger_type:
`AttributeError: 'numpy.float32' object has no attribute 'swagger_types'`
This is solved by either casting to `numpy.float64` or a plain python `float`.

- [ ] Implement a check before sending the active learning scores, that `numpy.float32` are cast to `float` or` numpy.float64`

MalteEbner:
close #259 

Problem: as `isinstance(np.ones((1, ) dtype=float32), float) == False`, serialising a np.float32 array to json fails, whereas np.float64 is a float. The error occurs as https://github.com/lightly-ai/lightly/blob/4b9401e91c555be6811dfe320ce46b22cb4cb976/lightly/openapi_generated/swagger_client/api_client.py#L196 is not entered for np.float32 arrays, causing Exceptions later on. The test in https://github.com/lightly-ai/lightly/blob/259-active-learning-scores-numpy-float32-me/tests/api/test_rest_parser.py shows the error directly as the unmocked functions are used. The changes in the other two test files reproduce the error, as the mocked function mimics the error throwing for np.float32.

I solved the problem by 
https://github.com/lightly-ai/lightly/blob/acb59f3416a33a2bc0b3eb00a55cd59639c127c4/lightly/api/api_workflow_sampling.py#L58-L59

The change in the documentation to SamplingMethod.CORESET is unrelated, but I changed it, as PyCharm showed me an error for it and it is easier for users to use the attributes, as they guide directly which methods are allowed.

### DIFF
--- a/lightly/active_learning/config/sampler_config.py
+++ b/lightly/active_learning/config/sampler_config.py
@@ -24,13 +24,13 @@ class SamplerConfig:
     Examples:
         >>> # sample 100 images with CORESET sampling
         >>> config = SamplerConfig(method=SamplingMethod.CORESET, n_samples=100)
-        >>> config = SamplerConfig(method='CORESET', n_samples=100)
+        >>> config = SamplerConfig(method=SamplingMethod.CORESET, n_samples=100)
         >>>
         >>> # give your sampling a name
-        >>> config = SamplerConfig(method='CORESET', n_samples=100, name='my-sampling')
+        >>> config = SamplerConfig(method=SamplingMethod.CORESET, n_samples=100, name='my-sampling')
         >>>
         >>> # use minimum distance between samples as stopping criterion
-        >>> config = SamplerConfig(method='CORESET', n_samples=-1, min_distance=0.1)
+        >>> config = SamplerConfig(method=SamplingMethod.CORESET, n_samples=-1, min_distance=0.1)
 
     """
     def __init__(self, method: SamplingMethod = SamplingMethod.CORESET, n_samples: int = 32, min_distance: float = -1,
--- a/lightly/active_learning/config/sampler_config.py
+++ b/lightly/active_learning/config/sampler_config.py
@@ -24,7 +24,6 @@ class SamplerConfig:
     Examples:
         >>> # sample 100 images with CORESET sampling
         >>> config = SamplerConfig(method=SamplingMethod.CORESET, n_samples=100)
-        >>> config = SamplerConfig(method=SamplingMethod.CORESET, n_samples=100)
         >>>
         >>> # give your sampling a name
         >>> config = SamplerConfig(method=SamplingMethod.CORESET, n_samples=100, name='my-sampling')
--- a/lightly/api/api_workflow_sampling.py
+++ b/lightly/api/api_workflow_sampling.py
@@ -55,6 +55,8 @@ class _SamplingMixin:
             if preselected_tag_id is None:
                 raise ValueError
             for score_type, score_values in al_scores.items():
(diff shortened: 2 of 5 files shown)


lightly/active_learning/config/sampler_config.py
```diff
@@ -24,13 +24,13 @@ class SamplerConfig:
     Examples:
         >>> # sample 100 images with CORESET sampling
         >>> config = SamplerConfig(method=SamplingMethod.CORESET, n_samples=100)
-        >>> config = SamplerConfig(method='CORESET', n_samples=100)
+        >>> config = SamplerConfig(method=SamplingMethod.CORESET, n_samples=100)
```
philippmwirth, line 27: Just remove the line as it's the same as above.

lightly/api/api_workflow_sampling.py
```diff
@@ -55,6 +55,8 @@ def sampling(self, sampler_config: SamplerConfig, al_scores: Dict[str, List[np.n
             if preselected_tag_id is None:
                 raise ValueError
             for score_type, score_values in al_scores.items():
+                if isinstance(score_values, np.ndarray):
+                    score_values = score_values.astype(np.float64)
```
philippmwirth, line 59: As a general question, would it not be more efficient to send float32 instead of float64 to the api?

tests/api_workflow/mocked_api_workflow_client.py
```diff
@@ -110,6 +110,8 @@ def get_tags_by_dataset_id(self, dataset_id, **kwargs):
 class MockedScoresApi(ScoresApi):
     def create_or_update_active_learning_score_by_tag_id(self, body, dataset_id, tag_id, **kwargs) -> \
             CreateEntityResponse:
+        if len(body.scores) > 0 and not isinstance(body.scores[0], float):
```
philippmwirth, line 113: How is this an `AttributeError`? Shouldn't it be a `ValueError` if the scores are not of type float?

philippmwirth (changes requested):

tests/api_workflow/mocked_api_workflow_client.py
philippmwirth: or a `TypeError`

lightly/api/api_workflow_sampling.py
MalteEbner: `float32` is not supported by th generated openapi REST client. However, we could do the parsing to a string ourselves and pass a string to be uploaded.

tests/api_workflow/mocked_api_workflow_client.py
MalteEbner: I agree, it is not an `AttributeError`. However, I threw the attribute error to mimic the behaviour of the generated client, which throughs an `AttributeError` if a `float32` is passed, as it tries to access the non-existing `obj.swagger_types` of it.

lightly/api/api_workflow_sampling.py
philippmwirth: ok, then let's go with the current solution

tests/api_workflow/mocked_api_workflow_client.py
philippmwirth: ok

philippmwirth (approved):

PR #296, opened 2021-04-16
download only newly added samples

### LINKED ISSUE #286: User can download the newly added samples for labeling after performing an AL iteration
When using lightly for active learning the user should be able to download the samples for labeling in an easy way.

Depends on https://github.com/lightly-ai/lightly-core/issues/445

MalteEbner:
close #286

## Changes:
- adapted bitmask to support `difference()`, `subset_a_list()`, `from_length()`, `__sub__()` and `__eq__()`
- `ActiveLearningAgent` now also computes the `added_set` and computes the set more efficiently with bitmasks
- the CLI `lightly-download` now has the flag `exclude_parent_tag`
- the `lightly-download(cfg)` function is unittested. The mocking for these tests made it necessary to rename the file `download_cli.py` to `download_cli_file.py` to be able to mock via `lightly.cli.download_cli_file.ApiWorkflowClient = MockedApiWorkflowClient`

## How to test:
- All python code changes are tested via unittests
- The new `lightly-download --help` can be called without errors in the GitHub action (needed adapting the `setup.py`).
- How to test the new `lightly-download` by hand:
```bash
# start the local server in lightly-core/appv2
npm run start:app-api
# test the new CLI
export BRANCH_NAME="286-tag-only-newly-added-samples-me" 
export LIGHTLY_SERVER_LOCATION="http://localhost:5000"
lightly-download tag_name='my-tag' dataset_id='your_dataset_id' token='your_access_token' exclude_parent_tag=True
```

- The manual tests worked successfully on my side, I checked the filenames in the download files for the tag with parent, tag without parent and parent and the three files fit.


lightly/active_learning/agents/agent.py
```diff
@@ -76,28 +76,29 @@ def _set_labeled_and_unlabeled_set(self, preselected_tag_data: TagData = None):
                 optional param, then it must not be loaded from the API
 
         """
-        if self.preselected_tag_id is None:
-            self.labeled_set = []
-        else:
+        self.bitmask_labeled_set = BitMask.from_hex("0x0")
```
philippmwirth, line 79: Maybe add a comment here what a bitmask from "0x0" means

lightly/active_learning/agents/agent.py
```diff
@@ -76,28 +76,29 @@ def _set_labeled_and_unlabeled_set(self, preselected_tag_data: TagData = None):
...
+        self.bitmask_added_set = BitMask.from_hex("0x0")
+        if self.preselected_tag_id is not None:
```
philippmwirth, line 81: Also a comment here please describing what happens if the preselected tag id is not None

lightly/active_learning/agents/agent.py
```diff
@@ -110,7 +111,7 @@ def query(self, sampler_config: SamplerConfig, al_scorer: Scorer = None) -> List
                 An instance of a class inheriting from Scorer, e.g. a ClassificationScorer.
 
         Returns:
-            The filenames of the samples in the new labeled_set.
+            The filenames of the samples in the new labeled_set and the added filenames.
```
philippmwirth, line 114: This description is too minimal, please elaborate. Remember, this is not internal but what's displayed in the documentation so ideally, a user would read this and understand immediately what the returned lists are.

lightly/active_learning/agents/agent.py
```diff
@@ -119,29 +120,29 @@ def query(self, sampler_config: SamplerConfig, al_scorer: Scorer = None) -> List
                            "including the current labeled set "
                            "(sampler_config.n_samples) "
                             "is smaller than the number of samples in the current labeled set.")
```
philippmwirth, line 122: We should add to this warning that the sampler will skip and just return the already labeled set.

lightly/active_learning/agents/agent.py
```diff
@@ -119,29 +120,29 @@ def query(self, sampler_config: SamplerConfig, al_scorer: Scorer = None) -> List
...
-            scores_dict = al_scorer._calculate_scores()
+
         else:
-            scores_dict = None
+            # calculate scores
+            if al_scorer is not None:
+                no_unlabeled_samples = len(self.unlabeled_set)
+                no_samples_with_predictions = len(al_scorer.model_output)
+                if no_unlabeled_samples != no_samples_with_predictions:
+                    raise ValueError(f"The scorer must have exactly as much samples as in the unlabeled set,"
```
philippmwirth, line 130: "as many"

lightly/active_learning/agents/agent.py
```diff
@@ -119,29 +120,29 @@ def query(self, sampler_config: SamplerConfig, al_scorer: Scorer = None) -> List
...
+                                     f"but there are {no_samples_with_predictions} predictions in the scorer,"
+                                     f"but {no_unlabeled_samples} in the unlabeled set.")
```
philippmwirth, line 132: "and" instead of but and remove the comma in the line above

lightly/active_learning/agents/agent.py
philippmwirth, line 130: Could you also add a test for this value error?

lightly/api/bitmask.py
```diff
@@ -68,6 +68,12 @@ def _intersection(x: int, y: int) -> int:
     return x & y
 
 
+def _difference(x: int, y: int) -> int:
+    """Uses difference to get the intersection of the two masks.
```
philippmwirth, line 72: Not the intersection but the difference

lightly/api/bitmask.py
```diff
@@ -173,6 +186,34 @@ def intersection(self, other):
         """
         self.x = _intersection(self.x, other.x)
 
+    def difference(self, other):
+        """Calculates the difference of two bit masks.
+        Examples:
+            >>> mask1 = BitMask.from_bin('0b0111')
+            >>> mask2 = BitMask.from_bin('0b1100')
+            >>> mask1.difference(mask2)
+            >>> # mask1.binstring is '0b0011'
```
philippmwirth, line 195: This implementation is highly buggy. I quickly found two breaking examples.

Example 1
```
x = 7  (0b0111)
y = 12 (0b1100)
expected result: 0b0011
actual result: -0b0101 ( = bin(-5))
``` 

Example 2
```
x = 23 (0b10111)
y = 12 (0b01100)
expected result: 0b10011
actual result: 0b01011
```

I am pretty sure that the difference of the representing integers is not the way to get the difference of the two sets. In my opinion, you should simply mask all bits in `x` which are set in `y` so it would be something like:
```
return _union(x, y) - y
```

(the above code worked for my examples)

lightly/api/bitmask.py
```diff
@@ -173,6 +186,34 @@ def intersection(self, other):
...
+        """
+        self.x = _difference(self.x, other.x)
+
+    def __sub__(self, other):
+        return BitMask(self.x - other.x)
```
philippmwirth, line 200: Same problem here as in `difference`. Also, please re-use existing function, if you implement difference above then use it here (or at least use the `_difference` function)

lightly/api/bitmask.py
```diff
@@ -173,6 +186,34 @@ def intersection(self, other):
...
+
+    def __eq__(self, other):
+        return self.to_bin() == other.to_bin()
+
+    def subset_a_list(self, list_: List):
```
philippmwirth, line 205: I would rename this function to something like `masked_select` or `masked_select_from_list`

philippmwirth (changes requested):
I like the changes you made but there are some problems, please take a look at my remarks.

About renaming the `download_cli` file: Are you 100% sure there's no other way? Because if there is, I'd prefer to keep the name of the file the same. If not, you will have to update the documentation as well, see:
- https://github.com/lightly-ai/lightly/blob/develop/docs/source/lightly.cli.rst

lightly/active_learning/agents/agent.py
```diff
@@ -111,38 +114,43 @@ def query(self, sampler_config: SamplerConfig, al_scorer: Scorer = None) -> Tupl
...
             warnings.warn("ActiveLearningAgent.query: The number of samples which should be sampled "
-                           "including the current labeled set "
-                           "(sampler_config.n_samples) "
-                            "is smaller than the number of samples in the current labeled set.")
-
+                          "including the current labeled set "
+                          "(sampler_config.n_samples) "
+                          "is smaller than the number of samples in the current labeled set."
+                          "Skipping the sampling and returning the old labeled_set and"
+                          "no ne filenames.")
```
philippmwirth, line 130: I'd prefer "Skipping the sampling and returning the previous labeled set"

philippmwirth: This would als get rid of the typo in line 130.

philippmwirth (approved):

PR #298, opened 2021-04-19
Bugfix: lightly-upload now uses loader.num_workers

### LINKED ISSUE #295: loader.num_workers=12 doesn't actually spawn multiple workers
I've noticed that when trying to create an embedding with the param `loader.num_workers` there seems to be no difference in CPU consumption;
`lightly-embed input_dir='./75000' trainer.max_epochs=1 loader.num_workers=12`

This could be that my images are ~8kb and there is some other bottleneck preventing the CPUs to max-out? in My ActivityMonitor (macOSX) I see that I have 80% idle CPU

System: MacBook Pro (16-inch, 2019), 2.6 GHz 6-Core Intel Core i7
OS: MacOSX 11.2.3

MalteEbner:
close #295 

## Changes:
- bugfix itself: look at code, very straightforward
- testing it: adapted the unmocked `test_api` function, it works

## How to test:
- run the new `test_api`function and see how 9 workers are used
- directly with CLI:
```bash
export BRANCH_NAME="295-lightly-upload-always-uses-8-workers-me" 
export LIGHTLY_SERVER_LOCATION="http://localhost:5000"
# create a new dataset on production and copy the CLI command, e.g.
lightly-upload token='your_token' dataset_id='your_dataset_id' input_dir='path_to_your_dataset' loader.num_workers=5
```

### DIFF
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -52,7 +52,7 @@ def _upload_cli(cfg, is_cli_call=True):
         mode = cfg['upload']
         dataset = LightlyDataset(input_dir=input_dir, transform=transform)
         api_workflow_client.upload_dataset(
-            input=dataset, mode=mode
+            input=dataset, mode=mode, max_workers=cfg['loader']['num_workers']
         )
 
     if path_to_embeddings:
--- a/lightly/api/api_workflow_upload_dataset.py
+++ b/lightly/api/api_workflow_upload_dataset.py
@@ -61,6 +61,7 @@ class _UploadDatasetMixin:
 
         # handle the case where len(dataset) < max_workers
         max_workers = min(len(dataset), max_workers)
+        max_workers = max(max_workers, 1)
 
         # upload the samples
         if verbose:
--- a/tests/UNMOCKED_end2end_tests/test_api.py
+++ b/tests/UNMOCKED_end2end_tests/test_api.py
@@ -4,6 +4,9 @@ import sys
 from typing import List, Tuple
 
 import numpy as np
+from hydra.experimental import initialize, compose
+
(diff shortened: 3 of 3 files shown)


lightly/cli/upload_cli.py
```diff
@@ -52,7 +52,7 @@ def _upload_cli(cfg, is_cli_call=True):
         mode = cfg['upload']
         dataset = LightlyDataset(input_dir=input_dir, transform=transform)
         api_workflow_client.upload_dataset(
-            input=dataset, mode=mode
+            input=dataset, mode=mode, max_workers=cfg['loader']['num_workers']
```
philippmwirth, line 55: Careful, we also do concurrent uploads, see [here](https://github.com/lightly-ai/lightly/blob/fc50052f9b2dc62dad9e4823783a471c17c58831/lightly/api/api_workflow_upload_dataset.py#L96). Did you test this?

MalteEbner: I do not get why to be careful, wasn't it intended to do it like this? Before, always 8 workers were used, see the issue.
Yes I tested it in both ways described in **How to test**

philippmwirth: Ah yes, my fault. I thought you had added it to the constructor of a `LightlyDataset`.

philippmwirth: Just safeguard it for `0` as this will break the `ThreadPoolExecutor` and is currently the default argument I believe and then I'll approve it :)

MalteEbner: I safeguarded it for <1.
The default argument was 1, so this was no problem.

philippmwirth (approved):

PR #301, opened 2021-04-19
renamed scorer.calculate_scores()

### LINKED ISSUE #300: Make private method _calculate_scores() of scorers public
It would help a lot to make the `_calculate_scores()` method public. The scores could be helpful for quick debugging and visualizing of samples where the model performs poorly without having to upload the dataset to the Lightly platform.

- make the private method public and check docstrings
- update documentation

MalteEbner:
close #300 

The renaming was done with IDE tools to all files at once, thus it should not be forgotten anywhere.

### DIFF
--- a/lightly/active_learning/scorers/scorer.py
+++ b/lightly/active_learning/scorers/scorer.py
@@ -9,12 +9,6 @@ class Scorer():
         self.model_output = model_output
 
     def calculate_scores(self) -> Dict[str, np.ndarray]:
-        """Calculates and returns the active learning scores
-
-        Which scores are calculated depends on the implementation
-        of this parent class by the child classes.
-        Returns:
-            A dictionary mapping from the score name (as string)
-            to the scores (as a single-dimensional numpy array).
+        """Calculates and returns active learning scores in a dictionary.
         """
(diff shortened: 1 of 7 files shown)


lightly/active_learning/scorers/scorer.py
```diff
@@ -8,5 +8,13 @@ class Scorer():
     def __init__(self, model_output):
         self.model_output = model_output
 
-    def _calculate_scores(self) -> Dict[str, np.ndarray]:
+    def calculate_scores(self) -> Dict[str, np.ndarray]:
+        """Calculates and returns the active learning scores
+
+        Which scores are calculated depends on the implementation
+        of this parent class by the child classes.
+        Returns:
```
philippmwirth, line 16: This function does not return anything so I'd remove the second part of your comment. IMO the following suffices:

"Calculates and returns active learning scores in a dictionary."

philippmwirth (commented):
Looks good, thanks for updating the docs. I left one comment about a docstring.

Can you run a `grep` with `_calculate_scores` to make sure you got all of them?

philippmwirth (approved):

PR #318, opened 2021-04-22
update scorers and documentation

### LINKED ISSUE #290: Update scorers and documentation
### Goal

Our implementations of scorers match with commonly used ones. I would follow the conventions from the papers. There are also good descriptions [here](https://modal-python.readthedocs.io/en/latest/content/query_strategies/uncertainty_sampling.html) or [here](https://towardsdatascience.com/uncertainty-sampling-cheatsheet-ec57bc067c0b). and from a [academic active learning literature survey](http://burrsettles.com/pub/settles.activelearning.pdf)
It should be very transparent how the scores are computed and how the different scorers work.

Make sure we have the following for the existing samplers (uncertainty and margin):

- [ ] implementation matches "commonly used standards"
- [ ] we have references to the papers in the docs if possible
- [ ] we describe the algorithms in the docs

MalteEbner:
closes #290

## Description
- [ ] My change is breaking
- updated the names of the scorers for uncertainty sampling for classification to follow the naming in literature
- added the new classification scorer: margin scorer, the scorer called margin beforehand was the least confidence scorer in reality. 
- added the classification scorers for detection, computing the classification scores on the classification of a detected object. Then the maximum of all these scores over an image is taken.
- updated tests and documentation.

## Tests
- [ ] I have manually tested the change. if_yes_describe_how

## Documentation
- [x] My change requires a change to the documentation ( `.rst` files).


docs/source/getting_started/active_learning.rst
```diff
@@ -180,13 +180,26 @@ Image Classification
...
-- **prediction-margin** uses the margin between 1.0 and the highest confidence 
-  prediction. Use this scorer to select images where the model is insecure.
+Currently we offer three uncertainty scorers,which are taken from
+http://burrsettles.com/pub/settles.activelearning.pdf, Section 3.1, page 12f
+and also explained in https://towardsdatascience.com/uncertainty-sampling-cheatsheet-ec57bc067c0b
+They all have in common, that the score is highest if all classes have the
+same confidence and are 0 if the model assigns 100% probability to a single class.
+The differ in the number of class confidences they take into account.
+
+- **uncertainty_least_confidence**:
```
IgorSusmelj, line 191: Why not just call it `least_confidence`?
I think that is precise enough and simpler to remember.

docs/source/getting_started/active_learning.rst
```diff
@@ -180,13 +180,26 @@ Image Classification
...
+    This score is 1 - the highest confidence prediction. It is high
+    when the confidence about the most probable class is low.
 
-- **prediction-entropy** computes the entropy of the prediction. 
-  All confidences are considered to compute the entropy of a sample.
+- **uncertainty_margin**
```
IgorSusmelj, line 195: maybe just `margin`?

docs/source/getting_started/active_learning.rst
```diff
@@ -180,13 +180,26 @@ Image Classification
...
+    This score is 1 - the margin between the highest confidence
+    and second highest confidence prediction. It is high when the model
+    cannot decide between the two most probable classes.
+
+- **uncertainty_entropy**
```
IgorSusmelj, line 200: maybe just `entropy`?

docs/source/getting_started/active_learning.rst
```diff
@@ -199,13 +212,27 @@ boxes.
...
-  perception in autonomous driving.
-
-- **prediction-margin** uses the margin between 1.0 and the mean of the highest 
-  confidence prediction. Use this scorer to select images where the model is insecure.
+- **object-frequency**
+  This score measures the number of objects in the image. Use this scorer if
+  you want scenes with lots of objects in them. This is suited for computer vision
+  tasks such as perception in autonomous driving.
+
+- **prediction-margin**
```
IgorSusmelj, line 220: maybe just `margin`?

docs/source/getting_started/active_learning.rst
```diff
@@ -199,13 +212,27 @@ boxes.
...
+  This score is 1 - the mean of the highest confidence prediction. Use this scorer
+  to select images where the model is insecure about both whether it found an object
+  at all and the class of the object.
+
+- **classification_scores**
```
IgorSusmelj, line 225: Wouldn't it make sense to postpone this? The goal was to simplify the setup and make it more transparent how the scorers work. This is a new scorer. I think it's a great idea but maybe it makes more sense to first work on the ones we have?

docs/source/getting_started/active_learning.rst
```diff
@@ -180,13 +180,26 @@ Image Classification
...
 Use this scorer when working on a classification problem (binary or multiclass).
 
-Currently, the following scorers are available:
 
-- **prediction-margin** uses the margin between 1.0 and the highest confidence 
-  prediction. Use this scorer to select images where the model is insecure.
+Currently we offer three uncertainty scorers,which are taken from
+http://burrsettles.com/pub/settles.activelearning.pdf, Section 3.1, page 12f
+and also explained in https://towardsdatascience.com/uncertainty-sampling-cheatsheet-ec57bc067c0b
+They all have in common, that the score is highest if all classes have the
```
IgorSusmelj, line 187: Max value is 1, no? So we could say: ...that the score is **1.0** if all...

lightly/active_learning/scorers/classification.py
```diff
@@ -25,20 +25,48 @@ def _entropy(probs: np.ndarray, axis: int = 1) -> np.ndarray:
...
     Currently supports the following scorers:
 
-        `prediction-margin`:
-            This scorer uses the margin between 1.0 and the highest confidence
-            prediction. Use this scorer to select images where the model is
-            insecure.
+        The following three uncertainty scores are taken from
+        http://burrsettles.com/pub/settles.activelearning.pdf, Section 3.1, page 12f
+        and also explained in https://towardsdatascience.com/uncertainty-sampling-cheatsheet-ec57bc067c0b
+        They all have in common, that the score is highest if all classes have the
```
IgorSusmelj, line 54: Would also here consider replacing `highest` with `1.0`

IgorSusmelj (changes requested):
Really nice changes. I like the new setup of the scorers which makes them more scalable :)
I left a few comments mainly regarding the naming. Would also consider moving the new sampler to another issue since this pull request should be all about simplification and transparency and not adding more scorers.

MalteEbner:
> I left a few comments mainly regarding the naming. 
I would definitely keep the prefix `uncertainty_` for three reasons:
- Uncertainty Sampling is the far most common and well-known term. All three scorers are basically three different variants of uncertainty sampling and have very similar properties, thus I would have similar names for them.
- The terms "least confidence" and especially "entropy" and "margin" are by far not precise enough to tell which score is used. In fact, there are other active learning strategies using "margin" in their name, e.g. sampling within the margin of a SVM. Entropy can be computed for many different things, e.g. one could also have the expected increase of the entropy of the class distribution over all samples as a score, which allows balancing the dataset and chooses completely different samples than when using the `uncertainty_entropy` score.
- The term "uncertainty" already tells a lot about the sampler, which helps users get a feeling for what a sampler does without needing to look it up. You know directly that the score uses the class predictions as a basis, is highest if the model is uncertain which class it is and lowest if the model is fairly certain. The term "margin" does not transport this knowledge at all.

> Max value is 1, no? So we could say: ...that the score is 1.0 if all...
The ranges for the scores are the following:
uncertainty_least_confidence: [0, 1 - 1 / num_classes]
uncertainty_margin: [0, 1]
uncertainty_entropy: [0, log(num_classes)]
I agree that I can write these ranges down. However, I would say that the exact height of the values is not that important, but rather the qualitative relationship between the prediction vector and the score, e.g. when is the score highest.

> Would also consider moving the new sampler to another issue since this pull request should be all about simplification and transparency and not adding more scorers.
I agree that I should have put that into a separate issue, sorry for that fault of mine. 
Should I still move it to another issue? I expect the separation of the code to another branch would take me 20-30 minutes of manual work, especially because e.g. the changes to the documentation are in the same file. Do you think this is worth it?

IgorSusmelj:
> The ranges for the scores are the following:
> uncertainty_least_confidence: [0, 1 - 1 / num_classes]
> uncertainty_margin: [0, 1]
> uncertainty_entropy: [0, log(num_classes)]

Great summary. Thanks, Malte. Let's keep the uncertainty_ prefix :)

Regarding the score values/ ranges. I would like to enforce numbers between 0 and 1. It makes it easier for the front end and the sampler. Otherwise, we will have to normalize them always on the fly and must not forget it.

lightly/active_learning/scorers/classification.py
```diff
@@ -64,25 +92,73 @@ class ScorerClassification(Scorer):
...
+        for score, score_name in scores_with_names:
+            score = np.nan_to_num(score)
+            scores[score_name] = score
+
+        if normalize_to_0_1:
+            scores = self.normalize_scores_0_1(scores)
+
+        return scores
+
+    def normalize_scores_0_1(self, scores: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
```
IgorSusmelj, line 138: I would like to investigate an option to only rescaling the max. E.g. making sure that max value is 1.0. But the min value could be 0.0 or 0.2 etc. 

I would make another (exploratory) issue for this. Do you think that makes sense @MalteEbner?

lightly/active_learning/scorers/classification.py
```diff
@@ -64,25 +92,73 @@ class ScorerClassification(Scorer):
...
+            scores[score_name] = np.interp(scores[score_name], interp_xp, interp_fp)
+
         return scores
 
-    def _get_prediction_margin_score(self):
-        uncertainties = np.array([1 - max(class_probabilities) for class_probabilities in self.model_output])
-        return uncertainties
+    def _get_scores_uncertainty_least_confidence(self):
+        scores = 1 - np.max(self.model_output, axis=1)
+        return scores, "uncertainty_least_confidence"
```
IgorSusmelj, line 156: Could you please make new issues for updating the database and API endpoints with the new scorer names? (also the sampling worker would need an update for the default scorer)

IgorSusmelj (approved):
Thank you for the changes. Looks very good :)

lightly/active_learning/scorers/classification.py
MalteEbner: The advantage of rescaling to the theoretical max instead of the max of the current samples is that it makes the scores comparable across different iterations.
Thus I think that there is no clearly better solution for this topic. Feel free to create an issue for outlining and discussing the different options. However, I feel that this is rather low-priority.

lightly/active_learning/scorers/classification.py
MalteEbner: We will discuss a solution in the meeting this afternoon.

PR #321, opened 2021-04-23
317 upload scores to query tag pw

### LINKED ISSUE #317: Adapt active learning score to predict for all and upload to query tag
## Description
Following our discussions around #482, we need to change two things to implement B1:
- [x] upload the scores to the query tag
- [x] compute the scores for the whole query tag, not only the unlabeled set
## Tasks
- [x] make the changes in the description
- [x] update the tutorials
- [x] update the documentation

## Acceptance Criteria
- [x] active learning tutorial still works

philippmwirth:
# 317 uploade scores to query tag

Closes #317.

- exposed `labeled_set`, `unlabeled_set`, `query_set`, and `added_set` to user
- refactored the active learning agent
- upload active learning scores to query tag instead of preselected tag
- updated the docstrings and comments
- updated the documentation and tutorial
- updated the tests

The unmocked tests and the active learning tutorial run through on `api-dev` 
@ `commit 57bb1660402d54b953f05a3bf3562bf301ed63f6 `

### DIFF
--- a/docs/source/getting_started/active_learning.rst
+++ b/docs/source/getting_started/active_learning.rst
@@ -98,18 +98,18 @@ Let's configure the sampling request and request an initial selection next:
    from lightly.active_learning.config import SamplerConfig
    from lightly.openapi_generated.swagger_client import SamplingMethod
 
-   # we want an initial pool of 100 images
-   config = SamplerConfig(n_samples=100, method=SamplingMethod.CORESET, name='initial-selection')
-   initial_selection, added_set = al_agent.query(config)
+   # we want an initial pool of 150 images
+   config = SamplerConfig(n_samples=150, method=SamplingMethod.CORESET, name='initial-selection')
+   al_agent.query(config)
+   initial_selection = al_agent.labeled_set
    
-   # initial_selection contains now 100 filenames
-   assert len(initial_selection) == 100
-   assert len(added_set) == 100
+   # initial_selection now contains 150 filenames
+   assert len(initial_selection) == 150
 
-The query returns the list of filenames corresponding to the initial selection and the list of samples which
-were added to the selection. In the first iteration, the two lists are equal. Additionally, you
-will find that a tag has been created in the web-app under the name "initial-selection".
-Head there to scroll through the samples and download the selected images before annotating them.
+The result of the query is a tag in the web-app under the name "initial-selection". The tag contains
+the images which were selected by the sampling algorithm. Head there to scroll through the samples and
+download the selected images before annotating them. Alternatively, you can access the filenames
+of the selected images via the attribute `labeled_set` as shown above.
 
 
 Active Learning Step
@@ -161,16 +161,20 @@ here is that the argument `n_samples` always refers to the total size of the lab
 
 .. code-block:: Python
 
-   # we want a total of 200 images after the first iteration
+   # we want a total of 200 images after the first iteration (50 new samples)
    # this time, we use the CORAL sampler and provide a scorer to the query
    config = SamplerConfig(n_samples=200, method=SamplingMethod.CORAL, name='al-iteration-1')
-   labeled_set_iteration_1, added_set = al_agent.query(sampler_config, scorer)
+   al_agent.query(sampler_config, scorer)
+
+   labeled_set_iteration_1 = al_agent.labeled_set
+   added_set_iteration_1 = al_agent.added_set
 
    assert len(labeled_set_iteration_1) == 200
-   assert len(added_set) == 100
+   assert len(added_set_iteration_1) == 50
 
-As before, you will receive the filenames of all the images in the labeled set and the filenames
-of the newly added images. Additionally, there will be a new tag named `al-iteration-1` visible in the web-app.
+As before, there will be a new tag named `al-iteration-1` visible in the web-app. Additionally, 
+you can access the filenames of all the images in the labeled set and the filenames which were
+added by this query via the attributes `labeled_set` and `added_set` respectively.
 You can repeat the active learning step until the model achieves the required accuracy.
 
 Scorers
--- a/docs/source/getting_started/active_learning.rst
+++ b/docs/source/getting_started/active_learning.rst
@@ -71,7 +71,7 @@ dataset_id and token.
    # use trainer.max_epochs=0 to skip training
    lightly-magic input_dir='path/to/your/raw/dataset' dataset_id='xyz' token='123' trainer.max_epochs=0
 
-Next, you will need to initialize the `ApiWorkflowClient` and the `ActiveLearningAgent`
+Then, in your Python script, you will need to initialize the `ApiWorkflowClient` and the `ActiveLearningAgent`
 
 .. code-block:: Python
 
@@ -89,7 +89,7 @@ Next, you will need to initialize the `ApiWorkflowClient` and the `ActiveLearnin
    it could be that a large portion of the images is blurry. In that case, it's 
    possible to create a tag in the web-app which only contains the sharp images
    and tell the `ActiveLearningAgent` to only sample from this tag. To do so, set
-   the `query_tag_name` argument in the constructor.
+   the `query_tag_name` argument in the constructor of the agent.
 
 Let's configure the sampling request and request an initial selection next:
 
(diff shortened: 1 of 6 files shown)


docs/source/getting_started/active_learning.rst
```diff
@@ -100,11 +100,14 @@ Let's configure the sampling request and request an initial selection next:
 
    # we want an initial pool of 100 images
    config = SamplerConfig(n_samples=100, method=SamplingMethod.CORESET, name='initial-selection')
-   initial_selection = al_agent.query(config)
+   initial_selection, added_set = al_agent.query(config)
```
IgorSusmelj, line 103: I don't like this example here. Would not use 100 for both. This just introduces confusion. Maybe start with 200 images and then add another 100. I also don't get the difference between initial selection and added set here

IgorSusmelj (changes requested):
Tbh I'm a bit confused. I thought the idea is to just switch to where we upload the scores. This PR is changing quite a bit of the whole concept. We return a tuple? Why? We use sets instead of tag_ids? Why?
Could you please elaborate?

lightly/active_learning/agents/agent.py
```diff
@@ -34,78 +38,142 @@ class ActiveLearningAgent:
...
         >>> sampler_config = SamplerConfig(n_samples=200, name='second-set')
-        >>> second_set = agent.query(sampler_config, scorer)
+        >>> second_set, added_set = agent.query(sampler_config, scorer)
 
     """
 
-    def __init__(self, api_workflow_client: ApiWorkflowClient, query_tag_name: str = None,
+    def __init__(self,
+                 api_workflow_client: ApiWorkflowClient,
+                 query_tag_name: str = None,
```
MalteEbner, line 60: We could make it even easier and cleaner with `query_tag_name: str = "initial-tag"`. Then no checks are necessary if the `query_tag_name is None` and it is documented in code to which tag the query tag defaults.

philippmwirth:
> Tbh I'm a bit confused. I thought the idea is to just switch to where we upload the scores. This PR is changing quite a bit of the whole concept. We return a tuple? Why? We use sets instead of tag_ids? Why?
> Could you please elaborate?

The only conecptual change is that we upload the scores to the query tag.

- We always returned a tuple - the docs just didn't reflect this. See
https://github.com/lightly-ai/lightly/blob/7314e93077d11fc905c4a5152ad2ab4663122b22/lightly/active_learning/agents/agent.py#L155
- Where do we use sets instead of tag ids? We always had helpers to get the `labeled_set` and `unlabeled_set`, see 
https://github.com/lightly-ai/lightly/blob/7314e93077d11fc905c4a5152ad2ab4663122b22/lightly/active_learning/agents/agent.py#L69 I just refactored it and added the `query_set` so that users can easily upload predictions for the whole query set instead of the unlabeled set. The refactoring makes the code much easier to understand IMO.
- If you check out the changes in the tutorial or the unmocked test script, you can see that the only change from a user perspective is that the predictions should now be made on the `query_set` instead of the `unlabeled_set`.

Can you elaborate what your confusion is about?

MalteEbner:
> Tbh I'm a bit confused. I thought the idea is to just switch to where we upload the scores. This PR is changing quite a bit of the whole concept. We return a tuple? Why? We use sets instead of tag_ids? Why?
> Could you please elaborate?

We also switched that the scores are not computed for the `unlabeled_set` anymore, but for the whole `query_set`, thus the large number of changes in the active learning agent.

Furthermore, the functions in the active learning agent to compute the labeled_set, unlabeled_set, added_set and query_set out of the respective bitmasks became quite a lot and not that easily understandable, thus Philipp refactored that part to a much nicer structure.

I agree that we could rethink whether we really want to return a tuple. Now I would prefer to just return nothing, as the returned values are also easily accessible as attributes. It would make it unnecessary for the user to look up, what the returned values are.

docs/source/getting_started/active_learning.rst
philippmwirth: There is no difference in the initial step which is a little inelegant. Should we just go back to having a single return value? @MalteEbner thoughts?

lightly/active_learning/agents/agent.py
```diff
@@ -120,36 +188,40 @@ def query(self, sampler_config: SamplerConfig, al_scorer: Scorer = None) -> Tupl
...
+            query_tag_id=self._query_tag_id
+        )
+
+        # update the old preselected_tag
+        self._old_preselected_tag_bitmask = self._preselected_tag_bitmask
+        # set the newly chosen tag as the new preselected_tag
+        self._preselected_tag_id = new_tag_data.id
+        self._preselected_tag_bitmask = self._get_preselected_tag_bitmask()
 
         return self.labeled_set, self.added_set
```
MalteEbner, line 227: Perhaps we should just return nothing and let the user access the sets by him/herself. Then no documentation for the return values is need.
It is also a bit unlogical, that the query_set, which is needed for predicting the new scores, is not returned.

MalteEbner (approved):
I love your refactoring of the setting of the sets in the ActiveLearningAgent. I did not like the function `_set_labeled_and_unlabeled_set` I had implemented, but did not see an elegant way how to make it better. Now you found one which is great!
You also added further tests, which I always like :)

The comments I made are minor propositions, feel free to follow them or not.

docs/source/getting_started/active_learning.rst
IgorSusmelj: Why should I get a tuple? That does not really make sense to me. I query --> want to get back what I specified in the query

lightly/active_learning/agents/agent.py
IgorSusmelj: Yes, could do that.

docs/source/getting_started/active_learning.rst
MalteEbner: I would go to have no return value at all, see my comments below.

IgorSusmelj:
> > Tbh I'm a bit confused. I thought the idea is to just switch to where we upload the scores. This PR is changing quite a bit of the whole concept. We return a tuple? Why? We use sets instead of tag_ids? Why?
> > Could you please elaborate?
> 
> The only conecptual change is that we upload the scores to the query tag.
> 
> * We always returned a tuple - the docs just didn't reflect this. See
>   https://github.com/lightly-ai/lightly/blob/7314e93077d11fc905c4a5152ad2ab4663122b22/lightly/active_learning/agents/agent.py#L155
> * Where do we use sets instead of tag ids? We always had helpers to get the `labeled_set` and `unlabeled_set`, see
>   https://github.com/lightly-ai/lightly/blob/7314e93077d11fc905c4a5152ad2ab4663122b22/lightly/active_learning/agents/agent.py#L69
>    I just refactored it and added the `query_set` so that users can easily upload predictions for the whole query set instead of the unlabeled set. The refactoring makes the code much easier to understand IMO.
> * If you check out the changes in the tutorial or the unmocked test script, you can see that the only change from a user perspective is that the predictions should now be made on the `query_set` instead of the `unlabeled_set`.
> 
> Can you elaborate what your confusion is about?

Let me test the PR first properly myself with the tutorial.

IgorSusmelj:
The other changes just confused me as I didn't expect any change besides where we store the scores.

Let's take the opportunity and streamline the workflow. There should be 1 way to do things. As Malte suggested, we could also drop the return value and streamline that the user just accesses the attribute of the al agent. I like that idea as it enforces consistency and simplifies the workflow.

IgorSusmelj:
The tutorials runs through also on my machine 👯

IgorSusmelj:
And thanks for all the clarifications

lightly/active_learning/agents/agent.py
```diff
@@ -14,14 +14,18 @@ class ActiveLearningAgent:
...
-            The filenames of the samples in the labeled set, List[str]
+            Set of filenames corresponding to samples in the labeled set.
+            Set to all samples in the preselected tag or to an empty list by default.
         unlabeled_set:
-            The filenames of the samples in the unlabeled set, List[str]
+            Set of filenames corresponding to samples which are in the query set
+            but not in the labeled set.
+        added_set:
+            Set of filenames corresponding to samples which were added to the 
+            labeled set in the last query.
```
MalteEbner, line 28: Optional: "added to the labeled set and removed from the unlabeled set in the last query"

lightly/active_learning/agents/agent.py
MalteEbner: @philippmwirth What do you think about that?

docs/source/getting_started/active_learning.rst
```diff
@@ -98,23 +98,26 @@ Let's configure the sampling request and request an initial selection next:
    from lightly.active_learning.config import SamplerConfig
    from lightly.openapi_generated.swagger_client import SamplingMethod
 
-   # we want an initial pool of 100 images
-   config = SamplerConfig(n_samples=100, method=SamplingMethod.CORESET, name='initial-selection')
-   initial_selection = al_agent.query(config)
+   # we want an initial pool of 150 images
+   config = SamplerConfig(n_samples=150, method=SamplingMethod.CORESET, name='initial-selection')
+   al_agent.query(config)
+   initial_selection = al_agent.labeled_set
```
IgorSusmelj, line 104: cold even drop this and directly to the print/ assert

IgorSusmelj (approved):
I like it a lot! I had some other ideas on how to improve but will summarize them in another issue :)

Issue #133, opened 2021-02-17
Pytest failed for tqdm version 4.41.1


IgorSusmelj:
When running pytest on my machine I got an error regarding the tqdm module:
```
==================================== ERRORS ====================================
______________ ERROR collecting tests/data/test_LightlyDataset.py ______________
ImportError while importing test module '/home/igor_lightly_ai/lightly/tests/data/test_LightlyDataset.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/conda/envs/lightly/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/data/test_LightlyDataset.py:12: in <module>
    from lightly.data._utils import check_images
lightly/data/_utils.py:10: in <module>
    import tqdm.contrib.concurrent as concurrent
E   ModuleNotFoundError: No module named 'tqdm.contrib.concurrent'
```

When uninstalling tqdm and installing the latest version (4.56.2) through pip it worked.

MalteEbner:
I could reproduce it, version 4.42.0 is needed at least.

PR #328, opened 2021-04-26
Reduce min tqdm version and trigger github tests for pull requests

### LINKED ISSUE #327: Reduce tqdm min version in requirements.txt
Google Colab and some other environments have tqdm 4.41 installed by default. Since there is no need to use 4.44 I would suggest reducing the minimum required version.

Is there a particular reason we use 4.44?

IgorSusmelj:
### Changes

closes #327 

- decrease minimum required tqdm version from 4.44 to 4.41 (4.41 seems to be the default on some systems and lightly just runs fine with it)
- make sure we trigger the github actions (tests) for pull requests too. This makes sure external contributors creating a pull request trigger the tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,6 @@
 name: Unit Tests
 
-on: [push, workflow_dispatch]
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


IgorSusmelj:
This issue is interfering with another issue we resolved. Thanks, @MalteEbner for pointing this out: https://github.com/lightly-ai/lightly/issues/133

The GitHub actions don't install the minimum version. The tqdm version can however not be reduce to 4.41 because of the contrib concurrent module we use for parallel processing.

I will therefore close the PR and close the issue.

PR #336, opened 2021-04-28
ScorerClassification has better handling of weird model_output + tests

### LINKED ISSUE #335: AL detection scorer fails for object detection
## Error
image
## Reproduction:
`np.partition([[1.]],-2, axis=1)` yields `ValueError: kth(=-1) out of bounds (1)`
## Possible cause of error:
https://github.com/lightly-ai/lightly/blob/b064ab7f077867ab1fe4ca1a3977090701856c26/lightly/active_learning/utils/object_detection_output.py#L1[…]27
tries to do a one-hot-encoding without knowing the number of classes. Thus if max(labels) is only 1, only one class is encoded and the probability vector only has a length of 1. Thus partitioning it into 2 fails.

### 1. Prevent the error cause
Create the class_probabilities vector correctly with a given number of classes.
### 2. Handle the error (MUST DO, We all agree on it)
In the ScorerClassification handle the case where there is only a single class
### 3. General solution
The class_probabilities do not make sense at all if the ObjectDetectionOutput is created from scores. Thus we could also not create the `class_probabilities` in this case and not compute any classification scores if the `class_probabilities` are unset.

Solution 1 and 3 are exclusive, solution 2 can be combined with the other solutions.

## How to work on this issue
- First write a unittest reproducing the error by using ClassificationScorer.calculate_scores()
- Next fix this unittest with solution 2.
- Do not work on Solution 3 yet (another issue)

MalteEbner:
closes #335 

## Description
- [ ] My change is breaking
- Following cases of model_output yield following output now:
- len(model_output) == 0: No error, returns empty list for every score
- otherwise: require model_output to have 2 dimensions, with the second dimension being >= 1. If the second dimension is zero, an error is raised.
- if the second dimension == 1, it is assumed to be a binary classification problem with the `model_output` vector denoting
the probability of class 0. Thus we set `prob_class_1 =  1 - prob_class_0` and concat it to the `model_output` to get `model_output.shape == (n, 2)`.
For the case of the detection model setting  `model_output` to a one-hot encoded vector of all ones (i.e. `model_output = np.ones(shape=(n,1))` ), this does not change the values of the scores, they are still all 0.
- Wrote unittests to test different number of dimensions (a.k.a. tensor order) and different sizes for each dimension. Also wrote tests for the binary classifier with hardcoded specific values.

## Tests
- [ ] I have manually tested the change.

## Documentation
- no documentation needed

### DIFF
--- a/lightly/active_learning/scorers/classification.py
+++ b/lightly/active_learning/scorers/classification.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import *
 
 import numpy as np
@@ -93,9 +94,21 @@ class ScorerClassification(Scorer):
 
     """
     def __init__(self, model_output: Union[np.ndarray, List[List[float]]]):
-        if isinstance(model_output, List):
+        if not isinstance(model_output, np.ndarray):
             model_output = np.array(model_output)
+
+        self.ensure_model_output_is_correct(model_output)
+
         super(ScorerClassification, self).__init__(model_output)
+
+    def ensure_model_output_is_correct(self, model_output: np.ndarray):
+        if len(model_output.shape) != 2:
+            raise ValueError("ScorerClassification model_output must be a 2-dimensional array")
+        if model_output.shape[0] == 0 or model_output.shape[1] == 0:
+            raise ValueError("ScorerClassification model_output must not have empty dimensions")
+        if model_output.shape[1] == 1:
+            warnings.warn("Trying to use the ScorerClassification on a prediction vector"
+                          "with only a single class. This does not make sense.")
 
     @classmethod
     def score_names(cls) -> List[str]:
@@ -156,7 +169,10 @@ class ScorerClassification(Scorer):
         return scores, "uncertainty_least_confidence"
 
     def _get_scores_uncertainty_margin(self):
-        scores = 1 - _margin_largest_secondlargest(self.model_output)
+        if self.model_output.shape[1] == 1:
+            scores = np.zeros_like(self.model_output)
+        else:
+            scores = 1 - _margin_largest_secondlargest(self.model_output)
         return scores, "uncertainty_margin"
 
     def _get_scores_uncertainty_entropy(self):
--- a/lightly/active_learning/scorers/classification.py
+++ b/lightly/active_learning/scorers/classification.py
@@ -97,20 +97,22 @@ class ScorerClassification(Scorer):
         if not isinstance(model_output, np.ndarray):
             model_output = np.array(model_output)
 
-        self.ensure_model_output_is_correct(model_output)
+        validated_model_output = self.ensure_valid_model_output(model_output)
 
-        super(ScorerClassification, self).__init__(model_output)
+        super(ScorerClassification, self).__init__(validated_model_output)
 
-    def ensure_model_output_is_correct(self, model_output: np.ndarray):
+    def ensure_valid_model_output(self, model_output: np.ndarray) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


lightly/active_learning/scorers/classification.py
```diff
@@ -93,10 +94,24 @@ class ScorerClassification(Scorer):
...
     def __init__(self, model_output: Union[np.ndarray, List[List[float]]]):
-        if isinstance(model_output, List):
+        if not isinstance(model_output, np.ndarray):
             model_output = np.array(model_output)
+
+        self.ensure_model_output_is_correct(model_output)
+
         super(ScorerClassification, self).__init__(model_output)
 
+    def ensure_model_output_is_correct(self, model_output: np.ndarray):
```
philippmwirth, line 104: not a huge thing but I'd say typically the name of functions like this is along the lines of `ensure_valid_model_output` and not a full sentence

lightly/active_learning/scorers/classification.py
```diff
@@ -156,7 +171,10 @@ def _get_scores_uncertainty_least_confidence(self):
         return scores, "uncertainty_least_confidence"
 
     def _get_scores_uncertainty_margin(self):
-        scores = 1 - _margin_largest_secondlargest(self.model_output)
+        if self.model_output.shape[1] == 1:
+            scores = np.zeros_like(self.model_output)
+        else:
+            scores = 1 - _margin_largest_secondlargest(self.model_output)
         return scores, "uncertainty_margin"
```
philippmwirth, line 178: In the first case, the scores will have shape `n x 1` and in the second case only `n` right? I would make this consistent

lightly/active_learning/scorers/classification.py
```diff
@@ -93,10 +94,24 @@ class ScorerClassification(Scorer):
...
         super(ScorerClassification, self).__init__(model_output)
 
+    def ensure_model_output_is_correct(self, model_output: np.ndarray):
+        if len(model_output) == 0:
+            return
+        if len(model_output.shape) != 2:
+            raise ValueError("ScorerClassification model_output must be a 2-dimensional array")
+        if model_output.shape[1] == 0:
+            raise ValueError("ScorerClassification model_output must not have an empty dimension 1")
+        if model_output.shape[1] == 1:
```
philippmwirth, line 111: while this is an acceptable solution, please consult with @IgorSusmelj whether we want to add this already or only later when the case is not as likely anymore (e.g. when we handle it correctly in the detection scorer). Because right now this can happen relatively often and it's not a good look for us.

IgorSusmelj: For now, I just want the tutorials to work o.0
I'm currently running it in background. @philippmwirth or @MalteEbner, could you test the kNN tutorial as well?

MalteEbner: I am still working on reworking the AL tutorial, I just postponed it every time we changed something on the active learning side (i.e. always).

MalteEbner: Regarding the warning: I though a little more about it and removed it, as calculating the scores like this makes sense if you have a binary classification problem and thus a vector with the probability of class 0 for each sample (shape: nx1).

lightly/active_learning/scorers/classification.py
MalteEbner: Did it

philippmwirth (approved):
Looks good

Issue #338, opened 2021-04-28
Make all active learning scorers work for object detection even without class probabilities


IgorSusmelj:
The current setup does not allow active learning scorers to work when we only have labels but no class probabilities in the output of an object detection model.

One way to solve the problem is that we always multiply the objectness probability and the classification probability. If there is no classification probability we use the constant `1.0` for the multiplication. Additionally, we turn the output into a probability distribution for which the selected class has the value `p` whereas the remaining samples have the values `(1 -p) / (nclasses - 1)`

Issue #337, opened 2021-04-28
Active Learning scorer does not return scores as numpy array


IgorSusmelj:
Some of the AL scorers return Python list instead of numpy arrays:

`'uncertainty_least_confidence', 'uncertainty_margin', 'uncertainty_entropy'` return Python lists

`'object_frequency', 'objectness_least_confidence'` return numpy arrays.

PR #339, opened 2021-04-28
Fix al scores is

IgorSusmelj:
### Changes

solves #337 and #338 

- Make sure AL scores are all NumPy arrays
- Make sure AL scores are not all 0 arrays if we only have class labels

### DIFF
--- a/docs/source/getting_started/active_learning.rst
+++ b/docs/source/getting_started/active_learning.rst
@@ -224,11 +224,11 @@ We expect the model predictions to contain
 - objectness_probability for each bounding box
 - classification_probabilities for each bounding box
 
-You find more about the format here: 
+You can find more about the format here: 
 :py:class:`lightly.active_learning.utils.object_detection_output.ObjectDetectionOutput`
 
-We also provide a helper method to work with the model output format of only
-consisting of a probability per bounding box and the associcated label.
+We also provide a helper method to work with the model output format consisting 
+of only a probability per bounding box and the associated label.
 :py:class:`lightly.active_learning.utils.object_detection_output.ObjectDetectionOutput.from_scores`
 
 
--- a/docs/source/getting_started/active_learning.rst
+++ b/docs/source/getting_started/active_learning.rst
@@ -215,7 +215,22 @@ For more information about how to use the classification scorer have a look here
 Object Detection
 ^^^^^^^^^^^^^^^^^^^^^
 Use this scorer when working on an object detection problem using bounding
-boxes.
+boxes. The object detection scorers require the input to be in 
+the `ObjectDetectionOutput` format.
+
+We expect the model predictions to contain
+
+- bounding boxes of shape (x0, y0, x1, y1)
+- objectness_probability for each bounding box
+- classification_probabilities for each bounding box
+
+You find more about the format here: 
+:py:class:`lightly.active_learning.utils.object_detection_output.ObjectDetectionOutput`
+
+We also provide a helper method to work with the model output format of only
+consisting of a probability per bounding box and the associcated label.
+:py:class:`lightly.active_learning.utils.object_detection_output.ObjectDetectionOutput.from_scores`
+
 
 Currently, the following scorers are available:
 
@@ -233,9 +248,9 @@ Currently, the following scorers are available:
   These scores are computed for each object detection per image out of
   the class probability prediction for this detection. Then, they are reduced
   to one score per image by taking the maximum. In particular we support:
-  - **classification_uncertainty_least_confidence**
-  - **classification_uncertainty_margin**
-  - **classification_uncertainty_entropy**
+  - **uncertainty_least_confidence**
+  - **uncertainty_margin**
+  - **uncertainty_entropy**
   The scores are computed using the scorer for classification.
 
 
--- a/lightly/active_learning/utils/object_detection_output.py
+++ b/lightly/active_learning/utils/object_detection_output.py
@@ -132,7 +132,7 @@ class ObjectDetectionOutput:
         class_probabilities = []
         for object_prob, label in zip(object_probabilities, labels):
             num_classes = max(max_label, 1)
-            c = [(1-object_prob)/num_classes] * (num_classes + 1)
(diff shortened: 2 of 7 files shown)


lightly/active_learning/utils/object_detection_output.py
```diff
@@ -126,9 +126,9 @@ def from_scores(cls,
         # create one-hot class probabilities
         max_label = max(labels) if len(labels) > 0 else 0
         class_probabilities = []
-        for label in labels:
-            c = [0.] * (max_label + 1)
-            c[label] = 1.
+        for object_prob, label in zip(object_probabilities, labels):
+            c = [(1-object_prob)/(max_label + 1)] * (max_label + 1)
+            c[label] = object_prob
```
philippmwirth, line 131: For this to sum to one you would need to use `max_label` instead of `max_label + 1`, or am I wrong? Can you test this somehow?

Also, I want to make you aware that the number of classes detected on an image now has an influence on the score because we evaluate the number of classes dynamically. For example, for two detections with a score of `0.7` on two different images, where there are ten classes detected on the first image and two classes on the second, the difference between the most confident class probability and the second most confident class probability will be different for the two detections.

IgorSusmelj: `max_label` can be `0`. I just rushed to get something working for all cases. I can add more tests (e.g. scores sum up to `1.0`. And yes, the scores now depend on the number of detected classes within images.

MalteEbner: This will make the the scorers "uncertainty_least_confidence" and "objectness_least_confidence" output exactly the same values, thus the advantages is zero.
If get the point of faking the tutorial, but having the uncertainty scores without probability vectors makes 0 sense from an active learning point of view. Thus I strongly oppose putting such fake distributions in our production code.

philippmwirth: 1. That's not true because the number of classes varies for each image (which is a problem on its own)
2. It's not faking. If the user can't provide probabilities we need to do the best we can.
3. We need to think about these kinds of problems before we implement major changes. Now we only have two options. Either we merge this fix and release today or we postpone the release and implement all the infrastructure need so that we can go past using default scores and have the user decide which score to use.

MalteEbner:
The active learning scores for the `ScorerClassification` might still be Lists. If you want to fix that, adapt the unittests to ensure they are failing, then fix the following line.
https://github.com/lightly-ai/lightly/blob/a5d4bb6fa5ca3b63f82d1d7f64b78a197e095a62/lightly/active_learning/scorers/classification.py#L137

lightly/active_learning/utils/object_detection_output.py
MalteEbner: This is not about "making it work" but about "faking it to make it seem as if works" which should not be part of the production code.

IgorSusmelj: This is not about faking or not. It's about allowing the user to use our AL workflow when has no class probabilities but only labels. We have to be transparent and tell the user that the scores are not as expected when using only labels and not class probabilities. I would also add to the documentation that we strongly advise against using the `.from_scores(..)` method. 
If possible users should provide class probabilities. But at the same time, we want to make sure, all users can benefit from AL without a lot of extra work.

docs/source/getting_started/active_learning.rst
```diff
@@ -215,7 +215,22 @@ For more information about how to use the classification scorer have a look here
...
+boxes. The object detection scorers require the input to be in 
+the `ObjectDetectionOutput` format.
+
+We expect the model predictions to contain
+
+- bounding boxes of shape (x0, y0, x1, y1)
+- objectness_probability for each bounding box
+- classification_probabilities for each bounding box
+
+You find more about the format here:
```
philippmwirth, line 227: You **can** find more...

docs/source/getting_started/active_learning.rst
```diff
@@ -215,7 +215,22 @@ For more information about how to use the classification scorer have a look here
...
+We expect the model predictions to contain
+
+- bounding boxes of shape (x0, y0, x1, y1)
+- objectness_probability for each bounding box
+- classification_probabilities for each bounding box
+
+You find more about the format here: 
+:py:class:`lightly.active_learning.utils.object_detection_output.ObjectDetectionOutput`
+
+We also provide a helper method to work with the model output format of only
```
philippmwirth, line 230: We also provide a helper method to work with the model output format consisting of only a probability per bounding box and the associated label.

(you had the of in there twice)

docs/source/getting_started/active_learning.rst
```diff
@@ -233,9 +248,9 @@ Currently, the following scorers are available:
   These scores are computed for each object detection per image out of
   the class probability prediction for this detection. Then, they are reduced
   to one score per image by taking the maximum. In particular we support:
-  - **classification_uncertainty_least_confidence**
-  - **classification_uncertainty_margin**
-  - **classification_uncertainty_entropy**
+  - **uncertainty_least_confidence**
+  - **uncertainty_margin**
+  - **uncertainty_entropy**
```
philippmwirth, line 253: I like the naming now :)

lightly/active_learning/utils/object_detection_output.py
```diff
@@ -126,9 +130,10 @@ def from_scores(cls,
         # create one-hot class probabilities
         max_label = max(labels) if len(labels) > 0 else 0
         class_probabilities = []
-        for label in labels:
-            c = [0.] * (max_label + 1)
-            c[label] = 1.
+        for object_prob, label in zip(object_probabilities, labels):
+            num_classes = max(max_label, 1)
+            c = [(1-object_prob)/num_classes] * (num_classes + 1)
```
philippmwirth, line 135: Minor thing: Google styleguide likes spaces between mathematical operations

philippmwirth (changes requested):
Almost good, just two typos and one minor change

philippmwirth (approved):

PR #340, opened 2021-04-29
refactor NTXent loss

### LINKED ISSUE #268: Refactor NTXentLoss
# Refactor NTXentLoss

The implementation of the NTXentLoss is cumbersome and overly complicated. A simple implementation is possible using `torch.einsum`. We should refactor the code to make it easier for our users to understand what's going on.

MalteEbner:
closes #268

## Description
- [ ] My change is breaking
- For the unittests: wrote down an alternative calculation of the NTXent loss following exactly the pseudocode in the paper.
https://github.com/lightly-ai/lightly/blob/05bb9bd6c93c537763e8be8dc1f776db7ef4b793/tests/loss/test_NTXentLoss.py#L30-L31
Then it is checked whether the torch implementation and manual implementations with for-loops and numpy yield the same results.
- Deleted unused functions in the NTXent loss class
- Made the code a bit simpler and calculated the similarity matrix with Einstein summation.
- Added a higher number of comments to make it easier to understand what is going on. 

## Performance Comparison
with `out0.shape == out1.shape == (256, 64)` and 10000 repetitions

- old implementation on MBP CPU: 36s
- new implementation on MBP CPU: 20s

Flaws of evaluation: did not test with back propagation/gradient

## Tests
- [ ] I have manually tested the change. if_yes_describe_how

## Documentation

### DIFF
--- a/lightly/loss/ntx_ent_loss.py
+++ b/lightly/loss/ntx_ent_loss.py
@@ -162,7 +162,7 @@ class NTXentLoss(MemoryBankModule):
 
             if self.mask_negative_samples is None \
                     or 2 * batch_size != self.mask_negative_samples.shape[0]:
-                #
+                # if the mask_negative_samples is not correct or has the wrong size
                 self.mask_negative_samples = \
                     self._torch_get_mask_negative_samples(batch_size, device=device)
 
@@ -177,7 +177,7 @@ class NTXentLoss(MemoryBankModule):
             # sim_pos is of shape (2 * batch_size, 1)
             # sim_neg is of shape (2 * batch_size, 2 * (batch_size -1))
 
-        # The goal is to maximize sim_pos in relation sim_neg.
+        # The goal is to maximize sim_pos in relation to sim_neg.
         # This goal can be targeted with the cross-entropy-loss
 
         logits = torch.cat([sim_pos, sim_neg], dim=1)
--- a/lightly/loss/ntx_ent_loss.py
+++ b/lightly/loss/ntx_ent_loss.py
@@ -4,7 +4,6 @@
 # All Rights Reserved
 
 import torch
-import numpy as np
 
 from lightly.loss.memory_bank import MemoryBankModule
 
@@ -22,8 +21,6 @@ class NTXentLoss(MemoryBankModule):
     Attributes:
         temperature:
             Scale logits by the inverse of the temperature.
-        use_cosine_similarity:
-            Whether to use cosine similarity over L2 distance.
         memory_bank_size:
             Number of negative samples to store in the memory bank. 
             Use 0 for SimCLR. For MoCo we typically use numbers like 4096 or 65536.
@@ -55,14 +52,30 @@ class NTXentLoss(MemoryBankModule):
         super(NTXentLoss, self).__init__(size=memory_bank_size)
         self.temperature = temperature
         self.cross_entropy = torch.nn.CrossEntropyLoss(reduction="mean")
-        self.correlated_mask = None
+        self.mask_negative_samples = None
         self.eps = 1e-8
 
         if abs(self.temperature) < self.eps:
             raise ValueError('Illegal temperature: abs({}) < 1e-8'
                              .format(self.temperature))
 
-    def _torch_get_correlated_mask(self, batch_size, device=None):
+    def _torch_get_mask_negative_samples(self, batch_size, device=None):
+        """ Creates a matrix being true at values of the similarity matrix for
+        negative samples.
+
+        It creates a square matrix of size (2*batch_size, 2*batch_size)
+        being True except on three diagonals. These three diagonals are
+        the main diagonal and the diagonals of the other two quadrants.
+
+        Example for batch_size=3 with x=False and _=True:
+        x _ _ x _  _
+        _ x _ _ x  _
+        _ _ x _ _  x
+        x _ _ x _  _
+        _ x _ _ x  _
+        _ _ x _ _  x
+
+        """
         diag = torch.eye(2 * batch_size, device=device)
         diag[batch_size:, :batch_size] += torch.eye(batch_size, device=device)
         diag[:batch_size, batch_size:] += torch.eye(batch_size, device=device)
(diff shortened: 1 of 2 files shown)


lightly/loss/ntx_ent_loss.py
```diff
@@ -117,8 +94,11 @@ def forward(self,
             Args:
                 out0:
                     Output projections of the first set of transformed images.
+                    Shape: (batch_size, embedding_size)
                 out1:
                     Output projections of the second set of transformed images.
+                    Shape: (batch_size, embedding_size)
+                    out0[i] and out1[i] are two different augmentation of the same image.
```
IgorSusmelj, line 101: Would drop this line. Whether we have two views of the same image or two different views is up to the user.

lightly/loss/ntx_ent_loss.py
```diff
@@ -136,37 +115,77 @@ def forward(self,
...
+            sim_pos = torch.cat([sim_pos_out0_out1, sim_pos_out1_out0]).view(2 * batch_size, 1)
+
+            if self.mask_negative_samples is None \
+                    or 2 * batch_size != self.mask_negative_samples.shape[0]:
+                # if the mask_negative_samples is not correct or has the wrong size
+                self.mask_negative_samples = \
+                    self._torch_get_mask_negative_samples(batch_size, device=device)
+
+            sim_neg = similarity_matrix[
+                self.mask_negative_samples].view(2 * batch_size, -1)
```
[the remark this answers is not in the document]
philippmwirth: (twice so that it's symmetric)

lightly/loss/ntx_ent_loss.py
```diff
@@ -136,37 +115,77 @@ def forward(self,
...
 
-            # filter out the scores from the positive samples
-            l_pos = torch.diag(similarity_matrix, batch_size)
-            r_pos = torch.diag(similarity_matrix, -batch_size)
-            l_pos = torch.cat([l_pos, r_pos]).view(2 * batch_size, 1)
+            # sim_neg each is of shape (batch_size, memory_bank_size) and sim_neg[i,j] denotes the similarity
+            # of the i-th sample to the j-th negative sample
+            sim_neg_out0 = torch.einsum('nc,ck->nk', out0, negatives.clone().detach())
+            sim_neg_out1 = torch.einsum('nc,ck->nk', out1, negatives.clone().detach())
+            sim_neg = sim_neg_out0 + sim_neg_out1
```
philippmwirth, line 140: The only problem I see here is that now, it's a symmetrized version of the loss before. Can you add a unit test to make sure that the output is the same for the original implementation of the loss and this one in the case of memory bank? I believe you will need to remove the symmetrization again to make it work... 

Maybe FIY: The implementation of the loss with memory bank comes directly from the MoCo reference implementation :)

philippmwirth (changes requested):
I like the implementation and the extensive test suite :) 

I left some comments about things we had discussed in person. Please take a look at them.

lightly/loss/ntx_ent_loss.py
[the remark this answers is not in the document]
MalteEbner: I don't think so, as the implementation you proposed does not include the loss within out0 and out1 respectively. Thus we would need to write it four times (for each of the four quadrants). One such block only computes it for one quadrant.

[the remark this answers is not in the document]
philippmwirth: good point, didn't notice that

lightly/loss/ntx_ent_loss.py
MalteEbner: I agree and changed it back to the original version.

lightly/loss/ntx_ent_loss.py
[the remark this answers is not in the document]
MalteEbner: However, I like the idea of computing the mask directly and not in an extra function, I took that idea and changed the code accordingly.

lightly/loss/ntx_ent_loss.py
```diff
@@ -136,37 +91,76 @@ def forward(self,
...
+            mask_base = torch.eye(batch_size, dtype=torch.bool, device=out0.device)
+            mask_0 = torch.zeros_like(mask_base, device=out0.device)
+
+            # the positive samples are on the diagonals of the upper right and lower left quadrant
+            mask_pos = concat_to_2x2(mask_0, mask_base, mask_base, mask_0)
+            sim_pos = similarity_matrix[mask_pos].view(2 * batch_size, -1)
+
+            # the negative samples are everywhere except on the diagonals of the 4 quadrants
+            mask_neg = ~concat_to_2x2(mask_base, mask_base, mask_base, mask_base)
+            sim_neg = similarity_matrix[mask_neg].view(2 * batch_size, -1)
```
philippmwirth, line 143: I think this part is unnecessarily slow and complicated (e.g. you use the `concat_to_2x2` twice here).

I'd like to propose a much shorter implementation which doesn't require masks:

```python
if negatives is not None:
    negatives = negatives.to(device)
    sim_pos = torch.einsum('nc,nc->n', out0, out1).unsqueeze(-1)
    sim_neg = torch.einsum('nc,ck->nk', out0, negatives.clone().detach())
    logits = torch.cat([sim_pos, sim_neg], dim=1)
    labels = torch.zeros(logits.shape[0]).long()
else:
    output = torch.cat([out0, out1], dim=0)
    logits = torch.einsum('nc,mc->nm', output, output)
    labels = torch.arange(batch_size)
    labels = torch.cat([labels + batch_size, labels])
```

This should result in the same output as your current implementation and is probably more efficient because we don't need the masked selects. What do you think? :)

(I left away the comments in my example but you should obviously keep them)

MalteEbner: The code for `if negatives is not None` is equal to what we have now.
The code for the else part does not seem correct to me for following reasons:
1. The similarities of samples to themselves, i.e. the values on the main diagonal are also part of the calculation. In the paper, however, they weren't. https://github.com/lightly-ai/lightly/blob/05bb9bd6c93c537763e8be8dc1f776db7ef4b793/tests/loss/test_NTXentLoss.py#L55
2. Consequently, the unittests testing against the other implementation fail.

The exact implementation I used was the following:
```python
output = torch.cat([out0, out1], dim=0)
logits = torch.einsum('nc,mc->nm', output, output)/self.temperature
labels = torch.arange(batch_size)
labels = torch.cat([labels + batch_size, labels]).long()
```

MalteEbner: If you want to speed up, the mask_pos and mask_neg could be saved and reused.

philippmwirth: Hm you are right but strangely, in an experiment I did locally, the result from my calculations was the same as the one for your calculations.

philippmwirth: Why don't we just follow the SimCLR implementation?

https://github.com/google-research/simclr/blob/3ad6700c1b139ee18e43f73546b7263a710de699/objective.py#L70

IgorSusmelj:
@MalteEbner, could you run a simple benchmark and compare the performance of the new implementation against the old one in terms of speed? If you run it on CUDA please make sure you use `torch.cuda.synchronize()` to measure the execution time properly. We should make sure the new implementation is not slower than the old one :)

MalteEbner:
After thinking a bit more about the solution proposed by Philip, I found a way to remove the similarities of samples to themselves. Now it is working with only a single mask and thus less masking as before. The tests are running through with it. 

```python
# use other samples from batch as negatives
output = torch.cat((out0, out1), axis=0)

# the logits are the similarity matrix divided by the temperature
logits = torch.einsum('nc,mc->nm', output, output) / self.temperature
# We need to removed the similarities of samples to themselves
logits = logits[~torch.eye(2*batch_size, dtype=torch.bool, device=out0.device)].view(2*batch_size, -1)

# The labels point of a sample in out_i to its equivalent in out_(1-i)
labels = torch.arange(batch_size)
labels = torch.cat([labels + batch_size - 1, labels]).long()
```

IgorSusmelj (approved):
I tested it locally and it works fine. Trained models with the old code and the new one. Results match :)

philippmwirth (approved):
Love the speed-up, good work @MalteEbner :)

PR #344, opened 2021-04-29
clean up of platform tutorial 2 active learning

### LINKED ISSUE #242: Clean up of platform tutorial 2: Active Learning
# Clean up of platform tutorial 2: Active Learning

Following external feedback, the following things should be improved in the active learning tutorial.
- [ ] Expect the user to have zero setup; list requirements (scikit-learn!), show the expected folder structure (images + embeddings)
- [ ] Make it clear whether the steps are to be followed in a notebook or terminal. I.e. make differences between code and CLI clearer.
- [ ] Fix the typo in the `lightly-train` command (missing ")
Dependency: #280 
- [ ] Make it clear that the checkpoint is not in the working directory in `lightly-embed`
- [ ] Alternatively, simplify and use `lightly-magic` instead of `lightly-train` & `lightly-embed`
- [ ] Fix punctuation and be more elaborate in the following screenshot:
grafik
- [ ] Make it clear that the `path_to_embeddings` must be set by the user
- [ ] Make it clear that the `try_get_token_and_id_from_env` function is only for internal testing reasons (can be ignored by user)

MalteEbner:
closes #242

## Description
- [ ] My change is breaking
- use LinearRegression instead of kNN and describe that this equals using the embedding model as backbone and only fine-tuning a classification head on top of it.
- use `lightly-magic` to make usage easier
- use always the same dataset_name everywhere
- more and better comments

pdf of generated output:
[Tutorial 3: Active learning for classification — lightly 1.0.0 documentation.pdf](https://github.com/lightly-ai/lightly/files/6433960/Tutorial.3.Active.learning.for.classification.lightly.1.0.0.documentation.pdf)

## Further issues
The setting of the variables and its alignment is in issue https://github.com/lightly-ai/lightly/issues/350


docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -33,58 +33,77 @@
...
+    #### Commands in terminal / command line
 
-    lightly-train input_dir="/datasets/clothing-dataset-small/train trainer.max_epochs=5
+    # Lightly is installed as a pip package
+    pip install lightly
 
-B. Create embeddings for the dataset, e.g. with
+    # Your personal token for accessing the Lightly Platform is defined. You can get it from
+    # the Lightly platform at https://app.lightly.ai under your username and then Preferences.
+    export LIGHTLY_TOKEN="YOUR_TOKEN"
```
IgorSusmelj, line 83: Would create a new **code** block. I would put each single line terminal command into a new code block.

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -33,58 +33,77 @@
...
 
-.. code::
-    
-    lightly-embed input_dir="/datasets/clothing-dataset-small/train" checkpoint=mycheckpoint.ckpt
+    # You set the name your dataset should have on the lightly platform.
+    # You can give it any name you want and change the default name.
+    export LIGHTLY_DATASET_NAME="clothing_dataset_small"
```
IgorSusmelj, line 87: Would drop this part and just hardcode the name of the dataset to 'active learning clothing data' or something like this. I would try to minimize the use/ set of environment variables and reduce the complexity for the user as much as possible.

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -33,58 +33,77 @@
...
 
-In this tutorial, we use a k-nearest-neighbor classifier that predicts the class of a sample
-based on the class of the k nearest samples in the labeled set.
-We use the euclidean distance between a sample's embeddings as the distance metric.
-The advantage of such a classifier compared to CNNs is that it is very fast and easily implemented.
+In this tutorial, we use a linear regression on top of the embeddings as classifier.
+This is the same as using the embedding model as a pretrained backbone
+and putting a single layer classification head on top of it, while finetuning only the classification head
+on the labeled dataset, but keeping the backbone frozen.
+As the embeddings as input to the classification head are already computed,
```
philippmwirth, line 40: Suggestion:

Since the embeddings are already computed, we can use them directly as input to the classification head. This saves computational time and makes the tutorial quick to walk through.

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -33,58 +33,77 @@
...
   We use the clothing-dataset-small. You can download it using
 
 .. code::
 
-    git clone https://github.com/alexeygrigorev/clothing-dataset-small.git
+    git clone https://github.com/alexeygrigorev/clothing-dataset-small.git ./clothing-dataset-small
+
+The dataset's images are RGB images with a few hundred pixels in width and height. They show clothes
+from 10 different classes, like dresses, hats or t-shirts. The dataset is already split into a train,
+test and validation set and all images for one class are put into one folder.
```
philippmwirth, line 60: Suggestion:

... and the images are grouped in folders by class.

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -33,58 +33,77 @@
...
 
 
 Creation of the dataset on the Lightly Platform with embeddings
 ---------------------------------------------------------------
 
-To perform samplings, we need to perform several steps, ideally with the CLI.
-More documentation on each step can be found here: :ref:`lightly-command-line-tool`.
-
-A. Train a model on the dataset, e.g. with
+To perform samplings, we need a dataset with embeddings on the Lightly platform.
```
philippmwirth, line 70: Maybe:

To do active learning...

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -33,58 +33,77 @@
...
+The first step for it is training an embedding model. Next, your dataset is embedded.
+Last, the dataset and the embeddings are uploaded to the Lightly platform.
+These three steps can all be done using our lightly pip package with the single lightly-magic command:
 
 .. code::
+    #### Commands in terminal / command line
 
-    lightly-train input_dir="/datasets/clothing-dataset-small/train trainer.max_epochs=5
+    # Lightly is installed as a pip package
```
philippmwirth, line 78: we refer to the pip package as `lightly` and to the platform/company as `Lightly`

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -33,58 +33,77 @@
...
 
 Creation of the dataset on the Lightly Platform with embeddings
 ---------------------------------------------------------------
 
-To perform samplings, we need to perform several steps, ideally with the CLI.
-More documentation on each step can be found here: :ref:`lightly-command-line-tool`.
-
-A. Train a model on the dataset, e.g. with
+To perform samplings, we need a dataset with embeddings on the Lightly platform.
+The first step for it is training an embedding model. Next, your dataset is embedded.
```
philippmwirth, line 71: Suggestion:

The first step for **this** is to train a self-supervised embedding model. Then, embed your dataset and lastly, upload the dataset and embeddings to the Lightly platform. These three steps can be done using a single command from the lightly pip package: `lightly-magic`.

docs/source/tutorials_source/platform/tutorial_active_learning.py
philippmwirth: I agree with @IgorSusmelj. We should find the simplest solution possible for the user...

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -33,58 +33,77 @@
...
-D. Upload the images to the platform, e.g. with
 
-.. code::
-    
-    lightly-upload input_dir="/datasets/clothing-dataset-small/train" token="yourToken" dataset_id="yourDatasetId"
+Optional:
+You can find out more about the CLI commands and their parameters at https://docs.lightly.ai/lightly.cli.html
+
+Optional:
+You can have a look at your dataset and embeddings by browsing to it on the lightly platform.
```
philippmwirth, line 106: browsing **through** it

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -94,39 +113,27 @@
 # -----------------
 #
 # Import the Python frameworks we need for this tutorial.
+# We need numpy, scikit-learn and lightly as non-standard packages.
```
philippmwirth, line 116: can you also show the commands for installation here? I think some people just want to follow instructions

philippmwirth (changes requested):
I like the changes in general although I have to agree with @IgorSusmelj that the use of environment variables makes it unnecessarily hard for people to follow the tutorial. 

My suggestion would be to follow a similar approach as before but document it better. I.e. we set the token, etc as constants in the code and then load them from the environment variables only when we run the tutorials. What do you think?

I also found a few typos and made some suggestions w.r.t wording.

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -94,38 +111,29 @@
...
+# If you don't have them install them via pip:
+# in terminal: pip install numpy
```
philippmwirth, line 116: code block? :)

MalteEbner: I put it here, as I assumed that most users will have the frameworks already installed.

MalteEbner:
About the environment variables:
Currently the user has to set two environment variables, which is not too much in my eyes.
What I like about using them is that our users only have to change code in terminal commands, but can leave the python code completely unchanged and do not need to copy anything into the python code.

docs/source/tutorials_source/platform/tutorial_active_learning.py
```diff
@@ -33,58 +33,80 @@
...
-- Make sure you are familiar with the command line tools described at :ref:`lightly-command-line-tool`.
 - To make the definition of the dataset for training the classifier easy,
-  we recommend using a dataset where the samples are grouped into folder according to their class.
+  we recommend using a dataset where the images are grouped in folders by class.
   We use the clothing-dataset-small. You can download it using
 
 .. code::
 
-    git clone https://github.com/alexeygrigorev/clothing-dataset-small.git
+    git clone https://github.com/alexeygrigorev/clothing-dataset-small.git ./clothing-dataset-small
```
IgorSusmelj, line 56: Do we need the `./clothing-dataset-small` argument? I thought `git clone ...` would create a folder with the slug which in this case would be `clothing-dataset-small`?

MalteEbner: No, we could remove it. I added it to show that our users can clone into a different folder, if they want to. Would you prefer to remove it?

IgorSusmelj: I would remove it. This has nothing to do with the actual tutorial. We just use git because we have to get the dataset. No need to make it look more complex than it is.

philippmwirth:
grafik

I found some problems in the screenshot above, otherwise lgtm:
- Following Commands -> The following commands
- The comments in the third code block look bad
- The links in the bottom are not formatted

philippmwirth (approved):
LGTM

PR #347, opened 2021-04-30
284 object detection tutorial

IgorSusmelj:
### Changes

- New tutorial showing how to use Lightly Active Learning to select the most interesting 100 images from Comma10k in order to improve a MOCO pretrained model.

I added a pdf of the generated tutorial here: 
[Tutorial 4_ Active Learning using Detectron2 on Comma10k — lightly 1.0.0 documentation.pdf](https://github.com/lightly-ai/lightly/files/6413785/Tutorial.4_.Active.Learning.using.Detectron2.on.Comma10k.lightly.1.0.0.documentation.pdf)

### DIFF
--- a/docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
+++ b/docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
@@ -1,0 +1,256 @@
+"""
+
+.. _lightly-tutorial-active-learning-detectron2:
+
+Tutorial 4: Active Learning using Detectron2 on Comma10k
+=========================================================
+
+This tutorial is available as a 
+`Google Colab Notebook<https://colab.research.google.com/drive/1r0KDqIwr6PV3hFhREKgSjRaEbQa5N_5I?usp=sharing>`_
+
+In this tutorial you will learn:
+
+*   how to use Lightly Active Learning together with the 
+    [detectron2](https://github.com/facebookresearch/detectron2) framework 
+    for object detection
+*   how to use the Lightly Platform to inspect the selected samples
+*   how to get the selected samples for labeling
+
+The tutorial will be divided into
+the following steps. 
+
+1. (optional) Installation of detectron2 and lightly
+1. Run predictions using a pretrained model
+1. Use lightly to compute active learning scores for the predictions
+1. Use the Lightly Platform to understand where our model struggles
+
+Requirements
+------------
+- Make sure you have the detectron2 framework installed on your machine. You can use
+  the following code to install detectron2:
+
+.. code::
+    pip install detectron2
+
+- In this tutorial we work with the comma10k dataset. The dataset consists of
+  10'000 images for autonomous driving and is available
+  [here on GitHub](https://github.com/commaai/comma10k)
+  We can download the dataset using `git clone`. We save the dataset locally
+  to `/datasets/`
+
+.. code::
+
+    git clone https://github.com/commaai/comma10k
+
+"""
+
+# Some basic setup:
+# Setup detectron2 logger
+import detectron2
+from detectron2.utils.logger import setup_logger
+setup_logger()
+
+# import some common libraries
+import numpy as np
+import os, json, cv2, random, glob
+import tqdm, gc
+import matplotlib.pyplot as plt
+
+# import some common detectron2 utilities
+from detectron2 import model_zoo
+from detectron2.engine import DefaultPredictor
+from detectron2.config import get_cfg
+from detectron2.utils.visualizer import Visualizer
+from detectron2.data import MetadataCatalog, DatasetCatalog
+
+# includes for lightly
+import lightly
+from lightly.active_learning.utils.bounding_box import BoundingBox
+from lightly.active_learning.utils.object_detection_output import ObjectDetectionOutput
+from lightly.active_learning.scorers import ScorerObjectDetection
+from lightly.api.api_workflow_client import ApiWorkflowClient
+from lightly.active_learning.agents import ActiveLearningAgent
+from lightly.active_learning.config import SamplerConfig
+from lightly.openapi_generated.swagger_client import SamplingMethod
+
+# %%
+# Upload dataset to lightly
+# -------------------------
+#
+# To work with the Lightly Platform and use the active learning feature we
+# need to uploda the dataset. 
+# 
+# First, head over to `the Lightly Platform <https://app.lightly.ai/>`_ and 
+# create a new dataset.
+#
+# We can now upload the data using using the command line interface. Don't forget
+# to adjust the input_dir to the location of your dataset.
+#
+# .. code:: 
+# 
+#     lightly-magic token="yourToken" dataset_id="yourDatasetId" \
+#         input_dir='/datasets/comma10k/imgs/' trainer.max_epochs=20 \
+#         loader.batch_size=64 loader.num_workers=2
+#
+# .. note::
+#
+#     In this tutorial we use the lightly-magic command which trains a model
+#     before embedding and uploading it to the Lightly Platform.
+
+
+YOUR_TOKEN = "yourToken"  # your token of the web platform
+YOUR_DATASET_ID = "yourDatasetId"  # the id of your dataset on the web platform
+DATASET_ROOT = '/datasets/comma10k/imgs/'
+
+# allow setting of token and dataset_id from environment variables
+def try_get_token_and_id_from_env():
+    token = os.getenv('TOKEN', YOUR_TOKEN)
+    dataset_id = os.getenv('AL_TUTORIAL_DATASET_ID', YOUR_DATASET_ID)
+    return token, dataset_id
+
+YOUR_TOKEN, YOUR_DATASET_ID = try_get_token_and_id_from_env()
+
+
+
+# %%
+# Inference on unlabeled data
+# ----------------------------
+#
+# In active learning we want to pick the new data for which our model struggles 
+# the most. If we have an image with a single car in it and our model has a very 
+# high confidence that there is a single car we don't gain a lot by including 
+# this example into our training data. However, if we focus on images where the 
+# model is not sure whether the object is a car or a building we want 
+# to include the image.
+#
+# So what we need to do is provide lightly with the model predictions. 
+# Wen can use the ApiWorkflowClient for this. Make sure that we use the 
+# right dataset_id and token.
+
+# create Lightly API client
+api_client = ApiWorkflowClient(dataset_id=YOUR_DATASET_ID, token=YOUR_TOKEN)
+al_agent = ActiveLearningAgent(api_client)
+
+# %%
+
+# we can access the images of the dataset we want to use for active learning using
+# the `al_agent.query_set` property
+
+# let's print the first 3 entries
+print(al_agent.query_set[:3])
+
+# %%
+
+# let's verify the length of the query_set. This should match the number of uploaded
+# images (1887)
+print(len(al_agent.query_set))
+
+# %%
+# Create our Detectron2 model
+# ----------------------------
+#
+#Then, we create a detectron2 config and a detectron2 `DefaultPredictor` to run inference on the newimages.
+# 
+# - We use a pre-trained Faster R-CNN with a ResNet-50 backbone
+# - We use a MS COCO pre-trained model from detectron2
+
+cfg = get_cfg()
+# add project-specific config (e.g., TensorMask) here if you're not running a model in detectron2's core library
+###cfg.merge_from_file(model_zoo.get_config_file("COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_3x.yaml"))
+cfg.merge_from_file(model_zoo.get_config_file("COCO-Detection/faster_rcnn_R_50_FPN_3x.yaml"))
+cfg.MODEL.ROI_HEADS.SCORE_THRESH_TEST = 0.5  # set threshold for this model
+# Find a model from detectron2's model zoo. You can use the https://dl.fbaipublicfiles... url as well
+###cfg.MODEL.WEIGHTS = model_zoo.get_checkpoint_url("COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_3x.yaml")
+cfg.MODEL.WEIGHTS = model_zoo.get_checkpoint_url("COCO-Detection/faster_rcnn_R_50_FPN_3x.yaml")
+predictor = DefaultPredictor(cfg)
+
+# %%
+
+def predict_and_overlay(model, fname):
+    # helper method to run the model on an image and overlay the predictions
+    im = cv2.imread(fname)
+    out = model(im)
+    # We can use `Visualizer` to draw the predictions on the image.
+    v = Visualizer(im[:, :, ::-1], MetadataCatalog.get(cfg.DATASETS.TRAIN[0]), scale=1.2)
+    out = v.draw_instance_predictions(out["instances"].to("cpu"))
+    plt.imshow(out.get_image()[:, :, ::-1])
+
+# %%
+
+def convert_bbox_detectron2lightly(outputs):
+    # convert detectron2 predictions into lightly format
+    height, width = outputs['instances'].image_size
+    boxes = []
+
+    for (bbox_raw, score, class_idx) in zip(outputs['instances'].pred_boxes.tensor, outputs['instances'].scores, outputs['instances'].pred_classes):
+        x0, y0, x1, y1 = bbox_raw.cpu().numpy()
+        x0 /= width
+        y0 /= height
+        x1 /= width
+        y1 /= height
+      
+        boxes.append(BoundingBox(x0, y0, x1, y1))
+    output = ObjectDetectionOutput.from_scores(boxes, outputs['instances'].scores.cpu().numpy(), outputs['instances'].pred_classes.cpu().numpy().tolist())
+    return output
+
+# %%
+# Run model predictions
+# ----------------------
+#
+# We now use the created model and iterate over the query_set and make predictions.
+# It's important that the predictions are in the same order as the filenames
+# in the query_set. Otherwise, we could upload a prediction to the wrong sample!
+
+obj_detection_outputs = []
+pbar = tqdm.tqdm(al_agent.query_set)
+for fname in pbar:
+  fname_full = os.path.join(DATASET_ROOT, fname)
+  im = cv2.imread(fname_full)
+  out = predictor(im)
+  obj_detection_output = convert_bbox_detectron2lightly(out)
+  obj_detection_outputs.append(obj_detection_output)
+
+# %%
+
+# now, we need to turn the predictions into scores
+# The scorer assigns scores between 0.0 and 1.0 to 
+# each sample and for each scoring method
+scorer = ScorerObjectDetection(obj_detection_outputs)
+score = scorer.calculate_scores()
+# %% 
+# let's have a look at the sample with the highest
+# prediction-margin score
+score = score['uncertainty_margin'].max()
+idx = score['uncertainty_margin'].argmax()
+print(f'Highest uncertainty_margin score found for idx {idx}: {score}')
+
+# %%
+# let's have a look at this particular example and show the model 
+# prediction
+fname = os.path.join(DATASET_ROOT, al_agent.query_set[idx])
+predict_and_overlay(predictor, fname)
+
+# %%
+config = SamplerConfig(
+  n_samples=100, 
+  method=SamplingMethod.CORAL, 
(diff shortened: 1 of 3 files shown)


docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+common dataset is MS COCO consisting of over 100'000 images containing 80
+different classes. Our goal is now to take a MS COCO pretrained model and optimize
+it for an autonomous driving task where we only care about pedestrians and cars
+(two classes present within MS COCO). We use the pretrained model and run
+predictions on a second dataset (Comma10k) which has been collected for
+autonomous driving. The predictions combined with self-supervised embeddings we 
+can get using the lighty framework allow us to find the 100 most important images
+through active learning. Active learning is a process where we take model predictions
+to find new images for annotation having a great positive impact on model accuracy.
+
```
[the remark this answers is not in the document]
philippmwirth: Maybe you can move the first sentence all the way to the top and make it a new paragraph? All in all I'm imagening something like this:


Tutorial 4: ....

Active learning is a process of using model predictions to find a new set of images to annotate. The images are chosen to have maximal impact on the model performance. In this tutorial, we will use a pre-trained object detection model to do active learning on a completely unlabeled set of images.

In machine learning, we often don't train a model from scratch. Instead, we start with an already pre-trained model. For object detection tasks, a common pre-training dataset is MS COCO consisting of over 100'000 images containing 80 different classes. Our goal is to take an MS COCO pre-trained model and optimize it for an autonomous driving task where we only care about pedestrians and cars (two classes present within MS COCO). We will proceed as follows: First, we will use the pre-trained model to make predictions on our task dataset (Comma10k) which has been collected for auonomous driving. Then, we use the predictions, self-supervised learning, and active learning with the lightly framework to find the 100 most informative images on which we can finetune our model.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+import matplotlib.pyplot as plt
+
+# import some common detectron2 utilities
+from detectron2 import model_zoo
+from detectron2.engine import DefaultPredictor
+from detectron2.config import get_cfg
+from detectron2.utils.visualizer import Visualizer
+from detectron2.data import MetadataCatalog, DatasetCatalog
+
+# includes for lightly
```
philippmwirth, line 77: technically, these are imports

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+from lightly.active_learning.utils.bounding_box import BoundingBox
+from lightly.active_learning.utils.object_detection_output import ObjectDetectionOutput
+from lightly.active_learning.scorers import ScorerObjectDetection
+from lightly.api.api_workflow_client import ApiWorkflowClient
+from lightly.active_learning.agents import ActiveLearningAgent
+from lightly.active_learning.config import SamplerConfig
+from lightly.openapi_generated.swagger_client import SamplingMethod
+
+# %%
+# Upload dataset to lightly
```
philippmwirth, line 88: Lightly

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+#
+# We can now upload the data using using the command line interface. Replace 
+# **yourToken** and *yourDatasetId** with the two provided values from the web app.
+# Don't forget to adjust the **input_dir** to the location of your dataset.
+#
+# .. code:: 
+# 
+#     lightly-magic token="yourToken" dataset_id="yourDatasetId" \
+#         input_dir='/datasets/comma10k/imgs/' trainer.max_epochs=20 \
+#         loader.batch_size=64 loader.num_workers=2
```
philippmwirth, line 105: How long does it take with 2 workers? Maybe we can increase the number?

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+#
+# .. note::
+#
+#     In this tutorial we use the lightly-magic command which trains a model
+#     before embedding and uploading it to the Lightly Platform.
```
philippmwirth, line 110: Can you add something like "To skip training, you can set `trainer.max_epochs=0`"

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+YOUR_TOKEN = "yourToken"  # your token of the web platform
+YOUR_DATASET_ID = "yourDatasetId"  # the id of your dataset on the web platform
+DATASET_ROOT = '/datasets/comma10k/imgs/'
+
+# allow setting of token and dataset_id from environment variables
+def try_get_token_and_id_from_env():
```
philippmwirth, line 118: Same problem as @MalteEbner has. Since we are working on two tutorials at the same time, let's decide for one approach to follow and then do it that way for all tutorials.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+
+# Some basic setup:
+# Setup detectron2 logger
+import detectron2
+from detectron2.utils.logger import setup_logger
+setup_logger()
+
+# import some common libraries
+import numpy as np
+import os, json, cv2, random, glob
```
philippmwirth, line 66: Here you require cv2 and it's not mentioned that you have to install it anywhere. Additionally, you say above that the first step will be to show how lightly is install yet there is no information on that.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+
+    for (bbox_raw, score, class_idx) in zip(outputs['instances'].pred_boxes.tensor, outputs['instances'].scores, outputs['instances'].pred_classes):
+        x0, y0, x1, y1 = bbox_raw.cpu().numpy()
+        x0 /= width
+        y0 /= height
+        x1 /= width
+        y1 /= height
+      
+        boxes.append(BoundingBox(x0, y0, x1, y1))
+    output = ObjectDetectionOutput.from_scores(boxes, outputs['instances'].scores.cpu().numpy(), outputs['instances'].pred_classes.cpu().numpy().tolist())
```
philippmwirth, line 222: why does this have to be a single line 😆

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+  obj_detection_outputs.append(obj_detection_output)
+
+# %%
+# Now, we need to turn the predictions into scores.
+# The scorer assigns scores between 0.0 and 1.0 to 
+# each sample and for each scoring method.
+scorer = ScorerObjectDetection(obj_detection_outputs)
+scores = scorer.calculate_scores()
+# %% 
+# Let's have a look at the sample with the highest
```
philippmwirth, line 249: It's now called uncertainty-margin so you should also adapt the text.

I would also appreciate a short box here (maybe a note) about what it means to have a high uncertainty-margin.

E.g.

Note: A high uncertainty margin means that the image contains at least one bounding box for which the model is unsure about the class of the object in the bounding box. Read more about how our active learning scores are calculated [here](link).

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+# prediction for it.
+fname = os.path.join(DATASET_ROOT, al_agent.query_set[idx])
+predict_and_overlay(predictor, fname)
+
+# %%
+# Finally, we can tell our agent to select the top 100 images to annotate and
+# improve our existing model. We pick the sampling method called `CORAL` which
+# is a combination of Coreset and Active Learning. Whereas Coreset maximizes
+# the image diversity based on the embeddings, active learning aims at selecting
+# images where our models struggles the most.
```
philippmwirth, line 266: One more sentence about why this is a good thing? :)

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+for i in range(5):
+  predict_and_overlay(predictor, to_label[i])
+
+# %%
+# What's Next?
+# -------------
+# 
+# We showed in this tutorial how you can use Lightly Active Learing to discover 
+# the images you should label next. You can close the loop by annotating 
+# the 100 images and re'train your model. Then start the next iteration
```
philippmwirth, line 294: re'train is a typo

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,306 @@
...
+  predict_and_overlay(predictor, to_label[i])
+
+# %%
+# What's Next?
+# -------------
+# 
+# We showed in this tutorial how you can use Lightly Active Learing to discover 
+# the images you should label next. You can close the loop by annotating 
+# the 100 images and re'train your model. Then start the next iteration 
+# by making new model predictions on the updated query_set (which will now
```
philippmwirth, line 295: The query set doesn't change. Maybe change this sentence to:

Then start the next iteration by making new model predictions with the updated model.

philippmwirth (changes requested):
This is nice. I left some ideas for improvements.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
IgorSusmelj: I have 2 here because Google Colab apparently only has two cores per machine and I use the same code. But I can change this.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
IgorSusmelj: Good catch. I'll add the info about installing opencv

IgorSusmelj:
New version after the second iteration:
[Uploading Tutorial 4_ Active Learning using Detectron2 on Comma10k — lightly 1.0.0 documentation.pdf…]()

philippmwirth:
> 
> 
> New version after the second iteration:
> Uploading Tutorial 4_ Active Learning using Detectron2 on Comma10k — lightly 1.0.0 documentation.pdf…

This link doesn't work for me :/ Can you post it again please?

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,340 @@
...
+# Let's show model predictions for the first 5 images.
+to_label = [os.path.join(DATASET_ROOT, x) for x in al_agent.added_set]
+for i in range(5):
+  predict_and_overlay(predictor, to_label[i])
+
+# %%
+# What's Next?
+# -------------
+# 
+# We showed in this tutorial how you can use Lightly Active Learing to discover
```
philippmwirth, line 320: Active Learing

Do you want to run it through some spell check?

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,340 @@
...
+  predict_and_overlay(predictor, to_label[i])
+
+# %%
+# What's Next?
+# -------------
+# 
+# We showed in this tutorial how you can use Lightly Active Learing to discover 
+# the images you should label next. You can close the loop by annotating 
+# the 100 images and re-train your model. Then start the next iteration 
+# by making new model predictions on the updated unlabeled_set (which will now
```
philippmwirth, line 323: But you start the next iteration by making predictions on the `query_set` and not the `unlabeled_set`.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,340 @@
...
+
+- In this tutorial we work with the comma10k dataset. The dataset consists of
+  10'000 images for autonomous driving and is available
+  `here on GitHub <https://github.com/commaai/comma10k>`_
+  We can download the dataset using `git clone`. We save the dataset locally
+  to `/datasets/`
+
+.. code::
+
+    git clone https://github.com/commaai/comma10k
```
MalteEbner, line 70: If we save the dataset locally to `/datasets/`, you could add the save location directly to the git clone, e.g.
```bash
git clone https://github.com/commaai/comma10k /datasets/
```

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,340 @@
...
+# - By combining the model predictions with the image embeddings we make sure we 
+#   don't select many similar images. Imagine the model being very bad at small 
+#   red cars and the 100 images therefore would only contain this set of images. 
+#   We might overfit the model because it suddenly has too many training examples 
+#   of small red cars.
+
+# %%
+# After retraining our model on the newly labeled 100 images 
+# we can do another active learning iteration with the updated unlabeld set.
+print(len(al_agent.unlabeled_set))
```
philippmwirth, line 340: Same here.

Maybe writing too little text here is a mistake. It could be worth to elaborate more. For example:

At any time during the active learning process, the active learning agent provides access to all images in the active learning pool, the unlabeled images, the labeled images, and the images which were selected in the last active learning query:

```Python
# get the list of filenames of all samples in your pool
al_agent.query_set

# get the list of filenames of all unlabeled samples
# (this is the query_set without the labeled_set)
al_agent.unlabeled_set

# get the list of filenames of all labeled_samples:
# (this is the query_set without the unlabeled_set)
al_agent.labeled_set

# get the list of filenames of all samples selected in the last query
# this is a subset of the labeled_set
al_agent.added_set
```

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
IgorSusmelj: I would leave it up for the user. Not sure everyone likes to store things into `/datasets`. I can add a line to clarify, that this is just the setup we have.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
MalteEbner: https://github.com/lightly-ai/lightly/blob/242-Clean-up-of-platform-tutorial-2-Active-Learning-me/docs/source/tutorials_source/platform/tutorial_active_learning.py
The use sets all variables once via command line as environment variables. Then the environment variables are read both in the lightly-magic command and in the python code. Thus the user only has to set each variable once and does not need to change anything in the python code.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
IgorSusmelj: I did not want to include this because it adds a lot of complexity. I'll think about a way to simplify.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,340 @@
...
+# %%
+# Inference on unlabeled data
+# ----------------------------
+#
+# In active learning we want to pick the new data for which our model struggles 
+# the most. If we have an image with a single car in it and our model has a very 
+# high confidence that there is a car we don't gain a lot by including 
+# this example into our training data. However, if we focus on images where the 
+# model is not sure whether the object is a car or a building we want 
+# to include the image to refine the decision boundary.
```
MalteEbner, line 153: `these images`, keep the plural

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,340 @@
...
+    output = ObjectDetectionOutput.from_scores(
+      boxes, outputs['instances'].scores.cpu().numpy(),
+      outputs['instances'].pred_classes.cpu().numpy().tolist())
+    return output
+
+# %%
+# Run model predictions
+# ----------------------
+#
+# We now use the created model and iterate over the `unlabeled_set` and make predictions.
```
MalteEbner, line 250: I would not use the `unlabeled_set`, but `query_set` here. I agree that `unlabeled_set` is a better name for this tutorial, but it could easily happen, that users will predict on the `unlabeled_set` also for the second iteration of AL. This will cause an error, as predictions are needed for the full `query_set`, no the `unlabeled_set`.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
MalteEbner: E.g. 
"To understand why it is helpful to combine Active Learning with CORESET, consider the following case: There are two very similar images, both of which have a high uncertainty score. Having very similar samples does not help the model to generalise well, thus it would be good, if only one of these samples was chosen.
 A pure Active Learning sampler, however, would choose both of them, whereas CORAL also takes the diversity of samples into account and thus only chooses one of them.

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
MalteEbner, line 340: Here we definitely need to use the query set, as in the earlier comment.
There is also a typo in `unlabeld`

IgorSusmelj:
[Active learning — lightly 1.0.0 documentation.pdf](https://github.com/lightly-ai/lightly/files/6417154/Active.learning.lightly.1.0.0.documentation.pdf)

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,339 @@
...
+# -------------------------
+#
+# To work with the Lightly Platform and use the active learning feature we
+# need to upload the dataset. 
+# 
+# First, head over to `the Lightly Platform <https://app.lightly.ai/>`_ and 
+# create a new dataset.
+#
+# We can now upload the data using the command line interface. Replace 
+# **yourToken** and *yourDatasetId** with the two provided values from the web app.
```
MalteEbner, line 114: `*yourDatasetId**` is missing a star in the beginning

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
```diff
@@ -0,0 +1,339 @@
...
+api_client = ApiWorkflowClient(dataset_id=YOUR_DATASET_ID, token=YOUR_TOKEN)
+al_agent = ActiveLearningAgent(api_client)
+
+# %%
+
+# we can access the images of the dataset we want to use for active learning using
+# the `al_agent.unlabeled_set` property
+
+# let's print the first 3 entries
+print(al_agent.unlabeled_set[:3])
```
philippmwirth, line 170: You still use the `unlabeled_set` here. Especially the sentence "The `unlabeled_set` is the set of images from which we want to query" below (line 176) is problematic. Why not always use `query_set`?

MalteEbner:
To align the setting of the variables with the other tutorials, there is a new issue: https://github.com/lightly-ai/lightly/issues/350

docs/source/tutorials_source/platform/tutorial_active_learning_detectron2.py
MalteEbner: I recommend doing a STRG-F for "unlabeled" and replace it with `query_set` nearly everywhere.

IgorSusmelj: I'll replace it with `query_set` and let's discuss this offline in a call. I switched it to unlabeled_set because if you re-run the code twice because you did two iterations you will have to deal with the preselected tag.

IgorSusmelj:
Added the last changes here
[Tutorial 4_ Active Learning using Detectron2 on Comma10k — lightly 1.0.0 documentation.pdf](https://github.com/lightly-ai/lightly/files/6420123/Tutorial.4_.Active.Learning.using.Detectron2.on.Comma10k.lightly.1.0.0.documentation.pdf)

philippmwirth (approved):
Looks good, let's finish this :)

PR #348, opened 2021-05-03
Added BYOL model and Test cases

pranavsinghps1:
- Added BYOL model (https://arxiv.org/abs/2006.07733)
- Added Test Cases for BYOL model


IgorSusmelj (approved):
This looks great :)

IgorSusmelj:
Made some minor changes. The model works now. It is not fully compatible with all the other lightly models. We can change that in another PR :)

IgorSusmelj (approved):

PR #356, opened 2021-05-05
printing dataset_id and important filenames/values in blue

### LINKED ISSUE #332: Return the dataset_id when creating a new dataset through the CLI
When using the lightly CLI to upload a new dataset it would be great to return the `dataset_id` of the newly created dataset.

Otherwise, the user has to go to the web app to get the `dataset_id`.

Also make sure that embedding path and checkpoint path are still printed.

MalteEbner:
closes #332 
closes https://github.com/lightly-ai/lightly-core/issues/352

## Description
- [ ] My change is breaking
- The `lightly-upload` command now also prints the `dataset_id` if the `new_dataset_name` was specified.
- Many important filenames/values are now printed in blue.
- The `lightly-magic` command now prints e.g. `########## Starting to embed your dataset.` before each of its three steps
- More unittests for `lightly-download` to also cover new printings in blue
- added helper to print missing cli api arguments as yellow warnings, but without text

`lightly-magic:` (cut off parts of the start)
image
image

`lightly-download:`
image
image

## Tests

If applicable, describe the manual test procedure, e.g:
```bash
pip uninstall lightly
export BRANCH_NAME="332-Return-the-dataset_id-when-creating-a-new-dataset-me"
lightly-magic new_dataset_name='new_dataset_name_xyz' token=MY_TOKEN trainer.max_epochs=1 input_dir='my_input_dir'
lightly-download token= MY_TOKEN dataset_id='was_in_blue_in_CLI_print' tag_name='initial-tag' input_dir='my_input_dir' output_dir='my_output_dir'
```

## Documentation
- [ ] I have added docstrings to all changed/added public functions/methods.
- [ ] The autodocs update the documentation accordingly.`

### DIFF
--- a/lightly/cli/download_cli.py
+++ b/lightly/cli/download_cli.py
@@ -34,6 +34,7 @@ def _download_cli(cfg, is_cli_call=True):
     if not tag_name or not token or not dataset_id:
         print_as_warning('Please specify all of the parameters tag_name, token and dataset_id')
         print_as_warning('For help, try: lightly-download --help')
+        return
 
     api_workflow_client = ApiWorkflowClient(
         token=token, dataset_id=dataset_id
--- a/lightly/cli/download_cli.py
+++ b/lightly/cli/download_cli.py
@@ -13,6 +13,7 @@ import shutil
 import warnings
 
 import hydra
+from torch.utils.hipify.hipify_python import bcolors
 from tqdm import tqdm
 
 import lightly.data as data
@@ -73,12 +74,13 @@ def _download_cli(cfg, is_cli_call=True):
     samples = [api_workflow_client.filenames_on_server[i] for i in chosen_samples_ids]
 
     # store sample names in a .txt file
-    with open(cfg['tag_name'] + '.txt', 'w') as f:
+    filename = cfg['tag_name'] + '.txt'
+    with open(filename, 'w') as f:
         for item in samples:
             f.write("%s\n" % item)
 
-    msg = 'The list of files in tag {} is stored at: '.format(cfg['tag_name'])
-    msg += os.path.join(os.getcwd(), cfg['tag_name'] + '.txt')
+    filepath = os.path.join(os.getcwd(), filename)
+    msg = f'The list of files in tag {cfg["tag_name"]} is stored at: {bcolors.OKBLUE}{filepath}{bcolors.ENDC}'
     print(msg, flush=True)
 
     if not cfg['input_dir'] and cfg['output_dir']:
@@ -89,7 +91,7 @@ def _download_cli(cfg, is_cli_call=True):
     elif cfg['input_dir'] and cfg['output_dir']:
         input_dir = fix_input_path(cfg['input_dir'])
         output_dir = fix_input_path(cfg['output_dir'])
-        print(f'Copying files from {input_dir} to {output_dir}.')
+        print(f'Copying files from {input_dir} to {bcolors.OKBLUE}{output_dir}{bcolors.ENDC}.')
 
         # create a dataset from the input directory
         dataset = data.LightlyDataset(input_dir=input_dir)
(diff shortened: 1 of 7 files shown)


tests/cli/test_cli_download.py
```diff
@@ -61,10 +73,30 @@ def test_download_tag_name_exclude_parent(self):
         self.parse_cli_string(cli_string)
         lightly.cli.download_cli(self.cfg)
```
japrescott, line 74: since every test has this line as well, it would call this multiple times, no?

lightly/cli/download_cli.py
```diff
@@ -73,12 +74,13 @@ def _download_cli(cfg, is_cli_call=True):
...
-    with open(cfg['tag_name'] + '.txt', 'w') as f:
+    filename = cfg['tag_name'] + '.txt'
+    with open(filename, 'w') as f:
         for item in samples:
             f.write("%s\n" % item)
 
-    msg = 'The list of files in tag {} is stored at: '.format(cfg['tag_name'])
-    msg += os.path.join(os.getcwd(), cfg['tag_name'] + '.txt')
+    filepath = os.path.join(os.getcwd(), filename)
+    msg = f'The list of files in tag {cfg["tag_name"]} is stored at: {bcolors.OKBLUE}{filepath}{bcolors.ENDC}'
```
japrescott, line 83: nice

japrescott (commented):
looks good with my limited knowledge :)

Questions;
- Can we use a custom blue (like our lightly blue? #2BA3D1 :)) )
- Can we color the upload/progress bar different from the rest of the information? e.g white?

Suggestions;
- Maybe we would also want to show warnings/errors in an orange/red color to make it more eye-popping when it occurs?

MalteEbner:
> * Can we use a custom blue (like our lightly blue? #2BA3D1 :)) )
> * Maybe we would also want to show warnings/errors in an orange/red color to make it more eye-popping when it occurs?

We can also have it blink, change the background and font and even more:
https://stackoverflow.com/questions/4842424/list-of-ansi-color-escape-sequences

tests/cli/test_cli_download.py
MalteEbner: Yes, it called this once for every test
We could put those two lines into an extra function which is called directly with the cli_string.

japrescott: nope, I just realised its called twice thats all. If that is intended behaviour 1) to set it up and 2) to then call it correctly, then all is good

IgorSusmelj (approved):
Looks great!

japrescott:
> However, I would put those ideas into a new issue, as they need a discussion about which design languages to follow. We should not overwhelm our users with a rainbow of colours.

I agree we shouldn't flood them with blinking rainbow colours. I was just thinking of adding warning/errors as they can be hard to spot in a flood of text (especially if it occurred before the last lines). I thought you wanted an open discussion of how we want to handle it, since you also added the ########## "styling" instead of just colouring info text differently.

I don't like to constantly move discussions to new issues/creating a meeting for everything. But as we agreed upon, it's in the hands of the PR author.

philippmwirth (approved):
Looks good! The only thing I don't like is that there seems to be no linebreak before printing the dataset id.

Other than that I would love it if you could add one print statement (or two) for the embedding upload.

MalteEbner:
> Looks good! The only thing I don't like is that there seems to be no linebreak before printing the dataset id.

This is due to tqdm, which always creates those artifacts. Due to race conditions, they are not reproducible that easily.

> Other than that I would love it if you could add one print statement (or two) for the embedding upload.
I added those statements: 
https://github.com/lightly-ai/lightly/blob/485163a36b0cce584b9ffc3f34cc13d48eba0911/lightly/cli/upload_cli.py#L73-L77

PR #367, opened 2021-05-16
Use pl callback from version 1.3 on

### LINKED ISSUE #361: Make compatible with pytorch-lightning >=1.3.0
# Make compatible with pytorch-lightning >=1.3.0

Pytorch-lightning 1.3.0 introduced breaking changes which we'll have to adapt to.

Release notes: https://github.com/PyTorchLightning/pytorch-lightning/releases/tag/1.3.0

IgorSusmelj:
### Changes

closes #361 

We introduced a hack for previous PyTorch Lightning versions to customize the way checkpoints are saved. For example, we wanted to get rid of the `checkpoint=1.ckpt` in the checkpoint filename.
With the new PyTorch Lightning version they introduce all sorts of handy changes which make our hack obsolete. This PR switches to a Lightning checkpoint callback.

- we can still load old checkpoints
- new checkpoints are saved with the same filename as the old version
- now it works with PyTorch Lightning 1.3 :)

### DIFF
--- a/lightly/embedding/_base.py
+++ b/lightly/embedding/_base.py
@@ -7,6 +7,7 @@ import copy
 
 import pytorch_lightning as pl
 import pytorch_lightning.core.lightning as lightning
+from pytorch_lightning.callbacks.model_checkpoint import ModelCheckpoint
 import torch.nn as nn
 
 from lightly.embedding._callback import CustomModelCheckpoint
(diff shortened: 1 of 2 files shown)


lightly/embedding/_base.py
```diff
@@ -114,11 +115,21 @@ def init_checkpoint_callback(self,
...
-        self.checkpoint_callback = CustomModelCheckpoint()
-        self.checkpoint_callback.save_last = save_last
-        self.checkpoint_callback.save_top_k = save_top_k
-        self.checkpoint_callback.monitor = monitor
-
-        dirpath = self.cwd if dirpath is None else dirpath
-        self.checkpoint_callback.dirpath = dirpath
+
+        if pl.__version__[:3] in ['1.0', '1.1', '1.2']:
+            # initialize custom model checkpoint
```
philippmwirth, line 120: Since this part is not covered by tests, I would maybe add a warning to tell the user that support for `pl<1.3` will be dropped in the near future.

philippmwirth: and then we can remove the hack in 2-3 releases

IgorSusmelj:
Added a follow-up issue to cleanup the hack: https://github.com/lightly-ai/lightly/issues/368

philippmwirth (approved):

PR #370, opened 2021-05-17
369 additional video formats is

### LINKED ISSUE #369: Improve video format support
Lightly only allows for video extensions such as `'.mp4', '.mov', '.avi'` even though other formats are also supported. 
Especially, formats such as `'.mpg', '.hevc', '.m4v', '.webm', '.mpeg'` which are quite common in mobile/ web could be supported too.

IgorSusmelj:
### Changes

closes #369 

- Add support for additional video formats
- Take care of case when we don't have a duration and use pyav

### DIFF
--- a/lightly/data/_video.py
+++ b/lightly/data/_video.py
@@ -175,8 +175,8 @@ def _make_dataset(directory,
     if is_valid_file is not None:
         _is_valid_file = is_valid_file
 
-    # find all instances (no subdirectories)
-    instances = []
+    # find all video instances (no subdirectories)
+    video_instances = []
     for fname in os.listdir(directory):
 
         # skip invalid files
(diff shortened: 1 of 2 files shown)


philippmwirth (approved):
Looks good, you could raise a warning instead of issuing the print statement but that's up to you.

lightly/data/_video.py
```diff
@@ -185,18 +185,25 @@ def _is_valid_file(filename):
...
             # This is a hacky solution to estimate the timestamps.
             # When using the video_reader this approach fails because the 
             # estimated timestamps are not correct.
             with av.open(instance) as av_video:
                 stream = av_video.streams.video[0]
+
+                # check if we can extract the video duration
+                if not stream.duration:
+                    print(f'Video {instance} has no timestamp and will be skipped...')
+                    video_instances.remove(instance) # remove from original list (not copy)
```
japrescott, line 204: unsure how this works in python if instance is a reference and only the "list" was copied. This removes it correctly?

lightly/data/_video.py
```diff
@@ -185,18 +185,25 @@ def _is_valid_file(filename):
 
         # keep track of valid files
         path = os.path.join(directory, fname)
-        instances.append(path)
+        video_instances.append(path)
 
     # get timestamps
     timestamps, fpss = [], []
-    for instance in instances:
+    for instance in video_instances[:]: # video_instances[:] creates a copy
```
japrescott, line 192: why even make a copy? so if you remove an instance the iterator doesnt blow up?

japrescott (commented):
just some non-python-pro questions :)

Issue #307, opened 2021-04-20
Self supervised models don't work with batch size = 1


guarin:
All self supervised models  apply currently a squeeze operation after passing an input batch through the backbone:
https://github.com/lightly-ai/lightly/blob/bd982c9513effd75c11d9b2671de09a4be6ce155/lightly/models/simclr.py#L94

But this breaks the batch structure if you have batch size = 1 as the batch dimension is removed. A possible fix for it is to replace squeeze with flatten as it takes a start dimension argument:
```
f0 = self.backbone(x0).flatten(start_dim=1)
```

philippmwirth:
Closed by #372

PR #372, opened 2021-05-18
Replace squeeze with flatten

guarin:
Addresses #307.

I did not update tests, benchmarks and tutorials.


IgorSusmelj (approved):
Looks good. Thanks for the contribution!

PR #385, opened 2021-05-24
349 add NNCRL model

### LINKED ISSUE #349: Implement NNCLR
Interesting paper which shows very good results with a rather simple change. Use the nearest neighbor for the positive sample of the contrastive loss. 

They add a queue (kind of like a memory bank) of the previous embedding. For new samples, they just look up the nearest neighbor.
Just switching from the current view to the nearest neighbor yields a nice boost: 
image

The same approach also works with SimSiam. If we implement the NN-queue in a good way we should also have a compatible version to some of our models :)
image

**ImageNet accuracy:**
image

See the paper: https://arxiv.org/pdf/2104.14548.pdf

IgorSusmelj:
### Changes

closes #349 

- Adds NNCLR model from the paper: https://arxiv.org/pdf/2104.14548.pdf
- Add nearest neighbour memory bank
- Add simple tests for NNCLR model (+ nn memory bank)
- Update docs

Once this PR is merged I'll update the benchmarks and run them on cifar10/ imagenette

### DIFF
--- a/lightly/models/modules/nn_memory_bank.py
+++ b/lightly/models/modules/nn_memory_bank.py
@@ -7,7 +7,7 @@ import torch
 from lightly.loss.memory_bank import MemoryBankModule
 
 
-class NNmemoryBankModule(MemoryBankModule):
+class NNMemoryBankModule(MemoryBankModule):
     """Nearest Neighbour Memory Bank implementation
 
     This class implements a nearest neighbour memory bank as described in the 
@@ -36,7 +36,7 @@ class NNmemoryBankModule(MemoryBankModule):
 
     """
     def __init__(self, size: int = 2 ** 16):
-        super(NNmemoryBankModule, self).__init__(size)
+        super(NNMemoryBankModule, self).__init__(size)
 
     def forward(self,
                 output: torch.Tensor,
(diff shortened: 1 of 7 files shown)


lightly/models/modules/nn_memory_bank.py
```diff
@@ -0,0 +1,62 @@
+""" Nearest Neighbour Memory Bank Module """
+
+# Copyright (c) 2021. Lightly AG and its affiliates.
+# All Rights Reserved
+
+import torch
+from lightly.loss.memory_bank import MemoryBankModule
+
+
+class NNmemoryBankModule(MemoryBankModule):
```
philippmwirth, line 10: It's a little off-looking to me that it's called "NNmemoryBank" instead of "NNMemoryBank" but we can go with it if you think yours is better.

philippmwirth (approved):
Only one comment and it's optional. Do you have any numbers from your internal benchmarks on how well this model works?

lightly/models/modules/nn_memory_bank.py
IgorSusmelj: I agree. I'll change that one. `NNMemoryBank` it is :)

IgorSusmelj:
Results from my early experiments on Cifar10. I did not tune parameters. I'm looking forward to ImageNette and ImageNet experiments :)


**Temperature: 0.1, 200 epochs**

NNCLR without NN: 0.8247
NNCLR with 2**13 NN: 0.8336
NNCLR with 2**16 NN: 0.8382


**Temperature: 0.1, 800 epochs**

NNCLR without NN: 0.8873

NNCLR with 2**16 NN: 0.8765

PR #399, opened 2021-06-07
390 auto num workers pw

### LINKED ISSUE #390: Automatically set number of workers to number of CPU cores when using the CLI
When using the lightly CLI to train and embed models we don't use any workers by default. We decided to go down this road because there were sometimes issues with the PyTorch data loaders when we started working on lightly. However, the system seems quite stable now.

I would suggest switching the default behaviour to use a number of workers = number of CPU cores.

**Todo:**

- [x] Find a simple helper function to determine the number of CPU cores
- [x] Add a new default value for the `loader.num_workers=-1` to automatically find the right number of workers
- [x] Make sure we use the number in
    - [x] `lightly-magic`
    - [x] `lightly-train`
    - [x] `lightly-embed`
    - [x] `lightly-upload`

philippmwirth:
# 390 Automatically set num_workers

Closes #390.

- Wrote a helper function to find the number of available CPUs.
- Used the helper function to automatically set `num_workers` if `cfg.loader.num_workers=-1`.
- Set the default value of `cfg.loader.num_workers` to `-1`

### DIFF
--- a/lightly/cli/_helpers.py
+++ b/lightly/cli/_helpers.py
@@ -27,6 +27,15 @@ def print_as_warning(message: str):
     warnings.formatwarning = old_format
 
 
+def available_cpu_count():
+    """Returns the number of CPUs which can be used by the process.
+
+    This number is not equivalent to the number of CPUs in the system.
+
+    """
+    return len(os.sched_getaffinity(0))
+
+
 def fix_input_path(path):
     """Fix broken relative paths.
 
--- a/lightly/cli/_helpers.py
+++ b/lightly/cli/_helpers.py
@@ -27,13 +27,13 @@ def print_as_warning(message: str):
     warnings.formatwarning = old_format
 
 
-def available_cpu_count():
-    """Returns the number of CPUs which can be used by the process.
+def cpu_count():
+    """Returns the number of CPUs which are present in the system.
 
-    This number is not equivalent to the number of CPUs in the system.
+    This number is not equivalent to the number of available CPUs to the process.
 
     """
-    return len(os.sched_getaffinity(0))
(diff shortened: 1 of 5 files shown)


lightly/cli/_helpers.py
```diff
@@ -27,6 +27,15 @@ def print_as_warning(message: str):
     warnings.formatwarning = old_format
 
 
+def available_cpu_count():
+    """Returns the number of CPUs which can be used by the process.
+
+    This number is not equivalent to the number of CPUs in the system.
+
+    """
+    return len(os.sched_getaffinity(0))
```
MalteEbner, line 36: I am unsure, whether this is the best function to use.
According to https://stackoverflow.com/questions/1006289/how-to-find-out-the-number-of-cpus-using-python,
it is only available on linux systems, but we should also target Mac and Windows.

Furthermore, this number might be higher than the actual number of CPU cores, which happens e.g. in the GitHub action:
> /opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/torch/utils/data/dataloader.py:477: UserWarning: This DataLoader will create 6 worker processes in total. Our suggested max number of worker in current system is 2, which is smaller than what this DataLoader is going to create. Please be aware that excessive worker creation might get DataLoader running slow or even freeze, lower the worker number to avoid potential slowness/freeze if necessary.

 https://github.com/lightly-ai/lightly/runs/2761447843?check_suite_focus=true,

MalteEbner (changes requested):
In general, this looks good.

However, I would look more closely at the best function to get the number of CPUs.

MalteEbner (approved):
This should work on most systems, well done!

PR #402, opened 2021-06-09
384 speed up dataset indexing pw

### LINKED ISSUE #384: Lightly Dataset is slow at indexing larger datasets
`LightlyDataset` is significantly slower than a simple `os.listdir` (500ms vs 50ms with a folder of 100k images). Let's figure out whether we can improve.

**Tasks:**

- [x] Figure out where we spend most of the time
- [x] Figure out whether we can improve

philippmwirth:
# 384 speed up dataset indexing

Closes #384.

I made use of `os.scandir` to get a speed-up of approximately 2.

| Dataset | Images | Before | After |
|----------|---------|---------|------|
| `bdd100k/train` | 70000 | 500ms | 270ms |

The biggest speed-up comes from using `os.scandir` when checking for subdirectories. As described [here](https://docs.python.org/3/library/os.html#os.scandir), `os.scandir` can "significantly increase the performance of code that also needs file type or file attribute information".

### DIFF
--- a/lightly/data/_helpers.py
+++ b/lightly/data/_helpers.py
@@ -33,9 +33,7 @@ def _contains_videos(root: str, extensions: tuple):
         True if root contains subdirectories else false.
     """
     with os.scandir(root) as scan_dir:
-        return any(
-            [f.name.lower().endswith(extensions) for f in scan_dir]
-        )
+        return any(f.name.lower().endswith(extensions) for f in scan_dir)
 
 
 def _is_lightly_output_dir(dirname: str):
@@ -62,9 +60,8 @@ def _contains_subdirs(root: str):
 
     """
     with os.scandir(root) as scan_dir:
-        return any(
-            [f.is_dir() for f in scan_dir if not _is_lightly_output_dir(f.name)]
-        )
+        return any(f.is_dir() for f in scan_dir \
+            if not _is_lightly_output_dir(f.name))
 
 
 def _load_dataset_from_folder(root: str, transform):
--- a/lightly/data/_helpers.py
+++ b/lightly/data/_helpers.py
@@ -60,8 +60,8 @@ def _contains_subdirs(root: str):
 
     """
     with os.scandir(root) as scan_dir:
-        return any(f.is_dir() for f in scan_dir \
-            if not _is_lightly_output_dir(f.name))
+        return any(not _is_lightly_output_dir(f.name) for f in scan_dir \
+            if f.is_dir())
 
 
 def _load_dataset_from_folder(root: str, transform):
--- a/lightly/data/_helpers.py
+++ b/lightly/data/_helpers.py
@@ -32,9 +32,9 @@ def _contains_videos(root: str, extensions: tuple):
     Returns:
         True if root contains subdirectories else false.
     """
-    list_dir = os.listdir(root)
+    scan_dir = os.scandir(root)
     is_video = \
-        [f.lower().endswith(extensions) for f in list_dir]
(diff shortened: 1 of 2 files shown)


lightly/data/_image.py
```diff
@@ -41,16 +41,15 @@ def _is_valid_file(filename):
...
-    for fname in os.listdir(directory):
+    for f in os.scandir(directory):
 
-        if not _is_valid_file(fname):
+        if not _is_valid_file(f.name):
             continue
 
-        path = os.path.join(directory, fname)
-        item = (path, 0)
+        item = (f.path, 0)
```
MalteEbner, line 49: I know this did not change with this PR, but could you specify why the target_i is always 0?
Furthermore, I would love to see the return type of this function specified as `List[str, int]`

MalteEbner (approved):
Great that you find the reason for the low performance and an easy way to speed it up!

Only one minor comment asking for clarification.

lightly/data/_helpers.py
```diff
@@ -61,10 +61,10 @@ def _contains_subdirs(root: str):
         True if root contains subdirectories else false.
 
     """
-    scan_dir = os.scandir(root)
-    is_dir = \
-        [f.is_dir() for f in scan_dir if not _is_lightly_output_dir(f.name)]
-    return any(is_dir)
+    with os.scandir(root) as scan_dir:
+        return any(
+            [f.is_dir() for f in scan_dir if not _is_lightly_output_dir(f.name)]
```
MalteEbner, line 66: I did not know this is a generator, but this allows for even more optimization: Could you remove the list brackets and change to `any(f.is_dir() for f in scan_dir if not _is_lightly_output_dir(f.name)`? Such an expression will return once the first directory is found, making it return instantly in the case the first element is a directory.

lightly/data/_helpers.py
```diff
@@ -32,10 +32,10 @@ def _contains_videos(root: str, extensions: tuple):
     Returns:
         True if root contains subdirectories else false.
     """
-    scan_dir = os.scandir(root)
-    is_video = \
-        [f.name.lower().endswith(extensions) for f in scan_dir]
-    return any(is_video)
+    with os.scandir(root) as scan_dir:
+        return any(
+            [f.name.lower().endswith(extensions) for f in scan_dir]
```
MalteEbner, line 37: Here you can remove the list brackets as well (see comment below).

MalteEbner (approved):
LGTM, you can merge this or make the checks even faster. However, they should not become too much faster, as the most time-consuming case (folder with many images) is not optimised.

lightly/data/_helpers.py
```diff
@@ -62,9 +60,8 @@ def _contains_subdirs(root: str):
 
     """
     with os.scandir(root) as scan_dir:
-        return any(
-            [f.is_dir() for f in scan_dir if not _is_lightly_output_dir(f.name)]
-        )
+        return any(f.is_dir() for f in scan_dir \
+            if not _is_lightly_output_dir(f.name))
```
MalteEbner, line 64: This line essentially does the following:
```
for f in scan_dir
    if not _is_lightly_output_dir(f.name):
        if f.is_dir:
            return True
return False
```
For a folder with many images, both if-conditions are checked every time. However, it could be made faster by reversing the order of the two if-conditions. It would cause the `_is_lightly_output_dir(f.name)` to be checked a maximum of two times.
Solution: `return any(f.is_dir() and not _is_lightly_output_dir(f.name) for fi in scan_dir)`

MalteEbner (approved):
LGTM (Let's get this merged)

PR #415, opened 2021-06-18
created unmocked scripts to test the api latency and upload performance

### LINKED ISSUE #410: Benchmark file upload speed
Write a benchmark that does a file upload and allows to see how much each of the different components contributed to the total upload time.

MalteEbner:
closes #410

## Script to test api latency
- It calls the `get_quota()` endpoint 200 times and reports the latency, also in comparison to pinging google.
image

## Script to benchmark the upload
- There is a small script to run an upload of a dataset to the server. I tried to implement general benchmarking code with cProfile and Yappi, but they did not provide good information that can be easily analysed. Thus I kept using the PyCharm profiler to get nice graphs.
image (3)

## Other
- Updated readme with instructions how to use these scripts

### DIFF
--- a/tests/UNMOCKED_end2end_tests/benchmark_upload.py
+++ b/tests/UNMOCKED_end2end_tests/benchmark_upload.py
@@ -1,0 +1,45 @@
+import sys
+
+from hydra.experimental import initialize, compose
+
+from lightly.api import ApiWorkflowClient
+from lightly.cli import upload_cli
+
+
+def create_new_dataset(path_to_dataset: str, token: str, dataset_name: str):
+    api_workflow_client = ApiWorkflowClient(token=token)
+
+    # create the dataset
+    api_workflow_client.create_new_dataset_with_unique_name(dataset_basename=dataset_name)
+
+    # upload to the dataset
+    initialize(config_path="../../lightly/cli/config", job_name="test_app")
+    cfg = compose(config_name="config", overrides=[
+        f"input_dir='{path_to_dataset}'",
+        f"token='{token}'",
+        f"dataset_id={api_workflow_client.dataset_id}",
+        f"loader.num_workers=12"
+        ])
+    upload_cli(cfg)
+
+    api_workflow_client.delete_dataset_by_id(api_workflow_client.dataset_id)
+
+
+
+def mock_dataset_upload_to_measure():
+    pass
+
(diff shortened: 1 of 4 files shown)


tests/UNMOCKED_end2end_tests/benchmark_upload.py
```diff
@@ -0,0 +1,34 @@
...
+    api_workflow_client.delete_dataset_by_id(api_workflow_client.dataset_id)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) == 1 + 2:
+        path_to_dataset, token = \
+            (sys.argv[1 + i] for i in range(2))
+    else:
+        raise ValueError("ERROR in number of command line arguments, must be 2."
+                         "Example: python test_api path/to/dataset TOKEN")
```
IgorSusmelj, line 32: This should be `python benchmark_upload.py ...`?

tests/UNMOCKED_end2end_tests/benchmark_upload.py
```diff
@@ -0,0 +1,34 @@
+import sys
+
+from hydra.experimental import initialize, compose
+
+from lightly.api import ApiWorkflowClient
+from lightly.cli import upload_cli
+
+def benchmark_upload(path_to_dataset, token):
```
IgorSusmelj, line 8: Could you add 2-3 lines explaining what happens and what we test?

tests/UNMOCKED_end2end_tests/test_api_latency.py
```diff
@@ -0,0 +1,30 @@
...
+
+from lightly.api import ApiWorkflowClient
+from lightly.openapi_generated.swagger_client import Configuration, ApiClient, QuotaApi
+
+if __name__ == "__main__":
+    token = os.getenv("TOKEN")
+    api_client = ApiWorkflowClient(token=token).api_client
+    quota_api = QuotaApi(api_client)
+
+    no_iters = 200
```
IgorSusmelj, line 15: I don't want to be picky here but I would use `n_iters` as its commonly done. `no_iters` sounds like `no iterations` but we do 200

IgorSusmelj (changes requested):
Just a few small things. Otherwise looks great!

IgorSusmelj (approved):

PR #417, opened 2021-06-18
make upload more performant

MalteEbner:
closes https://github.com/lightly-ai/lightly-core/issues/763

## Description
Formerly, the PIL image was saved twice as image again, causing a large overhead, especially for large images. This PR computes the file size on the image which is already on the disk and also uses it to be uploaded.
This increases the upload speed already x2 for small images and even more on larger images.

## Performance before this PR:
Uploading 241 images from the clothing dataset with 12 workers to prod takes 79seconds.
image
## Performance with this PR:
The same uploads takes 37seconds and is thus more than twice as fast.
image

### Before
1. image on disk -> `PIL_image`
2.1 `PIL_image` -> calculate metadata
2.2 `PIL_image` -> save in bytes (as png in quality 100) -> calculate size
2.3 `PIL_image` -> save in bytes (as png in quality 100) -> upload (yes, two separate save in bytes)

### after
1. image on disk -> PIL_image
2.1 `PIL_image` -> calculate metadata
2.2 image on disk -> get size with `os.path.getsize()`
2.3 image on disk -> open -> upload

With the new workflow, jpeg-images are uploaded as they are, without saving them as high-quality png. This reduces the actual file size, both in the metadata and of the uploaded file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ lightly_outputs
 .eggs
 tests/UNMOCKED_end2end_tests/call_test_api.py
 tests/UNMOCKED_end2end_tests/get_versions_all_apis.py
+tests/UNMOCKED_end2end_tests/call_benchmark_upload_private.py
--- a/lightly/__init__.py
+++ b/lightly/__init__.py
@@ -92,11 +92,11 @@ except NameError:
 
 
 if __LIGHTLY_SETUP__:
-    # setting up lightly
+    # setting up lightly
     msg = f'Partial import of {__name__}=={__version__} during build process.' 
     print(msg)
 else:
-    # see if prefetch_generator is available
+    # see if prefetch_generator is available
     try:
         import prefetch_generator
     except ImportError:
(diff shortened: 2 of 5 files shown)


japrescott (commented):
awesome stuff malte! 🥳 

just one question as @philippmwirth mentioned yesterday that calculating the filesize is not so easy; Did you check if the old and new method report the same filesizes?

IgorSusmelj (approved):
Great PR!

MalteEbner:
> awesome stuff malte! 🥳
> 
> just one question as @philippmwirth mentioned yesterday that calculating the filesize is not so easy; Did you check if the old and new method report the same filesizes?

Yes I checked that, they don't. I updated the workflow description in the PR description. The filesizez is calculated based on the image opened to be uploaded. If this is compressed (e.g. jpg), the compressed images is uploaded and the compressed file size is reported.

However, I also checked that the file sizes in the webapp and on my disk aligned, and they didn't. The reason was a run conflict when using overwrites in a multithreading environment. I fixed that in my latest commit. Now the file sizes on disk and in the webapp are the same.

IgorSusmelj (approved):

PR #421, opened 2021-06-24
Update ntx_ent_loss.py

guarin:
- Avoid unnecessary copy and detach of negatives from memory bank. The memory bank already handles this, see: https://github.com/lightly-ai/lightly/blob/1fd211ef5c82b410287012a9eb7bdc51e755cbc4/lightly/loss/memory_bank.py#L118
- Instantiate labels on correct device.


IgorSusmelj (approved):
Great, thanks for the PR!

PR #424, opened 2021-06-28
use endpoint to get both write urls at once

### LINKED ISSUE #420: Use endpoint to get both write URLs at once
When uploading both the thumbnails and the full images, currently the write URLs are requested with two separate API requests. We could use the new endpoint to get both URLs at once, which should increase the upload speed even further.

PR of new endpoint: https://github.com/lightly-ai/lightly-core/pull/783

MalteEbner:
closes #420

### Other change not related to the issue:
The lightly-crop feature is now more verbose and uses tqdm:
https://github.com/lightly-ai/lightly/blob/6e2378225dea4bdb85d2556d305a65d3865ce995/lightly/cli/crop_cli.py#L50-L51

## Description
This uses the new endpoint to get both write urls (thumbnail and full image) at once. It gives a slight performance boost (about 10%) both with 0 and 12 workers.

## Performance measurement with num_workers=0
- 241 images from clothing dataset
#### with this PR: 136s
image

#### before this PR: 156s
image

## Performance measurement with num_workers=12
- 241 images from clothing dataset
#### with this PR: 14.2s
image

#### before this PR: 15.7s
image


IgorSusmelj (approved):
Looks good and works!

lightly/openapi_generated/swagger_client/api/samples_api.py
```diff
@@ -461,6 +461,109 @@ def get_sample_image_write_url_by_id_with_http_info(self, dataset_id, sample_id,
...
+
+        return self.api_client.call_api(
+            '/v1/datasets/{datasetId}/samples/{sampleId}/writeurls', 'GET',
+            path_params,
+            query_params,
+            header_params,
+            body=body_params,
+            post_params=form_params,
+            files=local_var_files,
+            response_type='SampleWriteUrls',  # noqa: E501
```
japrescott, line 559: is this "normal" that we need to do this or did I mess up the openapi specs?

MalteEbner: This is autogenerated, thus I did not look at it. As long as it works, I'm happy.

PR #425, opened 2021-06-29
lightly upload fails with video dataset

### LINKED ISSUE #423: lightly-upload fails with video dataset
# lightly-upload fails with video dataset

Bug introduced in #417 which breaks image upload for video datasets because we try to determine the file size of each image on disk.

Relevant lines:
https://github.com/lightly-ai/lightly/blob/faff18868914f6884dab73a00ec4940e35106f48/lightly/api/api_workflow_upload_dataset.py#L143

Possible fix: Wrap the file size extraction in a try-catch and use the old method as a fallback.

MalteEbner:
closes #423 

# Description
- reproduces the error with a unittest
- fixes the error by writing the image on disk for using it later

### DIFF
--- a/lightly/data/dataset.py
+++ b/lightly/data/dataset.py
@@ -280,7 +280,7 @@ class LightlyDataset:
         for i, filename in zip(indices, filenames):
             _dump_image(self.dataset, output_dir, filename, i, fmt=format)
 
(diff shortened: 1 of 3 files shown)


IgorSusmelj:
Looks good. Tested it locally. For me, it worked for both, video and image datasets.

IgorSusmelj (approved):

PR #432, opened 2021-06-30
restricts filepath to 255 characters, which is a common limit

MalteEbner:
closes https://github.com/lightly-ai/lightly-core/issues/743

# Description
- restricts filepath to 255 characters
- 255 is a common limit by OS, file systems and also gcloud
- raise ValueError if limit is exceeded, causing this sample to be skipped (each sample upload is in a try-catch)

### DIFF
--- a/tests/api_workflow/test_api_workflow_upload_dataset.py
+++ b/tests/api_workflow/test_api_workflow_upload_dataset.py
@@ -28,15 +28,18 @@ class TestApiWorkflowUploadDataset(MockedApiWorkflowSetup):
                                                      image_size=(3, 32, 32))
 
         self.folder_path = tempfile.mkdtemp()
-        sample_names = [f'img_{i}.jpg' for i in range(n_data)]
+        image_extension = '.jpg'
+        sample_names = [f'img_{i}{image_extension}' for i in range(n_data)]
         for sample_idx in range(n_data):
             data = self.dataset[sample_idx]
             sample_name = sample_names[sample_idx]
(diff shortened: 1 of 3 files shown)


tests/api_workflow/test_api_workflow_upload_dataset.py
```diff
@@ -20,17 +22,22 @@ def setUp(self) -> None:
...
-        self.sample_names = sample_names
         for sample_idx in range(n_data):
             data = self.dataset[sample_idx]
-            path = os.path.join(self.folder_path, sample_names[sample_idx])
+            sample_name = sample_names[sample_idx]
+            path = os.path.join(self.folder_path, sample_name)
+
+            if length_of_filepath > len(path):
+                n_missing_chars = length_of_filepath - len(path)
+                path = path.replace('.jpg', 'x' * n_missing_chars + '.jpg')
```
japrescott, line 39: I know this is a "fake" thing but just to be sure when replacing things, ensure that it is where you expect it to be. E.g at the end of the filename. (as a file could be called `hello.jpg.png`). So I would suggest replacing it with a regex `/\.jpg$/` just incase someone would copy this code..

japrescott (commented):
looks nice. just a hint/info :)

tests/api_workflow/test_api_workflow_upload_dataset.py
MalteEbner: Good point, I safeguarded it.

philippmwirth (approved):

PR #442, opened 2021-07-20
441 upload custom metadata pw

### LINKED ISSUE #441: Allow upload of custom metadata
# Allow upload of custom metadata

Tasks:
- [x] Figure out a format in which the custom metadata should be provided.
- [x] Allow passing it to the cli-upload command and upload the custom metadata alongside the images.
- [x] Update the documentation
- [ ] Optional: Add a new cli-command to upload custom metadata after a dataset has been completed.

philippmwirth:
441 Upload custom metadata

Closes #441. 

- Wrote functions to upload custom metadata from Python
- Updated the `upload_cli` so that users can pass a path to the custom metadata
- Added the newly generated code from OpenAPI
- Updated the documentation

Example output from the CLI:
```bash
########## Starting to embed your dataset.
Compute efficiency: 0.16: 100%|██████████| 28/28 [00:01<00:00, 15.53it/s]
Embeddings are stored at /home/philipp_lightly_ai/datasets/aquarium/lightly_outputs/2021-07-20/07-41-13/embeddings.csv
########## Starting to upload your dataset to the Lightly platform.
Loading custom metadata from /home/philipp_lightly_ai/datasets/aquarium/train/_annotations.coco.metadata.json
Uploading images (with 8 workers).
100%|██████████|448/448 [00:40<00:00, 10.57imgs/s]Finished the upload of the dataset.
Starting upload of embeddings.
Finished upload of embeddings.
########## Finished
100%|██████████| 448/448 [00:41<00:00, 10.71imgs/s]
```


docs/source/getting_started/command_line_tool.rst
```diff
@@ -100,20 +100,62 @@ Lightly-magic is a singular command for training, embedding, and uploading to th
...
+            {
+                "file_name": "image0.jpg",
+                "id": 0,
+            },
+            {
+                "file_name": "image1.jpg",
+                "id": 1,
+            }
+        ],
+        "metadata" [
```
IgorSusmelj, line 138: typo: add `:`

docs/source/getting_started/command_line_tool.rst
```diff
@@ -100,20 +100,62 @@ Lightly-magic is a singular command for training, embedding, and uploading to th
...
+            {
+                "image_id": 0,  // this metadata belongs to "image0.jpg"
```
IgorSusmelj, line 140: JSON does not support comments. Leave it out.

docs/source/getting_started/command_line_tool.rst
```diff
@@ -100,20 +100,62 @@ Lightly-magic is a singular command for training, embedding, and uploading to th
...
     
+    .. _ref-upload-custom-metadata-lightly:
 
+Upload custom metadata using the CLI
+---------------------------------------
+    
+You can upload custom metadata along with your images. Custom metadata can be used
+to gain additional insights in the web-app. In order to upload the custom metadata,
+you need to save it to a `.json` file in the following COCO-like format:
+
```
IgorSusmelj, line 124: I would clarify two cases:
- Not every sample has metadata. Is that supported?
- Not every sample has the same kind of metadata. Is that supported?

docs/source/getting_started/command_line_tool.rst
```diff
@@ -100,20 +100,62 @@ Lightly-magic is a singular command for training, embedding, and uploading to th
...
+            },
+            {
+                "file_name": "image1.jpg",
+                "id": 1,
+            }
+        ],
+        "metadata" [
+            {
+                "image_id": 0,  // this metadata belongs to "image0.jpg"
+                "custom_metadata_entry_0": 0.5,
```
IgorSusmelj, line 141: I would give it a more meaningful example or make a story out of it.

For example, for many computer vision problems, we are interested if there is a person present in the image. We can pass this information as `custom_metadata`....

`"number_of_people": 3`

docs/source/getting_started/command_line_tool.rst
IgorSusmelj: Also, add information about what kind of values are allowed. 
- Strings?
- Integers?
- Float?
- Any range limitations? Can the numbers be negative?

lightly/api/api_workflow_upload_dataset.py
```diff
@@ -97,6 +111,10 @@ def lambda_(i):
                 return True
 
             filepath = dataset.get_filepath_from_filename(filename, image)
+
+            # get custom metadata (evaluates to None if there is none)
+            custom_metadata_ = filename_to_metadata.get(filename, None)
```
IgorSusmelj, line 116: Can we use another variable name here? I think we should not append `_`. Somehow it feels wrong xD
I might be wrong here. But if it's a private variable use `_custom_metadata`. Why can't we use `custom_metadata`?

lightly/api/api_workflow_upload_metadata.py
```diff
@@ -0,0 +1,196 @@
...
+                    },
+                    {
+                        "file_name": "image1.jpg",
+                        "id": 1
+                    }
+                ],
+                "metadata": [
+                    {
+                        "image_id": 0,  // this metadata belongs to "image0.jpg"
+                        "custom_metadata_entry_0": 0.5,
```
IgorSusmelj, line 150: Same here. Remove comment. Add more meaningful examples, please. An ideal example reflects all the constraints. If we support integer + string + float we should have examples for all in here. If we support that different samples have different metadata I would also show this here.

IgorSusmelj (changes requested):
Overall it looks good. I left a few comments mainly regarding the documentation. For me, lots of cases are unclear. What kind of metadata is supported? How strict is the format?
I'm trying to run the code locally.

lightly/api/api_workflow_upload_dataset.py
philippmwirth: You're right, trailing underscores are only for conflicts with built-in names, from PEP 8:
> "`single_trailing_underscore_`: used by convention to avoid conflicts with Python keyword, e.g."

docs/source/getting_started/command_line_tool.rst
```diff
@@ -100,20 +100,78 @@ Lightly-magic is a singular command for training, embedding, and uploading to th
...
+            }
+        ]
+    }
+
+
+.. note:: Make sure that the custom metadata is present for every image. The metadata
+          must not necessarily include the same keys for all images but it is strongly
+          recommended.
+
+.. note:: We support numbers, strings, booleans, and even nested objects for
```
IgorSusmelj, line 166: I would split numbers into integers and floats (like the types we have in Python). I think you want to use numbers here because of TypeScript?

lightly/api/api_workflow_upload_metadata.py
```diff
@@ -0,0 +1,202 @@
...
+
+from lightly.openapi_generated.swagger_client.models.sample_update_request import SampleUpdateRequest
+
+
+class _COCO_ANNOTATION_KEYS:
+    """Enum of coco annotation keys complemented with a key for custom metadata.
+    
+    """
+    # image keys
+    images: str = 'images'
```
IgorSusmelj, line 14: If we have such "fixed" keys I would maybe mention this in the documentation.

IgorSusmelj (changes requested):
Some smaller comments to round things up

IgorSusmelj (approved):

PR #444, opened 2021-07-21
524 fix masked select pw

philippmwirth:
# 524 Fix masked select

Closes [#524](https://github.com/lightly-ai/lightly-core/issues/524).

Fixes the buggy implementation of `masked_select` by utilizing the existing function `to_indices` which is implemented identically in the pip package and the api.

### DIFF
--- a/tests/api/test_BitMask.py
+++ b/tests/api/test_BitMask.py
@@ -140,11 +140,22 @@ class TestBitMask(unittest.TestCase):
         self.assertEqual(mask_a, mask_b)
 
     def test_subset_a_list(self):
-        list_ = [4, 7, 9, 1]
-        mask = BitMask.from_bin("0b0101")
-        target_masked_list = [7, 1]
-        masked_list = mask.masked_select_from_list(list_)
-        self.assertEqual(target_masked_list, masked_list)
+        n = 1000
+        list_ = [randint(0, 1) for _ in range(n - 2)] + [0, 1]
+        mask = BitMask.from_length(n)
+        for index, item_ in enumerate(list_):
+            if item_ == 0:
+                mask.unset_kth_bit(index)
+            else:
+                mask.set_kth_bit(index)
+
+        all_ones = mask.masked_select_from_list(list_)
+        mask.x = mask.x ^ (2 ** n - 1) # inverse
+        all_zeros = mask.masked_select_from_list(list_)
+        self.assertGreater(len(all_ones), 0)
+        self.assertGreater(len(all_zeros), 0)
+        self.assertTrue(all([item_ > 0 for item_ in all_ones]))
+        self.assertTrue(all([item_ == 0 for item_ in all_zeros]))
 
     def test_nonzero_bits(self):
 
--- a/tests/api/test_BitMask.py
+++ b/tests/api/test_BitMask.py
@@ -105,7 +105,7 @@ class TestBitMask(unittest.TestCase):
         self.assert_difference("0b0111", "0b1100", "0b0011")
         self.assert_difference("0b10111", "0b01100", "0b10011")
 
-    def random_bitsting(self, length: int):
+    def random_bitstring(self, length: int):
         bitsting = '0b'
         for i in range(length):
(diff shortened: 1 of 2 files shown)


tests/api/test_BitMask.py
```diff
@@ -140,11 +140,22 @@ def test_equal(self):
...
+        list_ = [randint(0, 1) for _ in range(n - 2)] + [0, 1]
+        mask = BitMask.from_length(n)
+        for index, item_ in enumerate(list_):
+            if item_ == 0:
+                mask.unset_kth_bit(index)
+            else:
+                mask.set_kth_bit(index)
+
+        all_ones = mask.masked_select_from_list(list_)
+        mask.x = mask.x ^ (2 ** n - 1) # inverse
```
japrescott, line 153: please use our inverse function we have on the BitMask itself

tests/api/test_BitMask.py
```diff
@@ -140,11 +140,22 @@ def test_equal(self):
...
+        all_zeros = mask.masked_select_from_list(list_)
+        self.assertGreater(len(all_ones), 0)
+        self.assertGreater(len(all_zeros), 0)
+        self.assertTrue(all([item_ > 0 for item_ in all_ones]))
```
japrescott, line 157: maybe add another test to ensure we actually select the correct ones. E.g basic mask 0b010101 selects the correct ones?

tests/api/test_BitMask.py
philippmwirth: grafik


It's not implemented yet... But I'll implement it and use it - good call.

tests/api/test_BitMask.py
philippmwirth: Like the one in the example? Sure :)

tests/api/test_BitMask.py
japrescott: I thought we already had the magic happening but great call to just do it :)

tests/api/test_BitMask.py
```diff
@@ -150,13 +150,41 @@ def test_subset_a_list(self):
...
+ 
+        #get inverse
+        mask = BitMask.from_bin(bitstring)
+        mask.invert(length)
+        inverted = mask.to_bin()
+
+        # remove 0b
+        inverted = inverted[2:]
+        bitstring = bitstring[2:]
+        for i in range(min(len(bitstring), len(inverted))):
```
japrescott, line 181: is there a scenario where one string could be shorter than the other? like a '11110000' turning in to a '1111' when inverting? (I had that in the typescript side) If so, we should ensure that the same length pops out again

tests/api/test_BitMask.py
```diff
@@ -150,13 +150,41 @@ def test_subset_a_list(self):
...
         all_zeros = mask.masked_select_from_list(list_)
         self.assertGreater(len(all_ones), 0)
         self.assertGreater(len(all_zeros), 0)
         self.assertTrue(all([item_ > 0 for item_ in all_ones]))
         self.assertTrue(all([item_ == 0 for item_ in all_zeros]))
 
+
+    def test_masked_select_from_list_example(self):
+        list_ = [1, 2, 3, 4, 5, 6]
+        mask = BitMask.from_bin('0b001101') # expected result is [1, 3, 4]
```
japrescott, line 163: still feels weird that it should be correct to be right to left.. I need to check this in typescript if we both are doing the same :D

japrescott (commented):
still feel a bit uncertain about the bitmask right to left thing... need to check that typescript side of things :)

japrescott (approved):

tests/api/test_BitMask.py
japrescott: I wrote more tests on the typescript side and we dot it the same way from right to left. Makes sense

tests/api/test_BitMask.py
japrescott: as discussed privately this isnt a real issue and only apparent when using the string form.

PR #474, opened 2021-08-23
Refactor projection and prediction heads

### LINKED ISSUE #378: Refactor Prediction/Projection Heads
# Refactor Prediction/Projection Heads

Depends on #377.

As a first step to refactoring the lightly models, we should refactor the implementation of the different prediction and projection heads. The refactored implementation should anticipate new models with currently unknown implementations of prediction/projection heads.

philippmwirth:
# Refactor projection and prediction heads

Closes #378.

The implementation of the prediction and projection head was abstracted in a single class `ProjectionHead`. Model specific heads are implemented by deriving from `ProjectionHead` and providing default parameters.

### DIFF
--- a/tests/models/test_ModelsSimSiam.py
+++ b/tests/models/test_ModelsSimSiam.py
@@ -29,14 +29,12 @@ class TestSimSiam(unittest.TestCase):
                 proj_hidden_dim=512,
                 pred_hidden_dim=128,
                 out_dim=512,
-                num_mlp_layers=2
             ),
             resnet50 = dict(
                 num_ftrs=2048,
                 proj_hidden_dim=2048,
                 pred_hidden_dim=512,
                 out_dim=2048,
-                num_mlp_layers=3
             )
         )
         self.batch_size = 2


lightly/models/modules/heads.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class ProjectionHead(nn.Module):
+    """TODO
+    
+    """
+
+    def __init__(self, layers: List[Tuple[int, int, nn.Module, nn.Module]]):
+
+        self.ffnn = []
```
IgorSusmelj, line 17: would give it a more meaningful variable name. E.g. `self.block`, `self.head`, `self.linear`?

IgorSusmelj:
Could you please print how a model looks like when adding the head?

create a resnet, add the head and then run `print(model)`

I'd like to see whether there is some weird output because of the layering.

Would it look like this?
```
(projectionHead)
    (linear)
    (bn)
    (relu)
    (linear)
```

lightly/models/modules/heads.py
MalteEbner: I agree, I don't know what ffnn stands for. Why not `self.layers`?

lightly/models/modules/heads.py
```diff
@@ -0,0 +1,117 @@
...
+import torch.nn as nn
+from torch.nn.modules.batchnorm import BatchNorm1d
+
+
+class ProjectionHead(nn.Module):
+    """TODO
+    
+    """
+
+    def __init__(self, layers: List[Tuple[int, int, nn.Module, nn.Module]]):
```
MalteEbner, line 15: Is each list element a layer or other a block of (linear layer, batch norm layer, non-linear layer). If yes, than I would call it blocks.

MalteEbner (commented):
I like the structure a lot! It is really great to see much more easily what makes up the difference between the different heads.

lightly/models/modules/heads.py
philippmwirth: feed forward neural network... but I'll change it 😄

lightly/models/modules/heads.py
philippmwirth: It's what you would call a block so I'll rename it.

philippmwirth:
> 
> 
> Could you please print how a model looks like when adding the head?
> 
> create a resnet, add the head and then run `print(model)`
> 
> I'd like to see whether there is some weird output because of the layering.
> 
> Would it look like this?
> 
> ```
> (projectionHead)
>     (linear)
>     (bn)
>     (relu)
>     (linear)
> ```

This is the projection head without ResNet:

```
MoCoProjectionHead(
  (ffnn): Sequential(
    (0): Linear(in_features=512, out_features=512, bias=True)
    (1): ReLU()
    (2): Linear(in_features=512, out_features=256, bias=True)
  )
)
```

I don't know what you mean by "Add the head" because we've never done this. Typically we had a wrapper class (like `MoCoModel`) which would have an attribute `backbone` (the ResNet) and an attribute `projection_head`  which would be the projection head.

Issue #396, opened 2021-06-04
Add PIRL 


ramdhan1989:
Is there any plan to add PIRL ?

philippmwirth:
Yes, we'd like to add PIRL to the framework. However, I can't provide a timeline on this since we're currently planning to do a refactoring of the existing models.

Would you be interested in working on PIRL yourself? It would be a great addition to lightly :)

ramdhan1989:
yeah, I would be happy to help.

philippmwirth:
Cool! I think a good first step would be to add the "jigsaw" augmentation to [`lightly.transforms`](https://github.com/lightly-ai/lightly/tree/master/lightly/transforms).

If you need any information on how to contribute, please take a look at our [contribution guide](https://github.com/lightly-ai/lightly/blob/master/CONTRIBUTING.md).

The reference paper is [here](https://arxiv.org/pdf/1912.01991.pdf).

Let me know if you there are any questions, I'm happy to help :)

shikharmn:
I'm not sure if this issue is still resolved. If not, can this be assigned to me? I'd love to work on this.

philippmwirth:
Hi @shikharmn, the issue is not resolved. You're very welcome to work on it. Can I assign the issue to you?

shikharmn:
Apologies for the delay, do check out #477 and let me know if any changes are needed.

philippmwirth:
Hi @shikharmn and thanks a lot for your contribution! I think the next step would be to write a collate function for PIRL (see [other collate functions](https://github.com/lightly-ai/lightly/blob/master/lightly/data/collate.py) for inspiration).

shikharmn:
Okay, sure! I'll start with that and get back to you if I face any issues.

shikharmn:
Hi @philippmwirth, is this still relevant? Can pick this up again now.

(sorry, was drowned with uni work)

philippmwirth:
Yes sure, I think we can move ahead with this one 👍

shikharmn:
Hi @philippmwirth, do let me know if #850 is okay.

shikharmn:
Ah no worries, I didn't know if the notification reached you, do take a look whenever you find the time! :)

shikharmn:
Thank you @guarin for approving the PR #850! I'll move on to writing the PIRL network and related loss function now to finish this up. Do let me know if there's any details regarding that task which I should keep in mind.

philippmwirth:
[...]

PR #477, opened 2021-08-30
Add Jigsaw trasformation

shikharmn:
Jigsaw transformation returns n_grid random crops from the input image. This augmentation is essential for multiple Self Supervised Learning tasks. Part of a adding PIRL to the framework, as per #396.

### DIFF
--- a/tests/transforms/test_Jigsaw.py
+++ b/tests/transforms/test_Jigsaw.py
@@ -4,7 +4,7 @@ import unittest
 from lightly.transforms import Jigsaw
 
 
-class TestGaussianBlur(unittest.TestCase):
+class TestJigsaw(unittest.TestCase):
 
     def test_on_pil_image(self):
         crop = Jigsaw()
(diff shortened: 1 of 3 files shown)


IgorSusmelj:
This looks great thanks! Could you add an example of how to use it?

shikharmn:
Yeah, sure. Where exactly should I add this example?

shikharmn:
I've added an example usage to the jigsaw.py file as I've seen in other parts of the code, do let me know if this works.

tests/transforms/test_Jigsaw.py
```diff
@@ -0,0 +1,12 @@
+from PIL import Image
+import unittest
+
+from lightly.transforms import Jigsaw
+
+
+class TestGaussianBlur(unittest.TestCase):
```
IgorSusmelj, line 7: rename to `TestJigsaw`

IgorSusmelj (changes requested):
Please rename the class for the test. Otherwise this looks great and we can merge it :)
Thanks a lot!

shikharmn:
Oops, done.

IgorSusmelj (approved):

PR #480, opened 2021-08-31
first draft of refactoring SSL models: implemented for SimCLR, MoCo

### LINKED ISSUE #379: Refactor Self-Supervised Models
# Refactor Self-Supervised Models

Depends on #378.

- Create a construction site branch to merge this issue into
- Refactor one model: e.g. moco + momentum_encoder separate
- discuss the prototype/model with others

MalteEbner:
closes #379

# STALLED FOR THE MOMENT
# go to https://github.com/lightly-ai/lightly/pull/484 instead

## Good article on features of SSL models
https://towardsdatascience.com/a-framework-for-contrastive-self-supervised-learning-and-designing-a-new-approach-3caab5d29619
### Base Embedding Model
The base embedding model does a forward pass for x0, x1 through the backbone and head.

https://github.com/lightly-ai/lightly/blob/5774fcc3d90d64acff3bcc1345e07907bd165d8e/lightly/models/base_embedding_model.py#L13-L20

https://github.com/lightly-ai/lightly/blob/5774fcc3d90d64acff3bcc1345e07907bd165d8e/lightly/models/base_embedding_model.py#L29-L40

It needs to distinguish four cases:
```python
out = model(x) 
out, f = model(x, return_features=True)
out0, out1 = model(x0, x1)
(out0, f0), (out1, f1) = model(x0, x1, return_features=True)
```
This does not work for SimSiam with lines like
https://github.com/lightly-ai/lightly/blob/5774fcc3d90d64acff3bcc1345e07907bd165d8e/lightly/models/simsiam.py#L106

### refactored SimCLR model
It becomes very small and easy to use:
https://github.com/lightly-ai/lightly/blob/813d1b39d987af1ee86642dd59ce65c5f504c1af/lightly/models/simclr.py#L30-L36

### refactored MoCo model
It is basically two models inside:
https://github.com/lightly-ai/lightly/blob/813d1b39d987af1ee86642dd59ce65c5f504c1af/lightly/models/moco.py#L51-L55

### DIFF
--- a/lightly/cli/_helpers.py
+++ b/lightly/cli/_helpers.py
@@ -181,10 +181,27 @@ def load_from_state_dict(model,
                          state_dict,
                          strict: bool = True,
                          apply_filter: bool = True,
-                         num_splits: int = 0):
+                         num_splits: int = 0,
+                         replace_projection_head_with_head: bool = True):
     """Loads the model weights from the state dictionary.
 
     """
+    # Step 0
+    if replace_projection_head_with_head and 'projection_head' in state_dict:
+        # Replace the state dict to prevent the error when loading a legacy
+        # embedding model
+        """
+        RuntimeError: Error(s) in loading state_dict for SimCLR: 32
+    
+        Missing key(s) in state_dict: "head.layers.0.weight", 
+        "head.layers.0.bias", "head.layers.2.weight", "head.layers.2.bias".33
+    
+        Unexpected key(s) in state_dict: "projection_head.layers.0.weight", 
+        "projection_head.layers.0.bias", "projection_head.layers.2.weight", 
+        "projection_head.layers.2.bias".
+        """
+        state_dict['head'] = state_dict['projection_head']
+        del state_dict['projection_head']
 
     # step 1: filter state dict
     if apply_filter:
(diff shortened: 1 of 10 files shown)


MalteEbner:
Proposal for improvements:
- The return value of the `BaseEmbeddingModel.forward()` is too complicated (4 different return formats) and is a tuple (Anna Susmelj told that she doesn't like that, I agree). Furthermore, this format does not scale to models with prediction and projection heads (e.g. SimSiam). Thus I propose to have a more general return value, possibly as a class with the attributes features_0, out0, features_1, out1, z0, z1, p0, p1 and more. The the return value is always this class and the return_features argument can be deprecated.

- Generally, there are two different use cases of a model (e.g. Moco):
    - run it with only x0 for predicting/evaluating. The "consumer" only needs the features.
    - run it with x0 and x1 for training. The "consumer" is the criterion / loss function, needing different outputs depending on the function.
    These cases are mixed into one forward() function, in which there is a large if-condition. Would it be possible, to make these two cases more separate? Then it could even be possible to add the criterion/loss function to the "model" and add the loss to the return object.

MalteEbner:
Closed, we now know more clearly what to do with which steps: #493

Issue #479, opened 2021-08-31
Lightly Documentation: Figure out how to create landing page and overviews targeting different user groups


MalteEbner:
Currently we have many different user groups for the lightly platform. Of course, some users may be part of multiple of these groups. 
A. potential customers, who want to evaluate our product offering
B. people with ML background who want to perform self-supervised learning
C. people who want some embeddings for further tasks (nearest-neighbour search, clustering, ...)
D. people who want to check out some cool ML stuff

We have many different subparts as part of the Lightly Ecosystem:
1. The SSL part, making up the core of the lightly package
2. The Lightly Platform + Web App
3. The Active Learning framework
4. The CLI for doing SSL and/or for using the Lightly Platform
5. The docker

Our user groups probably have their interest in the following parts:
A. potential customers: in all subparts, not necessarily in detail. E.g. they probably don't care that much about the building blocks of the SSL part (collate function, ...).
B. people for SSL: mostly 1 in detail, perhaps also 4 for the CLI
C. people for embeddings: mostly the CLI for SSL, partly 1.
D. people for cool ML stuff: 1-4, depending on their interest, tutorials are great.

How can we have a landing page and overview pages (e.g. main concepts) that 
a) attract all user groups
b) link these user groups to the subpages of their interest

sujanneeraj:
Hi,

Is this issue still offen. If yes, I would like to contribute to it. Could you guide me a bit? 

Thank You

MalteEbner:
Hey @sujanneeraj,

yes, this issue is still open. Thank you for wanting to work on it, we always welcome that! As it is a "Figure Out" issue, its result should be a concept of how to guide the users. This concept could be in form of a sketch on paper, or different drafts and comparisons between them. Then it should also include a description of what should be implemented, i.e. follow-up issues.

As a first step, you could think of yourself: Which user group (A-D) do you fit in best? And which subpages of lightly were of biggest interest to you? Which ones have you looked at first? And at which ones would you have liked to looked at first? What would have been the ideal guidance for you for the lightly documentation?

PR #484, opened 2021-09-02
ssl model refactoring remove models, provide full examples instead

### LINKED ISSUE #483: SSL model refactoring: remove models, provide full examples instead
## Current state:

Currently, we offer 3 levels of SSL parts:
- low-level: the separate building blocks such as a projection head, nn memory bank, backbone creation
https://github.com/lightly-ai/lightly/blob/8dde4d68c0598758bfd05c05f225bf754d9f1ed5/lightly/models/modules/heads.py#L12
- mid-level: the model (backbone + heads + forward functions)
https://github.com/lightly-ai/lightly/blob/8dde4d68c0598758bfd05c05f225bf754d9f1ed5/lightly/models/moco.py#L13
- high-level: the model, criterion, training loop put together as pytorch lightning module
https://github.com/lightly-ai/lightly/blob/8dde4d68c0598758bfd05c05f225bf754d9f1ed5/docs/source/getting_started/benchmarks/cifar10_benchmark.py#L144

## Target state:
We remove the mid-level representations (delete that code). We provide more examples on the high-level: One example pytorch lightning module for every implemented paper

## Advantages:
- As we don't offer mid-level blocks, we don't need to document and maintain them and try to keep them flexible.
- We offer full-fletched examples of how to put the building blocks together to a complete pytorch lightning module.
- By calling the high-level module an example, we don't need to make it configurable and can keep using hardcoded parameters. 
- The high-level example shows directly (not nested) all building blocks and how they play together, thus our users can easily adapt the example on all levels to their needs. They can also change some building blocks very easily and thus come up with new architectures.

MalteEbner:
closes #483

## Description:
- implemented high-level examples for MoCo and SimCLR as pytorch lightning modules
- put the MomentumEncoder is an independent class (no Mixin anymore)

## Open to-dos:

- [ ] write new momentum encoder (make it more flexible)
- [ ] make cli work 
- [ ] deprecation warning old momentum encoder
- [ ] implement high-level example for one new model (e.g. SwAV) https://github.com/lightly-ai/lightly/issues/492
- [ ] deprecation warnings for old models

- [ ] 3. delete mid-level models in lightly/lightly/models
- [ ] 4. ensure compatibility with old pretrained models
- [ ] 5. update documentation
- [ ] 6. implement high-level examples for all implemented papers based on base building blocks

### DIFF
--- a/lightly/models/examples/moco_system.py
+++ b/lightly/models/examples/moco_system.py
@@ -8,8 +8,8 @@ from lightly.models.modules.momentum_encoder import MomentumEncoder
 
 
 class MocoSystem(pl.LightningModule):
-    def __init__(self, dataloader_kNN, num_classes):
-        super().__init__(dataloader_kNN, num_classes)
+    def __init__(self):
+        super().__init__()
         # create a ResNet backbone and remove the classification head
         resnet = lightly.models.ResNetGenerator('resnet-18', num_splits=8)
         self.backbone = torch.nn.Sequential(
@@ -19,7 +19,6 @@ class MocoSystem(pl.LightningModule):
         # create a moco model based on ResNet
         num_ftrs = 512
         out_dim = 128
-        self.m = 0.999
         self.projection_head = MoCoProjectionHead(num_ftrs, num_ftrs, out_dim)
 
         # define the momentum encoded model
@@ -62,6 +61,6 @@ class MocoSystem(pl.LightningModule):
         optim = torch.optim.SGD(self.resnet_moco.parameters(), lr=6e-2,
                                 momentum=0.9, weight_decay=5e-4)
         max_epochs = 200
-        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optim,
-                                                               T_max=200)
+        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(
+            optim, T_max=max_epochs)
         return [optim], [scheduler]
--- a/lightly/models/examples/moco_system.py
+++ b/lightly/models/examples/moco_system.py
@@ -1,0 +1,84 @@
+import pytorch_lightning as pl
+import torch
+
+from lightly.models._momentum import _MomentumEncoderMixin
+from lightly.models.modules.heads import MoCoProjectionHead
+
+import lightly
+
+
+class MocoSystem(pl.LightningModule, _MomentumEncoderMixin):
+    def __init__(self, dataloader_kNN, num_classes):
+        super().__init__(dataloader_kNN, num_classes)
+        # create a ResNet backbone and remove the classification head
+        resnet = lightly.models.ResNetGenerator('resnet-18', num_splits=8)
+        self.backbone = torch.nn.Sequential(
+            *list(resnet.children())[:-1],
+            torch.nn.AdaptiveAvgPool2d(1),
+        )
+        # create a moco model based on ResNet
+        num_ftrs = 512
+        out_dim = 128
+        self.m = 0.999
+        self.projection_head = MoCoProjectionHead(num_ftrs, num_ftrs, out_dim)
+
+        # define the momentum encoded model
+        self.m = 0.999
+        self.batch_shuffle = True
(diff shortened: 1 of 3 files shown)


lightly/models/examples/moco_system.py
```diff
@@ -0,0 +1,67 @@
+import pytorch_lightning as pl
+import torch
+
+from lightly.models.modules.heads import MoCoProjectionHead
+
+import lightly
+from lightly.models.modules.momentum_encoder import MomentumEncoder
+
+
+class MocoSystem(pl.LightningModule):
```
MalteEbner, line 10: I would keep calling it MocoSystem for the moment as the name MoCo is already taken by https://github.com/lightly-ai/lightly/blob/8dde4d68c0598758bfd05c05f225bf754d9f1ed5/docs/source/getting_started/benchmarks/cifar10_benchmark.py#L144

IgorSusmelj:
That does look good to me. I'd need to play around a bit to be able to give more feedback.

lightly/models/examples/moco_system.py
philippmwirth: I am also in favor of **not** calling it `MoCoSystem`.

lightly/models/examples/moco_system.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self, dataloader_kNN, num_classes):
+        super().__init__(dataloader_kNN, num_classes)
+        # create a ResNet backbone and remove the classification head
+        resnet = lightly.models.ResNetGenerator('resnet-18', num_splits=8)
```
philippmwirth, line 14: I hate that the parameters are hard-coded here and below. IMO this is not an option except if we move the `examples` directory to the top-level (same level as `docs` and `tests`).

Alternatives:
- Pass them as arguments
- Pass `config` objects as arguments to avoid too many arguments
    > "If you have a procedure with 10 parameters, you probably missed some." ~ [Perlisims](http://www.cs.yale.edu/homes/perlis-alan/quotes.html)
- Pass projection head, backbone, loss, and optimizer as arguments to the constructor (they have their own default parameters which we could use)
- We brainstorm and try to find another solution

lightly/models/examples/moco_system.py
```diff
@@ -0,0 +1,67 @@
...
+import torch
+
+from lightly.models.modules.heads import MoCoProjectionHead
+
+import lightly
+from lightly.models.modules.momentum_encoder import MomentumEncoder
+
+
+class MocoSystem(pl.LightningModule):
+    def __init__(self, dataloader_kNN, num_classes):
```
philippmwirth, line 11: The `dataloader_kNN` and `num_classes` are just a copy-paste artefact, right?

lightly/models/examples/moco_system.py
MalteEbner: If we have some arguments or even a config object, we have the following disadvantages:
- more maintenance and documentation work
- It is less transparent which parameters are used where and why. E.g. will there be one config object for all models or one for each model? Can I pass a config object with `momentum=0.999` also to the SimCLR Model?
- We can hardly cover all parameters. E.g. what about the scheduler and its parameters? What about using or not using a symmetric loss?
- We become similar to pytorch lightning bolt. 

I agree that we can easily put the examples to the top-level or as an own subpart of the documentation. I like the idea of brainstorming a little more on that.

lightly/models/examples/moco_system.py
MalteEbner: Yes, sorry

lightly/models/examples/moco_system.py
MalteEbner: One general problem might be that we don't have a clear target group of the pip package yet. This is also covered in another issue: https://github.com/lightly-ai/lightly/issues/479

Perhaps we should have a discussion about who we want to target with our pip package.

MalteEbner:
Closed, we now know more clearly what to do with which steps: https://github.com/lightly-ai/lightly/issues/493

PR #486, opened 2021-09-06
Better check for invalid chars in filenames

MalteEbner:
closes https://github.com/lightly-ai/lightly-core/issues/1038
## Description:
- Puts the function to check if the filenames are valid into the utils. https://github.com/lightly-ai/lightly/blob/f73b4e05f2205cf2b17a75fc425c1f7dd26f3960/lightly/utils/io.py#L25
- Uses it when creating a `LightlyDataset`. https://github.com/lightly-ai/lightly/blob/41b0f943a5f66163ddc28e4e18fa0b306a5958af/lightly/data/dataset.py#L163
- Uses it when saving embeddings. https://github.com/lightly-ai/lightly/blob/f73b4e05f2205cf2b17a75fc425c1f7dd26f3960/lightly/utils/io.py#L61
- Uses it when reading embeddings to upload them. https://github.com/lightly-ai/lightly/blob/41b0f943a5f66163ddc28e4e18fa0b306a5958af/lightly/utils/io.py#L115
- some formatting improvements to be flake8 compliant

## Unittests:
- Create LightlyDataset with invalid characters in the filenames and assert that an error is thrown.
- Save embeddings with invalid characters in the filenames and assert that an error is thrown.
- Load embeddings with invalid characters in the filenames and assert that an error is thrown.

### DIFF
--- a/lightly/api/api_workflow_upload_embeddings.py
+++ b/lightly/api/api_workflow_upload_embeddings.py
@@ -1,20 +1,11 @@
 import csv
 from typing import List
 
-from lightly.openapi_generated.swagger_client.models.dataset_embedding_data import DatasetEmbeddingData
-from lightly.openapi_generated.swagger_client.models.write_csv_url_data import WriteCSVUrlData
-
-
-
-def _is_valid_filename(filename: str):
-    """Returns False if the filename is misformatted.
-
-    """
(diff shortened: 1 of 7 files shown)


japrescott (approved):
looks great!

PR #489, opened 2021-09-07
generate openapi client

MalteEbner:
- without Datasources and User Profiles


japrescott (approved):
Looks good. Only skeptical about the change from `TagBitMaskResponse` to `CreateEntityResponse`. Maybe its also just a git diff issue.

Do we have tests for tag arithmetics to ensure this didnt break?

MalteEbner:
> Looks good. Only skeptical about the change from `TagBitMaskResponse` to `CreateEntityResponse`. Maybe its also just a git diff issue.
> 
> Do we have tests for tag arithmetics to ensure this didnt break?

This specific TagArithmeticsRequest still gets a TagBitMaskRespons, as it does not set a new tag tag name:
https://github.com/lightly-ai/lightly-core/blob/dfafb4ebfaacbb05910af3b03a40c5af944b72b7/openapi/spec/api.yml#L1166-L1171

PR #491, opened 2021-09-08
inline cropping image to reduce memory footprint

### LINKED ISSUE #485: Lightly-Crop: memory leak
When using lightly-crop some users experience a memory leak.

- [ ] Try to reproduce it.
- [ ] Fix it

MalteEbner:
closes #485 

## Description
- do the cropping out of the image inline to prevent saving all images
- benchmarked it with a memory profiler: the memory footprint is much smaller

### DIFF
--- a/lightly/utils/cropping/crop_image_by_bounding_boxes.py
+++ b/lightly/utils/cropping/crop_image_by_bounding_boxes.py
@@ -9,9 +9,7 @@ from tqdm import tqdm
 from lightly.active_learning.utils import BoundingBox
 from lightly.data import LightlyDataset
 
-from memory_profiler import profile
 
-@profile
 def crop_dataset_by_bounding_boxes_and_save(dataset: LightlyDataset,
                                             output_dir: str,
                                             bounding_boxes_list_list: List[List[BoundingBox]],
--- a/lightly/utils/cropping/crop_image_by_bounding_boxes.py
+++ b/lightly/utils/cropping/crop_image_by_bounding_boxes.py
@@ -9,19 +9,9 @@ from tqdm import tqdm
 from lightly.active_learning.utils import BoundingBox
 from lightly.data import LightlyDataset
 
+from memory_profiler import profile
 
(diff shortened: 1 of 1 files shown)


lightly/utils/cropping/crop_image_by_bounding_boxes.py
```diff
@@ -71,19 +59,28 @@ def crop_dataset_by_bounding_boxes_and_save(dataset: LightlyDataset,
...
                 class_name = f"class{class_index}"
             cropped_image_last_filename = f'{index}_{class_name}{image_extension}'
             cropped_image_filepath = os.path.join(filepath_out_dir, cropped_image_last_filename)
+
+            w, h = image.size
+            crop_box = (w * bbox.x0, h * bbox.y0, w * bbox.x1, h * bbox.y1)
+            crop_box = tuple(int(i) for i in crop_box)
+            cropped_image = image.crop(crop_box)
             cropped_image.save(cropped_image_filepath)
+            del cropped_image
```
IgorSusmelj, line 83: I think this is not necessary here as we reuse the same variable in the next loop iteration anyway

IgorSusmelj:
Looks good. I would remove the `del` as I think it does not make a difference in this context.

As an alternative, you could also turn the `crop_image_by_bounding_boxes` method into a generator without touching the code of the `crop_dataset_by_bounding_boxes_and_save` method.

```python
def crop_image_by_bounding_boxes(image_filepath: str, bounding_boxes: List[BoundingBox]) -> Iterator[Image.Image]:
    image = Image.open(image_filepath)
    for bbox in bounding_boxes:
        w, h = image.size
        crop_box = (w * bbox.x0, h * bbox.y0, w * bbox.x1, h * bbox.y1)
        crop_box = tuple(int(i) for i in crop_box)
        cropped_image = image.crop(crop_box)
        yield cropped_image
```
Up to you which approach you prefer

IgorSusmelj (approved):

PR #496, opened 2021-09-15
Fix filename to index mapping

### LINKED ISSUE #363: Lightly download does not work with videos
`lightly-download` does not work when trying to copy frames from an `input_dir` containing videos.

**Feedback:** It copies all the images to an output folder. The file names are correct, but the images are not the right ones.
Does this only work if I work with single images? Then I just convert them

philippmwirth:
Fix filename to index mapping

Closes #363.

### DIFF
--- a/lightly/data/dataset.py
+++ b/lightly/data/dataset.py
@@ -4,6 +4,7 @@
 # All Rights Reserved
 
 import os
+import bisect
 import shutil
 import tempfile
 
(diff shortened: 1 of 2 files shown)


MalteEbner (approved):
It would be great to have a unittest ensuring this is really fixed.

If you think this is not necessary or makes too much work, I also approve it without.

PR #497, opened 2021-09-16
add external tutorials to lightly docs

MalteEbner:
closes https://github.com/lightly-ai/lightly-core/issues/1045

## Description
The external tutorials are linked directly in the toctree.

## How it looks
image


philippmwirth (commented):
I like the format. However, I'm not such a fan of the enumeration as it's redundant information (it's also part of the tutorial). I think it would be really cool to show a preview of the linked page (although not sure if this is possible).

philippmwirth (commented):
Cool! Does the link in the navigation bar work as well?

Just one question: Why did you put the ARM tutorial under package?

MalteEbner:
> Cool! Does the link in the navigation bar work as well?

Yes, they do.

> Just one question: Why did you put the ARM tutorial under package?

I thought it was not using the platform. I checked it and found it does, thus I moved it to platform.

PR #502, opened 2021-09-22
12 Add SwaV Loss

### LINKED ISSUE #12: Add SwAV loss
Let's add the loss from [SwAV](https://arxiv.org/pdf/2006.09882.pdf) to the package such that it can be used with and without a memory bank. 

We'll put it in a separate file at `lightly/loss/swav.py`.

Following closely our implementation of the NTXentropy loss, the backbone should look like this:

```python

class SwAVLoss(MemoryBankModule):

    def __init__(self, 
                      # whatever arguments we need
                      memory_bank_size: int = 3000):

        super(SwAVLoss, self).__init__(size=memory_bank_size)
        ...

    def forward(self, output: torch.Tensor, labels: torch.Tensor = None):
        
        output, negatives = super(SwAVLoss, self).forward(output)
        if negatives is None:
            # calculate loss from batch only
            ...
        else:
            # calculate loss from batch and negatives
            ...

        return loss
```

### LINKED ISSUE #492: Refactoring: implement high-level example for one new model (e.g. SwAV)
Implement a high-level example of a new model (e.g. SwaV) This model should also include the loss function and how to put all together. E.g. write the corresponding pytorch lightning module. For inspiration how it looks like, look at https://github.com/lightly-ai/lightly/blob/483-SSL-model-refactoring-remove-models%2C-provide-full-examples-instead-me/lightly/models/examples/moco_system.py

## Dependencies
- https://github.com/lightly-ai/lightly/issues/12

philippmwirth:
# 12 Add SwaV Loss

Closes #12. Closes #492.

- Added an implementation of the SwaV loss
- Added multi-crop and SwaV collate functions
- Added an example implementation of SwaV in the imagenette benchmarks


lightly/loss/swav.py
```diff
@@ -0,0 +1,133 @@
...
+                )
+
+            # compute subloss for each pair of crops
+            subloss = 0.
+            for v in range(len(high_resolution_outputs)):
+                if v != i:
+                    subloss += self.subloss(high_resolution_outputs[v], q)
+
+            for v in range(len(low_resolution_outputs)):
+                subloss += self.subloss(low_resolution_outputs[v], q)
```
MalteEbner, line 129: Why don't you go over the low_resolution_outputs directly:
```python
for low_resolution_output in low_resolution_outputs:
    subloss += self.subloss(low_resolution_output, q)
```

For the high resolution outputs you can do the same by enumerating over them. Then you can still check the index inequality.

lightly/models/modules/heads.py
```diff
@@ -215,3 +215,42 @@ def __init__(self,
...
+                 hidden_dim: int,
+                 output_dim: int):
+        super(SwaVProjectionHead, self).__init__([
+            (input_dim, hidden_dim, nn.BatchNorm1d(hidden_dim), nn.ReLU()),
+            (hidden_dim, output_dim, None, None),
+        ])
+
+
+class SwaVPrototypes(ProjectionHead):
+    """Prototypes used for SwaV.
```
MalteEbner, line 240: Could you explain this a little more?

lightly/loss/swav.py
```diff
@@ -0,0 +1,133 @@
...
+        Q /= K
+        # normalize columns
+        Q /= torch.sum(Q, dim=0, keepdim=True)
+        Q /= B
+
+    Q *= B
+    return Q.t()
+
+
+class SwaV(nn.Module):
```
MalteEbner, line 56: I would call this `SwaVLoss`, I confused it with a SwaV module at first sight.

lightly/loss/swav.py
```diff
@@ -0,0 +1,133 @@
+from typing import List
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.distributed as dist
+
+
+@torch.no_grad()
+def sinkhorn(out: torch.Tensor, iterations: int = 3, epsilon: float = 0.05):
```
MalteEbner, line 10: Do you have a source for this algorithm? It would be great to add it here

docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -54,7 +55,7 @@
 
 # benchmark
 n_runs = 1 # optional, increase to create multiple runs and report mean + std
-batch_sizes = [256]
```
IgorSusmelj, line 57: When you want to commit the general benchmarks please make sure to not change the parameters so we don't have to run all benchmarks again.

docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -465,11 +475,63 @@ def configure_optimizers(self):
...
+        x = self.projection_head(x)
+        x = nn.functional.normalize(x, dim=1, p=2)
+        return self.prototypes(x)
+
+    def training_step(self, batch, batch_idx):
+
+        with torch.no_grad():
+            self.prototypes.normalize_weights()
+
+        multi_crops, _, _ = batch
```
IgorSusmelj, line 506: I'm not a big fan of how we handle the whole data input. It's now a hack in a hack. One big list with images of different sizes...

If you want to keep this in here I would add a few comments on what exactly `multi_crops` is (e.g. shape). Also that the first 2 elements are the high-res ones and the other 4? are the low-res ones.

lightly/models/modules/heads.py
```diff
@@ -215,3 +215,42 @@ def __init__(self,
             (input_dim, hidden_dim, nn.BatchNorm1d(hidden_dim), nn.ReLU()),
             (hidden_dim, output_dim, None, None),
         ])
+
+
+class SwaVProjectionHead(ProjectionHead):
+    """Projection head used for SwaV.
+
+    "We implement in SwAV the improvements used in SimCLR, i.e. (...) and the
```
IgorSusmelj, line 223: Would drop

lightly/models/modules/heads.py
```diff
@@ -215,3 +215,42 @@ def __init__(self,
...
+    MLP projection head." [0]
+
+    [0]: SwaV, 2020, https://arxiv.org/abs/2006.09882
+
+    """
+    def __init__(self,
+                 input_dim: int,
```
IgorSusmelj, line 230: default parameters from paper?

lightly/models/modules/heads.py
```diff
@@ -215,3 +215,42 @@ def __init__(self,
...
+                 n_prototypes: int):
+        super(SwaVPrototypes, self).__init__([])
+        self.layers = nn.Linear(input_dim, n_prototypes, bias=False)
+
+    @torch.no_grad()
+    def normalize_weights(self):
+        """Normalizes the prototypes to the unit sphere.
+        
+        """
+        w = self.layers.weight.data.clone()
```
IgorSusmelj, line 254: do we really need the `clone` and `copy` part?

lightly/loss/swav.py
```diff
@@ -0,0 +1,133 @@
...
+            Number of sinkhorn iterations.
+        epsilon:
+            Temperature parameter.
+
+    Returns:
+        Soft codes Q assigning each feature to a prototype.
+    
+    """
+
+    # get the world size in the distributed case
```
IgorSusmelj, line 26: I would for now drop the distributed part

docs/source/getting_started/benchmarks/imagenette_benchmark.py
philippmwirth: Yes, it's just a draft, I will change the parameters back :)

lightly/models/modules/heads.py
philippmwirth: we don't have them for any of the other heads

lightly/loss/swav.py
philippmwirth: I think it's easier to understand the way it is and I wanted to be as close as possible to equation 6 from the paper:
grafik

lightly/loss/swav.py
philippmwirth: why?

lightly/models/modules/heads.py
philippmwirth: no, we could also use this:

```python
self.layers.weight.div_(
    torch.norm(self.layers.weight, dim=1, keepdim=True)
)
```

docs/source/getting_started/benchmarks/imagenette_benchmark.py
philippmwirth: After our discussion I will simply add comments here to explain everything. This will likely change again with the refactoring of the collate function.

docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -105,7 +114,8 @@ def get_data_loaders(batch_size: int):
         dataset_train_ssl,
         batch_size=batch_size,
         shuffle=True,
-        collate_fn=collate_fn,
+        collate_fn=collate_fn if not multi_crops else swav_collate_fn,
```
MalteEbner, line 117: Is this a specific SwavCollateFun or a more general MultiCropCollateFun? If it is the latter, I would call it like that.

docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -105,7 +114,8 @@ def get_data_loaders(batch_size: int):
...
+        #collate_fn=collate_fn,
```
MalteEbner, line 118: why keep this line?

docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -96,7 +105,7 @@
     transform=test_transforms
 )
 
-def get_data_loaders(batch_size: int):
+def get_data_loaders(batch_size: int, multi_crops: bool = False):
```
MalteEbner, line 108: I would make this a little more flexible by adding the argument collate_fun with a default value. Then `get_data_loaders` will support any collate function. Furthermore, the variable `swav_collate_fn` would not need to be defined this globally anymore, Instead, it could be defined e.g. in a helper model of the SwavModel.

tests/loss/test_SwaVLoss.py
```diff
@@ -0,0 +1,71 @@
...
+                criterion = SwaVLoss(sinkhorn_iterations=sinkhorn_iterations)
+                low_res = [torch.eye(n, 1) for i in range(n_low_res)]
+                
+                with self.subTest(msg=f'n_low_res={n_low_res}, sinkhorn_iterations={sinkhorn_iterations}'):
+                    loss = criterion(high_res, low_res)
+                    # loss should be almost zero for unit matrix
+                    self.assertGreater(0.5, loss.cpu().numpy())
+
+    def test_forward_pass_cuda(self):
+        if torch.cuda.is_available():
```
MalteEbner, line 58: This will cause the unittest to pass if Cuda is unavailable, which is not desired.

Instead use a conditional skip decorator, see https://docs.python.org/3/library/unittest.html.
```python
@unittest.skipUnless(torch.cuda.is_available(), "skip")
```

MalteEbner (commented):
I like your implementation, I would have written it very similar. Great that you wrote a lot of unittests!

I only have one medium-sized (where to define `swav_collate`) and 2 smaller comments.

lightly/loss/swav.py
MalteEbner: I agree, it's closer to the paper,

docs/source/getting_started/benchmarks/imagenette_benchmark.py
IgorSusmelj: Not sure about this. This should just be a very simple code for the purpose of benchmarking. I'd prefer to keep it very "minimal".

philippmwirth: Let's leave it then - ok? We can do a follow-up issue. At some point we'll have to refactor the benchmarks a little bit anyways.

MalteEbner (approved):
You fixed the two smaller comments, great.

I am really looking forward to have this merged. It will be a large step forward on the refactoring epic.

PR #509, opened 2021-09-24
Update benchmarks

IgorSusmelj:
### Changes

- Update benchmarks to use new low-level models for BYOL and MoCo
- Added missing BarlowTwins and SwAV model to cifar10 benchmark
- Added note to models init file with a link to the benchmark code for reference implementations

### DIFF
--- a/docs/source/getting_started/benchmarks/cifar10_benchmark.py
+++ b/docs/source/getting_started/benchmarks/cifar10_benchmark.py
@@ -101,8 +101,8 @@ collate_fn = lightly.data.SimCLRCollateFunction(
 
 # Multi crop augmentation for SwAV
 swav_collate_fn = lightly.data.SwaVCollateFunction(
-    crop_sizes=[128, 64],
-    crop_counts=[2, 4], # 2 crops @ 128x128px and 4 crops @ 64x64px
+    crop_sizes=[32],
+    crop_counts=[2], # 2 crops @ 32x32px
 )
 
 # No additional augmentations for the test set
--- a/docs/source/getting_started/benchmarks/cifar10_benchmark.py
+++ b/docs/source/getting_started/benchmarks/cifar10_benchmark.py
@@ -93,10 +93,16 @@ distributed_backend = 'ddp' if torch.cuda.device_count() > 1 else None
 path_to_train = '/datasets/cifar10/train/'
 path_to_test = '/datasets/cifar10/test/'
 
-# Use SimCLR augmentations, additionally, disable blur
+# Use SimCLR augmentations, additionally, disable blur for cifar10
 collate_fn = lightly.data.SimCLRCollateFunction(
     input_size=32,
     gaussian_blur=0.,
+)
+
+# Multi crop augmentation for SwAV
+swav_collate_fn = lightly.data.SwaVCollateFunction(
+    crop_sizes=[128, 64],
+    crop_counts=[2, 4], # 2 crops @ 128x128px and 4 crops @ 64x64px
 )
 
 # No additional augmentations for the test set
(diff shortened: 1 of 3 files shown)


docs/source/getting_started/benchmarks/cifar10_benchmark.py
```diff
@@ -280,6 +294,54 @@ def configure_optimizers(self):
...
+        )
+        # create a barlow twins model based on ResNet
+        self.resnet_barlowtwins = \
+            lightly.models.BarlowTwins(
+                self.backbone, 
+                num_ftrs=512,
+                proj_hidden_dim=2048,
+                out_dim=2048,
+            )
+        self.resnet_barlowtwins.projection_mlp = ProjectionHead([
```
philippmwirth, line 314: why don't you use the `BarlowTwinsProjectionHead`

docs/source/getting_started/benchmarks/cifar10_benchmark.py
```diff
@@ -339,23 +401,28 @@ def configure_optimizers(self):
         scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optim, max_epochs)
         return [optim], [scheduler]
 
-model_names = ['MoCo_128', 'SimCLR_128', 'SimSiam_128', 'BYOL_128',
-               'MoCo_512', 'SimCLR_512', 'SimSiam_512', 'BYOL_512']
-models = [MocoModel, SimCLRModel, SimSiamModel, BYOLModel]
+model_names = ['MoCo_128', 'SimCLR_128', 'SimSiam_128', 'BarlowTwinsModel_128', 'BYOL_128',
+               'MoCo_512', 'SimCLR_512', 'SimSiam_512', 'BarlowTwinsModel_512', 'BYOL_512']
```
philippmwirth, line 405: Now you didn't add SwaV?

philippmwirth (changes requested):

docs/source/getting_started/benchmarks/cifar10_benchmark.py
IgorSusmelj: I did just copy-paste the setup from the ImageNette benchmark. Barlow Twins follows the SimSiam model. So for cifar10 the results seem much better by just using two instead of three layers. I wouldn't worry too much about that at this point. The goal of this PR is to get the two benchmarks more or less in sync.

docs/source/getting_started/benchmarks/cifar10_benchmark.py
IgorSusmelj: Thanks for pointing that out. I pushed the PR too early

docs/source/getting_started/benchmarks/cifar10_benchmark.py
```diff
@@ -79,12 +93,18 @@
...
 
-# Use SimCLR augmentations, additionally, disable blur
+# Use SimCLR augmentations, additionally, disable blur for cifar10
 collate_fn = lightly.data.SimCLRCollateFunction(
     input_size=32,
     gaussian_blur=0.,
 )
 
+# Multi crop augmentation for SwAV
+swav_collate_fn = lightly.data.SwaVCollateFunction(
```
philippmwirth, line 103: We should change this to something like

```python
swav_collate_fn = lightly.data.SwaVCollateFunction(
    crop_sizes=[32],
    crop_counts=[2],
)
```
or just use the same collate function as for the other benchmarks in this case.

philippmwirth (approved):
Approved! But we will probably have to change some parameters of the SwaV Cifar10 benchmark once we run it (see my other comment).

PR #516, opened 2021-09-29
Refactor CLI (deprecate using SimCLR)

### LINKED ISSUE #512: Refactor CLI (deprecate high-level models)
# Refactor CLI (deprecate high-level models)

With `1.1.2` we will deprecate all mid- and high-level models (see #498). This issue is about removing the high-level model from the CLI and implementing an alternative.

- [ ] Copy `SimCLR` into the cli part as an internal module and use it from there
- [ ] Remove the dependence of the train and embed cli on the `models/simclr.py`
- [ ] Make sure old checkpoint still work

### LINKED ISSUE #515: ensure compatibility with old pretrained models (checkpoints)
# Duplicate of #512

MalteEbner:
closes #512

## Description
The `SimCLR` model was copied to the `cli` folder, reduced to be minimal and imported from there.
Thus the `embed_cli` and `train_cli` is not dependant on the `SimCLR` model with its deprecation warning anymore.

## Tests
- unittests for the lightly-train, and lightly-embed run through (and raised a deprecation warning before)
- when installed from the new branch, lightly-train and lightly-embed and lightly-magic run through without deprecation warnings.
```
(venv) user@machine % lightly-embed input_dir=/datasets/jungfraujoch_m
Compute efficiency: 1.00: 100%|██████████████████████| 23/23 [01:25<00:00,  3.73s/it]
Embeddings are stored at /lightly_outputs/2021-09-29/15-51-36/embeddings.csv
```

## Miscellaneous 
Added a bash script that installs lightly from the current branch. It is quite useful for testing.

### DIFF
--- a/lightly/cli/_checkpoint_cli_simclr.py
+++ b/lightly/cli/_checkpoint_cli_simclr.py
@@ -20,15 +20,14 @@ class _SimCLR(nn.Module):
     def __init__(self, backbone: nn.Module, num_ftrs: int = 32,
                  out_dim: int = 128):
 
-        super(SimCLR, self).__init__()
+        super(_SimCLR, self).__init__()
 
         self.backbone = backbone
         self.projection_head = SimCLRProjectionHead(num_ftrs, num_ftrs,
                                                     out_dim)
 
 
-    def forward(self, x0: torch.Tensor, x1: torch.Tensor = None,
-                return_features: bool = False):
+    def forward(self, x0: torch.Tensor, x1: torch.Tensor = None):
         """Embeds and projects the input images.
 
         """
@@ -37,9 +36,6 @@ class _SimCLR(nn.Module):
         f0 = self.backbone(x0).flatten(start_dim=1)
         out0 = self.projection_head(f0)
 
-        # append features if requested
-        if return_features:
-            out0 = (out0, f0)
 
         # return out0 if x1 is None
         if x1 is None:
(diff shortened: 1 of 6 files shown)


tests/install_current_branch.sh
```diff
@@ -0,0 +1,7 @@
+set -e
+
+# Use this script to install lightly at the current branch
```
philippmwirth, line 3: what does it mean to install "lightly at the current branch"?
- latest master?
- local branch?

lightly/cli/_checkpoint_cli_simclr.py
```diff
@@ -0,0 +1,49 @@
...
+import warnings
+
+import torch
+import torch.nn as nn
+
+from lightly.models.modules import SimCLRProjectionHead
+
+
+class _SimCLR(nn.Module):
+    """Implementation of the SimCLR architecture for using the checkpoint
```
philippmwirth, line 15: Just write "Implementation of SimCLR used by the command-line interface." and maybe add "Provides backwards compatability with old checkpoints.".

And please don't forget that punctuation is important 😉

philippmwirth (commented):
Well done! My only concern is the filename `_checkpoint_cli_simclr` which doesn't make too much sense to me. It's just the model. The fact that it is required for backwards compatability with checkpoints has nothing to do with the contents of the file. My suggestion: Rename the file to `_cli_simclr.py` or even just `_simclr.py`.

tests/install_current_branch.sh
MalteEbner: deleted this file

philippmwirth (approved):

PR #518, opened 2021-09-30
remove embeddings_sorted.csv

### LINKED ISSUE #514: Remove embeddings.csv and rename embeddings_sorted.csv in lightly_outputs
When running lightly-cli we create two files in the output folder `embeddings.csv` and `embeddings_sorted.csv`. Having two files is just confusing. We only use the `embeddings_sorted.csv` for upload to the API. So I'd suggest the following:

Don't save the `embeddings_sorted.csv` on the disk, but only in memory or delete it.

MalteEbner:
closes #514 

## Description
- saves the sorted embeddings in a temporary in-memory file (first a string file, then a bytes file)
- uploads directly from the in-memory file

## Miscellaneous
- fixes a bug in ordering filenames by other filenames, that had a quadratic time complexity. Now it is required, that every filename on the server must also be in the list of filenames to be ordered.
- fixes a bug that the upload of a file to a signed url was not mocked to succeed always.

## Test of filename ordering:
```
  File "/Users/malteebnerlightly/Documents/GitHub/lightly/lightly/api/api_workflow_client.py", line 134, in _order_list_by_filenames
    raise ValueError(
ValueError: All inputs (filenames_for_list,  list_to_order and self.filenames_on_server) must have the same length, but their lengths are: (2,3 and 3).
```

## Performance
Preparing an `embeddings.csv` for upload with 1M samples now takes 40s.
image

### DIFF
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -129,13 +129,13 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin,
             filenames_for_list.
 
         """
-        if not len(filenames_for_list) == len(list_to_order) or \
-                not len(filenames_for_list) == len(self.filenames_on_server):
+        if len(filenames_for_list) != len(list_to_order) or \
+                len(filenames_for_list) != len(self.filenames_on_server):
             raise ValueError(
                 f"All inputs (filenames_for_list,  list_to_order and "
                 f"self.filenames_on_server) must have the same length, "
-                f"but they have lengths of: ({len(filenames_for_list)},"
-                f"{len(list_to_order)} and {len(self.filenames_on_server)}."
+                f"but their lengths are: ({len(filenames_for_list)},"
+                f"{len(list_to_order)} and {len(self.filenames_on_server)})."
             )
         dict_by_filenames = dict(zip(filenames_for_list, list_to_order))
         list_ordered = [dict_by_filenames[filename] for filename in self.filenames_on_server]
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -129,8 +129,14 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin,
             filenames_for_list.
 
         """
-        assert len(filenames_for_list) == len(list_to_order)
-        assert len(filenames_for_list) == len(self.filenames_on_server)
+        if not len(filenames_for_list) == len(list_to_order) or \
+                not len(filenames_for_list) == len(self.filenames_on_server):
+            raise ValueError(
+                f"All inputs (filenames_for_list,  list_to_order and "
+                f"self.filenames_on_server) must have the same length, "
+                f"but they have lengths of: ({len(filenames_for_list)},"
+                f"{len(list_to_order)} and {len(self.filenames_on_server)}."
+            )
         dict_by_filenames = dict(zip(filenames_for_list, list_to_order))
         list_ordered = [dict_by_filenames[filename] for filename in self.filenames_on_server]
         return list_ordered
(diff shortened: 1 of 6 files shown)


lightly/api/api_workflow_client.py
```diff
@@ -119,14 +122,17 @@ def _order_list_by_filenames(self, filenames_for_list: List[str], list_to_order:
...
-            would put them in the order of the filenames in self.filenames_on_server
+            The list reordered.
+            The same reorder applied on the filenames_for_list would put them
+            in the order of the filenames in self.filenames_on_server.
+            every filename in self.filenames_on_server must be in the
+            filenames_for_list.
 
         """
         assert len(filenames_for_list) == len(list_to_order)
+        assert len(filenames_for_list) == len(self.filenames_on_server)
```
philippmwirth, line 133: I must have missed this the last time but don't use asserts here, rather raise an exception.

lightly/api/api_workflow_upload_embeddings.py
```diff
@@ -67,11 +68,21 @@ def upload_embeddings(self, path_to_embeddings_csv: str, name: str):
...
-        # upload the csv to the URL
-        with open(path_to_ordered_embeddings_csv, 'rb') as file_ordered_embeddings_csv:
-            self.upload_file_with_signed_url(file=file_ordered_embeddings_csv, signed_write_url=signed_write_url)
+        # save the csv rows in a temporary in-memory string file
+        # using a csv writer and then read them as bytes
+        with tempfile.SpooledTemporaryFile(mode="rw") as f:
+            writer = csv.writer(f)
+            writer.writerows(rows_csv)
+            f.seek(0)
+            embeddings_csv_as_bytes = f.read().encode('utf-8')
```
philippmwirth, line 77: why can we not directly use this file for the upload?

MalteEbner: I tried, but we need to write in a string file, but need a byte file for upload. The TemporarySpoolFile (ans similar, BytesIO, StringIO) don't allow mixing strings and bytes.

lightly/api/api_workflow_client.py
MalteEbner: I replaced with a ValueError and wrote a unittest for it.

lightly/api/api_workflow_upload_embeddings.py
MalteEbner: Anyway, as you see from the performance tree diagram, not this conversion (reading from StringFile, converting to bytes, writing to BytesFile) is taking long, but only the actual write with the csv writer. It might be further optimised by creating the strings for each line without the csv writer, but I don't think we need further peformance improvements at the moment.

lightly/api/api_workflow_client.py
```diff
@@ -129,8 +129,14 @@ def _order_list_by_filenames(
             filenames_for_list.
 
         """
-        assert len(filenames_for_list) == len(list_to_order)
-        assert len(filenames_for_list) == len(self.filenames_on_server)
+        if not len(filenames_for_list) == len(list_to_order) or \
```
philippmwirth, line 132: you know there's a sign for `not x == y`? 😄

lightly/api/api_workflow_client.py
```diff
@@ -129,8 +129,14 @@ def _order_list_by_filenames(
...
+                not len(filenames_for_list) == len(self.filenames_on_server):
+            raise ValueError(
+                f"All inputs (filenames_for_list,  list_to_order and "
+                f"self.filenames_on_server) must have the same length, "
+                f"but they have lengths of: ({len(filenames_for_list)},"
```
philippmwirth, line 137: I think you can leave the "of" away here

philippmwirth (changes requested):

lightly/api/api_workflow_upload_embeddings.py
philippmwirth: Ok, we can still check alternatives when this becomes a bottleneck

philippmwirth (approved):
Tested it and it worked, nice :)

PR #527, opened 2021-10-12
implement LightlyDataset.from_filenames()

### LINKED ISSUE #526: Add a `from_filenames` function to `LightlyDataset`
# Add a `from_filenames` function to `LightlyDataset`

This should be a helpful utility to create `LightlyDataset`s from a list of filenames.

Ideally, a user can pass the root directory and one of 
- a list of image filenames
- a list of video filenames

Unfortunately, this case is not covered by the `LightlySubset` because the `LightlySubset` requires a `LightlyDataset` which loads all files in the root directory. This is inefficient and unnecessary in cases where there are many images or videos in a directory but only few of them are of interest.

MalteEbner:
closes #526 

## Description
- Deletes `_load_dataset`, calls directly `_load_dataset_from_folder`
- Adapts the `LightlyDataset` constructor to allow passing filenames. If they are passed, it defines the `is_valid_file(filepath)` function to return only true if the `filepath` is in the `set of filepaths`. 
- This `is_valid_file(filepath)` function is passed to `_load_dataset_from_folder` and used for all three base datasets: the `VideoDataset`,  `datasets.ImageFolder` and `DatasetFolder`. 
- The `_make_dataset` functions for videos and image folders now support setting both extensions and the `is_valid_filepath`.

- Adds unittests.

### DIFF
--- a/lightly/data/_helpers.py
+++ b/lightly/data/_helpers.py
@@ -116,7 +116,6 @@ def _load_dataset_from_folder(
                          'dependencies. The error from the imported '
                          f'module was: {VIDEO_DATASET_ERRORMSG}')
 
-
     if contains_videos:
         # root contains videos -> create a video dataset
         dataset = VideoDataset(root,
--- a/tests/data/test_LightlyDataset.py
+++ b/tests/data/test_LightlyDataset.py
@@ -202,7 +202,7 @@ class TestLightlyDataset(unittest.TestCase):
 
     def test_filenames_dataset_no_samples(self):
         tmp_dir, folder_names, sample_names = self.create_dataset()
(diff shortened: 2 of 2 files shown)


MalteEbner:
@philippmwirth I did the following changes:
- adressed all your comments
- created the 'is_valid_filepath' in the LightlyDataset so that it is passed from there directly to the `_load_dataset_from_folder`. See also updated PR description.
- Wrote a unittest to assert the RuntimeError for the case that no files in the input directory match the filenames passed.

I do not want to write unittest for more (e.g. creating a dataset without extensions), as these would be tests of low-level functionalities we never use. In fact, we could also make the extensions non-optional, but required.

PR #545, opened 2021-11-16
embedding upload bugfix new dataset name

### LINKED ISSUE #544: embedding upload not working with new_dataset_name
When using lightly-magic or lightly-upload to upload embedding to a new dataset with `new_dataset_name`, the download of the already existing embeddings (none are there) fails:

```
(venv) user@guest lightly % lightly-magic input_dir=/Users/datasets/clothing-dataset-small/test/dress trainer.max_epochs=0 token=TOKEN new_dataset_name="blub"
########## Starting to embed your dataset.
Compute efficiency: 0.24: 100%|█████| 1/1 [00:09<00:00,  9.04s/it]
Embeddings are stored at /Users/GitHub/lightly/lightly_outputs/2021-11-15/16-03-22/embeddings.csv
########## Starting to upload your dataset to the Lightly platform.
Uploading images (with 12 workers).
 100%|███████▏| 15/15 [00:01<00:00, 13.48imgs/s]
 Finished the upload of the dataset.
Starting upload of embeddings.
  File "/Users/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/cli/lightly_cli.py", line 87, in lightly_cli
    return _lightly_cli(cfg)
  File "/Users/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/cli/lightly_cli.py", line 38, in _lightly_cli
    _upload_cli(cfg)
  File "/Users/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/cli/upload_cli.py", line 101, in _upload_cli
    embeddings = api_workflow_client.embeddings_api \
  File "/Users/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/openapi_generated/swagger_client/api/embeddings_api.py", line 55, in get_embeddings_by_dataset_id
    (data) = self.get_embeddings_by_dataset_id_with_http_info(dataset_id, **kwargs)  # noqa: E501
  File "/Users/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/openapi_generated/swagger_client/api/embeddings_api.py", line 115, in get_embeddings_by_dataset_id_with_http_info
    return self.api_client.call_api(
  File "/Users/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/openapi_generated/swagger_client/api_client.py", line 326, in call_api
    return self.__call_api(resource_path, method,
  File "/Users/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/openapi_generated/swagger_client/api_client.py", line 158, in __call_api
    response_data = self.request(
  File "/Users/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/openapi_generated/swagger_client/api_client.py", line 348, in request
    return self.rest_client.GET(url,
  File "/Users/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/openapi_generated/swagger_client/rest.py", line 234, in GET
    return self.request("GET", url,
  File "/Users/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/openapi_generated/swagger_client/rest.py", line 228, in request
    raise ApiException(http_resp=r)
lightly.openapi_generated.swagger_client.rest.ApiException: (400)
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json; charset=utf-8', 'x-cloud-trace-context': '3aa5b5da3e128a2a549edae34d0156cf/10076201567220533420;o=1', 'Vary': 'Origin, Accept-Encoding', 'Content-Security-Policy': "script-src https:/*.lightly.ai https:/*.youtube.com https:/*.gstatic.com https:/*.google.com https:/*.doubleclick.net https:/*.google-analytics.com https:/*.googletagmanager.com https:/*.fullstory.com;img-src *;default-src 'self';base-uri 'self';block-all-mixed-content;font-src 'self' https: data:;frame-ancestors 'self';object-src 'none';script-src-attr 'none';style-src 'self' https: 'unsafe-inline';upgrade-insecure-requests", 'X-DNS-Prefetch-Control': 'off', 'Expect-CT': 'max-age=0', 'X-Frame-Options': 'SAMEORIGIN', 'Strict-Transport-Security': 'max-age=15552000; includeSubDomains', 'X-Download-Options': 'noopen', 'X-Content-Type-Options': 'nosniff', 'X-Permitted-Cross-Domain-Policies': 'none', 'Origin-Agent-Cluster': '?1', 'X-XSS-Protection': '0', 'Cross-Origin-Opener-Policy': 'same-origin-allow-popups', 'Cross-Origin-Resource-Policy': 'same-site', 'ETag': 'W/"1d-RxIXLBAjGcBSG7XheTm9uqfrCcY"', 'Date': 'Mon, 15 Nov 2021 15:03:35 GMT', 'Server': 'Google Frontend', 'Content-Length': '29'})
HTTP response body: {
    "code": "BAD_REQUEST"
}

```

The same error occurs when using a non-exiting dataset id.

MalteEbner:
closes #544 

## description
- fixes the bug in #544 
- fixes the bug in https://github.com/lightly-ai/lightly/issues/544#issuecomment-969390020

## Other changes
- changed the MockedApiWorkflowClient to ensure that all datasets ids passed are valid. 
- adds much more unittests, especially for the CLI upload
- reworks some of the unittests to reuse functionality. E.g. `MockedApiWorkflowClient._get_csv_reader_from_read_url()` now allows all unittests to mock the returned embeddings file on the server (upload_cli, magic_clie, test of direct upload).

## Follow-up:
- https://github.com/lightly-ai/lightly/issues/546

### DIFF
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -40,8 +40,12 @@ def _upload_cli(cfg, is_cli_call=True):
         print_as_warning('Please specify your access token.')
         cli_api_args_wrong = True
 
-    dataset_id_ok = dataset_id and len(dataset_id) > 0
-    new_dataset_name_ok = new_dataset_name and len(new_dataset_name) > 0
+    if dataset_id and len(dataset_id) > 0:
+        dataset_id_ok = True
+    else:
+        dataset_id_ok = False
+        del dataset_id
+    new_dataset_name_ok = bool(new_dataset_name and len(new_dataset_name) > 0)
     if new_dataset_name_ok and not dataset_id_ok:
         api_workflow_client = ApiWorkflowClient(token=token)
         api_workflow_client.create_dataset(dataset_name=new_dataset_name)
@@ -99,7 +103,7 @@ def _upload_cli(cfg, is_cli_call=True):
         print('Starting upload of embeddings.')
         try:
             embeddings = api_workflow_client.embeddings_api \
-                .get_embeddings_by_dataset_id(dataset_id=dataset_id)
+                .get_embeddings_by_dataset_id(dataset_id=api_workflow_client.dataset_id)
             # use latest embedding first
             embeddings = sorted(
                 embeddings,
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -59,6 +59,7 @@ def _upload_cli(cfg, is_cli_call=True):
                 'Please specify either the dataset_id of an existing dataset '
                 'or a new_dataset_name.')
             cli_api_args_wrong = True
+    del dataset_id
 
     if cli_api_args_wrong:
         print_as_warning('For help, try: lightly-upload --help')
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -121,24 +121,14 @@ def _upload_cli(cfg, is_cli_call=True):
 
         if embedding is not None:
 
-            filenames_on_server = api_workflow_client.filenames_on_server
-            with open(path_to_embeddings, 'r') as f:
-                # count number of new embedding rows
-                n_rows = -1 + sum(
-                    1 for row in csv.reader(f)
-                    if list(row)[0] not in set(filenames_on_server)
-                )
-
-            if n_rows < len(filenames_on_server): 
-                # more filenames than rows in the embedding file
-                # -> append rows from server
-                print('Appending embeddings from server.')
-                api_workflow_client.append_embeddings(
-                    path_to_embeddings,
-                    embedding.id,
-                )
-                now = datetime.now().strftime('%Y%m%d_%Hh%Mm%Ss')
-                name = f'{name}_{now}'
+            # -> append rows from server
+            print('Appending embeddings from server.')
+            api_workflow_client.append_embeddings(
+                path_to_embeddings,
+                embedding.id,
+            )
+            now = datetime.now().strftime('%Y%m%d_%Hh%Mm%Ss')
+            name = f'{name}_{now}'
 
         api_workflow_client.upload_embeddings(
             path_to_embeddings_csv=path_to_embeddings, name=name
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -142,7 +142,7 @@ def _upload_cli(cfg, is_cli_call=True):
             verbose=True
         )
 
-    if new_dataset_name_ok:
+    if new_dataset_name:
         print(f'The dataset_id of the newly created dataset is '
               f'{bcolors.OKBLUE}{api_workflow_client.dataset_id}{bcolors.ENDC}')
 
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -59,6 +59,8 @@ def _upload_cli(cfg, is_cli_call=True):
                 'Please specify either the dataset_id of an existing dataset '
                 'or a new_dataset_name.')
             cli_api_args_wrong = True
+    # delete the dataset_id as it might be an empty string
+    # Use api_workflow_client.dataset_id instead
     del dataset_id
 
     if cli_api_args_wrong:
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -40,26 +40,25 @@ def _upload_cli(cfg, is_cli_call=True):
         print_as_warning('Please specify your access token.')
         cli_api_args_wrong = True
 
-    if dataset_id and len(dataset_id) > 0:
-        dataset_id_ok = True
+    if dataset_id:
+        if new_dataset_name:
+            print_as_warning(
+                'Please specify either the dataset_id of an existing dataset '
+                'or a new_dataset_name, but not both.'
+            )
+            cli_api_args_wrong = True
+        else:
+            api_workflow_client = \
+                ApiWorkflowClient(token=token,dataset_id=dataset_id)
     else:
-        dataset_id_ok = False
-        del dataset_id
-    new_dataset_name_ok = bool(new_dataset_name and len(new_dataset_name) > 0)
-    if new_dataset_name_ok and not dataset_id_ok:
-        api_workflow_client = ApiWorkflowClient(token=token)
-        api_workflow_client.create_dataset(dataset_name=new_dataset_name)
-    elif dataset_id_ok and not new_dataset_name_ok:
-        api_workflow_client = ApiWorkflowClient(
-            token=token,
-            dataset_id=dataset_id
-        )
-    else:
-        print_as_warning(
-            'Please specify either the dataset_id of an existing dataset or a '
-            'new_dataset_name.'
-        )
-        cli_api_args_wrong = True
+        if new_dataset_name:
+            api_workflow_client = ApiWorkflowClient(token=token)
+            api_workflow_client.create_dataset(dataset_name=new_dataset_name)
+        else:
+            print_as_warning(
+                'Please specify either the dataset_id of an existing dataset '
+                'or a new_dataset_name.')
+            cli_api_args_wrong = True
 
     if cli_api_args_wrong:
         print_as_warning('For help, try: lightly-upload --help')
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -49,7 +49,7 @@ def _upload_cli(cfg, is_cli_call=True):
             cli_api_args_wrong = True
         else:
             api_workflow_client = \
-                ApiWorkflowClient(token=token,dataset_id=dataset_id)
+                ApiWorkflowClient(token=token, dataset_id=dataset_id)
     else:
         if new_dataset_name:
             api_workflow_client = ApiWorkflowClient(token=token)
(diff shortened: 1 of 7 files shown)


tests/api_workflow/mocked_api_workflow_client.py
```diff
@@ -298,7 +336,32 @@ def upload_file_with_signed_url(
...
+            row.append(i)
+            rows_csv.append(row)
+
+        # save the csv rows in a temporary in-memory string file
+        # using a csv writer and then read them as bytes
+        f = tempfile.SpooledTemporaryFile(mode="rw")
+        writer = csv.writer(f)
+        writer.writerows(rows_csv)
+        f.seek(0)
+        buffer = io.StringIO(f.read())
```
japrescott, line 357: this seems like we are loading the CSV 3-4 times into memory?!
- first when building `rows_csv`
- second when we write it to the temporary file `f`
- third when we read it via `StringIO(f.read())` in to the buffer
- forth? when we call `csv.reader(`

That seems a bit of an overkill, no? Is not possible to directly make a buffer from a dict struct?

lightly/cli/upload_cli.py
```diff
@@ -40,8 +40,12 @@ def _upload_cli(cfg, is_cli_call=True):
         print_as_warning('Please specify your access token.')
         cli_api_args_wrong = True
 
-    dataset_id_ok = dataset_id and len(dataset_id) > 0
-    new_dataset_name_ok = new_dataset_name and len(new_dataset_name) > 0
+    if dataset_id and len(dataset_id) > 0:
```
philippmwirth, line 43: if `dataset_id` is a string this is actually redundant

lightly/cli/upload_cli.py
```diff
@@ -40,8 +40,12 @@ def _upload_cli(cfg, is_cli_call=True):
...
+        dataset_id_ok = True
+    else:
+        dataset_id_ok = False
+        del dataset_id
```
philippmwirth, line 47: what, why?

lightly/cli/upload_cli.py
```diff
@@ -40,8 +40,12 @@ def _upload_cli(cfg, is_cli_call=True):
...
+    new_dataset_name_ok = bool(new_dataset_name and len(new_dataset_name) > 0)
```
philippmwirth, line 48: why parse it to bool?

lightly/cli/upload_cli.py
```diff
@@ -99,7 +103,7 @@ def _upload_cli(cfg, is_cli_call=True):
         print('Starting upload of embeddings.')
         try:
             embeddings = api_workflow_client.embeddings_api \
-                .get_embeddings_by_dataset_id(dataset_id=dataset_id)
+                .get_embeddings_by_dataset_id(dataset_id=api_workflow_client.dataset_id)
```
philippmwirth, line 106: is there a reason you didn't adopt nathan's idea of adding a return value to the `create_dataset` routine?

tests/api_workflow/mocked_api_workflow_client.py
```diff
@@ -38,6 +43,14 @@
 from lightly.openapi_generated.swagger_client.models.write_csv_url_data import WriteCSVUrlData
 
 
+def _check_dataset_id(dataset_id: str):
+    assert isinstance(dataset_id, str)
+    assert len(dataset_id) > 0
```
philippmwirth, line 48: simply `assert dataset_id` would do here

tests/api_workflow/test_api_workflow_upload_dataset.py
```diff
@@ -144,7 +144,7 @@ def failing_upload_sample(*args, **kwargs):
 
     def test_upload_dataset_twice_with_overlap(self):
 
-        all_sample_names = [f'sample_{i}.jpg' for i in range(10)]
+        all_sample_names = [f'img_{i}.jpg' for i in range(10)]
```
philippmwirth, line 147: why did you change this?

philippmwirth, line 147: why did you change this?

lightly/cli/upload_cli.py
MalteEbner: Yes, it also was before. I'll change it.

lightly/cli/upload_cli.py
MalteEbner: I wanted to delete `dataset_id`, so that no one will use it anywhere after this point, as it is an empty string. This would have prevented the implementation of the error in the first place.

lightly/cli/upload_cli.py
MalteEbner: Otherwise it might be a string.

tests/api_workflow/mocked_api_workflow_client.py
MalteEbner: This is for mocking, so performance for the files with <100 rows does not matter. Furthermore there is the problem that you cannot directly write into a string buffer and read from it byte-wise. Thus I would keep this working implementation.

tests/api_workflow/mocked_api_workflow_client.py
MalteEbner: Yes, this is true. However, I like the more explicit version.

lightly/cli/upload_cli.py
MalteEbner: Yes: if a user then creates a new dataset or switches to another dataset using the `api_workflow_client`, the `api_workflow_client.dataset_id` and `dataset_id` are again different. Thus I would prefer that always the `api_workflow_client.dataset_id`, as it is always up to date.

tests/api_workflow/test_api_workflow_upload_dataset.py
MalteEbner: Because the name needs to fit those used elsewhere: 
https://github.com/lightly-ai/lightly/blob/36e1bf05c78f928b7a75b7e253e51c00e5c79240/tests/api_workflow/mocked_api_workflow_client.py#L157
https://github.com/lightly-ai/lightly/blob/4f3e7328539683fdd4990d463da0891aaf92cc4d/tests/api_workflow/mocked_api_workflow_client.py#L309

MalteEbner: see https://github.com/lightly-ai/lightly/pull/545/files/4f3e7328539683fdd4990d463da0891aaf92cc4d#r750983079

lightly/cli/upload_cli.py
philippmwirth: then just add a comment and it's ok

lightly/cli/upload_cli.py
philippmwirth: Ok, but in the following workflow this could lead to problems:
- create a dataset X
- create a dataset Y
- I want to do something with the dataset id of X

It's clear that the returned `dataset_id` is static and hence easy to debug. However, it's not clear that the `api_workflow_client.dataset_id` is *not* static. So IMO it would be good to return the dataset id.

tests/api_workflow/test_api_workflow_upload_dataset.py
philippmwirth: I think I used different names on purpose to test the case where new data is added to an existing dataset

lightly/cli/upload_cli.py
MalteEbner: If a user wants to work with two different datasets, then two different `ApiWorkflowClient`s must be created. Most methods of the `ApiWorkflowClient` or one of its `Mixins` use the self.dataset_id anyway.

I think it would be better to make all the apis within the `ApiWorkflowClient` private and thus disallow calling `api_workflow_client.embeddings_api.get_embeddings_by_dataset_id `. Instead, there should be a method `ApiWorkflowClient.get_embeddings()`, which does the same, but a) inserts the dataset_id itself and b) creates typehints

MalteEbner: Third Solution: put the whole code in
https://github.com/lightly-ai/lightly/blob/592d8f899a427dd0763b66ef856c8ffd74179357/lightly/cli/upload_cli.py#L99-L137
into 
https://github.com/lightly-ai/lightly/blob/592d8f899a427dd0763b66ef856c8ffd74179357/lightly/api/api_workflow_upload_embeddings.py#L71-L78

lightly/cli/upload_cli.py
MalteEbner: I did and reworked the if-clauses to distinguish the 4 possible cases better.

tests/api_workflow/test_api_workflow_upload_dataset.py
MalteEbner: I just saw that they do indeed not depend on the other image names, I changed them back to a unique name.

lightly/cli/upload_cli.py
MalteEbner: I put this into another issue: https://github.com/lightly-ai/lightly/issues/568

lightly/cli/upload_cli.py
```diff
@@ -40,22 +40,28 @@ def _upload_cli(cfg, is_cli_call=True):
...
+    if dataset_id:
+        if new_dataset_name:
+            print_as_warning(
+                'Please specify either the dataset_id of an existing dataset '
+                'or a new_dataset_name, but not both.'
+            )
+            cli_api_args_wrong = True
+        else:
+            api_workflow_client = \
+                ApiWorkflowClient(token=token,dataset_id=dataset_id)
```
philippmwirth, line 52: space after comma

philippmwirth (approved):
nice! I like the new workflow a lot

PR #548, opened 2021-11-17
Detectron2 pre-training

### LINKED ISSUE #533: pretraining Detectron2 FPN models
Hi Lightly-team,

Is there a way to use Lightly for self-supervised pretraining of Detectron2 models?
I think that would be valuable for many people.
The "Lightly Backbone Playground.ipynb" colab shows the concept of using a Lightly implemented SimCLR to pretrain a torchvision resnet50.
However, the most effective Detectron2 models are resnet-FPNs (Feature Pyramid Networks).

The detectron standard Faster RCNN with resnet-FPN contains a backbone: detmodel.backbone
a RPN (region proposal network): detmodel.proposal_generator
and a ROI (region of interest): detmodel.head roi_heads
The backbone itself is a resnet-FPN and I assume it is easiest to 
train the bottom-up part of the resnet-FPN ( detmodel.backbone.bottom_up), as that is basically a resnet.

It seems doable except that the feature-shapes are mismatched.
I created a minimal working example in colab:
https://colab.research.google.com/drive/11AIdCGkHC1loQjvQnj51kXCD9TKNAib5?usp=sharing

Is there a way to get the SimCLR working with this backbone?

The colab above contains a runnable example, but I tried to capture the essence in the code below:
```
# Get original resnet model
resnet50 = torchvision.models.resnet50()
# Remove linear head
rbackbone = nn.Sequential(*list(resnet50.children())[:-1],)

# Toy input
toy_input_view_a = torch.randn((11,3,256,256)).to(device)

# Output shape of original resnet model
np.shape( rbackbone(toy_input_view_a) )
# Out: torch.Size([11, 2048, 1, 1])

# In the forward function of the file lightly/models/simclr.py:88
# f0 = self.backbone(x0).flatten(start_dim=1)
# which leads to the shape torch.Size([11, 2048]) for rbackbone's output

#However, with the detectron model
detmodel = build_model(<config_file>)
backbone = nn.Sequential(*list(detmodel.backbone.bottom_up.children()),)

# We get a different output shape to start with
np.shape( backbone(toy_input_view_a) )
# Out: torch.Size([11, 2048, 8, 8])

# Which, after the flatten in lightly simclr model lightly/models/simclr.py:88
# leads to the shape torch.Size([11, 131072])

# Thus 
simCLR = lightly.models.SimCLR(backbone, num_ftrs=2048)
out = simCLR(toy_input_view_a, toy_input_view_a)
# Leads to the error
# RuntimeError: mat1 and mat2 shapes cannot be multiplied (11x131072 and 2048x2048)
```

philippmwirth:
# Detectron2 pre-training

Closes #533.

Tutorial on how to pre-train a Detectron2 model on a retail object detection dataset and store the checkpoint.

@IgorSusmelj I didn't find any concrete reason as to why Detectron2 uses BGR. We can do a quick benchmark to check.

pdf (updated again): [Tutorial 6 Pre-train a Detectron2 Backbone with Lightly.pdf](https://github.com/lightly-ai/lightly/files/7580697/Tutorial.6.Pre-train.a.Detectron2.Backbone.with.Lightly.pdf)

### DIFF
--- a/docs/source/tutorials/package.rst
+++ b/docs/source/tutorials/package.rst
@@ -19,4 +19,4 @@ you might want to have a look at the two frameworks to understand basic concepts
    package/tutorial_simclr_clothing.rst
    package/tutorial_simsiam_esa.rst
    package/tutorial_custom_augmentations.rst
-   package/tutorial_simclr_detectron2.rst
+   package/tutorial_pretrain_detectron2.rst
--- a/docs/source/tutorials/package.rst
+++ b/docs/source/tutorials/package.rst
@@ -19,3 +19,4 @@ you might want to have a look at the two frameworks to understand basic concepts
    package/tutorial_simclr_clothing.rst
    package/tutorial_simsiam_esa.rst
    package/tutorial_custom_augmentations.rst
+   package/tutorial_simclr_detectron2.rst
--- a/docs/source/tutorials_source/package/Base-RCNN-FPN.yaml
+++ b/docs/source/tutorials_source/package/Base-RCNN-FPN.yaml
@@ -1,0 +1,42 @@
+MODEL:
+  META_ARCHITECTURE: "GeneralizedRCNN"
+  BACKBONE:
+    NAME: "build_resnet_fpn_backbone"
+  RESNETS:
+    OUT_FEATURES: ["res2", "res3", "res4", "res5"]
+  FPN:
+    IN_FEATURES: ["res2", "res3", "res4", "res5"]
+  ANCHOR_GENERATOR:
+    SIZES: [[32], [64], [128], [256], [512]]  # One size for each in feature map
+    ASPECT_RATIOS: [[0.5, 1.0, 2.0]]  # Three aspect ratios (same for all in feature maps)
+  RPN:
+    IN_FEATURES: ["p2", "p3", "p4", "p5", "p6"]
+    PRE_NMS_TOPK_TRAIN: 2000  # Per FPN level
+    PRE_NMS_TOPK_TEST: 1000  # Per FPN level
+    # Detectron1 uses 2000 proposals per-batch,
+    # (See "modeling/rpn/rpn_outputs.py" for details of this legacy issue)
+    # which is approximately 1000 proposals per-image since the default batch size for FPN is 2.
+    POST_NMS_TOPK_TRAIN: 1000
+    POST_NMS_TOPK_TEST: 1000
+  ROI_HEADS:
+    NAME: "StandardROIHeads"
+    IN_FEATURES: ["p2", "p3", "p4", "p5"]
+  ROI_BOX_HEAD:
+    NAME: "FastRCNNConvFCHead"
+    NUM_FC: 2
+    POOLER_RESOLUTION: 7
+  ROI_MASK_HEAD:
+    NAME: "MaskRCNNConvUpsampleHead"
+    NUM_CONV: 4
+    POOLER_RESOLUTION: 14
+DATASETS:
+  TRAIN: ("coco_2017_train",)
+  TEST: ("coco_2017_val",)
+SOLVER:
+  IMS_PER_BATCH: 16
+  BASE_LR: 0.02
+  STEPS: (60000, 80000)
+  MAX_ITER: 90000
+INPUT:
+  MIN_SIZE_TRAIN: (640, 672, 704, 736, 768, 800)
+VERSION: 2
--- a/docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
+++ b/docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
@@ -63,6 +63,9 @@ from detectron2.checkpoint import DetectionCheckpointer
 # We use a batch size of 64 and an input size of 128 in order to fit everything
 # on the available amount of memory on our GPU (16GB). The number of features
 # is set to the default output size of the ResNet50 backbone.
+#
+# We only train for 5 epochs because the focus of this tutorial is on the
+# integration with Detectron2.
 
 num_workers = 8
 batch_size = 64
@@ -71,7 +74,7 @@ num_ftrs = 2048
 lr = 0.001
 
 seed = 1
-max_epochs = 20
+max_epochs = 5
 
 # use cuda if possible
 device = 'cuda' if torch.cuda.is_available() else 'cpu'
@@ -85,7 +88,7 @@ ImageFile.LOAD_TRUNCATED_IMAGES = True
 # Set the path to the dataset accordingly. Additionally, make sure to set the
 # path to the config file of the Detectron2 model you want to use.
 # We will be using an RCNN with a feature pyramid network (FPN).
-data_path = '/datasets/SKU110K_fixed/resized'
+data_path = '/datasets/SKU110K_fixed/images'
 cfg_path = './Base-RCNN-FPN.yaml'
 
 # %%
--- a/docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
+++ b/docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
@@ -82,7 +82,7 @@ input_size = 128
 num_ftrs = 2048
 
 seed = 1
-max_epochs = 5
+max_epochs = 0
 
 # use cuda if possible
 device = 'cuda' if torch.cuda.is_available() else 'cpu'
@@ -141,6 +141,15 @@ simclr_backbone = torch.nn.Sequential(
     # res5 has shape bsz x 2048 x 4 x 4
     torch.nn.AdaptiveAvgPool2d(1),
 ).to(device)
+
+# %%
+#
+#
+#.. note::
+#
+#   The Detectron2 ResNet is missing the average pooling layer used to get a tensor of shape bsz x 2048.
+#   Therefore, we add an average pooling as in the `PyTorch ResNet <https://github.com/pytorch/pytorch/blob/1022443168b5fad55bbd03d087abf574c9d2e9df/benchmarks/functional_autograd_benchmark/torchvision_models.py#L147>`_.
+#
 
 # %%
 # Finally, let's build SimCLR around the backbone as shown in the other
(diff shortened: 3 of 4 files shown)


docs/source/tutorials/package.rst
```diff
@@ -19,3 +19,4 @@ you might want to have a look at the two frameworks to understand basic concepts
    package/tutorial_simclr_clothing.rst
    package/tutorial_simsiam_esa.rst
    package/tutorial_custom_augmentations.rst
+   package/tutorial_simclr_detectron2.rst
```
IgorSusmelj, line 22: I would suggest renaming the file to `tutorial_pretrain_detectron2.rst`

philippmwirth: @IgorSusmelj I changed the name and also provided the correct path to the dataset.

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
```diff
@@ -0,0 +1,254 @@
...
+- How to do self-supervised learning with the backbone
+- How to store the backbone in a checkpoint file which can be used by Detectron2
+
+Prerequisites:
+----------------
+For the purpose of this tutorial you will need to:
+
+- Install Lightly: Follow the `instructions <https://docs.lightly.ai/getting_started/install.html>`_.
+- Install Detectron2: Follow the `instructions <https://detectron2.readthedocs.io/en/latest/tutorials/install.html>`_.
+- Download a dataset for pre-training (we will use the `SKU-110K <https://github.com/eg4000/SKU110K_CVPR19>`_ dataset).
```
IgorSusmelj, line 32: Maybe mention the direct link to download the dataset we use? `http://trax-geometry.s3.amazonaws.com/cvpr_challenge/SKU110K_fixed.tar.gz`

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
```diff
@@ -0,0 +1,254 @@
...
+# path to the config file of the Detectron2 model you want to use.
+# We will be using an RCNN with a feature pyramid network (FPN).
+data_path = '/datasets/SKU110K_fixed/images'
+cfg_path = './Base-RCNN-FPN.yaml'
+
+# %%
+# Initialize the Detectron2 Model
+# --------------------------------
+# 
+# Since the output of the Detectron2 ResNet50 backbone is a dictionary with the keys
```
IgorSusmelj, line 98: It's not clear why we need this layer and which features we will pick. Maybe we can give a reference to the detectron2 docs? Or maybe we can mention that the different keys `res1` to `res5` correspond to the different stages of the model?
Also mention here already that we only care about the last stage `res5` in this example.

docs/source/tutorials_source/package/Base-RCNN-FPN.yaml
```diff
@@ -0,0 +1,42 @@
+MODEL:
```
IgorSusmelj, line 1: This config here does not differ from the official one from detectron2 (see https://github.com/facebookresearch/detectron2/blob/main/configs/Base-RCNN-FPN.yaml). Can we drop ours? and add somewhere the link to the exact config we use?

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
```diff
@@ -0,0 +1,254 @@
...
+num_ftrs = 2048
+lr = 0.001
+
+seed = 1
+max_epochs = 5
+
+# use cuda if possible
+device = 'cuda' if torch.cuda.is_available() else 'cpu'
+
+# some images in the dataset are truncated, use this to prevent errors
```
IgorSusmelj, line 82: This appears rather random. I would make sure all imports are at the beginning of the file. And can we elaborate more about what it means if images are truncated? 
How about using another dataset?

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
```diff
@@ -0,0 +1,254 @@
...
+
+    print(f'[Epoch {e:2d}] Mean Loss = {mean_loss:.2f}')
+
+
+# %%
+# Storing the checkpoint
+# -----------------------
+# Now, we can use the pre-trained backbone from the Detectron2 model. The code
+# below shows how to save it as a Detectron2 checkpoint called `my_model.pth`.
+detmodel.backbone.bottom_up = backbone[0]
```
IgorSusmelj, line 214: I would try to make it more clear that this line just uses the first module in backbone which is the original detectron2 backbone.

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
```diff
@@ -0,0 +1,254 @@
...
+# you can use the `train_net.py` script in the Detectron2 `tools`:
+#
+#
+
+# %%
+#.. code-block:: none
+#
+#   python train_net.py --config-file ../configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml \
+#       MODEL.WEIGHTS path/to/my_model.pth \
+#       MODEL.PIXEL_MEAN 123.675,116.280,103.530 \
```
IgorSusmelj, line 234: Why do we need to pass here pixel mean and std? Where are the numbers coming from?

docs/source/tutorials_source/package/Base-RCNN-FPN.yaml
philippmwirth: It says in the text that you need to download the config file. I just downloaded it permanently so the tutorial runs through.

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
philippmwirth: We could use a different dataset, I'll check.

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
philippmwirth: It's in the "note" right below

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
```diff
@@ -0,0 +1,265 @@
...
+# %%
+# Next, we can build the Detectron2 model and extract the ResNet50 backbone as
+# follows:
+
+detmodel = modeling.build_model(cfg)
+
+backbone = torch.nn.Sequential(
+    detmodel.backbone.bottom_up,
+    SelectStage('res5'),
+    torch.nn.AdaptiveAvgPool2d(1),
```
MalteEbner, line 142: Why do we need this?

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
```diff
@@ -0,0 +1,265 @@
...
+# The checkpoint from above can now be used by any Detectron2 script. For example,
+# you can use the `train_net.py` script in the Detectron2 `tools`:
+#
+#
+
+# %%
+#.. code-block:: none
+#
+#   python train_net.py --config-file ../configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml \
+#       MODEL.WEIGHTS path/to/my_model.pth \
```
MalteEbner, line 242: The `path/to/my_model.pth` should match the one you set in line 225/226 in my eyes (`./my_model.pth`).
Perhaps you could use e.g. `path/to/checkpoint` as `save_dir` in line 225.

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+# %%
+# Storing the checkpoint
+# -----------------------
+# Now, we can use the pre-trained backbone from the Detectron2 model. The code
+# below shows how to save it as a Detectron2 checkpoint called `my_model.pth`.
+
+# get the first module from the backbone (i.e. the detectron2 ResNet)
+detmodel.backbone.bottom_up = backbone[0]
```
MalteEbner, line 223: Could you give an overview of how the detmodel in total looks like? It is really hard to imagine for me, thus I don't understand why you have to choose which subpart of the detmodel at some points. I guess it is similar to
```
detmodel
  L backbone
      L bottom_up
        L res1
        L res2
        ...
        L res5
  L SelectStage
  L AdaptiveAvgPool2d
  L SimCLRProjectionHead
```
However, what I don't get is wether there is a difference between the `detmodel.backbone` and `backbone`. Perhaps you could rename to make this clearer. I assume that`backbone`refers to the `simclr_backbone`, which is made up of the `detmodel.backbone` and the 2 additional layers.

MalteEbner (commented):
I can follow the tutorial well on the high level, but there are some low-level points that are not so clear to me.

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
philippmwirth: the `train_net.py` is in the detectron repository so it doesn'tm ake sense to use the same path twice.

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
philippmwirth: I'll clarify

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
IgorSusmelj: Ok, so the numbers come from the lightly augmentations that use ImageNet normalization? (that's the part I would add to make it clear)

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
MalteEbner: It would still rename `backbone` to `simclr_backbone`, as they are different.

MalteEbner (commented):
You clarified parts of my questions, but some are still unanswered.

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
philippmwirth: Actually I was wrong. Now I just added a note that they are from ImageNet and can be replaced depending on the dataset.

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
philippmwirth: Done

MalteEbner (commented):
I still don't know why the `AdaptiveAvgPool2d` is needed, see my question in https://github.com/lightly-ai/lightly/pull/548#discussion_r754871840

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
MalteEbner: Do I get this right? The AdaptiveAvgPool2d takes the bsz x 2048 x 4 x 4 input and averages it to a bsz x 2048 output. Why not do a flatten here or why not take subset? Why do the average pooling?

philippmwirth:
> I still don't know why the `AdaptiveAvgPool2d` is needed, see my question in [#548 (comment)](https://github.com/lightly-ai/lightly/pull/548#discussion_r754871840)

This was clarified, right?

MalteEbner (approved):
Why the `AdaptiveAvgPool2d` was clarified to me verbally, but not in the code yet. 

You could tell something like "The detectron resnet is missing the average pooling layer used to get a tensor of shape bsz x 2048. We use the same average pooling as in the standard pytorch resnet: https://github.com/pytorch/pytorch/blob/1022443168b5fad55bbd03d087abf574c9d2e9df/benchmarks/functional_autograd_benchmark/torchvision_models.py#L147"

PR #566, opened 2021-11-18
precompute VideoDataset.length() to prevent squared complexity

### LINKED ISSUE #562: LightlyDataset for videos has squared complexity
## Description
When creating/using a video dataset, it calls the function `VideoDataset.get_filename(self, index)`/`VideoDataset__getitem__(self, index)` for every frame. Both of these methods do an index check and thus call `self.__len__()`. 
https://github.com/lightly-ai/lightly/blob/592d8f899a427dd0763b66ef856c8ffd74179357/lightly/data/_video.py#L334-L338

However, this method iterates over all `video_timestamps`. Thus, the index check has a squared complexity w.r.t the number of videos. 

I could reproduce this squared complexity with measurements on datasets and with profiling.

## Task
fix the squared complexity

MalteEbner:
closes #562 

## Description
precomputes the VideoDataset.length() as it is used everytime a single frame is accessed.

## Profiling results
Measured the duration of creating a `LightlyDataset()` on an input directory with videos with 91 frames each.

| number of videos  | 100  | 300  | 1000  | 3000  |
|-----------------------------------|------|------|-------|-------|
| duration before fix               | 2.33 | 6.88 | 28.96 | 128.6 |
| duration after fix                | 2.3  | 6.63 | 23.04 | 76.87 |

## Profiling with 1000 images before bugfix
image

### DIFF
--- a/lightly/data/_video.py
+++ b/lightly/data/_video.py
@@ -334,8 +334,18 @@ class VideoDataset(datasets.VisionDataset):
     def __len__(self):
         """Returns the number of samples (frames) in the dataset.
 
+        This can be precomputed, because self.video_timestamps is only
+        set in the __init__
         """
-        return sum((len(ts) for ts in self.video_timestamps))
+        try:
+            #raise AttributeError
+            return self._precomputed_length
+        except AttributeError:
+            self._precomputed_length = sum((
+                len(ts) for ts
+                 in self.video_timestamps
+            ))
+            return self._precomputed_length
 
     def get_filename(self, index):
         """Returns a filename for the frame at index.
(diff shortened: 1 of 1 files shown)


lightly/data/_video.py
```diff
@@ -271,6 +271,9 @@ def __init__(self,
 
         self.videos = videos
         self.video_timestamps = video_timestamps
+        self._precomputed_length = sum((
```
philippmwirth, line 274: I would just call it `_length` the "precomputed" does not add any information.

Also, in terms of formatting, I'd prefer either
```python
self._length = sum((
    len(ts) for ts in self.video_timestamps))
```
or
```python
self._length = sum((
    len(ts) for ts in self.video_timestamps
))
```

philippmwirth (approved):
I think `_length` is the better name but other than that this makes sense to me.

PR #577, opened 2021-11-22
Update custom augmentations tutorial with low-level blocks

### LINKED ISSUE #556: Rewrite Custom Augmentations Tutorial with Low-Level Building Blocks
Rewrite the Custom Augmentations tutorial with low-level building blocks as we want to remove mid-level blocks (#493)
Specifically:

- [x] Remove dependency on `lightly.models.MoCo` and add new model
- [x] Remove dependency on `lightly.embedding.SelfSupervisedEmbedding`
  - [x] Switch to pytorch lightning for training
  - [x] Add code to generate embeddings
- [x] Check that text is consistent with changes

guarin:
Closes #556 

- Add MoCo model
- Remove SelfSupervisedEmbedding dependency
- Add PyTorch Lightning trainer
- Add embedding generation code

[custom_augmentations.pdf](https://github.com/lightly-ai/lightly/files/7579381/custom_augmentations.pdf)

### DIFF
--- a/docs/source/tutorials_source/package/tutorial_custom_augmentations.py
+++ b/docs/source/tutorials_source/package/tutorial_custom_augmentations.py
@@ -45,6 +45,13 @@ from sklearn.preprocessing import normalize
 from PIL import Image
 import numpy as np
 import pandas
+import copy
+
+from lightly.models.modules.heads import MoCoProjectionHead
+from lightly.models.utils import deactivate_requires_grad
+from lightly.models.utils import update_momentum
+from lightly.models.utils import batch_shuffle
+from lightly.models.utils import batch_unshuffle
 
 # %%
 # Configuration
@@ -63,13 +70,13 @@ max_epochs = 50
 
 # %%
 # Let's set the seed for our experiments
+
 pl.seed_everything(seed)
 
 # %%
 # Set the path to our dataset
 
 path_to_data = '/datasets/vinbigdata/train_small'
-
 
 # %%
 # Setup custom data augmentations
@@ -127,7 +134,6 @@ class GaussianNoise:
         noise = torch.normal(torch.zeros(sample.shape), sigma)
         return sample + noise
 
-
 # %%
 # Now that we have implemented our custom augmentations, we can combine them
 # with available augmentations from the torchvision library to get to the same
@@ -139,7 +145,6 @@ class GaussianNoise:
 # the single color channel three times. The reason for this is that our ResNet expects
 # a three color channel input. This step can be skipped if a different backbone network
 # is used.
-
 
 # compose the custom augmentations with available augmentations
 transform = torchvision.transforms.Compose([
@@ -199,7 +204,6 @@ collate_fn = lightly.data.BaseCollateFunction(transform)
 #   grayscale image is loaded that way, all pixel values above 255 are simply clamped.
 #   Therefore, we overwrite the default image loader with our custom one.
 
-
 def tiff_loader(f):
     """Loads a 16-bit tiff image and returns it as a numpy array.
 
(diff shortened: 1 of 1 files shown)


docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -371,6 +431,7 @@ def plot_knn_multilabels(embeddings, multilabels, samples_idx, n_neighbors=50):
         bars1 = multilabels[idx]
         bars2 = np.mean(multilabels[indices[idx]], axis=0)
 
+        plt.title(filenames[idx])
```
MalteEbner, line 434: Why don't I see the title in the pdf?

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -380,4 +441,6 @@ def plot_knn_multilabels(embeddings, multilabels, samples_idx, n_neighbors=50):
 # plot the distribution of the multilabels of the k nearest neighbors of 
 # the three example images at index 4111, 3340, 1796
 k = 20
-plot_knn_multilabels(embeddings, multilabels, [4111, 3340, 1796], n_neighbors=k)
+plot_knn_multilabels(
+    embeddings, multilabels, [4111, 3340, 1796], fnames, n_neighbors=k
+)
```
MalteEbner, line 446: Cited from the tutorial:
_[We want to] evaluate how good the learned representations really are._

But after the plots there is no description/evaluation at all. I know this was missing in the tutorial already before, but it would be nice as a reader to follow the interpretation/evaluation of the author to find out whether the augmentations are good or bad.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -380,4 +441,6 @@ def plot_knn_multilabels(embeddings, multilabels, samples_idx, n_neighbors=50):
 # plot the distribution of the multilabels of the k nearest neighbors of 
 # the three example images at index 4111, 3340, 1796
 k = 20
```
MalteEbner, line 443: If you hardcode the sample indices anyway, you can also inline the `k=20` directly as `n_neighbors=20`

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -371,6 +431,7 @@ def plot_knn_multilabels(embeddings, multilabels, samples_idx, n_neighbors=50):
         bars1 = multilabels[idx]
```
MalteEbner, line 431: line 367/427: The samples are not picked randomly, they are hardcoded.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
```diff
@@ -317,10 +359,28 @@ def tiff_loader(f):
     num_workers=num_workers
 )
 
-# get vector representations for all images in the dataset
-encoder.to('cuda')
-embeddings, _, fnames = encoder.embed(dataloader_test, device='cuda')
-embeddings = normalize(embeddings)
+# next we add a small helper function to generate embeddings of our images
+def generate_embeddings(model, dataloader):
```
MalteEbner, line 363: Just a brainfart: If we use `generate_embeddings()` in every tutorial, why not have it somewhere as a common function?

Furthermore, I would then change the input argument `model` to `backbone`, to not rely on the argument `model` having the attribute `backbone`.

MalteEbner (commented):
If this was only about updating the tutorial so that it works, it would be an approve.

However, I think we should use this PR to also make the tutorial a bit nicer, thus some ideas on that.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
guarin: Ah, the pdf did not yet include this change. Will update it.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
guarin: I agree. We should maybe also consider changing the evaluation in the first place. It is not very easy to understand.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
guarin: True, done :)

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
guarin: Thanks, changed it.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
guarin: > Just a brainfart: If we use generate_embeddings() in every tutorial, why not have it somewhere as a common function?

Depends a bit on the goal of the tutorials. I think it is ok to show the user how to generate embeddings in detail. Especially now that we switched to low-level building blocks. But I agree that it generates a lot of copy paste code.

> I would then change the input argument model to backbone, to not rely on the argument model having the attribute backbone.

I agree that it would be nicer to pass only the backbone but it would add more complexity elsewhere. The update code would be:

```
model.eval()
backbone = model.backbone
device = model.device # backbone has no device attribute
generate_embeddings(backbone, dataloader, device)
```

The issue is that backbone is juste a `nn.Sequence` and has not the usual attributes like `.eval()` and `.device`. This would add even more copy paste code to the tutorials.

MalteEbner: Good point, the model should be kept.

docs/source/tutorials_source/package/tutorial_custom_augmentations.py
guarin: Ok thanks!

MalteEbner (approved):
I love how you progress so fast!

PR #580, opened 2021-11-22
Malte lig 301 fix + refactor api_workflow_client

### LINKED ISSUE #568: Refactor api_workfkow_client.upload_embeddings() and make apis private
## Description
The bug in https://github.com/lightly-ai/lightly/issues/544
was caused by two problems in the api connection which could be solved easily:

1. 
It would be better to make all the apis within the `ApiWorkflowClient` private and thus disallow calling `api_workflow_client.embeddings_api.get_embeddings_by_dataset_id` . Instead, there should be a method `ApiWorkflowClient.get_embeddings()`, which does the same, but a) inserts the `dataset_id` itself and b) creates typehints

2. 
The `ApiWorkflowClient.upload_embeddings()` should support a full workflow including
- checking if an embedding with the same name already exists
- appending the server embeddings to the client embeddings
- uploading the (appended) embeddings

Thus, put the whole code in
https://github.com/lightly-ai/lightly/blob/9b148e55ab6db9e05a8c56ec64065265718646b9/lightly/cli/upload_cli.py#L103-L131
into 
https://github.com/lightly-ai/lightly/blob/592d8f899a427dd0763b66ef856c8ffd74179357/lightly/api/api_workflow_upload_embeddings.py#L71-L78

### LINKED ISSUE #576: Make api_workflow_client.filenames_on_server a method again and remove the squared complexity
When changing
`api_workflow_client.filenames_on_server` to not save the filenames anymore, but load them from the server again, we introduced two problems:

- Developers will think that this is a cheap operation and thus use it too much.
- It already cause `lightly-download` to have an operation with squared complexity: for every sample in the tag, all filenames are downloaded:
https://github.com/lightly-ai/lightly/blob/dc74e2fde10b0c0c90887e9362baaf98b7a51b62/lightly/cli/download_cli.py#L69

TODO:
- Rename `api_workflow_client.filenames_on_server` to `api_workflow_client.download_filenames_on_server()` to make it clearer that this is an expensive operation
- Change all usages accordingly and try to remove further squared complexities.

MalteEbner:
close #568 
closes #576 

## Description
- Made all `api_workflow_client` apis private (e.g. rename `tags_api` to `_tags_api`).
- Renamed `filenames_on_server` to `download_filenames_from_server()` to make clear that this is not a cheap property, but an expensive download. This had also caused some squared complexities.
- Created a `_TagsMixin` with a method `get_tag_and_filenames`. It raises an error if the `tag_name` or `tag_id` is invalid.
- Put the bitmask when downloading a tag out of the `download_cli` inside the` api_workflow_client. get_tag_and_filenames`.
- Put the upload embeddings when appending stuff out of the `upload_cli` inside the` api_workflow_client.upload_embeddings`.
- Wrote a test class for the `ApiWorkflowTags`

## Tests
- unmocked tests run through: https://github.com/lightly-ai/lightly/actions/runs/1517025903

Open questions:
- How to handle invalid CLI arguments in general? Warning and return or ValueError? Currently, this is inconsistent.
- How to cache the `get_filenames()` better? E.g. in the `active_learning_agent`, each call of the properties (e.g. `query_set`) makes a new download.

### DIFF
--- a/lightly/active_learning/agents/agent.py
+++ b/lightly/active_learning/agents/agent.py
@@ -70,7 +70,7 @@ class ActiveLearningAgent:
 
         # build lookup table for tag_name to tag_id
         tag_name_id_dict = {}
-        for tag in self.api_workflow_client._get_all_tags():
+        for tag in self.api_workflow_client.get_all_tags():
             tag_name_id_dict[tag.name] = tag.id
         # use lookup table to set ids
         self._query_tag_id = tag_name_id_dict[query_tag_name]
--- a/lightly/active_learning/agents/agent.py
+++ b/lightly/active_learning/agents/agent.py
@@ -120,7 +120,7 @@ class ActiveLearningAgent:
 
         """
         return self._query_tag_bitmask.masked_select_from_list(
-            self.api_workflow_client.download_filenames_from_server()
+            self.api_workflow_client.get_filenames()
         )
 
     @property
@@ -129,7 +129,7 @@ class ActiveLearningAgent:
 
         """
         return self._preselected_tag_bitmask.masked_select_from_list(
-            self.api_workflow_client.download_filenames_from_server()
+            self.api_workflow_client.get_filenames()
         )
 
     @property
@@ -140,7 +140,7 @@ class ActiveLearningAgent:
         # unlabeled set is the query set minus the preselected set
         unlabeled_tag_bitmask = self._query_tag_bitmask - self._preselected_tag_bitmask
         return unlabeled_tag_bitmask.masked_select_from_list(
-            self.api_workflow_client.download_filenames_from_server()
+            self.api_workflow_client.get_filenames()
         )
 
     @property
@@ -157,7 +157,7 @@ class ActiveLearningAgent:
         # added set is new preselected set minus the old one
         added_tag_bitmask = self._preselected_tag_bitmask - self._old_preselected_tag_bitmask
         return added_tag_bitmask.masked_select_from_list(
-            self.api_workflow_client.download_filenames_from_server()
+            self.api_workflow_client.get_filenames()
         )
 
 
--- a/lightly/active_learning/agents/agent.py
+++ b/lightly/active_learning/agents/agent.py
@@ -89,7 +89,7 @@ class ActiveLearningAgent:
 
         """
         # get query tag from api and set bitmask accordingly
-        query_tag_data = self.api_workflow_client.tags_api.get_tag_by_tag_id(
+        query_tag_data = self.api_workflow_client._tags_api.get_tag_by_tag_id(
             self.api_workflow_client.dataset_id,
             tag_id=self._query_tag_id
         )
@@ -106,7 +106,7 @@ class ActiveLearningAgent:
             preselected_tag_bitmask = BitMask.from_hex('0x0')
         else:
             # get preselected tag from api and set bitmask accordingly
-            preselected_tag_data = self.api_workflow_client.tags_api.get_tag_by_tag_id(
+            preselected_tag_data = self.api_workflow_client._tags_api.get_tag_by_tag_id(
                 self.api_workflow_client.dataset_id,
                 tag_id=self._preselected_tag_id
             )
@@ -120,7 +120,7 @@ class ActiveLearningAgent:
 
         """
         return self._query_tag_bitmask.masked_select_from_list(
-            self.api_workflow_client.filenames_on_server
+            self.api_workflow_client.download_filenames_from_server()
         )
 
     @property
@@ -129,7 +129,7 @@ class ActiveLearningAgent:
 
         """
         return self._preselected_tag_bitmask.masked_select_from_list(
-            self.api_workflow_client.filenames_on_server
+            self.api_workflow_client.download_filenames_from_server()
         )
 
     @property
@@ -140,7 +140,7 @@ class ActiveLearningAgent:
         # unlabeled set is the query set minus the preselected set
         unlabeled_tag_bitmask = self._query_tag_bitmask - self._preselected_tag_bitmask
         return unlabeled_tag_bitmask.masked_select_from_list(
-            self.api_workflow_client.filenames_on_server
+            self.api_workflow_client.download_filenames_from_server()
         )
 
     @property
@@ -157,7 +157,7 @@ class ActiveLearningAgent:
         # added set is new preselected set minus the old one
         added_tag_bitmask = self._preselected_tag_bitmask - self._old_preselected_tag_bitmask
         return added_tag_bitmask.masked_select_from_list(
-            self.api_workflow_client.filenames_on_server
+            self.api_workflow_client.download_filenames_from_server()
         )
 
 
--- a/lightly/api/api_workflow_download_dataset.py
+++ b/lightly/api/api_workflow_download_dataset.py
@@ -88,7 +88,8 @@ class _DownloadDatasetMixin:
 
         indices = BitMask.from_hex(tag.bit_mask_data).to_indices()
         sample_ids = [sample_ids[i] for i in indices]
-        filenames = [self.filenames_on_server[i] for i in indices]
+        filenames_on_server = self.download_filenames_from_server()
+        filenames = [filenames_on_server[i] for i in indices]
 
         if verbose:
             print(f'Downloading {len(sample_ids)} images:', flush=True)
--- a/lightly/api/api_workflow_download_dataset.py
+++ b/lightly/api/api_workflow_download_dataset.py
@@ -72,7 +72,7 @@ class _DownloadDatasetMixin:
             )
 
         # check if tag exists
-        available_tags = self._get_all_tags()
+        available_tags = self.get_all_tags()
         try:
             tag = next(tag for tag in available_tags if tag.name == tag_name)
         except StopIteration:
--- a/lightly/api/api_workflow_download_dataset.py
+++ b/lightly/api/api_workflow_download_dataset.py
@@ -88,7 +88,7 @@ class _DownloadDatasetMixin:
 
         indices = BitMask.from_hex(tag.bit_mask_data).to_indices()
         sample_ids = [sample_ids[i] for i in indices]
-        filenames_on_server = self.download_filenames_from_server()
+        filenames_on_server = self.get_filenames()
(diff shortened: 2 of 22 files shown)


lightly/api/api_workflow_download_dataset.py
```diff
@@ -81,22 +81,23 @@ def download_dataset(self,
...
         # get sample ids
-        sample_ids = self.mappings_api.get_sample_mappings_by_dataset_id(
+        sample_ids = self._mappings_api.get_sample_mappings_by_dataset_id(
             self.dataset_id,
             field='_id'
         )
 
         indices = BitMask.from_hex(tag.bit_mask_data).to_indices()
         sample_ids = [sample_ids[i] for i in indices]
-        filenames = [self.filenames_on_server[i] for i in indices]
```
MalteEbner, line 91: This was a squared complexity.

tests/api_workflow/test_api_workflow_upload_embeddings.py
```diff
@@ -88,30 +88,21 @@ def test_upload_comma_filenames(self):
...
-        embedding_name = self.api_workflow_client.embeddings_api.embeddings[0].name
+        embedding_name = self.api_workflow_client._embeddings_api.embeddings[0].name
         self.api_workflow_client.set_embedding_id_by_name(embedding_name)
 
     def test_set_embedding_id_failure(self):
         embedding_name = "blibblabblub"
         with self.assertRaises(ValueError):
             self.api_workflow_client.set_embedding_id_by_name(embedding_name)
 
-    def test_upload_existing_embedding(self):
```
MalteEbner, line 99: Removed, because already covered by the following tests (testing with appending).

lightly/active_learning/agents/agent.py
```diff
@@ -157,7 +157,7 @@ def added_set(self):
         # added set is new preselected set minus the old one
         added_tag_bitmask = self._preselected_tag_bitmask - self._old_preselected_tag_bitmask
         return added_tag_bitmask.masked_select_from_list(
-            self.api_workflow_client.filenames_on_server
+            self.api_workflow_client.download_filenames_from_server()
```
philippmwirth, line 160: would call it `get_filenames_from_server` and not `download_filenames...`

philippmwirth: or even `get_filenames` which would be consistent with the names of the other functions, e.g. `get_datasets`

lightly/api/api_workflow_tags.py
```diff
@@ -0,0 +1,77 @@
...
+from lightly.openapi_generated.swagger_client import TagData, \
+    TagArithmeticsRequest, TagArithmeticsOperation, TagBitMaskResponse
+
+
+class _TagsMixin:
+
+    def _get_all_tags(self) -> List[TagData]:
+        """ Gets all tags on the server
+
+        Returns:
```
philippmwirth, line 13: missing return

philippmwirth: and there's no need to make this function private

lightly/api/api_workflow_tags.py
```diff
@@ -0,0 +1,77 @@
...
+
+        """
+        return self._tags_api.get_tags_by_dataset_id(self.dataset_id)
+
+    def get_tag_and_filenames(
```
philippmwirth, line 18: call this `get_tag_with_filenames`

philippmwirth: and I would just remove the argument "filenames_on_server" and simplify the workflow

philippmwirth: actually, I would split the function into three:

```python
def get_tag_by_id(self, tag_id: str):
    pass

def get_tag_by_name(self, tag_name: str):
    # convert tag_name to tag_id here
    return self.get_tag_by_id(tag_id)

def get_filenames_in_tag(self, tag):
    pass
```

philippmwirth: If you do it this way you can also keep the optional `filenames_on_server` argument because then the function isn't such a beast

lightly/api/api_workflow_upload_embeddings.py
```diff
@@ -70,21 +71,29 @@ def upload_embeddings(self, path_to_embeddings_csv: str, name: str):
...
-            return
+            self._embeddings_api.get_embeddings_by_dataset_id(
+                dataset_id=self.dataset_id
+            )
+        embeddings_on_server_dict = {
+            embedding.name: embedding
+            for embedding in embeddings_on_server
+        }
+
+        if name in embeddings_on_server_dict.keys():
```
philippmwirth, line 82: This check probably won't work correctly. Did you try and upsize a dataset with this?

Because the name of the embedding in the API will be something like `default_DATE_TIME` whereas the name you are passing is only `default`

lightly/cli/upload_cli.py
```diff
@@ -103,35 +103,6 @@ def _upload_cli(cfg, is_cli_call=True):
...
-            embeddings = sorted(
-                embeddings,
-                key=lambda x: x.created_at,
-                reverse=True
-            )
-            # find the latest embedding that starts with the name
-            # e.g. default_20211018_12h00m00s
-            embedding = next(
-                embedding for embedding in embeddings
-                if embedding.name.startswith(name)
```
philippmwirth, line 119: see, here I check with `startswith`

philippmwirth (changes requested):

lightly/api/api_workflow_upload_embeddings.py
MalteEbner: I reworked it completely to use the startswith

lightly/api/api_workflow_tags.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def get_tag_by_id(self, tag_id: str) -> TagData:
+        tag_data = self._tags_api.get_tag_by_tag_id(self.dataset_id, tag_id)
+        return tag_data
+
+    def get_tag_by_name(self, tag_name: str) -> TagData:
+        tag_name_id_dict = {tag.name: tag.id for tag in self.get_all_tags()}
+        tag_id = tag_name_id_dict.get(tag_name, None)
+        if tag_id is None:
+            raise ValueError(f'Your tag_name is invalid: {tag_name}.')
```
philippmwirth, line 27: The `tag_name` is not really invalid but it doesn't exist

lightly/api/api_workflow_tags.py
```diff
@@ -0,0 +1,72 @@
...
+                    body=tag_arithmetics_request, dataset_id=self.dataset_id
+                )
+            bit_mask_data = bit_mask_response.bit_mask_data
+        else:
+            bit_mask_data = tag_data.bit_mask_data
+
+        if not filenames_on_server:
+            filenames_on_server = self.get_filenames()
+
+        chosen_samples_ids = BitMask.from_hex(bit_mask_data).to_indices()
```
philippmwirth, line 69: there's a function in BitMask for exactly this: https://github.com/lightly-ai/lightly/blob/006891e67f819a7d1250db2e324c3fc1a163c45b/lightly/api/bitmask.py#L202

lightly/api/api_workflow_upload_embeddings.py
```diff
@@ -14,6 +15,8 @@
 from lightly.utils.io import check_filenames
 
 
+class EmbeddingDoesNotExistError(ValueError):
```
philippmwirth, line 18: Here you define an error for when an embedding does not exist but for the tag you don't - let's try to be consistent

lightly/api/api_workflow_upload_embeddings.py
```diff
@@ -29,29 +32,37 @@ def _get_csv_reader_from_read_url(self, read_url: str):
 
         return reader
 
+    def set_embedding_id(self):
```
philippmwirth, line 35: The naming is confusing here: It says "set" but I can't set it to anything and I also don't know to what it's being set

philippmwirth: What do we need this function for?

lightly/api/api_workflow_upload_embeddings.py
```diff
@@ -68,23 +79,25 @@ def upload_embeddings(self, path_to_embeddings_csv: str, name: str):
...
-            self.embeddings_api.get_embeddings_by_dataset_id(dataset_id=self.dataset_id)
-        names_embeddings_on_server = [embedding.name for embedding in embeddings_on_server]
 
-        if name in names_embeddings_on_server:
-            print(f"Aborting upload, embedding with name='{name}' already exists.")
-            self.embedding_id = next(embedding for embedding in embeddings_on_server if embedding.name == name).id
-            return
+        # Try to append the embeddings on the server, if they exist
+        try:
+            embedding = self.get_embedding_by_name(name)
```
philippmwirth, line 85: explicitly set the `ignore_suffix` flag here to prevent bugs in the future

philippmwirth: and for readibility

philippmwirth (changes requested):

lightly/api/api_workflow_upload_embeddings.py
MalteEbner: We need it for sampling:
https://github.com/lightly-ai/lightly/blob/072fe3fd8d2244540bcc4bebef22de9fbcdc2101/lightly/api/api_workflow_sampling.py#L84-L93

lightly/api/api_workflow_upload_embeddings.py
MalteEbner: I had added one for the embedding to catch this specific error later. Now I added one for the tag.

lightly/api/api_workflow_upload_embeddings.py
MalteEbner: I have renamed the method to `set_latest_embedding_id` and added a docstring.

philippmwirth: Your renaming doesn't resolve the problem, it's still a setter without an argument which is really strange. Now it just seems like I can "set the latest id" but I want to set the local id to the latest id on the server. For example "set_embedding_id_to_latest"

philippmwirth (approved):
Please test it thoroughly before merging

Issue #586, opened 2021-11-23
Use structured configs to enforce types in command-line tool


philippmwirth:
# Use structured configs to enforce types in command-line tool

Using structured configs is a better solution for #575. However, it requires `dataclasses` which is only available from Python 3.7. There is a backport for dataclasses so we could also try this but we should be careful and test thoroughly.

Another option would be to "hack" something together with different config files for Python 3.6 and Python >= 3.7.

Docs for structured configs with hydra: https://hydra.cc/docs/tutorials/structured_config/schema/

For my experiments I used the following code which enforced types for the `download_cli`:

`lightly/cli/config/config.yaml`:
```yaml

#defaults:
# - base_config

### i/o
# The following arguments specify input and output locations
# of images, embeddings, and checkpoints.
input_dir: ''                 # Path to input directory which holds images.
output_dir: ''                # Path to directory which should store downloads.
embeddings: ''                # Path to csv file which holds embeddings.
checkpoint: ''                # Path to a model checkpoint. If left empty, a pre-trained model
                              # will be used.
label_dir: ''                 # Path to the input directory which holds the labels.
label_names_file: ''          # Path to a yaml file having the label names under the value 'names'
custom_metadata: ''           # Path to a json file in COCO format containing additional metadata

### Lightly platform
# The following arguments are required for requests to the
# Lightly platform.
token: ''                     # User access token to the Lightly platform.
dataset_id: ''                # Identifier of the dataset on the Lightly platform.
new_dataset_name: ''          # Name of the new dataset to be created on the Lightly platform
upload: 'full'                # Whether to upload full images, thumbnails only, or metadata only.
                              # Must be one of ['full', 'thumbnails', 'none']
resize: -1                    # Allow resizing of the images before uploading, usage =-1, =x
embedding_name: 'default'     # Name of the embedding to be used on the Lightly platform.
emb_upload_bsz: 32            # Number of embeddings which are uploaded in a single batch.
tag_name: 'initial-tag'       # Name of the requested tag on the Lightly platform.
exclude_parent_tag: False     # If true, only the samples in the defined tag, but without the parent tag, are taken.

### training and embeddings
pre_trained: True             # Whether to use a pre-trained model or not
crop_padding: 0.1             # The padding to use when cropping

# model namespace: Passed to lightly.models.ResNetGenerator.
model:
  name: 'resnet-18'           # Name of the model, currently supports popular variants:
[...]

PR #587, opened 2021-11-23
lightly download fails for integer tag

### LINKED ISSUE #575: lightly-download fails for integer tag names
lightly-download fails for integer tag names

For the `tag_name` `1000`, the following warning appears:
> Possible bug: I get `warnings.warn(f'The specified tag {tag_name} does not exist`

The source of the problem is probably in this line:
https://github.com/lightly-ai/lightly/blob/db33b15de6f77e50b0c815c4c405a8fb371d22e7/lightly/cli/download_cli.py#L44

Current guess: Either the api sends the string as a number or the command-line tool parses the string as a number which makes the lookup fail.

philippmwirth:
# lightly download fails for integer tag name

Closes #575. A better solution is proposed and up for discussion, see #586.

- Adds a unit test to reproduce #575.
- Fixes the problem by converting the `tag_name` to string

### DIFF
--- a/lightly/cli/download_cli.py
+++ b/lightly/cli/download_cli.py
@@ -27,9 +27,10 @@ from lightly.openapi_generated.swagger_client import TagData, TagArithmeticsRequ
 
 
 def _download_cli(cfg, is_cli_call=True):
-    tag_name = cfg['tag_name']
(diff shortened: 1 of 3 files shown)


guarin (approved):
Looks good :)

PR #589, opened 2021-11-24
Bump version to 1.2.0

guarin:


lightly/__init__.py
```diff
@@ -74,8 +74,8 @@
 # Copyright (c) 2020. Lightly AG and its affiliates.
 # All Rights Reserved
 
-__name__ = 'lightly'
-__version__ = '1.1.22'
+__name__ = "lightly"
+__version__ = "1.2.0"
```
philippmwirth, line 78: We should decide for the whole project whether we want to do double quotes or single quotes.

philippmwirth:
Can we make sure all the issues which need to be done are done before merging this? Are there any open questions?

MalteEbner (approved):

PR #599, opened 2021-11-29
Fix SimSiam Tutorial

### LINKED ISSUE #595: Update SimSiam module in SimSiam tutorial
The SimSiam module in the SimSiam tutorial requires currently `x0` and `x1` in the forward pass but they are independent. We can simplify the forward function to accept only a single `x`.

guarin:
Closes #595 

Improves SimSiam model in tutorial and fixes bug where arguments were not correctly passed to loss function.


guarin:
Includes examples by accident. Will create a new pr.

PR #611, opened 2021-11-30
use the same caching and a pip install for all CI

MalteEbner:
## Description
see issue and code

## Duration after caching:
- check setup.py from 2-3minutes down to 1min 4s
- unittests: from 4min down to 2min 41 s

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,11 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python }}
-    - name: Install Dependencies
+    - uses: actions/cache@v2
+      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -23,7 +23,11 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python }}
-    - name: Install Dependencies
+    - uses: actions/cache@v2
+      with:
+        path: ${{ env.pythonLocation }}
+        key: cache_v1_${{ env.pythonLocation }}-${{ hashFiles('requirements/**') }}
```
japrescott, line 29: very interesting!

japrescott (approved):
awesome!

PR #618, opened 2021-12-02
Update docs version

guarin:
Update docs version and link copyright in docs footer to lightly website

Closes #618


docs/source/conf.py
```diff
@@ -24,7 +24,7 @@
 author = 'Philipp Wirth, Igor Susmelj'
 
 # The full version, including alpha/beta/rc tags
-release = '1.0.0'
+release = '1.2.0'
```
philippmwirth, line 27: You could try to get this on from `../lightly/__init__.py`

guarin: Works :)

docs/source/conf.py
```diff
@@ -15,16 +15,19 @@
 sys.path.insert(0, os.path.abspath('../..'))
 
 import sphinx_rtd_theme
+import lightly
```
philippmwirth, line 18: are you sure this doesn't import an installed version of lightly? which might mess up the version... so the procedure would be 
1. install lightly from source
2. release the docs

docs/source/conf.py
```diff
@@ -15,16 +15,19 @@
...
 project = 'lightly'
-copyright = '2020, Lightly AG'
+copyright_year = '2020'
+copyright = "Lightly AG"
+website_url = 'https://www.lightly.ai/'
 author = 'Philipp Wirth, Igor Susmelj'
 
 # The full version, including alpha/beta/rc tags
-release = '1.0.0'
+release = lightly.__version__
```
japrescott, line 30: very nice

japrescott (commented):
good work with the footer!

docs/source/conf.py
guarin: The `sys.path.insert(0, os.path.abspath('../..'))` makes sure that lightly is imported from the local directory and not the installed package.

philippmwirth: ah cool, didn't see that

philippmwirth (approved):
Looks good, did you try it out and build the docs?

guarin:
Yes, did not completely rebuild docs but the changes work as expected. If I change the lightly version without reinstalling the package it successfully updates the version displayed in the docs.

Issue #608, opened 2021-11-30
Unmocked tests: cover more cases


MalteEbner:
## Current state
The unmocked tests for CLI commands also used in the CI cover already quite some cases.
https://github.com/lightly-ai/lightly/blob/master/tests/UNMOCKED_end2end_tests/run_all_unmocked_tests.sh

We also have CI script for the CLI commands magic and upload, which need the API:
https://github.com/lightly-ai/lightly/blob/e18656578886f7a3eacfa6a052ef5d537bef5460/.github/workflows/tests_unmocked.yml#L21-L24

## Description
The unmocked should ideally cover all CLI examples for all the API-related commands. Currently, it does not cover the download command and not all upload commands, due to the `dataset_id` and `embeddings_path` not being accessible.
https://github.com/lightly-ai/lightly/blob/e18656578886f7a3eacfa6a052ef5d537bef5460/lightly/cli/upload_cli.py#L201-L228
https://github.com/lightly-ai/lightly/blob/e18656578886f7a3eacfa6a052ef5d537bef5460/lightly/cli/download_cli.py#L131-L141
 
## Dependencies
- #607

guarin:
CLI upload and download are deprecated.

PR #622, opened 2021-12-03
setting paths to CLI artifacts as environment variables

MalteEbner:
## Descriptions
- Adds `environment_variable_names` to cfg to make it configurable where the paths are saved.
- When using the CLI for train, embed, upload, the paths to the artifacts are set at the corresponding locations.
- CLI unittests check that the paths are really set.

## Follow-Up Issue
- https://github.com/lightly-ai/lightly/issues/608


japrescott (approved):
nice

PR #623, opened 2021-12-06
Clean Docs

### LINKED ISSUE #594: Update pytorch lightning metrics dependency
`pytorch_lightning.metrics` is now an independent package: https://github.com/PyTorchLightning/metrics

We currently use the metrics module in the pizza filter tutorial and should remove it or add the new dependency.

guarin:
This pr is an effort to fix many small issues in the docs. It does the following:

* Add `make html-noplot` option to generate docs without running .py files
* Enable the sphinx napoleon extension so that the api docs are properly formatted
* Fix formatting/referencing issues
* Add [torchmetrics](https://torchmetrics.readthedocs.io/en/latest/) dependency for `dev` as the pytorch lightning metrics module was moved into an independent package. Closes #594.
* Download aquarium dataset automatically as it has to be included in the docs

---

The docs should now build with only the following warnings:
```
WARNING: The config value `plot_gallery' has type `str', defaults to `bool'.
```
This is due to the `make html-noplot` command and afaik cannot be changed. See https://github.com/sphinx-gallery/sphinx-gallery/issues/302

```
WARNING: Failed to create a cross reference. A title or caption not found
```
in `docs/source/tutorials/platform/index.rst`. This is in a file automatically generated by sphinx gallery and I could not find any way to fix this error.


philippmwirth (approved):
nicely done!

PR #625, opened 2021-12-09
Guarin lig 364 add download urls function to lightlyapi

guarin:
Draft for downloading samples with readurl. I included an iterable dataset for completeness.

The following code takes 20s to load 1k samples:
```
import torch
from tqdm import tqdm

from lightly.data.iterable_dataset import ImageIterableDataset
from lightly.api import ApiWorkflowClient

client = ApiWorkflowClient(
    token="token",
    dataset_id="dataset id"
)

samples = client.download_raw_samples(client.dataset_id)
dataset = ImageIterableDataset(samples)
dataloader = torch.utils.data.DataLoader(
    dataset,
    num_workers=8,
    batch_size=8
)

for image, filename, frame_idx in tqdm(dataloader):
    pass
```

The same code for videos runs in 10s for 5 videos with 300 frames each using only 2 workers.


guarin:
Some updates:

* All downloading primitives are now in `lightly.api.download`
* Using the new swagger generated api

Downloading and saving 2000 images to disk takes <1 sec if you use enough workers.

It is currently much faster if you download many files with a lot of workers and then load them from disk than load the files within the dataloader (where workers are limited to the number of cores on your system).

guarin:
Added tests. The `lightly.api.download.download_video_frame` is currently untested because the `cv2.VideoWriter` does not write the `cv2.CAP_PROP_FRAME_COUNT` flag when saving a synthetic video which results in an error in the code.

lightly/data/iterable_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+        # flatten videos so we have a single iterator
+        # over all frames of all videos
+        return itertools.chain.from_iterable(videos)
+
+    def _load_video(self, file_info):
+        """Generates (frame, filename, frame_idx) tuples for a video"""
+        filename, url = file_info
+        frames = download.download_all_video_frames(url)
+        for frame_index, frame in enumerate(frames):
+            yield frame, filename, frame_index
```
philippmwirth, line 89: Two things here:
- This shouldn't return the `frame_index` but rather the `_target`. Maybe we can change the definition of `_target` for the `VideoIterableDataset` to take the video filename instead of the frame filename and then define the class based on the video filename (i.e. all frames from the same class have the same weak label).
- The filename shouldn't be the filename of the video but rather something like `{video_filename}_{frame_index}_{video_format}.{image_format}`

philippmwirth (changes requested):
I like how you structured the code! This should give us all the tools we need :)

There are still some things which we should look into in detail. I left comments.

philippmwirth:
Oh, and please add docstrings to all your functions and classes!

lightly/api/api_workflow_datasources.py
```diff
@@ -0,0 +1,65 @@
...
+
+import time
+
+def _convert_raw_data_to_list(data: List[DatasourceRawSamplesDataRow]):
+    return [(d.file_name, d.read_url) for d in data]
+
+
+class _DatasourcesMixin:
+
+    def download_raw_samples(self, _from: int = 0, to: int = None):
```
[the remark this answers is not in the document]
guarin: I agree that `from_` is the correct name. I used `_from` because swagger generated it like this, but I will change it to the nicer version.

lightly/api/api_workflow_datasources.py
```diff
@@ -0,0 +1,65 @@
...
+                dataset_id=self.dataset_id, cursor=cursor
+            )
+            cursor = response.cursor
+            new_data = _convert_raw_data_to_list(response.data)
+            data.extend(new_data)
+        return data
+
+    def download_new_raw_samples(self):
+        response = self.get_processed_until_timestamp()
+        _from = int(response)
```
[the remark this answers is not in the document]
guarin: I think I discussed this with @japrescott. Maybe he can confirm to be sure.

lightly/api/download.py
```diff
@@ -0,0 +1,186 @@
+import concurrent.futures
```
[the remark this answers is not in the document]
guarin: The functions do not depend on the api or the client so I separated them from the workflow code. I considered them more as utils functions. But we could definitely also add them / call them directly from a workflow.

lightly/api/download.py
```diff
@@ -0,0 +1,186 @@
...
+            Whether to return the frame as PIL.Image
+
+    Returns:
+        generator:
+            Loads and returns a single frame per step.
+            Frames are of type PIL.Image if `as_pil_image` is `True`,
+            otherwise a np.array is returned
+
+    """
+    video = cv2.VideoCapture(url)
```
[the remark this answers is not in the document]
guarin: Yes I am aware. cv2 is the only library that I am aware of that allows this workflow. I would love to avoid it if possible. Maybe it would make more sense to download the video in the background to disk and then use the videoreader from torchvision. This would be more consistent with the non-iterable dataset.

lightly/api/download.py
```diff
@@ -0,0 +1,186 @@
...
+            yield cv2_to_pil(frame)
+        else:
+            yield frame
+    video.release()
+
+
+def download_video_frame(url: str, frame_index: int, as_pil_image: int = True):
+    """Retrieves a specific frame from a video stored at `url`
+
+    Warning: This is pretty slow
```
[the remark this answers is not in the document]
guarin: I got ~3s/frame.

lightly/api/api_workflow_datasources.py
```diff
@@ -0,0 +1,65 @@
...
+        cursor = response.cursor
+        data = _convert_raw_data_to_list(response.data)
+        while response.has_more:
+            response: DatasourceRawSamplesData = self._datasources_api.get_list_of_raw_samples_from_datasource_by_dataset_id(
+                dataset_id=self.dataset_id, cursor=cursor
+            )
+            cursor = response.cursor
+            new_data = _convert_raw_data_to_list(response.data)
+            data.extend(new_data)
+        return data
```
MalteEbner, line 37: Several points here:
- `data` is a bad variable name
- What is the reason to change the `List[DatasourceRawSamplesDataRow]` to `List[Tuple[str, str]]`?
- The type hint for the return value is missing.
- What is the reason to `_convert_raw_data_to_list` both `data` and `new_data` instead of once data after the while loop? As this would call the function to only be called once, it could be inlined in this case and deleted.

lightly/api/api_workflow_datasources.py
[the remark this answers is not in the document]
MalteEbner: I would call it something like `start_time`, `end_time`

lightly/api/api_workflow_datasources.py
MalteEbner: In general, I don't like returning `List[Tuple[str, str]]` over multiple instances, as you never know what the strings stand for. However, a `List[DatasourceRawSamplesDataRow]` is always fully clear, as you can use the intellisense to get the attribute names like `element.file_name` for each element.

lightly/api/api_workflow_datasources.py
```diff
@@ -0,0 +1,65 @@
...
+            # Add 1 because the samples with timestamp == _from
+            # have already been processed
+            _from += 1
+
+        to = int(time.time())
+        data = self.download_raw_samples(_from=_from, to=to)
+        self.update_processed_until_timestamp(timestamp=to)
+        return data
+
+    def get_processed_until_timestamp(self):
```
MalteEbner, line 54: I don't see a reason for having a one-line function that is only referenced by a single other function.

lightly/api/api_workflow_datasources.py
MalteEbner: Alternatively, you could define a small class called `FileInfo`, with the `filename` and `url`. Or use a named tuple.

lightly/data/iterable_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+import torch
+
+from lightly.api import download
+from lightly.data import _helpers
+
+def _filter_extensions(samples, extensions):
+    return [s for s in samples if s[0].endswith(extensions)]
+
+
+class IterableDataset(torch.utils.data.IterableDataset):
```
MalteEbner, line 12: I would call this `LightlyIterableDataset` to prevent name conflicts.

lightly/data/iterable_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+        """Get weak label based on filename"""
+        raise NotImplementedError
+
+
+class ImageIterableDataset(IterableDataset):
+
+    def __init__(self, samples):
+        """
+        Args:
+            samples: list(tuple)
```
MalteEbner, line 45: Type hints belong to the function declaration, not the docstring.

lightly/api/api_workflow_datasources.py
MalteEbner: If you have it as an explicit element, you also don't need to write down that the tuple must be (filename, url) in the docstrings of every function it is used.

MalteEbner:
As a general point: Many functions are missing typehints for the return value.

MalteEbner:
Regarding the usage: I like the workflow you proposed very much, except for one point:
```
samples = client.download_raw_samples(client.dataset_id)
dataset = ImageIterableDataset(samples)
```
Why not make it much simpler?
```
dataset = ImageIterableDataset(client)
```

lightly/data/iterable_dataset.py
MalteEbner: The filename should be the same as here:
https://github.com/lightly-ai/lightly/blob/5bdc307a69b8d0f7cafded685321e2e7e4e582a2/lightly/data/_video.py#L391

lightly/api/api_workflow_datasources.py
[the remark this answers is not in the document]
japrescott: Yes it will default to 0

lightly/api/api_workflow_datasources.py
guarin: I do not return `List[DatasourceRawSamplesDataRow]` because I wanted to avoid having api objects in different parts of the codebase. And I did not use an extra class because I have not seen it used elsewhere in the code. For example all torch modules expect a batch to be a tuple of images, labels, and filenames without using an extra class. I completely agree that a class would make the code more readable. Another issue is that I am not sure if there is currently any module where it would make most sense to define such classes.

lightly/api/api_workflow_datasources.py
guarin: Could you elaborate? I think it makes sense to give the client the option to get/change the timestamp to his liking.

guarin:
> Regarding the usage: I like the workflow you proposed very much, except for one point:
> 
> ```
> samples = client.download_raw_samples(client.dataset_id)
> dataset = ImageIterableDataset(samples)
> ```
> 
> Why not make it much simpler?
> 
> ```
> dataset = ImageIterableDataset(client)
> ```

Maybe for clarification: I only included the IterableDataset in this pr to showcase how to use it. I discussed with @philippmwirth that it probably makes sense to only have it in the docker first and maybe later move it to the pip package.

My goal was to make the dataset independent from the api. Currently you could also use the dataset to load samples from any public url, without the need for the client. Avoiding the client in the dataset also makes sure that we have no surprises with multiprocessing in the dataloaders. Finally, it makes testing much easier.

The following steps are separated on purpose:
> samples = client.download_raw_samples(client.dataset_id)
> dataset = ImageIterableDataset(samples)

This is because we want to be able to optionally download all data like this:
```
samples = client.download_raw_samples(client.dataset_id)
if len(samples) < MAX_NUM_SAMPLES_TO_DOWNLOAD:
    download.download_and_write_all_files(samples, output_dir)
    dataset = LightlyDataset(samples)
else:
    dataset = ImageIterableDataset(samples)
```
Downloading all samples is much more efficient, so we should do it if possible. The same is true if we want to do training as training with an iterable dataset is not possible. 


> The `api_workflow_client` is designed to handle the connection to a single dataset. Thus you should not need to pass its `dataset_id` to any of its methods, they should fetch it internally.

I don't think there is a function that requires the dataset_id?

> Furthermore I think that we could abstract the whole "samples" away from the user. In the end, they want a dataset from and to a certain point or one with only new samples. They don't care that the data is represented internally in a step inbeween as samples as `List[Tuple[str, str]]`.

I don't think this is possible. We somewhere need a list of the samples because they will be reused for cropping, sampling, ...

> The from, to, or `only_new_samples` could be passed alongside the client to the `ImageIterableDataset`.
> 
> Alternatively, the `api_workflow_client` could have a method directly returning an `ImageIterableDataset`:
> 
> ```python
> def get_iterable_dataset(self, from, to, only_new_samples: bool=False) -> ImageIterableDataset:
> ```

As stated above I don't think we should mix api with datasets.

lightly/api/api_workflow_datasources.py
```diff
@@ -1,23 +1,27 @@
...
-    def download_raw_samples(self, from_: int = 0, to: int = None):
+    def download_raw_samples(
+        self, from_: int = 0, to: int = None
+    ) -> List[Tuple[str, str]]:
+        """Downloads all filenames and read urls from the datasource between `from_` and `to`.
+
+        Samples which have timestamp == `from_` or timestamp == `to` will also be included.
+        
+        Args:
+            from_ (int): Unix timestamp from which on samples are downloaded.
```
MalteEbner, line 18: Don't include the type in the docstring. It is sufficient to have it in the declaration.

lightly/api/api_workflow_datasources.py
```diff
@@ -26,17 +30,27 @@ def download_raw_samples(self, from_: int = 0, to: int = None):
...
-    def download_new_raw_samples(self):
+    def download_new_raw_samples(self) -> List[Tuple[str, str]]:
+        """Downloads filenames and read urls of unprocessed samples from the datasource.
+
+        All samples after the timestamp of `ApiWorkflowClient.get_processed_until_timestamp()` are 
+        fetched. After downloading the samples the timestamp is updated to the current time.
+        This function can be repeatedly called to retrieve new samples from the datasource.
+        
+        Returns:
+            Samples (list(tuple)): Each tuple contains (filename, url) of the respective sample
```
MalteEbner, line 51: Don't include the type in the docstring. It is sufficient to have it in the declaration.

lightly/api/download.py
```diff
@@ -34,10 +28,12 @@ def download_image(url: str, session: requests.Session = None):
...
     req = requests if session is None else session
     response = req.get(url, stream=True)
-    return Image.open(response.raw)
+    return PIL.Image.open(response.raw)
 
 
-def download_all_video_frames(url: str, as_pil_image: int = True):
+def download_all_video_frames(
+    url: str, as_pil_image: int = True
+) -> Iterable[Union[PIL.Image.Image, av.VideoFrame]]:
```
MalteEbner, line 36: Nice typehint!

tests/api/test_download.py
```diff
@@ -69,10 +69,45 @@ def test_download_image_with_session(self):
...
+    def test_download_video_frame(self):
+        with tempfile.NamedTemporaryFile(suffix='.avi') as file:
+            original = _generate_video(file.name)
+            for timestamp, orig in enumerate(original):
+                frame = download.download_video_frame(file.name, timestamp)
+                assert _images_equal(frame, orig)
+
+    def test_download_video_frame_fps(self):
+        for fps in range(1, 5):
+            with tempfile.NamedTemporaryFile(suffix='.avi') as file:
```
MalteEbner, line 85: Add a subtest here:
```python
with self.subTest(fps=fps):
```
It allows you to see directly in the unittesting framework which of the subtest failed (e.g. all 5 or just one of them and which).
Documentation: https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest

The same applies also to the other unittests.

lightly/api/download.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def download_image(url: str, session: requests.Session = None) -> PIL.Image.Image:
+    """Downloads an image from an url.
+
+    Args:
+        url: 
+            Url where image is stored.
+        session: 
+            Make request using the given session.
```
MalteEbner, line 24: This wording is a little ambiguous, as the session is not a make request. I would rather explain it e.g. as
`An existing session to perform the download. If it is not passed, a new one is created.`

lightly/api/download.py
```diff
@@ -0,0 +1,192 @@
...
+from lightly.api import utils
+import PIL
+
+
+def download_image(url: str, session: requests.Session = None) -> PIL.Image.Image:
+    """Downloads an image from an url.
+
+    Args:
+        url: 
+            Url where image is stored.
```
MalteEbner, line 22: where THE image

lightly/api/download.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def download_all_video_frames(
+    url: str, as_pil_image: int = True
+) -> Iterable[Union[PIL.Image.Image, av.VideoFrame]]:
+    """Lazily retrieves all frames from a video.
+
+    Args:
+        url: 
+            Url where video is stored.
```
MalteEbner, line 42: where THE video

lightly/api/download.py
MalteEbner: Furthermore, the term "stored" is ambigous. Later int this file, you use the the it for describing the OUTPUT filepath. 
Thus I would rather call it "URL where the image is downloaded from." 
Compare also with lines 112-115. 

The same applies to other docstrings of urls in this file.

lightly/api/download.py
MalteEbner: The same applies to other docstrings of sessions in this file.

lightly/api/download.py
```diff
@@ -0,0 +1,192 @@
...
+    container.close()
+    if as_pil_image:
+        return frame.to_image()
+    return frame
+
+
+def download_and_write_file(
+    url: str, output_path: str, session: requests.Session = None
+) -> None:
+    """Downloads a file from url and saves it to disk
```
MalteEbner, line 109: from A url
saves it to disk POINT

MalteEbner (approved):
This looks good to me, I am looking forward to use it.

The docstrings could be improved a little in my eyes.

lightly/api/download.py
philippmwirth: Or just call it read url to avoid any confusion

lightly/api/download.py
philippmwirth: Or: "Session object to persist certain parameters across requests." (this is pretty much how the requests library describes a session).

philippmwirth: > If it is not passed, a new one is created.

This is not true

lightly/api/download.py
```diff
@@ -0,0 +1,192 @@
...
+        as_pil_image: 
+            Whether to return the frame as PIL.Image.
+
+    Returns:
+        A generator that loads and returns a single frame per step.
+
+    """
+    container = av.open(url)
+    stream = container.streams.video[0]
+    stream.thread_type = 'AUTO'
```
philippmwirth, line 52: What are the alternatives here? I would add it as an argument to the function with default value or use a constant outside the function.

lightly/api/download.py
```diff
@@ -0,0 +1,192 @@
...
+    for frame in container.decode(stream):
+        if as_pil_image:
+            yield frame.to_image()
+        else:
+            yield frame
+    container.close()
+
+
+def download_video_frame(
```
philippmwirth, line 61: why did you change from access by index to access by timestamp?

lightly/api/download.py
philippmwirth: from A url, I believe :)

philippmwirth (commented):
Some more questions and comments but it's taking on shape! :)

lightly/api/download.py
guarin: You cannot seek to a specific frame index in pyav. The problem is that videos do not necessarily store frames in fixed intervals. You would have to read the complete video first to map the frame index to a timestamp. Torchvision also avoids frame indices and uses timestamps instead.

guarin:
> Or just call it read url to avoid any confusion
The functions can load images/videos from pretty much any url.

> Or: "Session object to persist certain parameters across requests." (this is pretty much how the requests library describes a session).
Changed it :)

philippmwirth (approved):

PR #627, opened 2021-12-13
correctly name samples_api

japrescott:
fixes lig-384
- custom_metadata upload when dataset was already created failed

### DIFF
--- a/lightly/api/api_workflow_upload_metadata.py
+++ b/lightly/api/api_workflow_upload_metadata.py
@@ -183,7 +183,7 @@ class _UploadCustomMetadataMixin:
                     custom_meta_data=metadata
                 )
                 # send the request to the api
-                self.samples_api.update_sample_by_id(
+                self._samples_api.update_sample_by_id(
                     update_sample_request,
                     self.dataset_id,
                     sample.id
--- a/tests/api/test_rest_parser.py
+++ b/tests/api/test_rest_parser.py
@@ -17,16 +17,16 @@ class TestRestParser(unittest.TestCase):
             [12.0] * 999
         )
         api_client = ApiClient()
-        self.scores_api = ScoresApi(api_client)
+        self._scores_api = ScoresApi(api_client)
         for i, score_values in enumerate(score_value_tuple):
             with self.subTest(i=i, msg=str(type(score_values))):
                 body = ActiveLearningScoreCreateRequest(score_type=SamplingMethod.CORESET, scores=list(score_values))
                 if isinstance(score_values[0], float):
                     with self.assertRaises(ApiException):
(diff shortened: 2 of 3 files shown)


tests/UNMOCKED_end2end_tests/create_custom_metadata_from_input_dir.py
```diff
@@ -0,0 +1,24 @@
...
+from lightly.data import LightlyDataset
+from lightly.utils import save_custom_metadata
+
+if __name__ == "__main__":
+    if len(sys.argv) == 1 + 2:
+        input_dir, metadata_filename= \
+            (sys.argv[1 + i] for i in range(2))
+    else:
+        raise ValueError("ERROR in number of command line arguments, must be 2."
+                         "Example: python create_custom_metadata_from_input_dir input_dir metadata_filename")
```
guarin, line 12: I think you missed the .py file extension

tests/UNMOCKED_end2end_tests/delete_datasets_test_unmocked_cli.py
```diff
@@ -8,7 +8,7 @@
             (sys.argv[1 + i] for i in range(2))
     else:
         raise ValueError("ERROR in number of command line arguments, must be 2."
-                         "Example: python test_api 6 TOKEN")
+                         "Example: python delete_datasets_test_unmocked_cli 6 TOKEN")
```
guarin, line 11: Missed .py extension

tests/api_workflow/test_api_workflow_upload_custom_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+            data = self.dataset[sample_idx]
+            sample_name = sample_names[sample_idx]
+            path = os.path.join(self.folder_path, sample_name)
+            data[0].save(path)
+
+        coco_json = dict()
+        coco_json['images'] = [{'id': i, 'file_name': fname} for i, fname in
+                               enumerate(sample_names)]
+        coco_json['metadata'] = [{'id': i, 'image_id': i, 'custom_metadata': 0}
+                                 for i, _ in enumerate(sample_names)]
```
guarin, line 40: Maybe use `COCO_ANNOTATION_KEYS.images` instead of `'images'`  for consistency. Same for `'metadata'`

guarin (approved):
Looks good :) I suppose you remove the cli test because it is covered by the unmocked + api workflow tests?

MalteEbner:
> I suppose you remove the cli test because it is covered by the unmocked + api workflow tests?

I removed it, because it fails after the fix `Ensures that there is a value error if there are filenames in the custom metadata file that are not on the server.`. Implementing it there is hard, because the current tests do not allow keeping a stateful API between CLI commands.

PR #633, opened 2021-12-17
Fix: Extract timestamp from response

guarin:
This makes the `ApiWorkflowClient.get_processed_until_timestamp()` method actually return an `int`. Before it simply returned the api response object by accident.

I also updated tests (including adding types) and updated an unrelated line that caused a warning when running tests here:

https://github.com/lightly-ai/lightly/blob/f96b5de7054ec94d1d6d51981c8662600500e130/tests/api_workflow/mocked_api_workflow_client.py#L169

### DIFF
--- a/tests/api_workflow/mocked_api_workflow_client.py
+++ b/tests/api_workflow/mocked_api_workflow_client.py
@@ -8,6 +8,7 @@ import json
 
 import numpy as np
 from requests import Response
+from lightly.openapi_generated.swagger_client.models.datasource_processed_until_timestamp_response import DatasourceProcessedUntilTimestampResponse
 
 from lightly.openapi_generated.swagger_client.models.tag_creator import TagCreator
 from lightly.openapi_generated.swagger_client.models.dataset_create_request import DatasetCreateRequest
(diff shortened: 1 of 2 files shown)


tests/api_workflow/mocked_api_workflow_client.py
```diff
@@ -165,7 +166,7 @@ def get_tags_by_dataset_id(self, dataset_id, **kwargs):
 
     def perform_tag_arithmetics(self, body: TagArithmeticsRequest, dataset_id, **kwargs):
         _check_dataset_id(dataset_id)
-        if body.new_tag_name is None or body.new_tag_name is '':
+        if (body.new_tag_name is None) or (body.new_tag_name == ''):
```
japrescott, line 169: is this your linter kicking in or does this change, change anything?

japrescott (approved):

tests/api_workflow/mocked_api_workflow_client.py
guarin: Pytest shows a warning if you compare strings with `is`. The issue is that `is` tests for reference equality and should therefore not used with strings.

PR #636, opened 2021-12-21
Guarin lig 429 speed up getting filenames from video dataset

guarin:
This pr adds a `get_filenames` function directly to the the `VideoDataset` class. This avoids getting the filename for every frame by searching the corresponding video from the frame index.

I tested it with 20k videos and a total of 10M frames. For me this reduces the `get_filenames` operation from ~5 hours to 10 seconds. This also speeds up the time it takes to initialise `LightlyDataset` as it internally calls `get_filenames`. Using this pr it takes me 2min to initialise the dataset.

I added a test to make sure that the new method returns the same results as the old one.

### DIFF
--- a/lightly/data/_video.py
+++ b/lightly/data/_video.py
@@ -373,41 +373,74 @@ class VideoDataset(datasets.VisionDataset):
             i = i - 1
 
         # get filename of the video file
-        filename = self.videos[i]
-        filename = os.path.relpath(filename, self.root)
-
-        # get video format and video name
-        splits = filename.split('.')
-        video_format = splits[-1]
-        video_name = '.'.join(splits[:-1])
+        video = self.videos[i]
+        video_name, video_format = self._video_name_format(video)
 
         # get frame number
         frame_number = index - self.offsets[i]
-        if i < len(self.offsets) - 1:
-            n_frames = self.offsets[i+1] - self.offsets[i]
-        else:
-            n_frames = self.__len__() - self.offsets[i]
         
-        return f'{video_name}-{frame_number:0{len(str(n_frames))}}-{video_format}.png'
+        n_frames = self._video_frame_count(i)
+        zero_padding = len(str(n_frames))
+
+        return self._format_filename(
+            video_name=video_name,
+            video_format=video_format,
+            frame_number=frame_number,
+            zero_padding=zero_padding,
+        )
 
     def get_filenames(self) -> List[str]:
         """Returns a list filenames for all frames in the dataset.
         
         """
         filenames = []
-        for i, filename in enumerate(self.videos):
-            filename = os.path.relpath(filename, self.root)
-            splits = filename.split('.')
-            video_format = splits[-1]
-            video_name = '.'.join(splits[:-1])
+        for i, video in enumerate(self.videos):
+            video_name, video_format = self._video_name_format(video)
+            n_frames = self._video_frame_count(i)
 
-            if i < len(self.offsets) - 1:
-                n_frames = self.offsets[i+1] - self.offsets[i]
-            else:
-                n_frames = len(self) - self.offsets[i]
-
+            zero_padding = len(str(n_frames))
             for frame_number in range(n_frames):
                 filenames.append(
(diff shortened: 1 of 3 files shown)


lightly/data/_video.py
```diff
@@ -388,4 +388,26 @@ def get_filename(self, index):
...
+            video_name = '.'.join(splits[:-1])
+
+            if i < len(self.offsets) - 1:
+                n_frames = self.offsets[i+1] - self.offsets[i]
+            else:
+                n_frames = len(self) - self.offsets[i]
+
+            for frame_number in range(n_frames):
+                filenames.append(
+                    f'{video_name}-{frame_number:0{len(str(n_frames))}}-{video_format}.png'
```
japrescott, line 411: I guess you could precompute `len(str(n_frames))` as well

tests/data/test_LightlyDataset.py
```diff
@@ -339,3 +339,19 @@ def test_transform_setter(self, dataset: LightlyDataset = None):
...
+
+    @unittest.skipUnless(VIDEO_DATASET_AVAILABLE, "PyAV or CV2 is/are not installed")
+    def test_dataset_get_filenames(self):
+        self.create_video_dataset()
+        dataset = LightlyDataset(input_dir=self.input_dir)
+        
+        # get filenames using VideoDataset.get_filenames
+        video_dataset_filenames = dataset.dataset.get_filenames()
+        
+        # get filenames using calls to VideoDataset.get_filename(index)
```
japrescott, line 351: maybe make it clear that this "removes" the optimisation

japrescott (approved):
absolutely sensational! <3

tests/data/test_LightlyDataset.py
```diff
@@ -339,3 +339,19 @@ def test_transform_setter(self, dataset: LightlyDataset = None):
...
+        get_filenames = VideoDataset.get_filenames
+        del VideoDataset.get_filenames
+        lightly_dataset_filenames = dataset.get_filenames()
+        VideoDataset.get_filenames = get_filenames
```
MalteEbner, line 355: This is a great way to test it, good idea!

I think the code would be easier understandable with
`lightly_dataset_filenames = super(VideoDataset, dataset).get_filenames()`.

lightly/data/_video.py
MalteEbner, line 411: This line is duplicate to line 391.

japrescott: for that extra nanosecond saved ;)

lightly/data/_video.py
```diff
@@ -388,4 +388,26 @@ def get_filename(self, index):
...
+    def get_filenames(self) -> List[str]:
+        """Returns a list filenames for all frames in the dataset.
+        
+        """
+        filenames = []
+        for i, filename in enumerate(self.videos):
+            filename = os.path.relpath(filename, self.root)
+            splits = filename.split('.')
+            video_format = splits[-1]
+            video_name = '.'.join(splits[:-1])
```
MalteEbner, line 402: These 4 lines are duplicate to lines 377-382.

lightly/data/_video.py
guarin: Added it :)

tests/data/test_LightlyDataset.py
guarin: I don't think this call to super works. But I made the call more explicit.

lightly/data/_video.py
guarin: Moved it into a function.

lightly/data/_video.py
guarin: Added a function for this as well.

tests/data/test_LightlyDataset.py
guarin: Added some more comments :)

tests/data/test_LightlyDataset.py
MalteEbner: Thanks, that should work as well.

MalteEbner (approved):
This looks great, I am looking forward to seeing it in action.

PR #637, opened 2021-12-21
Add frame counts from video methods

guarin:
* Add video_frame_count to find number of frames in a single video
* Add all_video_frame_counts to find total and per video number of frames
* Add tests

This is needed to determine the number of frames and the filenames (are inferred from the number of frames) in the dataset.

### DIFF
--- a/lightly/api/download.py
+++ b/lightly/api/download.py
@@ -290,12 +290,15 @@ def all_video_frame_counts(
     """
 
     def job(url):
-        return utils.retry(
-            video_frame_count, 
-            url=url,
-            video_channel=video_channel,
-            thread_type=thread_type,
-        )
+        try:
+            return utils.retry(
+                video_frame_count, 
+                url=url,
+                video_channel=video_channel,
+                thread_type=thread_type,
+            )
+        except RuntimeError:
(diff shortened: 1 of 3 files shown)


lightly/api/download.py
```diff
@@ -222,3 +221,100 @@ def job(*args):
...
+                    f"url: {url}"
+                )
+
+        if num_failed > 0:
+            raise RuntimeError(
+                f"Could not find the number of frames for {num_failed} videos!"
+            )
+
+        total_frames = sum(frame_counts)
+        return total_frames, frame_counts
```
MalteEbner, line 320: Why not only return the `frame_counts`? I think this would make the return type easier to understand. Summing it up can be easily done by the caller of the function if the total number of frames is needed.

MalteEbner (approved):
Well-understandable structure, correct docstrings and typehints, many unittests with subtests, 100% code coverage. This is great!

I only had one proposal about the return type.

lightly/api/download.py
guarin: Good point! I changed it to only return frame_counts :)

PR #640, opened 2021-12-22
Guarin lig 431 add time unit argument to download_video_frame function

guarin:
Depends on #637


lightly/api/download.py
```diff
@@ -110,24 +113,32 @@ def download_video_frame(
...
+    if time_unit not in ('sec', 'pts'):
+        raise ValueError(f"time_unit must be 'sec' or 'pts' but is {time_unit}")
 
     container = av.open(url)
     stream = container.streams.video[video_channel]
     stream.thread_type = thread_type
-    offset = int(timestamp / stream.time_base)
+    
+    if time_unit == 'sec':
+        offset = int(timestamp / stream.time_base) + 1
```
IgorSusmelj, line 124: Could you elaborate on the `+ 1` part? Is this required to make the offsets compatible?

IgorSusmelj (changes requested):
This looks great. I just left one question. If you could elaborate on the difference between the seconds and pts and why we have the `- 1` :)
You can also directly add it as a comment to the source code to help others understand the reason for the offset.

IgorSusmelj (approved):

PR #642, opened 2021-12-22
Docs: example python script to create custom metadata

MalteEbner:
the python script creates metadata from a pandas dataframe

## Before reviewing: consider the example/comments before the actual script:
https://docs.lightly.ai/getting_started/platform.html#custom-metadata

# Open question:
Which option to use of the 2?
- Option 1 ?
- Option 2 ?
- Both ?

### DIFF
--- a/docs/source/getting_started/platform.rst
+++ b/docs/source/getting_started/platform.rst
@@ -199,32 +199,62 @@ you can use a simple script to translate it to the coco format:
 
 .. code-block:: python
 
-    import numpy as np
+    ## OPTION 1: pandas dataframe with named columns
+    # This works well if you have many columns and column names
+    import pandas as pd
+    from lightly.utils import save_custom_metadata
+
+    # Define the pandas dataframe
+    column_names = ["filename", "number_of_pedestrians", "scenario", "temperature"]
+    rows = [
+        ["image0.jpg", 3, "cloudy", 20.3],
+        ["image1.jpg", 1, "rainy", 15.0]
+    ]
+    df = pd.DataFrame(rows, columns=column_names)
+
+    # create a list of pairs of (filename, metadata)
+    # metadata must be a dictionary
+    custom_metadata = []
+    for data in df.itertuples():
+        filename = data.filename
+        metadata = data._asdict()
+        del metadata['filename']
+        del metadata['Index']
+        custom_metadata.append((filename, metadata))
+
+    # save custom metadata in the correct json format
+    output_file = "custom_metadata.json"
+    save_custom_metadata(output_file, custom_metadata)
+
+
+    ## OPTION 2: pandas dataframe without named columns
+    # As the metadata has to be defined by hand anyway,
+    # it can easily put into nested dictionaries
     import pandas as pd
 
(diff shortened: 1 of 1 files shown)


docs/source/getting_started/platform.rst
```diff
@@ -194,13 +194,47 @@ need to look like this:
...
+    import numpy as np
+    import pandas as pd
+
+    from lightly.utils import save_custom_metadata
+
+    # Let's assume you have your data in form of a pandas dataframe
+    df = pd.DataFrame({
+        "image0.jpg": [3, "cloudy", 20.3],
+        "image1.jpg": [1, "rainy", 15.0]
+    })
```
guarin, line 211: In most cases you have a pandas dataframe with column names (for example when you load it from a csv). I would include column names here as well and then use them to set the metadata names as well.

docs/source/getting_started/platform.rst
```diff
@@ -194,13 +194,47 @@ need to look like this:
         ]
     }
 
+If you don't have your data in coco format yet, but e.g. as a pandas dataframe,
+you can use a simple script to translate it to the coco format:
```
guarin, line 198: Maybe add an introductory sentence that tells you that we use coco format for metadata. Otherwise it is pretty confusing if you just start with coco format in a tutorial on metadata. A link to some coco format resource would be nice as well.

guarin (commented):
Looks good :) Added some small suggestions.

docs/source/getting_started/platform.rst
MalteEbner: Good point, I added that.

docs/source/getting_started/platform.rst
MalteEbner: Look at the docs before this change, it is included there.

docs/source/getting_started/platform.rst
```diff
@@ -194,13 +194,77 @@ need to look like this:
...
+    # Define the pandas dataframe
+    column_names = ["filename", "number_of_pedestrians", "scenario", "temperature"]
+    rows = [
+        ["image0.jpg", 3, "cloudy", 20.3],
+        ["image1.jpg", 1, "rainy", 15.0]
+    ]
+    df = pd.DataFrame(rows, columns=column_names)
+
+    # create a list of pairs of (filename, metadata)
+    # metadata must be a dictionary
```
philippmwirth, line 216: I liked it better before, why not simply do this:

```python
custom_metadta = []
for row in df.itertuples():
    filename = data.filename
    metadata = {
        "number_of_pedestrians": int(data.number_of_pedestrians),
        "weather": {
             "scenario": str(data.scenario),
             "temperature": float(data.temperature),
        }
    }
    custom_metadata.append((filename, metadata))
```

This is easy to read and allows for all the nesting we want.

MalteEbner: I followed your suggestion and used
- the named dataframe from option 1
- the named and nested metadata creation you just proposed

philippmwirth (approved):

PR #643, opened 2021-12-23
Guarin lig 442 add dataset type to pip api

guarin:
* Adds new generated api code
* Adds a `dataset_type` attribute to the api client
* Updates tests to reflect changes in `SampleData` model


lightly/api/api_workflow_datasets.py
```diff
@@ -7,6 +7,23 @@
...
+        return dataset.type # type: ignore
+
+    def _get_current_dataset(self):
+        """Returns the dataset with id == self.dataset_id.
+        
+        """
+        dataset_id = self.dataset_id
+        datasets: List[DatasetData] = self._datasets_api.get_datasets()
+        datasets = [dataset for dataset in datasets if dataset.id == dataset_id]
+        return datasets[0]
```
IgorSusmelj, line 25: Maybe use next instead of the first item from the list comprehension? 
see https://stackoverflow.com/questions/9868653/find-first-sequence-item-that-matches-a-criterion

IgorSusmelj (approved):
Look good to me. I just left a comment regarding the list comprehension. But this won't have any relevant impact. Feel free to either stick with the list or use next :)

lightly/api/api_workflow_datasets.py
guarin: Ah good catch, thanks! Changed it :)

PR #646, opened 2021-12-27
Datasource docs azure

philippmwirth:
# Datasource docs azure

Adds documentation on how to use Azure blob storage with Ligthly:
[Create a dataset from Azure blob storage — lightly 1.2.1 documentation.pdf](https://github.com/lightly-ai/lightly/files/7792449/Create.a.dataset.from.Azure.blob.storage.lightly.1.2.1.documentation.pdf)

I added the images to our gcloud bucket so they're not part of the repo.

### DIFF
--- a/docs/source/getting_started/dataset_creation/dataset_creation_azure_storage.rst
+++ b/docs/source/getting_started/dataset_creation/dataset_creation_azure_storage.rst
@@ -16,8 +16,6 @@ You have the following three options:
 
 - You want to use thumbnails, but don't have them yet. Then you need to give
   Lightly write access to your bucket to create the thumbnails there for you.
-  The write access can be configured not to allow overwriting and
-  deleting, thus existing data cannot get lost.
 - You already have thumbnails in your bucket with a consistent name scheme, e.g.
   an image called `img.jpg` has a corresponding thumbnail called `img_thumb.jpg`.
   In this case, a read access to your bucket is sufficient.
@@ -33,18 +31,6 @@ Setting up Azure
 For the purpose of this guide we assume you have a storage account called `lightlydatalake`.
 We further assume the container you want to use with lightly is called `farm-animals` and already contains images.
 If you don't have a storage account or container yet follow the instructions `here <https://docs.microsoft.com/en-us/azure/storage/common/storage-account-create?tabs=azure-portal>`_.
-
-1. In the `Azure Portal <https://portal.azure.com/#home>`_ navigate to `Storage Accounts`.
-2. Select your storage account and go to `Settings > Resource sharing (CORS)`. Allow at least `LIST` and `GET`. For write access also allow `PUT` and `DELETE`.
-   
-    .. figure:: ../resources/AzureResourceSharing.jpg
-        :align: center
-        :alt: Azure Resource Sharing (CORS)
-
-        Allow resource sharing by enabling CORS (in this example all methods are allowed).
-
-3. Go to `Security + networking > Access keys`. Copy the `Key` and store it in a secure location.
-
 Head to the next section to see how you can configure the Lightly dataset.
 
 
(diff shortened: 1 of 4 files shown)


docs/source/getting_started/dataset_creation/dataset_creation_azure_storage.rst
```diff
@@ -1,6 +1,102 @@
...
+
+Setting up Azure
+^^^^^^^^^^^^^^^^^
+
+For the purpose of this guide we assume you have a storage account called `lightlydatalake`.
+We further assume the container you want to use with lightly is called `farm-animals` and already contains images.
+If you don't have a storage account or container yet follow the instructions `here <https://docs.microsoft.com/en-us/azure/storage/common/storage-account-create?tabs=azure-portal>`_.
+
+1. In the `Azure Portal <https://portal.azure.com/#home>`_ navigate to `Storage Accounts`.
+2. Select your storage account and go to `Settings > Resource sharing (CORS)`. Allow at least `LIST` and `GET`. For write access also allow `PUT` and `DELETE`.
```
MalteEbner, line 37: Several points here:
- In the newest azure version, there is no "LIST". 
-  I don't understand why you added all 8 methods if only 2 or 4 are need.
- Earlier you tell `The write access can be configured not to allow overwriting and deleting, thus existing data cannot get lost.`. How is this possible here? Doesn't this mean that I don't need DELETE?
- Is is possible to restrict access to only one container? Or will lightly always have access to the full storage account including all of its containers?

docs/source/getting_started/dataset_creation/dataset_creation_azure_storage.rst
```diff
@@ -1,6 +1,102 @@
...
+
+.. code-block::
+    :caption: Sync your data with AzCopy. Example from `Microsoft's documentation <https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-blobs-synchronize?toc=/azure/storage/blobs/toc.json>`_
+
+    azcopy sync './farm-animals' 'https://lightlydatalake.blob.core.windows.net/farm-animals' --recursive
+
+
+To add the images to the dataset use `lightly-magic` or `lightly-upload` with the following parameters:
+
+- Use `input_dir=/local/projects/wild-animals`
```
MalteEbner, line 99: Use the same path here as in the azcopy command, e.g `/local/lightlydatalake/farm-animals` for both `azcopy` and the `input_dir`

MalteEbner (commented):
The tutorial looks quite good, it has a similar structure to the other ones making it easier to follow.

Unfortunately, I could only test the workflow partly:
- Our azure storage account already has CORS configured and I did not want to deactivate the existing to make sure the new one needs to be used.
- Configuring the dataset in the webapp worked, all lights are green.
- The azcopy command failed to me even after successfully logging in. The was an error 401.
- Azure does not allow to download the files easily from a container, thus I could not download them manually.

docs/source/getting_started/dataset_creation/dataset_creation_azure_storage.rst
```diff
@@ -1,6 +1,96 @@
...
+
+One decision you need to make first is whether you want to use thumbnails.
+Using thumbnails makes the Lightly Platform more responsive, as it's enough to
+load the thumbnails instead of the full images in many cases.
+As a drawback, the thumbnails will be stored in your bucket and thus need storage.
+You have the following three options:
+
+
+- You want to use thumbnails, but don't have them yet. Then you need to give
+  Lightly write access to your bucket to create the thumbnails there for you.
```
MalteEbner, line 18: How do I distinguish between read/write access to the bucket when setting it up? If I cannot, leave it away here.

docs/source/getting_started/dataset_creation/dataset_creation_azure_storage.rst
```diff
@@ -1,6 +1,96 @@
...
+      Then leave the thumbnail suffix undefined/empty.
+
+    .. figure:: ../resources/LightlyEditAzure.jpg
+        :align: center
+        :alt: Lightly Azure Blob Storage config.
+        :width: 60%
+
+        Lightly Azure Blob Storage config.
+
+6. Press save and ensure that at least the lights for List and Read turn green.
```
MalteEbner, line 69: All lights should turn green.

docs/source/getting_started/dataset_creation/dataset_creation_azure_storage.rst
```diff
@@ -1,6 +1,96 @@
...
+        To edit your dataset click the `Edit` button on the top.
+
+3. As your container name enter `farm-animals`.
+4. Enter the storage account name and storage account key from the previous step.
+5. The thumbnail suffix depends on the option you chose in the first step
+   
+    - You want Lightly to create the thumbnail for you.
+      Then choose the naming scheme to your liking.
+    - You have already generated thumbnails in your bucket.
+      Then choose the thumbnail suffix such that it reflects you naming scheme.
```
MalteEbner, line 58: youR naming scheme

MalteEbner (commented):
This looks quite good!

However, if we cannot distinguish between read and read+write access, then leave it away everywhere.

MalteEbner (approved):
LGTM

PR #647, opened 2021-12-28
DRAFT of openapi-generated client

MalteEbner:
# DRAFT for writing down findings

- OK: imports needed to be changed, now the import is directly from the file 
- BAD: the allowed values of an enum are not easily accessible anymore, see https://github.com/OpenAPITools/openapi-generator/issues/10752
- MIXED: the basic input types are often not anymore a string, but e.g. a VersionNumber or DatasetName.
   - GOOD: This allows partly for validation on the client side, e.g. a DatasetName has a minimum length of 3, which is checked for while creating.
   - BAD: Lots of code changes are needed, both before sending and after receiving.
- GOOD: The api methods now have named arguments and not only args and kwargs. This was introduced in 2021. The fact that it was missing was one of the reasons we chose `swagger-codegen` instead of `openapitools-codegen`.


philippmwirth:
Let's close the PR for the moment - you can keep the branch :)

spacether:
You may also want to look into the [python-experimental](https://github.com/OpenAPITools/openapi-generator/blob/master/docs/generators/python-experimental.md#metadata) generator which has more robust composition and type hints built in.
In it enums are [also accessible as class attributes](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python-experimental/petstore_api/model/string_enum.py#L94).

MalteEbner:
Thank you @spacether, I will try it out.

PR #654, opened 2022-01-04
Igor lig 468 why are there different backbones across

### LINKED ISSUE #650: Why are there different backbones across SSL methods in CIFAR10 benchmark?
Hi, thank you so much for developing this great library, I found it extremely useful for my work!

I noticed that in the [CIFAR10 benchmark script](https://github.com/lightly-ai/lightly/blob/master/docs/source/getting_started/benchmarks/cifar10_benchmark.py), there are different backbones across the SSL methods. 

The two differences I've noticed are:
1. some use `torchvision.models.resnet18()`, others use `lightly.models.ResNetGenerator('resnet-18')`
    * I understand that for the CIFAR datasets, we want to use a variant that's slightly different than the ones in torchvision (which is typically used for ImageNet)
    * what I don't understand is why you did not use one variant consistently across all SSL methods 
2. some have an additional `nn.AdaptiveAvgPool2d(1)` layer at the end, some don't. 
    * my guess would be that this has something to do with the respective projection heads
    * to be safe, I hope you don't mind me asking

I'd highly appreciate any help!

IgorSusmelj:
### Changes

- benchmark code works again
  - updated PyTorch lightning code
  - updated SwaV collate parameters
- make sure all models use the same backbone

closes #650

### DIFF
--- a/docs/source/getting_started/benchmarks/cifar10_benchmark.py
+++ b/docs/source/getting_started/benchmarks/cifar10_benchmark.py
@@ -231,7 +231,7 @@ class SimCLRModel(BenchmarkModule):
         resnet = lightly.models.ResNetGenerator('resnet-18')
         self.backbone = nn.Sequential(
             *list(resnet.children())[:-1],
-            nn.AdaptiveAvgPool2d(1),
+            nn.AdaptiveAvgPool2d(1)
         )
         # create a simclr model based on ResNet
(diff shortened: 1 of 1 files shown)


guarin (changes requested):
Looks good :)

guarin (approved):

PR #660, opened 2022-01-12
Malte lig 454 docker fails when files dont have read

MalteEbner:
## Description
- os.walk used for the video dataset now throws all errors it encounters including permission errors

## New tests
Added many tests for the following cases, all throwing permission errors.

#### Types of datasets for reference, there is one test for each of them.
- datasets.ImageFolder ( from torch for nested dirs)
- DatasetFolder (self-implemented, for plain dirs)
- VideoDataset (self.implemented, for plain and nested dirs)

#### With a nested image folder created:
- the images in the subdirs have no read rights (fails when iterating over the datasets.ImageFolder)
- subdirs have no read rights (fails in `_contains_subdirs()` containing `os.scandir(root)`)
- root has no read rights (fails in `_contains_subdirs()` containing `os.scandir(root)`)
#### With a plain image folder created:
- the images in root have no read rights (fails when iterating over the DatasetFolder)
- root has no read rights (fails in `_contains_subdirs()` containing `os.scandir(root)`)
#### With a video dataset created
- the videos in the subdirs have no read rights (fails when creating the VideoDataset, as it cannot read the video to know the number of frames)
- subdirs have no read rights (fails in `_contains_subdirs()` containing `os.scandir(root)`)
- root has no read rights (fails in `_contains_subdirs()` containing `os.scandir(root)`)

## NOT TESTED
- Image dataset with sub-sub-directories. I expect it to silently ignore all sub-sub-directories without read rights, as `os.walk` used but `datasets.ImageFolder` does that.
- Video dataset with videos in plain folder. I expect it to fail during creation just like for the other dataset, as it uses `os.walk` with error throwing internally.
- Video dataset with videos in sub-sub-directories. I expect it to fail during creation just like for the other video dataset, as it uses `os.walk` with error throwing internally.

### DIFF
--- a/lightly/data/_helpers.py
+++ b/lightly/data/_helpers.py
@@ -25,11 +25,6 @@ VIDEO_EXTENSIONS = ('.mp4', '.mov', '.avi', '.mpg',
                     '.hevc', '.m4v', '.webm', '.mpeg')
 
 
-def walk_dir_not_ignoring_erors(root):
-    def on_error(error):
-        raise error
-    return os.walk(root, onerror=on_error)
-
 def _dir_contains_videos(root: str, extensions: tuple):
     """Checks whether directory contains video files.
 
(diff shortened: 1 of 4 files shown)


philippmwirth (changes requested):
Could you add the same tests for video datasets and "flat" image datasets? Then we cover all the cases.

philippmwirth (approved):

PR #662, opened 2022-01-17
Malte lig 529 m01 doesnt work with multiple workers

MalteEbner:
## Description
- Ensures that `SelfSupervisedEmbedding.embed(dataloader)` returns the embeddings, labels and filenames exactly in the same order as in the dataset. This prohibits errors in later steps, where the embeddings are used.
- Removed `SelfSupervisedEmbedding.embed(to_numpy)` argument, as it is not used anywhere.
- Added unittest to ensure that the order is independent of the number of workers.
- Reworked `embed_cli` by extracting `get_model_from_config()` as a separate function, which is used by the unittest.
- fixed wrong typehint (unrelated)

The formatting of `embed_cli.py` and `test_embedding.py` follow black format.

### DIFF
--- a/lightly/embedding/embedding.py
+++ b/lightly/embedding/embedding.py
@@ -71,8 +71,9 @@ class SelfSupervisedEmbedding(BaseEmbedding):
             model, criterion, optimizer, dataloader, scheduler
         )
 
-    def embed(
-        self, dataloader: torch.utils.data.DataLoader, device: torch.device = None
+    def embed(self,
+              dataloader: torch.utils.data.DataLoader,
+              device: torch.device = None
     ) -> Tuple[np.ndarray, np.ndarray, List[str]]:
         """Embeds images in a vector space.
 
@@ -83,7 +84,7 @@ class SelfSupervisedEmbedding(BaseEmbedding):
                 Selected device (`cpu`, `cuda`, see PyTorch documentation)
 
         Returns:
-            The order of the return values is determined by the order of
+            Tuple of (embeddings, labels, filenames) ordered by the
             samples in the dataset of the dataloader.
                 embeddings:
                     Embedding of shape (n_samples, embedding_feature_size).
@@ -105,7 +106,8 @@ class SelfSupervisedEmbedding(BaseEmbedding):
 
         if lightly._is_prefetch_generator_available():
             pbar = tqdm(
-                BackgroundGenerator(dataloader, max_prefetch=3), total=len(dataloader)
+                BackgroundGenerator(dataloader, max_prefetch=3),
+                total=len(dataloader)
             )
         else:
             pbar = tqdm(dataloader, total=len(dataloader))
@@ -115,7 +117,7 @@ class SelfSupervisedEmbedding(BaseEmbedding):
         labels = []
         with torch.no_grad():
 
-            start_time = time.time()
+            start_timepoint = time.time()
             for (img, label, fname) in pbar:
 
                 img = img.to(device)
@@ -123,7 +125,7 @@ class SelfSupervisedEmbedding(BaseEmbedding):
                 fnames += [*fname]
 
                 batch_size = img.shape[0]
-                prepare_time = time.time()
+                prepared_timepoint = time.time()
 
                 emb = self.model.backbone(img)
                 emb = emb.detach().reshape(batch_size, -1)
@@ -131,11 +133,15 @@ class SelfSupervisedEmbedding(BaseEmbedding):
                 embeddings.append(emb)
                 labels.append(label)
 
-                process_time = time.time()
+                finished_timepoint = time.time()
 
-                efficiency = (process_time - prepare_time) / (process_time - start_time)
+                data_loading_time = prepared_timepoint - start_timepoint
+                inference_time = finished_timepoint - prepared_timepoint
+                total_batch_time = data_loading_time + inference_time
+
+                efficiency = inference_time / total_batch_time
                 pbar.set_description("Compute efficiency: {:.2f}".format(efficiency))
-                start_time = time.time()
+                start_timepoint = time.time()
 
             embeddings = torch.cat(embeddings, 0)
             labels = torch.cat(labels, 0)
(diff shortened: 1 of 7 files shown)


lightly/embedding/embedding.py
```diff
@@ -56,33 +58,41 @@ class SelfSupervisedEmbedding(BaseEmbedding):
...
-            model, criterion, optimizer, dataloader, scheduler)
+            model, criterion, optimizer, dataloader, scheduler
+        )
 
-    def embed(self,
-              dataloader: torch.utils.data.DataLoader,
-              device: torch.device = None,
-              to_numpy: bool = True):
+    def embed(
+        self, dataloader: torch.utils.data.DataLoader, device: torch.device = None
```
philippmwirth, line 75: Let's do function definitions as outlined in the [style guide](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods): One argument a line, and with typehints:
```python
def embed(self,
          dataloader: torch.utils.DataLoader,
          device: torch.device = None
) -> Tuple[np.ndarray, np.ndarray, List[str]]:
    """DOCSTRING

    """
```

grafik

lightly/embedding/embedding.py
```diff
@@ -56,33 +58,41 @@ class SelfSupervisedEmbedding(BaseEmbedding):
...
                 A PyTorch dataloader.
             device:
                 Selected device (`cpu`, `cuda`, see PyTorch documentation)
-            to_numpy:
-                Whether to return the embeddings as numpy array.
 
         Returns:
-            A tuple consisting of a tensor or ndarray of embeddings
-            with shape n_images x num_ftrs and labels, fnames
+            The order of the return values is determined by the order of
```
philippmwirth, line 86: This sentence is the most important to you since you just worked on this issue but if someone else reads the docstrings I think it should say what it returns first and then go into detail about the order of the returned values (if they are even important). So something like:

```
Returns:
    Embeddings, labels, and filenames of the embedded images ordered by the
    samples in the dataset of the dataloader.
```

lightly/embedding/embedding.py
```diff
@@ -94,12 +104,13 @@ def embed(self,
         embeddings, labels, fnames = None, None, []
 
         if lightly._is_prefetch_generator_available():
-            pbar = tqdm(BackgroundGenerator(dataloader, max_prefetch=3),
-                        total=len(dataloader))
+            pbar = tqdm(
+                BackgroundGenerator(dataloader, max_prefetch=3), total=len(dataloader)
```
philippmwirth, line 108: I think it's much more readable if the arguments are on two lines

lightly/embedding/embedding.py
```diff
@@ -123,16 +133,26 @@ def embed(self,
 
                 process_time = time.time()
 
-                efficiency = \
-                    (process_time - prepare_time) / (process_time - start_time)
-                pbar.set_description(
-                    "Compute efficiency: {:.2f}".format(efficiency))
+                efficiency = (process_time - prepare_time) / (process_time - start_time)
```
philippmwirth, line 136: While you're at it, I have the feeling that the efficiency calculation is buggy... Do you think it makes sense like this?

lightly/embedding/embedding.py
```diff
@@ -123,16 +133,26 @@ def embed(self,
...
+                pbar.set_description("Compute efficiency: {:.2f}".format(efficiency))
                 start_time = time.time()
 
             embeddings = torch.cat(embeddings, 0)
             labels = torch.cat(labels, 0)
-            if to_numpy:
```
philippmwirth, line 134: Why did you remove the `to_numpy`?

lightly/embedding/embedding.py
```diff
@@ -123,16 +133,26 @@ def embed(self,
...
-                embeddings = embeddings.cpu().numpy()
-                labels = labels.cpu().numpy()
 
-        return embeddings, labels, fnames
+            embeddings = embeddings.cpu().numpy()
+            labels = labels.cpu().numpy()
+
+        list_to_order = zip(embeddings, labels)
```
philippmwirth, line 146: this is not a list

lightly/embedding/embedding.py
```diff
@@ -123,16 +133,26 @@ def embed(self,
...
+        dict_by_filenames = dict(zip(fnames, list_to_order))
```
philippmwirth, line 147: the [style guide](https://google.github.io/styleguide/pyguide.html#3161-names-to-avoid) recommends to avoid names which include the type of the variable.

philippmwirth (changes requested):

lightly/embedding/embedding.py
MalteEbner: For 2 reasons:
- It is not used.
- The following lines of sorting and stacking with `np.stack` rely on numpy arrays.

I would not carry on functionality that is not used, makes the function and understanding it more complex and causes extra effort to maintain.

lightly/embedding/embedding.py
MalteEbner: I think that the calculation was correct. However, I refactored it a little to make it better understandable what is done.

lightly/embedding/embedding.py
philippmwirth: > I would not carry on functionality that is not used, makes the function and understanding it more complex and causes extra effort to maintain.

But what if people use it in their code base and everything breaks?

lightly/embedding/embedding.py
```diff
@@ -94,45 +105,60 @@ def embed(self,
...
             labels = torch.cat(labels, 0)
-            if to_numpy:
-                embeddings = embeddings.cpu().numpy()
-                labels = labels.cpu().numpy()
 
-        return embeddings, labels, fnames
+            embeddings = embeddings.cpu().numpy()
+            labels = labels.cpu().numpy()
+
+        to_order = zip(embeddings, labels)
```
philippmwirth, line 152: Sorry I didn't mention that before:
- I think you are trying to reduce redundant code (which is great) by using zip (which is a little hard to read)
- I think a cleaner way would be to use a function which orders any list by any other list:

```python
def sort_items_by_keys(
    keys: List[any],
    items: List[any],
    sorted_keys: List[any]
):
    lookup = { key_: item_ for key_, item_ in zip(keys, items) }
    sorted_ = [lookup[key_] for key_ in sorted_keys]
    return sorted_

Comment (philippmwirth) [reply lightly/embedding/embedding.py:152]: And then 
```python
sorted_filenames = dataloader.dataset.get_filenames()
sorted_embeddings = sort_items_by_keys(fnames, embeddings, sorted_filenames)
sorted_labels = sort_items_by_keys(fnames, labels, sorted_filenames)
return ....
```

philippmwirth (approved):

lightly/embedding/embedding.py
MalteEbner: We probably break working undocumented stuff all the time without getting to know about it. 
If someone really uses that extra argument, they will know how to fix it themselves by casting the numpy tensor to a torch tensor again. I think there are places were we can invest our developer time such that more users value is provided.

philippmwirth (approved):
I added the missing docstring

MalteEbner:
> I added the missing docstring

Thank you!

PR #663, opened 2022-01-18
Guarin lig 528 make metadata upload faster

guarin:
Make metadata upload requests concurrent.

I could not use the `asyn_req` parameter provided by swagger to make async requests because we cannot easily limit the number of concurrent requests.

Failed uploads are now retried.

Benchmark with cifar10 test set (10k images, 10 metadata entries per image):
Old: ~1h
New (8 workers): 3:57 min
New (16 workers): 2:13 min
New (32 workers): 1:39 min

### DIFF
--- a/lightly/api/api_workflow_upload_metadata.py
+++ b/lightly/api/api_workflow_upload_metadata.py
@@ -205,5 +205,4 @@ class _UploadCustomMetadataMixin:
             if verbose:
                 results = tqdm(results, total=len(sample_requests))
             # iterate over results to make sure they are completed
-            for _ in results:
-                pass
+            list(results)
--- a/lightly/api/api_workflow_upload_metadata.py
+++ b/lightly/api/api_workflow_upload_metadata.py
@@ -1,8 +1,11 @@
+import concurrent
+from concurrent.futures import ThreadPoolExecutor
 from typing import Dict, List
 from bisect import bisect_left
 
-import tqdm
+from tqdm import tqdm
 
+from lightly.api.utils import retry
 from lightly.openapi_generated.swagger_client.models.sample_update_request import \
     SampleUpdateRequest
 from lightly.utils.io import COCO_ANNOTATION_KEYS
@@ -112,7 +115,8 @@ class _UploadCustomMetadataMixin:
 
     def upload_custom_metadata(self,
                                custom_metadata: Dict,
-                               verbose: bool = False):
+                               verbose: bool = False,
+                               max_workers: int = 8):
         """Uploads custom metadata to the Lightly platform.
 
         The custom metadata is expected in a format similar to the COCO annotations:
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_upload_metadata.py
```diff
@@ -172,22 +178,34 @@ def upload_custom_metadata(self,
...
+            return retry(
+                self._samples_api.update_sample_by_id,
+                request,
+                dataset_id=self.dataset_id,
+                sample_id=sample.id,
+            )
+
+        # limit number of concurrent requests
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
+            futures = []
```
MalteEbner, line 192: What is the reason you used this futures implementation of parallel upload? It is unknown for me and thus hard to review.
I think it would be better to use the same workflow as for uploading the samples: 
https://github.com/lightly-ai/lightly/blob/4daa75625082a36511e45a3bf7f82c9512eea36b/lightly/api/api_workflow_upload_dataset.py#L196-L198

lightly/api/api_workflow_upload_metadata.py
```diff
@@ -172,22 +178,34 @@ def upload_custom_metadata(self,
...
+            )
+                if metadata is not None:
+                    future = executor.submit(
+                        upload_sample_metadata,
+                        update_sample_request,
+                        sample,
+                    )
+                    futures.append(future)
+            
+            completed_futures = concurrent.futures.as_completed(futures)
```
MalteEbner, line 206: Could you somehow tell that this is an iterator?

MalteEbner: In general, some comments here would be nice.

MalteEbner (commented):
It seems correct to me, but I am new to futures.

lightly/api/api_workflow_upload_metadata.py
guarin: Oh, I just named it `futures` instead of `results` but it is the same thing :)

guarin: Nvm, I see what you mean. I'll change it. Thanks a lot :)

lightly/api/api_workflow_upload_metadata.py
guarin: Added more comments

lightly/api/api_workflow_upload_metadata.py
```diff
@@ -179,33 +179,31 @@ def upload_custom_metadata(self,
...
+            # get iterator over results
+            results = executor.map(upload_sample_metadata, sample_requests)
             if verbose:
-                completed_futures = tqdm(completed_futures, total=len(futures))
-
-            for future in completed_futures:
-                future.result()
\ No newline at end of file
+                results = tqdm(results, total=len(sample_requests))
+            # iterate over results to make sure they are completed
```
MalteEbner, line 207: Can't you do that simpler by just creating a list out of them, just like in the sample upload?
https://stackoverflow.com/questions/3790848/fastest-way-to-convert-an-iterator-to-a-list

MalteEbner (commented):
This looks good and well-understandable to me. However, I have two points left.
- using `list` to get the elements of the iterator
- have you tested that tqdm works nicely? In the sample upload tqdm is more complex and using a lock. Perhaps there is a reason for that.
If tqdm works here ok-ish that is fine for me, I don't need it to work perfectly, but it should give a rough guesstimate of how long something takes.

lightly/api/api_workflow_upload_metadata.py
guarin: Good catch :)

guarin:
> * have you tested that tqdm works nicely? In the sample upload tqdm is more complex and using a lock. Perhaps there is a reason for that.
>   If tqdm works here ok-ish that is fine for me, I don't need it to work perfectly, but it should give a rough guesstimate of how long something takes.

The sample upload is different because tqdm is called within the function that runs on a different thread, thats why it uses a lock. Here tqdm is called in the main thread and it works well.

MalteEbner (approved):
LGTM

PR #666, opened 2022-01-19
Guarin lig 136 add multi gpu support

guarin:


lightly/loss/swav_loss.py
```diff
@@ -22,23 +27,33 @@ def sinkhorn(out: torch.Tensor, iterations: int = 3, epsilon: float = 0.05):
...
     K = Q.shape[0] # number of prototypes
 
-    for i in range(iterations):
+    for _ in range(iterations):
         # normalize rows
         sum_of_rows = torch.sum(Q, dim=1, keepdim=True)
+        if world_size > 1:
+            dist.all_reduce(sum_of_rows)
         Q /= sum_of_rows
         Q /= K
```
MalteEbner, line 58: This line might be redundant.

MalteEbner (commented):
Looks good to me, only some comments could be improved.

lightly/loss/swav_loss.py
MalteEbner: Q /= K is redundant, as the columns are normalised in the line below.

MalteEbner: If you insert `Q = Q / K` into `Q = Q / torch.sum(Q, dim=0, keepdim=True)` then you have
`Q = (Q / K) / torch.sum((Q / K), dim=0, keepdim=True)` and the K cancels itself out.

guarin: I agree with you :)

MalteEbner (approved):
LGTM

PR #668, opened 2022-01-19
Fix tests

guarin:
Pass device to embed function
Lower required tolerance in assert because gpu is not that precise


MalteEbner (approved):
This makes sense to me.

PR #677, opened 2022-01-28
Guarin lig 596 add dinoloss

guarin:


lightly/loss/dino_loss.py
```diff
@@ -0,0 +1,206 @@
...
+        super().__init__()
+        self.warmup_teacher_temp_epochs = warmup_teacher_temp_epochs
+        self.teacher_temp = teacher_temp
+        self.student_temp = student_temp
+        self.center_momentum = center_momentum
+        
+        self.register_buffer("center", torch.zeros(1, out_dim))
+        # we apply a warm up for the teacher temperature because
+        # a too high temperature makes the training instable at the beginning
+        self.teacher_temp_schedule = np.linspace(
```
philippmwirth, line 104: could use torch linspace here over numpy

tests/loss/test_DINOLoss.py
```diff
@@ -0,0 +1,260 @@
...
+import numpy as np
+import torch
+from torch import nn
+import torch.nn.functional as F
+
+from lightly.loss import dino_loss
+from lightly.loss import DINOLoss
+
+
+class FacebookDINOLoss(nn.Module):
```
philippmwirth, line 13: maybe add a comment that you copy pasted this from somewhere so it's very clear what the purpose is

philippmwirth (commented):
Looks mostly good 👍

lightly/loss/dino_loss.py
```diff
@@ -0,0 +1,206 @@
...
+    """Concatenates multiple outputs from the student model into a single 
+    tensor.
+    
+    Args:
+        batch_size:
+            The batch size B where B = number of images.
+        out_dim:
+            The dimensions of the model output.
+        student_out:
+            A single tensor with shape (B * V, D) or a list of tensors where
```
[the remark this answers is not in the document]
guarin: Yes it is the number of views. Added a comment about it.

lightly/loss/dino_loss.py
guarin: Ah good point!

philippmwirth (approved):

tests/loss/test_DINOLoss.py
```diff
@@ -0,0 +1,269 @@
...
```
IgorSusmelj, line 13: Would drop the `Facebook` part here. Or are there various implementations of the same loss?

philippmwirth (approved):
Looks good!

PR #680, opened 2022-02-01
Guarin lig 597 add dinocollatefunction

guarin:


lightly/data/collate.py
```diff
@@ -476,4 +478,187 @@ def __init__(self,
...
+            view = torch.stack([transform(img) for img, _, _ in batch])
+            views.append(view)
+        # list of labels
+        labels = torch.LongTensor([label for _, label, _ in batch])
+        # list of filenames
+        fnames = [fname for _, _, fname in batch]
+        return views, labels, fnames
+
+
+class DINOCollateFunction(MultiViewCollateFunction):
```
philippmwirth, line 519: What's the reason you're not inheriting from the `MultiCropCollateFunction`?

guarin: Because dino uses different transforms for some of the views. For example the first global view has no solarization while the second one has solarization. This is not supported by the `MultiCropCollateFunction` and I did not want to refactor it and make breaking changes.

philippmwirth: I see...

guarin: I think it would make sense to refactor the SwAV collate function to use the MultiView collate function in the future as it is more generic and supports any number of views and transforms.

philippmwirth: As a first step we could make the `MultiCropCollate` function inherit from the `MultiViewCollate`. Then the concepts are related at least.

philippmwirth (approved):

PR #682, opened 2022-02-01
Guarin lig 599 add dinoexamples

guarin:
Adds examples for dino. Fixes some small issues with other pytorch lightning models.

### DIFF
--- a/examples/pytorch/dino.py
+++ b/examples/pytorch/dino.py
@@ -55,7 +55,7 @@ collate_fn = DINOCollateFunction()
 
 dataloader = torch.utils.data.DataLoader(
     dataset,
-    batch_size=128,
+    batch_size=64,
     collate_fn=collate_fn,
     shuffle=True,
     drop_last=True,
@@ -69,7 +69,7 @@ criterion = DINOLoss(
 # move loss to correct device because it also contains parameters
 criterion = criterion.to(device)
 
-optimizer = torch.optim.SGD(model.parameters(), lr=0.06)
+optimizer = torch.optim.Adam(model.parameters(), lr=0.001)
 
 print("Starting Training")
 for epoch in range(10):
--- a/examples/pytorch/dino.py
+++ b/examples/pytorch/dino.py
@@ -75,6 +75,8 @@ print("Starting Training")
 for epoch in range(10):
     total_loss = 0
     for views, _, _ in dataloader:
+        update_momentum(model.student_backbone, model.teacher_backbone, m=0.99)
+        update_momentum(model.student_head, model.teacher_head, m=0.99)
         views = [view.to(device) for view in views]
         global_views = views[:2]
         teacher_out = [model.forward_teacher(view) for view in global_views]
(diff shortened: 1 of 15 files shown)


examples/pytorch/dino.py
```diff
@@ -0,0 +1,90 @@
...
+        y = self.teacher_backbone(x).flatten(start_dim=1)
+        z = self.teacher_head(y)
+        return z
+
+resnet = torchvision.models.resnet18()
+backbone = nn.Sequential(*list(resnet.children())[:-1])
+input_dim = 512
+# instead of a resnet you can also use a vision transformer backbone as in the
+# original paper (you might have to reduce the batch size in this case):
+# backbone = torch.hub.load('facebookresearch/dino:main', 'dino_vits16', pretrained=False)
```
IgorSusmelj, line 38: Would add an example using timm models (see https://github.com/rwightman/pytorch-image-models) here in the comments. Either additionally or instead of the facebook research one. Just simply because the repo has tons of more models and is well maintained :)

guarin: I just tested it and the vision transformers in timm expect all input images to have a fixed size. But DINO makes global and local crops with various sizes :(

examples/pytorch/dino.py
```diff
@@ -0,0 +1,90 @@
...
+        views = [view.to(device) for view in views]
+        global_views = views[:2]
+        teacher_out = [model.forward_teacher(view) for view in global_views]
+        student_out = [model.forward(view) for view in views]
+        loss = criterion(teacher_out, student_out, epoch=epoch)
+        total_loss += loss.detach()
+        loss.backward()
+        optimizer.step()
+        optimizer.zero_grad()
+        update_momentum(model.student_backbone, model.teacher_backbone, m=0.99)
```
philippmwirth, line 87: here you are doing the momentum update at the end of the step and in the lightning example you do it in the beginning. can we be consistent or is there a specific reason for this?

philippmwirth (changes requested):
I think you forgot to add the model to the index file as I can only access it by the direct link in the docs.

Did you try it with a vision transformer? If yes does it work? Do you have any numbers as to how good the model is if you run the example?

guarin:
> I think you forgot to add the model to the index file as I can only access it by the direct link in the docs.
Added it :)

> Did you try it with a vision transformer?
Yes

> If yes does it work? 
The loss decreases over many epochs so it looks ok.

> Do you have any numbers as to how good the model is if you run the example?
There is no validation code in the examples. I'll update and run the benchmarks to see how well it works.

examples/pytorch/dino.py
guarin: I moved it now to the top of the training loop. In general I would prefer it to be at the bottom of the loop but this does not work for pytorch lightning. This is because the optimizer step is executed after the forward step and thus it would no make sense to update the momentum at the end of the forward step.

philippmwirth (approved):
lgtm 🥳

PR #683, opened 2022-02-01
Retry downloads

philippmwirth:
# Retry downloads

Handle 5XX errors according to best practices ([link](https://aws.amazon.com/de/premiumsupport/knowledge-center/http-5xx-errors-s3/)).


MalteEbner (approved):
LGTM

PR #686, opened 2022-02-08
allow ignoring metadata and restarts

philippmwirth:
# allow ignoring metadata and restarts

Implements two features for video download:
- Restart the download of a video at a given timestamp in `download_all_video_frames`
- Count the number of frames instead of relying on the metadata in `all_video_frame_counts`

### DIFF
--- a/lightly/api/download.py
+++ b/lightly/api/download.py
@@ -45,6 +45,7 @@ def download_image(url: str, session: requests.Session = None) -> PIL.Image.Imag
 
 def download_all_video_frames(
     url: str, 
+    timestamp: Union[int, None] = None,
     as_pil_image: int = True, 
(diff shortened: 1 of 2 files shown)


lightly/api/download.py
```diff
@@ -68,10 +73,32 @@ def download_all_video_frames(
 
     """
     _check_av_available()
+    timestamp = 0 if timestamp is None else timestamp
+    if timestamp < 0:
+        raise ValueError(f"Negative timestamp is not allowed: {timestamp}")
```
guarin, line 78: I am actually not sure if negative timestamps are not allowed. I did not add the check because we only use the function with timestamps we already retrieved from the frames, so they should be correct.

guarin (approved):
Looks good, added one comment :)

PR #687, opened 2022-02-08
Guarin lig 648 fix subset get filenames

guarin:


MalteEbner (approved):
Great unittest, it really shows the problem is solved.

I was just wondering, why this error existed at all and found that these two lines added recently had caused the error: https://github.com/lightly-ai/lightly/blob/dad20be3de82a5222cbc7a4705449a49c1a50d84/lightly/data/dataset.py#L263-L264
If they are removed, the` Subset.get_filenames()` method you added would not have been necessary. Thus my takeaway is that we need to be more careful with such changes and/or test more.

guarin:
> If they are removed, the Subset.get_filenames() method you added would not have been necessary. Thus my takeaway is that we need to be more careful with such changes and/or test more.

They were added to support a custom get_filenames function for datasets. This is why loading videos got 100x faster. There is sadly not much we can do there. The error happened because I forgot to overwrite get_filenames for the LightlySubset in the first place.

PR #691, opened 2022-02-08
upload: set default number of workers to at least 8 

MalteEbner:
## Description
- when not setting the number of workers when uploading, use `max(8, number cores)`.

## Benchmark Results
Benchmarked the upload to dev with a cloud machine with 8 virtual cores. The cores restricted via taskset and checked via htop, that worked fine.

### High level results / interpretation
- The more virtual cores the better.
- More physical cores don't help a lot when the number of virtual cores stays the same.
- The more upload workers, the better usually. However, more that 8 workers rarely help.
- The bigger the images, the slower the upload.

### Table
|                      |                     |                     | Dataset with 364 images (jungfraujoch_m, 1380x450) |                                | Dataset with 372 images (clothing small test) | Dataset with 3068 images (clothing small train) |
|----------------------|---------------------|---------------------|----------------------------------------------------|--------------------------------|-----------------------------------------------|-------------------------------------------------|
| # physical cpu cores | # virtual cpu cores | # uploading workers | Speed [img/s]                                      | Speed [img/s], 2nd measurement | Speed [img/s]                                 | Speed [img/s]                                   |
| 4                    | 8                   | 32                  | 24.04                                              | 20.29                          |                                               |                                                 |
| 4                    | 8                   | 16                  | 17.78                                              | 20.01                          |                                               |                                                 |
| 4                    | 8                   | 8                   | 17.77                                              |                                |                                               |                                                 |
| 4                    | 8                   | 4                   | 10.32                                              |                                |                                               |                                                 |
|                      |                     |                     |                                                    |                                |                                               |                                                 |
| 4                    | 4                   | 32                  | 15.83                                              |                                |                                               |                                                 |
| 4                    | 4                   | 16                  | 14.18                                              |                                |                                               |                                                 |
| 4                    | 4                   | 8                   | 13.78                                              |                                |                                               |                                                 |
| 4                    | 4                   | 4                   | 9.64                                               |                                |                                               |                                                 |
|                      |                     |                     |                                                    |                                |                                               |                                                 |
| 2                    | 4                   | 32                  | 14.37                                              | 14.56                          | 29.82                                         | 28.55                                           |
| 2                    | 4                   | 16                  | 17.26                                              |                                | 24.22                                         |                                                 |
| 2                    | 4                   | 8                   | 18.02                                              |                                | 21.13                                         |                                                 |
| 2                    | 4                   | 4                   | 10.50                                              |                                | 13.67                                         |                                                 |
| 2                    | 4                   | 2                   | 5.34                                               |                                | 6.79                                          |                                                 |
|                      |                     |                     |                                                    |                                |                                               |                                                 |
| 2                    | 2                   | 32                  | 13.18                                              |                                | 22.90                                         |                                                 |
| 2                    | 2                   | 16                  | 10.67                                              | 11.73                          | 22.95                                         |                                                 |
| 2                    | 2                   | 8                   | 12.80                                              | 12.08                          | 20.80                                         |                                                 |
| 2                    | 2                   | 4                   | 9.66                                               |                                | 13.37                                         |                                                 |
| 2                    | 2                   | 2                   | 5.27                                               |                                | 7.10                                          |                                                 |
|                      |                     |                     |                                                    |                                |                                               |                                                 |
| 1                    | 2                   | 32                  |                                                    |                                | 15.13                                         | 24.4                                            |
| 1                    | 2                   | 16                  | 13.22                                              |                                | 21.41                                         |                                                 |
| 1                    | 2                   | 8                   | 12.62                                              |                                | 20.30                                         |                                                 |
| 1                    | 2                   | 4                   | 7.85                                               |                                | 12.71                                         |                                                 |
| 1                    | 2                   | 2                   | 5.20                                               |                                |                                               |                                                 |
|                      |                     |                     |                                                    |                                |                                               |                                                 |
| 1                    | 1                   | 16                  | 6.96                                               |                                |                                               |                                                 |
| 1                    | 1                   | 8                   | 6.53                                               |                                |                                               |                                                 |
| 1                    | 1                   | 4                   | 5.77                                               |                                |                                               |                                                 |
| 1                    | 1                   | 2                   | 4.23                                               |                                |                                               |                                                 |
| 1                    | 1                   | 1                   | 2.66                                               |                                |                                               |                                                 |

### DIFF
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -79,9 +79,10 @@ def _upload_cli(cfg, is_cli_call=True):
         with open(path_to_custom_metadata, 'r') as f:
             custom_metadata = json.load(f)
 
-    # determine the number of available cores
+
     if cfg['loader']['num_workers'] < 0:
-        cfg['loader']['num_workers'] = cpu_count()
+        # set the number of threads to use to minimum 8
+        cfg['loader']['num_workers'] = max(8, cpu_count())
 
     size = cfg['resize']
     if not isinstance(size, int):
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -79,9 +79,10 @@ def _upload_cli(cfg, is_cli_call=True):
         with open(path_to_custom_metadata, 'r') as f:
             custom_metadata = json.load(f)
 
-
+    # set the number of workers if unset
     if cfg['loader']['num_workers'] < 0:
-        # set the number of threads to use to minimum 8
+        # set the number of workers to the number of CPUs available,
+        # but minimum of 8
         cfg['loader']['num_workers'] = max(8, cpu_count())
 
     size = cfg['resize']
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -83,7 +83,9 @@ def _upload_cli(cfg, is_cli_call=True):
     if cfg['loader']['num_workers'] < 0:
         # set the number of workers to the number of CPUs available,
         # but minimum of 8
-        cfg['loader']['num_workers'] = max(8, cpu_count())
+        num_workers = max(8, cpu_count())
+        num_workers = min(32, num_workers)
+        cfg['loader']['num_workers'] = num_workers
 
     size = cfg['resize']
     if not isinstance(size, int):
--- a/lightly/cli/config/config.yaml
+++ b/lightly/cli/config/config.yaml
@@ -67,7 +67,9 @@ collate:
 loader:
   batch_size: 16              # Batch size for training / inference.
   shuffle: True               # Whether to reshuffle data each epoch.
-  num_workers: -1             # Number of workers pre-fetching batches (-1 == number of available cores, minimum of 8 for upload).
+  num_workers: -1             # Number of workers pre-fetching batches.
+                              # -1 == number of available cores,
+                              # if -1, minimum of 8, maximum of 32 workers for upload.
   drop_last: True             # Wether to drop the last batch during training.
 
 # trainer namespace: Passed to pytorch_lightning.Trainer.
--- a/lightly/cli/config/config.yaml
+++ b/lightly/cli/config/config.yaml
@@ -67,7 +67,7 @@ collate:
 loader:
   batch_size: 16              # Batch size for training / inference.
   shuffle: True               # Whether to reshuffle data each epoch.
-  num_workers: -1             # Number of workers pre-fetching batches (-1 == number of available cores).
+  num_workers: -1             # Number of workers pre-fetching batches (-1 == number of available cores, minimum of 8 for upload).
   drop_last: True             # Wether to drop the last batch during training.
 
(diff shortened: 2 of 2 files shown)


lightly/cli/upload_cli.py
```diff
@@ -79,9 +79,11 @@ def _upload_cli(cfg, is_cli_call=True):
         with open(path_to_custom_metadata, 'r') as f:
             custom_metadata = json.load(f)
 
-    # determine the number of available cores
+    # set the number of workers if unset
     if cfg['loader']['num_workers'] < 0:
-        cfg['loader']['num_workers'] = cpu_count()
+        # set the number of workers to the number of CPUs available,
+        # but minimum of 8
+        cfg['loader']['num_workers'] = max(8, cpu_count())
```
guarin, line 86: I think this line should not be in the `if` clause. If we move it outside and change it to: `cfg['loader']['num_workers'] = max(8, cfg['loader']['num_workers'])` the upload will always use at least 8 workers. If we don't do this then uploads might be slow if you manually set the number of workers to a low number (and I guess it breaks if you set it to 0 no?).

guarin (commented):
Left a comment :)

Also, I would maybe limit the number of workers to avoid overwhelming the api, something like 16 or 32 could be reasonable.

lightly/cli/upload_cli.py
MalteEbner: I would do it differently:
- If the user sets `loader.num_workers`, it is taken and used. This is the idea of a config parameter. As a user I would want to have my parameters used just as I set them if I set them.
- Only if the user does not set the parameter, Lightly should recommend a sensible value, here `max(8, cpu_count())`.

guarin: I agree that this is the better approach. Could you just verify that it works with 0 workers and I would still add an upper limit.

MalteEbner: > Could you just verify that it works with 0 workers

0 workers had always worked, it means not to use a separate thread, but use the same thread as the main program. I have also tested it, it works, just as before.

> I would still add an upper limit.

I don't see the sense of it as it
- Prevents the customer from specifying something they want conveniently.
- Does not prevent that customers overwrite it by changing the python code.
- Makes the upload for customers slower if they need it.

guarin: > 0 workers had always worked, it means not to use a separate thread, but use the same thread as the main program. I have also tested it, it works, just as before.
 
Ok perfect!

> I don't see the sense of it as it

The idea of the upper limit is to avoid accidents that the user did not intend to happen. One of them is to overload the api by accident. The other one is to use many more cores on his machine than he intended to. If you use `cpu_count()` on a cluster you will run into trouble. This is also why the `ThreadPoolExecutor` class itself limits the numbers of worker threads if they are not specified:

https://github.com/python/cpython/blob/9b23f8f78fdb0d7eba016616ae7a97abbfc65aa6/Lib/concurrent/futures/thread.py#L134-L142

MalteEbner: @guarin You convinced me to limit it to 32, I implemented it.

guarin (approved):

PR #692, opened 2022-02-08
Add new method for creating tags from a list of filenames

IgorSusmelj:
## Changes

- Added the `create_tag_from_filenames` method
- I tested it and it seems to work fine :)
- Usage example:
image

### DIFF
--- a/lightly/api/api_workflow_tags.py
+++ b/lightly/api/api_workflow_tags.py
@@ -76,14 +76,14 @@ class _TagsMixin:
 
     def create_tag_from_filenames(
         self,
-        list_of_filenames: List[str],
+        fnames_new_tag: List[str],
         new_tag_name: str,
         parent_tag_id: str = None
     ) -> TagData:
         """Creates a new tag from a list of filenames.
 
         Args:
-            list_of_filenames:
+            fnames_new_tag:
                 A list of filenames to be included in the new tag.
             new_tag_name:
                 The name of the new tag.
@@ -111,24 +111,25 @@ class _TagsMixin:
         tot_size = tags[-1].tot_size
 
         # get list of filenames from tag
-        fnames_tag = self.get_filenames_in_tag(tags[-1])
+        fnames_server = self.get_filenames()
 
         # create new bitmask 
         bitmask = BitMask(tot_size)
-        for i, fname in enumerate(fnames_tag):
+        for i, fname_server in enumerate(fnames_server):
             bitmask.unset_kth_bit(i)
-            for fname_selected in list_of_filenames:
-                if fname_selected == fname:
-                    bitmask.set_kth_bit(i)
-                    break
+            if fname_server in fnames_new_tag:
+                bitmask.set_kth_bit(i)
+
         
         # quick sanity check
-        num_selected_samples = len(bitmask.masked_select_from_list(fnames_tag))
-        if num_selected_samples != len(list_of_filenames):
+        num_selected_samples = len(bitmask.to_indices())
+        if num_selected_samples != len(fnames_new_tag):
             raise RuntimeError(
-                f'An error occured when creating the new subset!'
+                f'An error occured when creating the new subset! '
                 f'Found {num_selected_samples} samples newly selected '
-                f'instead of {len(list_of_filenames)}.'
+                f'instead of {len(fnames_new_tag)}. '
+                f'Make sure you use the correct filenames. '
+                f'Valid filename example from the dataset: {fnames_server[0]}'
                 )
 
         # create new tag
--- a/lightly/api/api_workflow_tags.py
+++ b/lightly/api/api_workflow_tags.py
@@ -106,12 +106,11 @@ class _TagsMixin:
 
         # fallback to initial tag if no parent tag is provided
         if parent_tag_id is None:
-            parent_tag_id = tags[-1].id
-
-        tot_size = tags[-1].tot_size
+            parent_tag_id = next(tag.id for tag in tags if tag.name=='initial-tag')
 
         # get list of filenames from tag
         fnames_server = self.get_filenames()
+        tot_size = len(fnames_server)
 
         # create new bitmask for the new tag
         bitmask = BitMask(0)
@@ -125,8 +124,9 @@ class _TagsMixin:
         if num_selected_samples != len(fnames_new_tag):
             raise RuntimeError(
                 f'An error occured when creating the new subset! '
-                f'Found {num_selected_samples} samples newly selected '
-                f'instead of {len(fnames_new_tag)}. '
+                f'Out of the {len(fnames_new_tag)} filenames you provided '
+                f'to create a new tag, only {num_selected_samples} have been '
+                f'found on the server. '
                 f'Make sure you use the correct filenames. '
                 f'Valid filename example from the dataset: {fnames_server[0]}'
                 )
--- a/lightly/api/api_workflow_tags.py
+++ b/lightly/api/api_workflow_tags.py
@@ -73,3 +73,75 @@ class _TagsMixin:
             masked_select_from_list(filenames_on_server)
 
         return filenames_tag
+
+    def create_tag_from_filenames(
+        self,
+        list_of_filenames: List[str],
+        new_tag_name: str,
+        parent_tag_id: str = None
+    ) -> TagData:
+        """Creates a new tag from a list of filenames.
+
+        Args:
+            list_of_filenames:
+                A list of filenames to be included in the new tag.
+            new_tag_name:
+                The name of the new tag.
+            parent_tag_id:
+                The tag defining where to sample from, default: None resolves to the initial-tag.
+
+        Returns:
+            The newly created tag.
+
+        Raises:
+            RuntimeError
+        """
+        
+        # make sure the tag name does not exist yet
+        tags = self.get_all_tags()
+        if new_tag_name in [tag.name for tag in tags]:
+            raise RuntimeError(f'There already exists a tag with tag_name {new_tag_name}.')
+        if len(tags) == 0:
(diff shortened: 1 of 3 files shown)


lightly/api/api_workflow_tags.py
```diff
@@ -73,3 +73,75 @@ def get_filenames_in_tag(
...
+            raise RuntimeError('There exists no initial-tag for this dataset.')
+
+        # fallback to initial tag if no parent tag is provided
+        if parent_tag_id is None:
+            parent_tag_id = tags[-1].id
+
+        tot_size = tags[-1].tot_size
+
+        # get list of filenames from tag
+        fnames_tag = self.get_filenames_in_tag(tags[-1])
```
japrescott, line 114: since you are not getting the filenames for a special tag (but from the initialTag), you can directly use self.get_filenames() and skip computing the names

lightly/api/api_workflow_tags.py
```diff
@@ -73,3 +73,75 @@ def get_filenames_in_tag(
...
+            bitmask.unset_kth_bit(i)
+            for fname_selected in list_of_filenames:
+                if fname_selected == fname:
+                    bitmask.set_kth_bit(i)
+                    break
+        
+        # quick sanity check
+        num_selected_samples = len(bitmask.masked_select_from_list(fnames_tag))
+        if num_selected_samples != len(list_of_filenames):
+            raise RuntimeError(
```
japrescott, line 128: maybe you can give a list of filenames that were not found?

japrescott (changes requested):
looks nice. Would be cool if you would write a mocked test for this

lightly/api/api_workflow_tags.py
```diff
@@ -73,3 +73,75 @@ def get_filenames_in_tag(
...
+        if parent_tag_id is None:
+            parent_tag_id = tags[-1].id
+
+        tot_size = tags[-1].tot_size
+
+        # get list of filenames from tag
+        fnames_tag = self.get_filenames_in_tag(tags[-1])
+
+        # create new bitmask 
+        bitmask = BitMask(tot_size)
```
MalteEbner, line 117: This does not seem the correct way of creating a bitmask, see the docs.
Rather use `Bitmask.from_length(tot_size)`. This is initialised by 1.
 
 I would rather do
 ```
 filenames_server = self.get_filenames()
 
bitmask = BitMask.from_length(tot_size)
filenames_new_tag = set(filenames_new_tag)
for i, fname in enumerate(filenames_server):
    if fname not in filenames_new_tag:
        bitmask.unset_kth_bit(i)
```
This also removes the squared complexity

File: lightly/api/api_workflow_tags.py [hunk 4]
```diff
@@ -73,3 +73,75 @@ def get_filenames_in_tag(
...
+        for i, fname in enumerate(fnames_tag):
+            bitmask.unset_kth_bit(i)
+            for fname_selected in list_of_filenames:
+                if fname_selected == fname:
+                    bitmask.set_kth_bit(i)
+                    break
+        
+        # quick sanity check
+        num_selected_samples = len(bitmask.masked_select_from_list(fnames_tag))
```
MalteEbner, line 126: Use `len(bitmask.to_indices) `instead of `len(bitmask.masked_select_from_list(fnames_tag))`, it is cheaper.

lightly/api/api_workflow_tags.py
```diff
@@ -73,3 +73,75 @@ def get_filenames_in_tag(
...
+        if num_selected_samples != len(list_of_filenames):
+            raise RuntimeError(
+                f'An error occured when creating the new subset!'
```
MalteEbner, line 129: This sanity check should tell better the reason for the error: that `filenames_new_tag` includes filename not found in `filenames_server`.

lightly/api/api_workflow_tags.py
```diff
@@ -73,3 +73,75 @@ def get_filenames_in_tag(
             masked_select_from_list(filenames_on_server)
 
         return filenames_tag
+
+    def create_tag_from_filenames(
+        self,
+        list_of_filenames: List[str],
```
MalteEbner, line 79: I would rename `list_of_filenames` to `filenames_new_tag` because 
- a) It is discouraged to use the type in the variable name
- b) It makes it clearer which filenames are meant.

lightly/api/api_workflow_tags.py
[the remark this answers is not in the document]
MalteEbner: Furthermore, I would rename `fnames_tag` to `filenames_server`.

MalteEbner (changes requested):
The overall approach looks fine to me, but I would improve several substeps.

Furthermore, unittests are missing.

lightly/api/api_workflow_tags.py
japrescott: good suggestion. For what its worth: ` if fname not in filenames_new_tag:` would always search till the end and not terminate early causing it to be n*m instead of n*m/2

IgorSusmelj: Hmmm, that was intentional. The fastest way to create a bitmask is the normal init. You have to loop through both arrays anyway. Not sure how you can go sub "quadratic" or NxM as Jeremy pointed it out :) 
I did rewrite the part though to not use a for loop but a simple `if .. in ..` check.

philippmwirth: @MalteEbner suggests using a set of filenames which reduces the lookup speed to "constant" time. Cleanest solution would probably be to create an index of filename -> index in linear time and then create an empty bitmask and iterate over the filenames in the tag and set the bit at the index.

IgorSusmelj:
Thanks for the great feedback! I did rework the PR and made sure everything still works. Please have another look :)

lightly/api/api_workflow_tags.py
MalteEbner: Yes, the solution I proposed is linear, because the set lookup is constant and not linear.

MalteEbner: Furthermore, `Bitmask(x)` is still unequal to `Bitmask.from_length(x)`. However, the initial length does not matter, as the bits are unset anyway. You can just use` Bitmask(0)` instead, this makes it clear that it does not matter what is inserted during the initialization.

Thus my proposal can be made easier to

```
filenames_server = self.get_filenames()

bitmask = BitMask(0) //initialize empty bitmask
filenames_new_tag = set(filenames_new_tag)
for i, fname in enumerate(filenames_server):
   if fname in filenames_new_tag: # constant time, as it is a set lookup.
       bitmask.set_kth_bit(i)
```

IgorSusmelj: I see, you're right. I missed the `set(...)` part :)

lightly/api/api_workflow_tags.py
```diff
@@ -73,3 +73,75 @@ def get_filenames_in_tag(
...
+        if new_tag_name in [tag.name for tag in tags]:
+            raise RuntimeError(f'There already exists a tag with tag_name {new_tag_name}.')
+        if len(tags) == 0:
+            raise RuntimeError('There exists no initial-tag for this dataset.')
+
+        # fallback to initial tag if no parent tag is provided
+        if parent_tag_id is None:
+            parent_tag_id = tags[-1].id
+
+        tot_size = tags[-1].tot_size
```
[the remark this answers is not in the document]
MalteEbner: I would prefer searching for the tag named 'initial-tag' and take this as the parent tag. Otherwise we might run into subtle bugs we do not find.
It is easy: `parent_tag_id = next(tag.id for tag in tags if tag.name=='initial-tag')`

tests/api_workflow/mocked_api_workflow_client.py
```diff
@@ -181,6 +181,13 @@ def upsize_tags_by_dataset_id(self, body, dataset_id, **kwargs):
         _check_dataset_id(dataset_id)
         assert body.upsize_tag_creator == TagCreator.USER_PIP
 
+    def create_tag_by_dataset_id(self, body, dataset_id, **kwargs):
+        _check_dataset_id(dataset_id)
+        tag = TagData(id='inital_tag_id', dataset_id=dataset_id, prev_tag_id=None,
+                      bit_mask_data="0xF", name=body['name'], tot_size=10,
```
MalteEbner, line 187: You could also use 'bit_mask_data=body['bit_mask_data']' here just like for the name. The same goes for the `prev_tag_id`.

lightly/api/api_workflow_tags.py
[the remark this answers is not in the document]
MalteEbner: And for `tot_size` I would use` len(frames_server)`, then you don't need to rely on the last tag.

lightly/api/api_workflow_tags.py
```diff
@@ -73,3 +73,75 @@ def get_filenames_in_tag(
...
+        for i, fname in enumerate(fnames_server):
+            if fname in fnames_new_tag:
+                bitmask.set_kth_bit(i)
+        
+        # quick sanity check
+        num_selected_samples = len(bitmask.to_indices())
+        if num_selected_samples != len(fnames_new_tag):
+            raise RuntimeError(
+                f'An error occured when creating the new subset! '
+                f'Found {num_selected_samples} samples newly selected '
```
MalteEbner, line 128: `"newly selected"` is weird. What/who did a selection of what from what?

I would rather say:
f"Out of the {len(fnames_new_tag)} filenames you provided to create a new tag, only {num_selected_samples} have been found on the server. Make sure you use the correct filenames. Valid filename example from the dataset: {fnames_server[0]}'"

MalteEbner (commented):
This looks almost perfect, just some minor comments to make it more secure and the error message better understandable.

PR #693, opened 2022-02-10
before upload to server: check embeddings and remove additional columns

MalteEbner:
## Reason for issue
There embedding file uploaded to the server should not include additional columns for some reasons:
- It makes the file smaller.
- It disallows uploading "selected" and "masked" via the embedding file and assuming that they mean something instead of relying on the "preselected_tag" and "query_tag"
- It can cause errors when appending embedding files if the columns are not the same.

For more details and discussion, see the issue https://linear.app/lightly/issue/LIG-565/make-sure-docker-only-uploads-embedding-files-without-additional

## Description
- Added check of embedding file format before uploading embeddings.
- Uploading embeddings now removes additional columns, but keeps the standard ones (filenames, embeddings_x, labels)

### DIFF
--- a/lightly/utils/io.py
+++ b/lightly/utils/io.py
@@ -36,7 +36,7 @@ def check_filenames(filenames: List[str]):
         raise ValueError(f'Invalid filename(s): {invalid_filenames}')
 
 
-def check_embeddings(path: str):
+def check_embeddings(path: str, remove_additional_columns: bool=False):
     """Raises an error if the embeddings csv file has not the correct format
     
     Use this check whenever you want to upload an embedding to the Lightly 
@@ -47,11 +47,14 @@ def check_embeddings(path: str):
     Args:
         path:
             Path to the embedding csv file
+        remove_additional_columns:
+            If True, all additional columns
+            which are not in {filenames, embeddings_x, labels} are removed.
+            If false, they are kept unchanged.
 
     Raises:
         RuntimeError
     """
-    header = []
     with open(path, 'r', newline='') as csv_file:
         reader = csv.reader(csv_file, delimiter=',')
         header: List[str] = next(reader)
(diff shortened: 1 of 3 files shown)


lightly/utils/io.py
```diff
@@ -91,6 +94,18 @@ def check_embeddings(path: str):
                     f'Found empty row on line {i}.'
                     )
 
+    if remove_additional_columns:
+        new_rows = []
+        with open(path, 'r', newline='') as csv_file:
```
guarin, line 99: Why do you reopen the same file? Can you not just continue reading the already opened file and add the header to new_rows?

lightly/utils/io.py
```diff
@@ -91,6 +94,18 @@ def check_embeddings(path: str):
...
+            reader = csv.reader(csv_file, delimiter=',')
+            for row in reader:
+                row = row[:header_labels_idx+1]
```
guarin, line 102: This assumes that the first columns always contain (filenames, embeddings, labels) and no one added another column in between. Is this assumption always correct? Or should we add a check to verify this?

If this is the assumption we take please state it in the docstring. Because right now I would assume that other columns are dropped regardless of their position.

lightly/utils/io.py
MalteEbner: No, I cannot, as the file was already fully read by iterating over all rows.

lightly/utils/io.py
MalteEbner: This is not assumed, but checked for in the lines above.

lightly/utils/io.py
guarin: Oh wow, bad review from my side. You could, however, make the row length check at the same time as creating the `new_rows` list to avoid reading twice.

guarin (approved):

PR #701, opened 2022-02-22
make sure train_model_and_embed_images returns everything

### LINKED ISSUE #696: _lightly_cli is missing it's return statement
as a side effect `train_model_and_embed_images` is not working as expected.

MalteEbner:
## Description
- wrote unittest reproducing the issue
- fixed it in the `train_model_and_embed_images()` function than now uses the `_train_cli()` and `_embed_cli()` instead of the `_lightly_cli()`
- added comments to the `_embed_cli()` to make the return type much clearer
- added a unittest to the lightly-cli to prevent lines from becoming uncovered

closes #696

### DIFF
--- a/lightly/cli/embed_cli.py
+++ b/lightly/cli/embed_cli.py
@@ -9,8 +9,10 @@ command-line interface.
 # All Rights Reserved
 
 import os
+from typing import Union, Tuple, List
 
 import hydra
+import numpy as np
 import torch
 import torchvision
 from torch.utils.hipify.hipify_python import bcolors
@@ -24,7 +26,17 @@ from lightly.cli._helpers import fix_input_path
 from lightly.cli._helpers import cpu_count
 
 
-def _embed_cli(cfg, is_cli_call=True):
+def _embed_cli(cfg, is_cli_call=True) -> \
+    Union[
+        Tuple[List[np.ndarray], List[int], List[str]],
+        str
+    ]:
+    """ See embed_cli() for usage documentation
+
+        is_cli_call:
+            If True, saves the embeddings as file and returns the filepath.
+            If False, returns the embeddings, labels, filenames as tuple.
+    """
     input_dir = cfg['input_dir']
     if input_dir and is_cli_call:
         input_dir = fix_input_path(input_dir)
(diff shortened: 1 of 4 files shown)


lightly/cli/embed_cli.py
```diff
@@ -24,7 +26,17 @@
 from lightly.cli._helpers import cpu_count
 
 
-def _embed_cli(cfg, is_cli_call=True):
+def _embed_cli(cfg, is_cli_call=True) -> \
+    Union[
+        Tuple[List[np.ndarray], List[int], List[str]],
```
philippmwirth, line 31: It's not a list of array, it's just an np.ndarray

lightly/core.py
```diff
@@ -73,7 +76,9 @@ def _add_kwargs(cfg, kwargs):
     return cfg
 
 
-def train_model_and_embed_images(config_path: str = None, **kwargs):
+def train_model_and_embed_images(config_path: str = None, **kwargs) -> Tuple[
+    List[np.ndarray], List[int], List[str]
```
philippmwirth, line 80: same here, or am I wrong?

tests/cli/test_cli_magic.py
```diff
@@ -48,13 +48,29 @@ def parse_cli_string(self, cli_words: str):
...
+                pass
+
+            key_subparts = key.split('.')
+            if len(key_subparts) == 1:
+                self.cfg[key] = value
+            elif len(key_subparts) == 2:
+                self.cfg[key_subparts[0]][key_subparts[1]] = value
+            else:
+                raise ValueError(
+                    f"Keys with more than 2 subparts are not supported,"
```
philippmwirth, line 63: Above you used single quotes and here you use double quotes. I would always use single quotes.

philippmwirth (commented):
nice one! just a few remarks

lightly/cli/embed_cli.py
MalteEbner: Thanks, good that we add the typehints :)

tests/cli/test_cli_magic.py
```diff
@@ -60,8 +60,8 @@ def parse_cli_string(self, cli_words: str):
                 self.cfg[key_subparts[0]][key_subparts[1]] = value
             else:
                 raise ValueError(
-                    f"Keys with more than 2 subparts are not supported,"
-                     "but you entered {key}."
+                    f'Keys with more than 2 subparts are not supported,'
+                     'but you entered {key}.'
```
philippmwirth, line 64: I just noticed that the `f` is in the wrong place here

philippmwirth (changes requested):

philippmwirth (approved):

PR #702, opened 2022-02-22
Guarin lig 692 fix video multiprocessing

guarin:


MalteEbner (approved):
LGTM

guarin:
I benchmarked the following 4 possible implementations for keeping track of VideoLoader instances in the dataset:

* current: What is now on the master branch
* none: Create a new instance in every `__getitem__` call
* single: Cache the last used VideoLoader instance
* all: Cache all used VideoLoader instances (similar to current)

```
| method        | backend      | dataset init | sequential 1 worker | sequential 8 worker | train 8 worker |
|               |              |        (sec) |            (im/sec) |            (im/sec) |       (im/sec) |
|---------------|--------------|-------------:|--------------------:|--------------------:|---------------:|
| current       | video reader |           22 |                 205 |                 244 |          error |
| none          | video reader |           19 |                   6 |                  46 |             36 |
| single        | video reader |           19 |                 175 |                 211 |             37 |
| list          | video reader |           22 |                 207 |                 215 |             42 |
| current       | pyav         |            2 |                   7 |                  34 |             43 |
| none          | pyav         |            2 |                   7 |                  34 |             42 |
| single        | pyav         |            2 |                   7 |                  34 |             42 |
| list          | pyav         |            2 |                   7 |                  34 |             40 |
```

* **current** is the fastest implementation but breaks when using the video_backend + training (or using the same dataloader for multiple iterations). 
* **none** is very slow when using the video_reader backend
* **single** is only slightly slower than **list** and **current** and uses only a constant amount of memory
* **list** is the fastest non-breaking solution but uses memory proportional to the number of videos

I would go for the **single** implementation as this should also fix crashes due to large dataset size when using the video_reader backend.

MalteEbner:
> I would go for the **single** implementation as this should also fix crashes due to large dataset size when using the video_reader backend.

These are great results, thanks! I agree, I would also go with "single"

One question about the `dataset init` with the video_loader: Why does it still need 19s with the video_loader backend? Shouldn't it also need 2s?

lightly/data/_video.py
```diff
@@ -283,6 +303,18 @@ def __init__(self,
         self.offsets = offsets
         self.fps = fps
 
+        # Current VideoLoader instance and the corresponding video index.
```
MalteEbner, line 306: Could you add a link to this PR and/or explain why we use this? So that whoever wants to change this code in the future knows why we implemented it like this.

guarin:
> One question about the dataset init with the video_loader: Why does it still need 19s with the video_loader backend? Shouldn't it also need 2s?

This is because different methods are used to get the timestamps: https://github.com/lightly-ai/lightly/blob/520a9b0ebc06493541ed822110331c4af163028a/lightly/data/_video.py#L196-L218

lightly/data/_video.py
guarin: Added a comment with the link.

MalteEbner (approved):
Great, can't wait to have this published and used in the docker.

PR #705, opened 2022-02-25
Guarin lig 723 pip package include docker endpoints

guarin:


MalteEbner (approved):
You only added the user_id to the mocked_api_workflow_client, right?

guarin:
> You only added the user_id to the mocked_api_workflow_client, right?

Yes I updated the openapi generated code and had then to fix the tests, which I did by adding the user_id.

PR #706, opened 2022-02-25
Malte lig 652 add cli argument for dataset upsizing

MalteEbner:
## Description
- Add the cli argument
- By default, appending samples to a dataset will only be tried if you have specified `append=True`. If append=False and there are already tags, the cli command aborts.
- By default, /appending embeddings to a dataset will only be tried if you have specified `append=True`. If append=False and there is already an embedding file with the specified name, the cli command aborts.
- Wrote unittest that checks that behaviour with all 2^5 = 32 possible combinations:
```python
        for input_dir in [True, False]:
            for existing_dataset in [True, False]:
                for embeddings_path in [True, False]:
                    for existing_embedding in [True, False]:
                        for append in [True, False]:
                            self.check_upload_dataset_and_embedding(
                                input_dir=input_dir,
                                existing_dataset=existing_dataset,
                                embeddings_path=embeddings_path,
                                existing_embedding=existing_embedding,
                                append=append,
                            )

```
## Reworked the upload of metadata
To understand the behaviour, it is good to know that there are 5 sets in total
```python
image_annotation_ids = [annotation['image_id`] for annotation in custom_metadata['annotations']]
image_ids = [image_info['id`] for image_info in custom_metadata['images']]
metadata_filenames = [image_info['file_name`] for image_info in custom_metadata['images']]
local_filenames = dataset.get_filenames()
server_filenames = [sample.file_name for sample in api_workflow_client.get_samples()]
```
These `image_id` sets might be equal, overlapping, one the subset of the other or exclusive. It is required that `image_ids` is a superset of `image_annotation_ids`, but not necessarily the other way round, as there might be images without custom metadata.
The 3 `filename` sets might be equal, overlapping, one the subset of the other(s) or exclusive.

#### Uploading metadata is done together with uploading samples
It is assumed that the user mainly wants to upload the samples. The metadata should only be uploaded if any is found for the local samples to be uploaded.
It is searched for metadata for every sample to be uploaded. Implemented here, the behaviour is as follows:
- If there is no `image_info['filename']` for a filename in `local_filenames`, there is NO error. We just assume that the user does not have metadata for that sample. This happens if `metadata_filenames` is not a superset of `local_filenames`. 
- If there is no `annotation['image_id]` corresponding to the `image_info['id']` in the metadata, there is NO error. We just assume that the user does not have metadata for that sample. This happens if `image_annotation_ids` is not equal to `image_ids`. 

#### Uploading metadata separately
It is assumed that the user wants to upload each and every metadata specified in `custom_metadata['annotations']`. If that fails, a `ValueError` is raised. This might happen if
- There is no `image_info['id']` corresponding to the `annotation['image_id]` in the metadata. This happens if `image_annotation_ids` is not equal to `image_ids`, which is clearly an error in the `custom_metadata` file.
- There is no filename in the `server_filenames` corresponding to the `image_info['file_name`]`. Thus the user had tried to upload custom metadata for a sample that does not exist.

- This behaviour is checked for with 4 different sets for each of the `image_annotation_ids`, `image_ids/metadata_filenames` and `server_filenames`, thus 4³ = 64 test cases.

### Completely unrelated
- Fixed docstring and some arguments of `api_workflow_client.upload_dataset()`.

### DIFF
--- a/lightly/api/api_workflow_upload_metadata.py
+++ b/lightly/api/api_workflow_upload_metadata.py
@@ -47,74 +47,6 @@ class _UploadCustomMetadataMixin:
             COCO_ANNOTATION_KEYS.custom_metadata, custom_metadata
         )
 
-    def index_custom_metadata_by_filename(self,
-                                          filenames: List[str],
-                                          custom_metadata: Dict):
-        """Creates an index to lookup custom metadata by filename.
-
-        Args:
-            filenames:
-                List of filenames.
-            custom_metadata:
-                Dictionary of custom metadata, see upload_custom_metadata for
-                the required format.
-
-        Returns:
-            A dictionary containing custom metdata indexed by filename.
-
-        """
-        custom_metadata_images = custom_metadata[COCO_ANNOTATION_KEYS.images]
-        custom_metadata_metadata = custom_metadata[COCO_ANNOTATION_KEYS.custom_metadata]
-
-        # sort images by filename
-        custom_metadata[COCO_ANNOTATION_KEYS.images] = sorted(
-            custom_metadata_images,
-            key=lambda x: x[COCO_ANNOTATION_KEYS.images_filename]
-        )
-
-        # sort metadata by image id
-        custom_metadata[COCO_ANNOTATION_KEYS.custom_metadata] = sorted(
-            custom_metadata_metadata,
-            key=lambda x: x[COCO_ANNOTATION_KEYS.custom_metadata_image_id]
-        )
-
-        # get a list of filenames for binary search
-        image_filenames = [
-            image[COCO_ANNOTATION_KEYS.images_filename] for image in
-            custom_metadata_images
-        ]
-
-        # get a list of image ids for binary search
-        metadata_image_ids = [
-            data[COCO_ANNOTATION_KEYS.custom_metadata_image_id] for data in
-            custom_metadata_metadata
-        ]
-
-        # map filename to metadata in O(n * logn)
-        filename_to_metadata = {}
-        for filename in filenames:
-
-            image_index = bisect_left(image_filenames, filename)
-            if image_index == len(image_filenames):
-                raise RuntimeError(
-                    f'Image with filename {filename} does not exist in custom metadata!'
-                )
-
-            image = custom_metadata[COCO_ANNOTATION_KEYS.images][image_index]
-            image_id = image[COCO_ANNOTATION_KEYS.images_id]
-
-            metadata_index = bisect_left(metadata_image_ids, image_id)
-            if metadata_index == len(metadata_image_ids):
-                raise RuntimeError(
-                    f'Image with id {image_id} has no custom metadata!'
-                )
-
-            metadata = custom_metadata[COCO_ANNOTATION_KEYS.custom_metadata][
-                metadata_index]
-            filename_to_metadata[filename] = metadata
-
-        return filename_to_metadata
-
     def upload_custom_metadata(self,
                                custom_metadata: Dict,
                                verbose: bool = False,
@@ -217,9 +149,7 @@ class _UploadCustomMetadataMixin:
                 sample_id=sample_id,
             )
 
-        # The
-
-        # limit number of concurrent requests
+        # Upload in parallel with a limit on the number of concurrent requests
         with ThreadPoolExecutor(max_workers=max_workers) as executor:
             # get iterator over results
             results = executor.map(upload_sample_metadata, upload_requests)
--- a/lightly/api/api_workflow_upload_metadata.py
+++ b/lightly/api/api_workflow_upload_metadata.py
@@ -1,6 +1,6 @@
 import concurrent
 from concurrent.futures import ThreadPoolExecutor
-from typing import Dict, List
+from typing import Dict, List, Union
 from bisect import bisect_left
 
 from tqdm import tqdm
@@ -47,7 +47,8 @@ class _UploadCustomMetadataMixin:
             COCO_ANNOTATION_KEYS.custom_metadata, custom_metadata
         )
 
-    def index_custom_metadata_by_filename(self, custom_metadata: Dict):
+    def index_custom_metadata_by_filename(self, custom_metadata: Dict)\
+            -> Dict[str, Union[Dict, None]]:
         """Creates an index to lookup custom metadata by filename.
 
         Args:
@@ -56,12 +57,15 @@ class _UploadCustomMetadataMixin:
                 the required format.
 
         Returns:
-            A dictionary containing custom metadata indexed by filename.
+            A dictionary mapping from filenames to custom metadata.
+            If there are no annotations for a filename, the custom metadata
+            is None instead.
 
         """
 
         # Developer note:
-        # the mapping is filename -> image_id -> custom_metadata
+        # The mapping is filename -> image_id -> custom_metadata
+        # This mapping is created in linear time.
         filename_to_image_id = {
             image_info[COCO_ANNOTATION_KEYS.images_filename]:
                 image_info[COCO_ANNOTATION_KEYS.images_id]
--- a/lightly/api/api_workflow_upload_metadata.py
+++ b/lightly/api/api_workflow_upload_metadata.py
@@ -47,6 +47,42 @@ class _UploadCustomMetadataMixin:
             COCO_ANNOTATION_KEYS.custom_metadata, custom_metadata
         )
 
+    def index_custom_metadata_by_filename(self, custom_metadata: Dict):
+        """Creates an index to lookup custom metadata by filename.
+
+        Args:
+            custom_metadata:
+                Dictionary of custom metadata, see upload_custom_metadata for
+                the required format.
+
+        Returns:
+            A dictionary containing custom metadata indexed by filename.
+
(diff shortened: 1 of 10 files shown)


guarin:
This looks good!

The only problem I see is that the check is performed after training + embedding if you are using lightly-magic which can potentially take a long time. Maybe you can put it in a function and make lightly-cli call it in the beginning?

MalteEbner:
> The only problem I see is that the check is performed after training + embedding if you are using lightly-magic which can potentially take a long time. Maybe you can put it in a function and make lightly-cli call it in the beginning?

We rather implemented that lightly-magic is all 3 cli calls done separately and the argument check is inside each of them. Otherwise, we would also need to put the check of the token and dataset_id before, see: https://github.com/lightly-ai/lightly/blob/416883b7801ac7e9d68979568fcc82a57d657073/lightly/cli/upload_cli.py#L39-L69

I don't see that as a big problem, if the upload fails, you can still use the embedding file from the last step and thus don't need to do everything agin.

Thus I would keep it like that.

> 
> Also, why did you disable the append check embeddings and custom metadata?

Where did I do that? They are both still possible, either by not specifying the input_dir or by specifying append=True. 

However, the issue is quite unclear on wether and how the append prevention should be applied on embeddings and metadata.

guarin:
> Otherwise, we would also need to put the check of the token and dataset_id before, see...

> I don't see that as a big problem, if the upload fails, you can still use the embedding file from the last step and thus don't need to do everything agin.

Good points, let's leave it like this.

> Where did I do that?

The check is only performed if input_dir is specified.

> However, the issue is quite unclear on wether and how the append prevention should be applied on embeddings and metadata.

Maybe just add the check for all the cases so it is very clear to the user that he adds something to an existing dataset.

lightly/api/api_workflow_upload_metadata.py
```diff
@@ -168,48 +142,62 @@ def upload_custom_metadata(self,
...
+        }
+
+        upload_requests = []
+        for metadata in custom_metadata[COCO_ANNOTATION_KEYS.custom_metadata]:
+            image_id = metadata[COCO_ANNOTATION_KEYS.custom_metadata_image_id]
+            filename = image_id_to_filename.get(image_id, None)
+            if filename is None:
+                raise ValueError(f"Your custom_metadata file is malformatted. "
+                                 f"There is an annotation with an image_id of "
+                                 f"{{{image_id}}},"
```
guarin, line 170: Why are there three `{{{`?

lightly/cli/lightly_cli.py
```diff
@@ -9,16 +9,43 @@
...
+
+    Returns:
+        Wether the config is valid
+
+    """
+    valid = True
+    if cfg['trainer']['max_epochs'] > 0 and cfg['append']:
+        print_as_warning('When appending to an existing dataset you must '
+                         'use the same embedding model. Thus specify'
+                         'trainer.max_epochs=0. If you had trained your own model,'
```
guarin, line 36: Missing whitespaces at the end of both lines.

lightly/cli/upload_cli.py
```diff
@@ -10,20 +10,35 @@
...
+    """
+
+    Returns:
+        if no errors were encountered:
+            SUCCESS_RETURN_VALUE
+        else:
+            None
+
+    """
+    print(cfg)
```
guarin, line 41: I think you missed to remove this print

lightly/cli/upload_cli.py
```diff
@@ -10,20 +10,35 @@
...
 import torchvision
 from torch.utils.hipify.hipify_python import bcolors
 
+from lightly.api.api_workflow_upload_embeddings import \
+    EmbeddingDoesNotExistError
 from lightly.cli._helpers import fix_input_path, print_as_warning, cpu_count
 
 from lightly.api.api_workflow_client import ApiWorkflowClient
 from lightly.data import LightlyDataset
+SUCCESS_RETURN_VALUE = "Success"
```
guarin, line 28: This is only used for testing right?

guarin (commented):
This looks good :) There are only some whitespaces missing.

lightly/cli/upload_cli.py
MalteEbner: Yes, it is needed to ensure that the "working case" tests actually work and do not exit earlier with warning.

lightly/api/api_workflow_upload_metadata.py
```diff
@@ -47,71 +47,45 @@ def verify_custom_metadata_format(self, custom_metadata: Dict):
...
-
-            image = custom_metadata[COCO_ANNOTATION_KEYS.images][image_index]
-            image_id = image[COCO_ANNOTATION_KEYS.images_id]
-
-            metadata_index = bisect_left(metadata_image_ids, image_id)
-            if metadata_index == len(metadata_image_ids):
-                raise RuntimeError(
-                    f'Image with id {image_id} has no custom metadata!'
-                )
+        # Developer note:
```
philippmwirth, line 66: what's a developer note 😄

lightly/api/api_workflow_upload_metadata.py
philippmwirth: I think we can write shorter and much more precise error messages:
```python
InvalidCustomMetadataError(
    f'No image found for annotation with image_id {image_id}`
)
```

lightly/api/api_workflow_upload_metadata.py
```diff
@@ -168,48 +142,62 @@ def upload_custom_metadata(self,
...
+            image_id = metadata[COCO_ANNOTATION_KEYS.custom_metadata_image_id]
+            filename = image_id_to_filename.get(image_id, None)
+            if filename is None:
+                raise ValueError(f"Your custom_metadata file is malformatted. "
+                                 f"There is an annotation with an image_id of "
+                                 f"{{{image_id}}},"
+                                 f"but the list of images does not include this id.")
+            sample_id = filename_to_sample_id.get(filename, None)
+            if sample_id is None:
+                raise ValueError(f"Your custom_metadata file "
```
philippmwirth, line 174:
```python
InvalidCustomMetadataError(
    f' Sample with filename {filename} does not exist on the server'
)
```

philippmwirth (commented):
I left some suggestions especially w.r.t. how we write errors. I think we can be much more precise.

Regarding your questions:
> How to handle samples not having any metadata?
How to handle metadata without a corresponding sample?
Do you handle it differently for upload of metadata together with samples and upload of metadata separately from samples?

Samples without custom metadata should simply have an empty custom metadata entry. This should already be handled correctly by the UI.

You can either raise a warning or simply ignore metadata without corresponding sample. Most likely it's an error from the user side though so I think a warning would be appropriate.

No, I wouldn't make a difference between the two workflows.

lightly/api/api_workflow_upload_metadata.py
MalteEbner: I removed it.

lightly/api/api_workflow_upload_metadata.py
MalteEbner: And tell the user that this annotations is skipped.

MalteEbner: This is how the implementation looks like:


image

philippmwirth: > Your proposals are a bit too short in my opinion. I would also try to hint the user more at how to solve the problem.
Fair point, maybe we can have a tech talk about how to write good exceptions sometime, I bet there are guidelines 🙂

philippmwirth: Only thing I could find so far:
https://google.github.io/styleguide/pyguide.html#3102-error-messages

lightly/api/api_workflow_upload_metadata.py
```diff
@@ -168,48 +147,73 @@ def upload_custom_metadata(self,
...
+            sample_id = filename_to_sample_id.get(filename, None)
+            if sample_id is None:
+                print_as_warning(
+                    f'You tried to upload custom metadata for a sample with '
+                    f'filename {{{filename}}}, '
+                    f'but a sample with this filename '
+                    f'does not exist on the server. '
+                    f'This custom metadata annotation is skipped. '
+                    f'Please upload the sample to the server first or fix '
+                    f'your custom metadata file.',
```
guarin, line 190: Not necessary.

lightly/api/api_workflow_upload_metadata.py
```diff
@@ -168,48 +147,73 @@ def upload_custom_metadata(self,
...
+        upload_requests = []
+        for metadata in custom_metadata[COCO_ANNOTATION_KEYS.custom_metadata]:
+            image_id = metadata[COCO_ANNOTATION_KEYS.custom_metadata_image_id]
+            filename = image_id_to_filename.get(image_id, None)
+            if filename is None:
+                print_as_warning(
+                    f'No image found for custom metadata annotation '
+                    f'with image_id {image_id}. '
+                    f'This custom metadata annotation is skipped. '
+                    f'Please fix your custom metadata file.',
```
guarin, line 177: Not necessary

guarin (approved):
Please remove the additional messages before merging. Otherwise looks good.

PR #712, opened 2022-03-02
Add compute worker api workflows

guarin:
* Add functions to update s3, gcs, azure, local datasource configs
* Add register/delete compute worker
* Add create/schedule new runs
* Add delete tag by id
* Make `client.dataset_id` settable (I added this because right now you can only get a dataset by its name but not its id)
* Add tests
* Update some generated api functions
* Did NOT add deleting of embedding files because it is not possible right now


lightly/api/api_workflow_client.py
```diff
@@ -122,6 +126,24 @@ def dataset_id(self) -> str:
...
+
+        Raises:
+            ValueError if the dataset id does not exist.
+        """
+        all_datasets: List[DatasetData] = self._datasets_api.get_datasets()
+        for dataset in all_datasets:
+            if dataset.id == dataset_id:
+                self._dataset_id = dataset_id
+                return
+        raise ValueError(f"A dataset with the id {dataset_id} does not exist on the web platform.")
```
philippmwirth, line 145: we should probably go for a consistent name for the "web platform". @MalteEbner used server.

lightly/api/api_workflow_compute_worker.py
```diff
@@ -0,0 +1,100 @@
...
+
+        """
+        request = CreateDockerWorkerRegistryEntryRequest(
+            name=name, 
+            worker_type=DockerWorkerType.FULL
+        )
+        response = self._compute_worker_api.register_docker_worker(request)
+        return response.id
+
+    def delete_compute_worker(self, worker_id: str):
```
philippmwirth, line 31: do we have the functions to implement a getter? could be very helpful when workers are already registered.

guarin: Added it :)

lightly/api/api_workflow_client.py
guarin: I changed it to "Lightly Platform" because we use this everywhere in the docs.

philippmwirth (approved):
well done

PR #714, opened 2022-03-03
Update api code

guarin:
* Adds the "AccessRole" model which is now required when accessing the datasets api


japrescott (approved):
very nice :)

PR #720, opened 2022-03-04
Adapt Azure docs to using SAS Token

philippmwirth:
# Adapt Azure docs to using SAS Token

- Updates the text of the guide to configure Azure datasource
- The image was updated in the cloud bucket


MalteEbner (approved):
Great that you adapted the tutorial!

PR #722, opened 2022-03-07
Make shared datasets accessible

philippmwirth:
# Make shared datasets accessible

Allows users to access datasets which were shared with them from the PIP package.


lightly/api/api_workflow_datasets.py
```diff
@@ -19,11 +19,32 @@ def _get_current_dataset(self) -> DatasetData:
...
+
+        """
+        datasets = self._datasets_api.get_datasets(shared=shared)
+        return datasets
+
+    def get_all_datasets(self) -> List[DatasetData]:
+        """Returns all datasets the user has access to. """
+        owned_datasets = self.get_datasets(shared=False)
+        shared_datasets = self.get_datasets(shared=True)
+        owned_datasets.extend(shared_datasets)
```
japrescott, line 45: hahaha nice :D

lightly/api/api_workflow_client.py
```diff
@@ -118,7 +118,7 @@ def dataset_id(self) -> str:
         try:
             return self._dataset_id
         except AttributeError:
-            all_datasets: List[DatasetData] = self._datasets_api.get_datasets()
+            all_datasets: List[DatasetData] = self.get_datasets()
```
japrescott, line 121: I guess this should be `get_all_datasets` as well?

lightly/api/api_workflow_datasets.py
```diff
@@ -19,11 +19,32 @@ def _get_current_dataset(self) -> DatasetData:
...
+            shared:
+                If True, only returns the datasets which have been shared with
+                the user.
+
+        """
+        datasets = self._datasets_api.get_datasets(shared=shared)
+        return datasets
+
+    def get_all_datasets(self) -> List[DatasetData]:
+        """Returns all datasets the user has access to. """
```
japrescott, line 42: as this function is called quite often (as it seems) to only check if the supplied id the user is passing is actually a valid datasetId, can we cache this information somehow?

lightly/api/api_workflow_client.py
```diff
@@ -137,7 +137,7 @@ def dataset_id(self, dataset_id: str):
         Raises:
             ValueError if the dataset id does not exist.
         """
-        all_datasets: List[DatasetData] = self._datasets_api.get_datasets()
+        all_datasets: List[DatasetData] = self.get_all_datasets()
```
japrescott, line 140: Is it actually really needed to get ALL and then see if one has it? Wouldn't it be better to just try and getDatasetById and if its there, its there otherwise fail?

lightly/api/api_workflow_client.py
```diff
@@ -118,7 +118,7 @@ def dataset_id(self) -> str:
         try:
             return self._dataset_id
         except AttributeError:
-            all_datasets: List[DatasetData] = self._datasets_api.get_datasets()
+            all_datasets: List[DatasetData] = self.get_datasets()
             datasets_sorted = sorted(all_datasets, key=lambda dataset: dataset.last_modified_at)
             last_modified_dataset = datasets_sorted[-1]
```
japrescott, line 123: this is kinda weird and cant imagine a situation where we want to auto-default to the the dataset that was last_modified_at?

japrescott (commented):
looks good! just some critique about historical code

lightly/api/api_workflow_client.py
philippmwirth: It depends, this is a bit of a weird function anyways... but I chose to only return the owned datasets so as to not accidentally set the dataset id to a shared dataset. I'd leave it like this if that's ok with you.

lightly/api/api_workflow_client.py
philippmwirth: I will not remove this from the code although I understand your confusion.

lightly/api/api_workflow_datasets.py
philippmwirth: Let's not add caching just yet as this may lead to many bugs. I will instead change the check to simply request the dataset by id.

tests/api_workflow/mocked_api_workflow_client.py
```diff
@@ -307,6 +307,12 @@ def reset(self):
     def get_datasets(self, **kwargs):
         return self.datasets
 
+    def get_all_datasets(self, **kwargs):
+        return self.get_datasets()
+
+    def dataset_exists(self, dataset_id: str):
+        return dataset_id in [d.id for d in self.default_datasets]
```
japrescott, line 314: I would use `self.get_all_datasets()`

japrescott (approved):

PR #726, opened 2022-03-08
Igor lig 724 write python script to test milestone 01

IgorSusmelj:
## Changes

- Remove `.` from datasource tutorial title
- Modify trigger docker from api tutorial
  - Add sample code to trigger a job through Python 
  - Rename to **trigger from UI or code** to make it more clear what this is about

PR #727, opened 2022-03-08
Igor lig 801 check embeddings fails with masked

IgorSusmelj:
## Changes

- Make sure we remove all special columns such as `masked`, `selected` (also if they appear earlier)
- Add new test case


philippmwirth:
Looks like you didn't checkout from `master` but from your local branch instead (this PR contains all the changes of #726 as far as I understand).

MalteEbner (commented):
Could we remove the `remove_additional_columns` argument and handle as if it is always true?  Or is there a case where we don't want to remove additional columns?

IgorSusmelj:
> Could we remove the `remove_additional_columns` argument and handle as if it is always true? Or is there a case where we don't want to remove additional columns?

We can look into this in a follow up issue

lightly/utils/io.py
```diff
@@ -98,12 +101,25 @@ def check_embeddings(path: str, remove_additional_columns: bool=False):
...
             for row in reader:
-                row = row[:header_labels_idx+1]
-                new_rows.append(row)
+                # apply mask to only use filenames, embedding_ or labels
+                new_rows.append(list(compress(row, col_mask)))
+
         with open(path, 'w', newline='') as csv_file:
             writer = csv.writer(csv_file, delimiter=',')
             writer.writerows(new_rows)
+            print(new_rows)
```
MalteEbner, line 122: Do we need this print?

tests/utils/test_io.py
```diff
@@ -100,4 +100,26 @@ def test_embeddings_extra_rows(self):
             for row_read, row_original in zip(csv_reader, rows):
                 self.assertListEqual(row_read, row_original[:-2])
 
+    def test_embeddings_extra_rows_special_order(self):
```
MalteEbner, line 103: Great unittest!

MalteEbner (approved):
This looks great! I would only remove the print.

lightly/utils/io.py
IgorSusmelj: Nooooo, good catch!

PR #729, opened 2022-03-09
Only run tests if python files changed

guarin:
* Update pre-commit hook to only run python tests if python files changed


MalteEbner:
Could you add documentation on the stuff you changed, especially on the regex? And can you link the issue this solves?

japrescott (approved):

PR #733, opened 2022-03-10
LIG-673: ApiWorkflowClient can receive LIGHTLY_TOKEN env variable

shaundaley39:
- ApiWorkflowClient can be initialized with or without a token argument (alternative: LIGHTLY_TOKEN can be provided)

### DIFF
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -81,6 +81,10 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin,
         configuration.host = getenv('LIGHTLY_SERVER_LOCATION', 'https://api.lightly.ai')
         if token is None:
             token = getenv('LIGHTLY_TOKEN', None)
+            if token is None:
+                raise ValueError(f"Either provide a 'token' argument or export"
+                                 f" a LIGHTLY_TOKEN environment variable")
+                
         configuration.api_key = {'token': token}
         api_client = ApiClient(configuration=configuration)
         self.api_client = api_client
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -73,12 +73,14 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin,
             but used by a workflow, the newest embedding is taken by default
     """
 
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_client.py
```diff
@@ -73,12 +73,14 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin,
...
 
-    def __init__(self, token: str, dataset_id: str = None, embedding_id: str = None):
+    def __init__(self, token: str = None, dataset_id: str = None, embedding_id: str = None):
 
         self.check_version_compatibility()
 
         configuration = Configuration()
         configuration.host = getenv('LIGHTLY_SERVER_LOCATION', 'https://api.lightly.ai')
+        if token is None:
+            token = getenv('LIGHTLY_TOKEN', None)
```
IgorSusmelj, line 83: What happens if there is no env variable? Maybe throw a `ValueError` then?

IgorSusmelj (changes requested):
Looks good. But let's make sure there exists an env variable and that we cover the case if there is none.

MalteEbner (commented):
And please add a unittest for creating an ApiWorkflowClient() without a token, but with an env variable.

lightly/api/api_workflow_client.py
shaundaley39: Sure, thrown

shaundaley39:
@MalteEbner unit tests added

MalteEbner (approved):
This is great! And you see how the code coverage of your newly added lines is 100%.

lightly/api/api_workflow_client.py
```diff
@@ -73,12 +73,18 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin,
...
+            if token is None:
+                raise ValueError(f"Either provide a 'token' argument or export"
```
IgorSusmelj, line 85: FYI. There is no need to use an f-string in python if you don't include variables. 

https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python

IgorSusmelj (approved):

PR #739, opened 2022-03-14
Update Docs Makefile to download required datasets

MalteEbner:
## Description
- allows to build the docs without needing to manually download all datasets to the correct location
- Developed mostly by @shaundaley39 

image

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -13,6 +13,13 @@ PLATFORMSOURCE 			= source/tutorials_source/platform
 DOCKERSOURCE   			= source/docker
 GETTING_STARTED_IMAGES 	= source/getting_started/resources
 
+
+DATADIR_FREIBURG = /datasets/freiburg_groceries_dataset
+DATADIR_CLOTHING = /datasets/clothing-dataset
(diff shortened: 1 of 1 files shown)


japrescott:
I guess it would be good if we would remove the archives after downloading them? And maybe check if the folder is there before just downloading everything?

shaundaley39 (changes requested):
Agreed with @japrescott 
- it would be better to check whether a dataset directory exists before downloading it (same as with the Clothing dataset, etc)
- it would be better to remove archives after their content is extracted, just to keep things cleaner

MalteEbner:
> * it would be better to remove archives after their content is extracted, just to keep things cleaner

This is done with `make clean` deleting `_data`. I would keep this for now.

shaundaley39 (approved):

shaundaley39:
Great, I like that this also drops datasets no longer needed for tutorials (and checks for preexistence! And of course, ensures existence of datasets needed for building the docs!).

IgorSusmelj (approved):
LGTM

PR #748, opened 2022-03-23
Show number of images in progress bar

philippmwirth:
# Show number of images in progress bar

Show the number of images instead of the number of batches when embedding.

### DIFF
--- a/lightly/embedding/embedding.py
+++ b/lightly/embedding/embedding.py
@@ -107,12 +107,13 @@ class SelfSupervisedEmbedding(BaseEmbedding):
         embeddings, labels, fnames = None, None, []
 
         if lightly._is_prefetch_generator_available():
-            pbar = tqdm(
-                BackgroundGenerator(dataloader, max_prefetch=3),
-                total=len(dataloader)
-            )
-        else:
-            pbar = tqdm(dataloader, total=len(dataloader))
+            dataloader = BackgroundGenerator(dataloader, max_prefetch=3),
+        
+        pbar = tqdm(
+            dataloader,
+            total=len(dataloader.dataset),
+            unit='imgs'
+        )
 
         efficiency = 0.0
         embeddings = []
@@ -120,7 +121,9 @@ class SelfSupervisedEmbedding(BaseEmbedding):
         with torch.no_grad():
 
             start_timepoint = time.time()
-            for (img, label, fname) in pbar:
+            for (img, label, fname) in dataloader:
+
+                bsz = img.shape[0]
 
                 # this following 2 lines are needed to prevent a file handler leak,
                 # see https://github.com/lightly-ai/lightly/pull/676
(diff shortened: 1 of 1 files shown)


lightly/embedding/embedding.py
```diff
@@ -107,20 +107,23 @@ def embed(self,
...
         efficiency = 0.0
         embeddings = []
         labels = []
         with torch.no_grad():
 
             start_timepoint = time.time()
-            for (img, label, fname) in pbar:
+            for (img, label, fname) in dataloader:
+
+                bsz = img.shape[0]
```
japrescott, line 126: why are you pulling out the with of an image? 🤔

guarin: `img` is a batch of images so `img.shape[0]` gives the batch size. But this line is actually not necessary as `batch_size = img.shape[0]` is already defined a bit further down (line 132)

lightly/embedding/embedding.py
```diff
@@ -107,20 +107,23 @@ def embed(self,
...
+        )
 
         efficiency = 0.0
         embeddings = []
         labels = []
         with torch.no_grad():
 
             start_timepoint = time.time()
-            for (img, label, fname) in pbar:
+            for (img, label, fname) in dataloader:
```
MalteEbner, line 124: Rename this to images, labels, fnames.

lightly/embedding/embedding.py
MalteEbner: Because img is actually a list of images (one batch), but badly named.

lightly/embedding/embedding.py
```diff
@@ -107,20 +107,23 @@ def embed(self,
...
         if lightly._is_prefetch_generator_available():
-            pbar = tqdm(
-                BackgroundGenerator(dataloader, max_prefetch=3),
-                total=len(dataloader)
-            )
-        else:
-            pbar = tqdm(dataloader, total=len(dataloader))
+            dataloader = BackgroundGenerator(dataloader, max_prefetch=3),
+        
+        pbar = tqdm(
```
MalteEbner, line 112: You don't need to provide the dataloader as argument, see the example here:
https://github.com/tqdm/tqdm#manual

lightly/embedding/embedding.py
MalteEbner: Furthermore, this line is duplicate to `batch_size = img.shape[0]` a few lines below -> delete it.

MalteEbner (commented):
The logic is fine, but the naming can be improved.

lightly/embedding/embedding.py
```diff
@@ -107,20 +107,23 @@ def embed(self,
...
+            dataloader,
```
MalteEbner, line 113: You don't need to provide the dataloader as argument, see the example here:
https://github.com/tqdm/tqdm#manual

MalteEbner (approved):
LGTM

PR #750, opened 2022-03-28
Philipp add api methods to download json files

philippmwirth:
# Philipp add api methods to download json files

Allow exporting your dataset to `Labelstudio` or `LabelBox` using the `ApiWorkflowClient`.


MalteEbner (commented):
Looks mostly good to me.

However, I would additionally check for the return value in the unittests. We already had bugs where it got missing, but the unittest did not check for it. E.g.
```
tasks = self.api_workflow_client.export_label_studio_tasks_by_tag_name('initial-tag')
self.assertAll(is instance(task, dict) for task in tasks))
```

MalteEbner (approved):

PR #752, opened 2022-03-29
Malte lig 887 add helpers to store predictions

MalteEbner:
## Description
- puts the documentation on storing predictions in the .ligthly format in a datasource into its own tutorial
- adds an example script to generate the .lightly folder from a coco annotations file.  

## New tutorial
[Add Predictions to a datasource — lightly 1.2.12 documentation.pdf](https://github.com/lightly-ai/lightly/files/8399238/Add.Predictions.to.a.datasource.lightly.1.2.12.documentation.pdf)


docs/source/docker/advanced/datasource_predictions.rst
```diff
@@ -0,0 +1,332 @@
+.. _ref-docker-datasource-predictions:
+
+Add Predictions to a datasource
+===============================
+
+Lightly can not only use images you provided in a datasource, but also predictions of a ML model on your images.
+They can be used e.g. for using Lightly for Active Learning or for using Lightly on object-level.
```
philippmwirth, line 7: used e.g. for using -> replace by "can be used for active learning for selecting images based on the objects in them."

docs/source/docker/advanced/datasource_predictions.rst
```diff
@@ -0,0 +1,332 @@
...
+saves them in the format just outlined. You can either save the predictions first on your local machine
+and then upload them for your datasource or upload them directly to your datasource.
+
+As an example, the following script takes an object detection annotations file in the coco format.
+It needs the path to the annotations file and the output directory
+where the .lightly folder should be created as input.
+
+.. code-block:: python
+
+    import json
```
philippmwirth, line 280: the script is great but I would move everything that's configurable to the top and make it clear that it needs to be changed (we could even use argparse, your call 🙂 )

philippmwirth (changes requested):
Great changes! Some minor improvement ideas

docs/source/docker/advanced/datasource_predictions.rst
MalteEbner: I put them at the top.

However, I would not add any argparse or other further helpers, as our users might use it very differently. E.g. they could copy the code to their prediction script, or change the parameters via env variables.

philippmwirth (changes requested):

philippmwirth (approved):

PR #754, opened 2022-04-01
Malte lig 868 add download video frames subset

MalteEbner:
## Description
- Created the function `download_video_frames_at_timestamps` return a generator over frames as specified timestamps.
- Wrote a unittest for it.
- Wrote a skipped unittest class for benchmarking the different download options.

## Timings 
- with 100Mbit Internet connection
- video has 18k frames, 116 MB, 12min length, 25fps

#### with 5 frames equally distributed
- test_download_full: 18s, 1000 iterations per second
- test_download_at_timestamps_for_loop: 3.6s, 1.3 iterations per second
- test_download_at_timestamps: 10.2s, 0.5 iterations per second

#### with 50 frames equally distributed
- test_download_full: 19.4s (should be equal to the other, so this is the measurement variance)
- test_download_at_timestamps_for_loop: 38s, again 1.3 iterations per second
- test_download_at_timestamps: 10.5s, 4.8 iterations per second

### DIFF
--- a/lightly/api/download.py
+++ b/lightly/api/download.py
@@ -105,66 +105,16 @@ def download_all_video_frames(
                 yield frame
 
 
-def download_video_frame(
-    url: str, 
-    timestamp: int, 
-    as_pil_image: int = True,
-    thread_type: av.codec.context.ThreadType = av.codec.context.ThreadType.AUTO,
-    video_channel: int = 0,
-) -> Union[PIL.Image.Image, av.VideoFrame, None]:
-    """Retrieves a specific frame from a video stored at the given url.
-
-    Finds the first frame in the video that has a timestamp equal or larger than
-    the timestamp argument.
-
-    Args:
-        url: 
-            The url where the video is downloaded from.
-        timestamp:
-            Timestamp in pts from the start of the video at which the frame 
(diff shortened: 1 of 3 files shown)


lightly/api/download.py
```diff
@@ -366,3 +316,82 @@ def download_prediction_file(
...
+                for details.
+            video_channel:
+                The video channel from which frames are loaded.
+
+        Returns:
+            A generator that loads and returns a single frame per step.
+
+        """
+        _check_av_available()
+
```
guarin, line 352: I think you have to add a test here if `timestamps` is empty and return an empty iterator in this case. Otherwise we will get exceptions later on in the function.

guarin (approved):
Code looks great! 

I would maybe change the `*args` and `**kwargs` arguments but I'll let you decide and approve already :)

guarin (changes requested):
Oh wait, I actually found something that you should change first. See the comment about zero length `timestamps` lists.

lightly/api/download.py
MalteEbner: I added that test/fix and the according unittest.

guarin (approved):

PR #764, opened 2022-04-11
Rename task_description to task_type

guarin:
Updates the docs for the `task_description` -> `task_type` renaming.


japrescott (approved):
we should not forget to update the new docs

guarin:
> we should not forget to update the new docs

I think we'll move/copy a lot of the docs over so this should not be a problem. But we should definitely make sure that everything is updated.

PR #766, opened 2022-04-14
Philipp add second datasource option docs

philippmwirth:
# Philipp add second datasource option docs

Update documentation to include information about how to configure an "output" datasource.

### DIFF
--- a/docs/source/getting_started/dataset_creation/dataset_creation_aws_bucket.rst
+++ b/docs/source/getting_started/dataset_creation/dataset_creation_aws_bucket.rst
@@ -121,31 +121,25 @@ Create and configure a dataset
    depending on the type of files in your S3 bucket.
 2. Edit your dataset and select S3 as your datasource
 
-    .. figure:: ../resources/LightlyEdit1.png
-        :align: center
-        :alt: Edit your dataset.
-
-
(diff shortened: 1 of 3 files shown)


MalteEbner (commented):
A general question: Do the lights for "list, read, write" apply to the input or output datasource? Shouldn't there be
If only one datasource:
- list, read, write for it
If two datasources:
- list, read for input
- list, read, write for output

MalteEbner:
The last steps (7-9) are the following for the 3 datasources:
> Press save and ensure that at least the lights for List and Read turn green. If you added permissions for writing, this light should also turn green.

> Press save and ensure that all lights turn green.

> After closing the pop-up by clicking the X, you should be on the dataset creation page again.

It would be nice to have this the same for all three datasources

MalteEbner (changes requested):
The text itself is nice, well done!

I am missing an addition to the `make download-noplot` to also get the images under `resources/resources_datasource_configure` downloaded.

philippmwirth:
> 

I have added it to the existing folders so if you do `make clean` and then `make html-noplot` you will get the new images.

MalteEbner:
> I have added it to the existing folders so if you do `make clean` and then `make html-noplot` you will get the new images.

Thanks, I just tried it out and it worked.

PR #768, opened 2022-04-19
Malte lig 455 use openapi generator instead of swagger pip side

MalteEbner:
generation branch: https://github.com/lightly-ai/lightly-core/compare/malte-lig-455-use-openapi-generator-instead-of-swagger-api-part

## Changes
- generates the new specs from openAPI
- adapts the corresponding imports, e.g. to`from lightly.openapi_generated.swagger_client.model.docker_run_data import DockerRunData`

## Partly changed
- Partly: changes the attributes from snake_case to camelCase. This is necessary, see https://github.com/OpenAPITools/openapi-generator/issues/12121 It needs quite some time to do that by hand.
- Partly: adapts the unittests to create valid attributes (e.g. a valid MongoObjectId with 24 chars in [a-f0-9]). This is necessary, otherwise the validator complains.

The partly changes are enough to make the AL unittests run through. However, the other unittests would need the same adaptations in other parts of the code.

# Evaluation

## OK
- imports needed to be changed, now the import is directly from the file 
- MIXED: the basic input types are often not anymore a string, but e.g. a VersionNumber or DatasetName.
   - GOOD: This allows partly for validation on the client side, e.g. a DatasetName has a minimum length of 3, which is checked for while creating.
   - BAD: Lots of code changes are needed, both before sending and after receiving.

## GOOD
- support openapi specs v3 with OneOf (not tried out yet)
- does client-side validation of some types (e.g. you cannot pass a string when a number is required).

## BAD
- the allowed values of an enum are not easily accessible anymore, see https://github.com/OpenAPITools/openapi-generator/issues/10752

## Problems (still existing as of 16.08.2022 with openapi-generator-cli 6.1 and python-experimental)
- the functions do not provide any arguments at all, only body This makes it very hard to find out what to pass and how.
- The version check is not working at all because of `NotImplementedError: Deserialization of text/html; charset=utf-8 has not yet been implemented`
- The connection to the API is not working at all because of` NotImplementedError: Deserialization of application/json; charset=utf-8 has not yet been implemented`


japrescott:
most of the problems could be fixed in the handlebars
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python-experimental/api_client.handlebars

MalteEbner:
We are now using the openapi generator with the python-experimental branch becoming non-experimental.

PR #771, opened 2022-04-21
lig-1007: Embedding stops before all frames are embedded

guarin:
Video timestamps are not necessarily in order. This lead to dropped frames when using seek as the seek operation filters out any frames with timestamp < seek timestamp. This pr solves the problem by explicitly seeking if the timestamp decreases between two subsequent frames.

With the fix the correct number of frames are found using different batch sizes and worker counts.

### DIFF
--- a/lightly/data/_video.py
+++ b/lightly/data/_video.py
@@ -143,7 +143,7 @@ class VideoLoader(threading.local):
                 #    has timestamp >= seek timestamp and filters out all frames
                 #    with a smaller timestamp. Future calls to next(self.reader)
                 #    can then miss the frames with the smaller timestamps,
-                #    resulting in accidentally dropped frames. See LIG-1007.
+                #    resulting in accidentally dropped frames.
                 if (
                     self.current_timestamp_idx != self.last_timestamp_idx + 1
                     or timestamp < self.timestamps[self.last_timestamp_idx]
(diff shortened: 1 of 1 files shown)


lightly/data/_video.py
```diff
@@ -119,16 +119,35 @@ def read_frame(self, timestamp = None):
...
-                # Calling seek is slow. If we read next frame we can skip it!
-                if self.timestamps.index(timestamp) != self.last_timestamp_idx + 1:
+                # Calling seek is slow. We avoid seeking unless:
+                # 1) We do not access frames sequentially.
+                # 2) The timestamps between subsequent frames decrease. This is
+                #    necessary because the seek looks for the first frame that
+                #    has timestamp >= seek timestamp and filters out all frames
+                #    with a smaller timestamp. Future calls to next(self.reader)
+                #    can then miss the frames with the smaller timestamps,
+                #    resulting in accidentally dropped frames. See LIG-1007.
```
philippmwirth, line 146: remove the reference to lig-1007

philippmwirth: If you have a reference to where you got the information about the behavior of seek then just put the link instead.

philippmwirth (changes requested):

lightly/data/_video.py
guarin: I removed the reference :) I initially wanted to keep a reference to the issue which has some additional information, like the video which created the error but we can always get it using a git blame.

philippmwirth (approved):

PR #774, opened 2022-04-25
Check for non-increasing timestamps

guarin:
* Mark all non-increasing timestamps on dataset loading
* Raise by default an exception if frame with non-increasing timestamp should be loaded
* Add non-increasing timestamp tests

### DIFF
--- a/lightly/data/_video.py
+++ b/lightly/data/_video.py
@@ -139,7 +139,8 @@ class VideoLoader(threading.local):
             ValueError: 
                 If provided timestamp is not in self.timestamps.
             RuntimeError:
-                If loaded frame has wrong shape.
+                If loaded frame has wrong shape or seeking to the provided
+                timestamp failed.
 
         """
         if not self.timestamps:
@@ -213,7 +214,14 @@ class VideoLoader(threading.local):
                 # Accidentally reached the end of the video, let's seek back to
                 # the correction position. This can happen due to imprecise seek.
                 self.reader.seek(timestamp)
-                frame_info = next(self.reader)
+                try:
+                    frame_info = next(self.reader)
+                except StopIteration as ex:
+                    # Seeking to this timestamp simply doesn't work.
+                    raise RuntimeError(
+                        f'Cannot seek to frame with timestamp {float(timestamp)} '
+                        f'in {self.path}.'
+                    ) from ex
 
             if (
                 frame_info['pts'] < timestamp - self.eps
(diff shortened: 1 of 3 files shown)


lightly/data/_video.py
```diff
@@ -112,61 +127,125 @@ def read_frame(self, timestamp = None):
...
-            if (
-                self.current_timestamp_idx + 1 < len(self.timestamps)
-                and self.timestamps[self.current_timestamp_idx + 1] == timestamp
-            ):
-                # Timestamp is timestamp of next frame. Set current timestamp
-                # index to next frame.
-                self.current_timestamp_idx += 1
+        if not self.timestamps:
+            # Empty video.
+            raise StopIteration()
```
philippmwirth, line 147: we shouldn't call `StopIteration` in `read_frame` this will stop the dataloader!

lightly/data/_video.py
```diff
@@ -112,61 +127,125 @@ def read_frame(self, timestamp = None):
...
+
+        if timestamp is None:
+            # Try to read next frame.
+            if self.current_index is None:
+                # Beginning of video.
+                index = 0
+                timestamp = self.timestamps[index]
+            elif self.current_index >= len(self.timestamps):
+                # Reached end of video.
+                raise StopIteration()
```
philippmwirth, line 157: same concerns about raising `StopIteration` as above

lightly/data/_video.py
guarin: What should we do instead if the video is empty? A call to `next(self.reader)` will also raise StopIteration.

lightly/data/_video.py
guarin: This only happens if you run `read_frame(None)` and we are at the end of the video. I think in this case we want to raise StopIteration or what would you suggest to do otherwise?

lightly/data/_video.py
```diff
@@ -112,61 +136,133 @@ def read_frame(self, timestamp = None):
...
             A PIL Image
 
+        Raises:
+            StopIteration:
+                If video is empty or end of video is reached and timestamp is None.
+            ValueError: 
+                If provided timestamp is not in self.timestamps.
+            RuntimeError:
+                If loaded frame has wrong shape or seeking to the provided
+                timestamp failed.
```
philippmwirth, line 152: Can you add the `UnseekableTimestampError` here?

lightly/data/_video.py
philippmwirth: Mid- to long-term we'll probably have to add an `EmptyVideoError` or similar which we can catch.

philippmwirth (changes requested):
Last round

philippmwirth (approved):

PR #792, opened 2022-05-10
Malte lig 1082 worker crashes on open images dataset

MalteEbner:
## Description
1. Finding the root cause for any such problems is hard, as the retry catches the real exception -> solution: Let the `retry` raise the `RuntimeError` from the real exception.

2. New `download_image` function uses
```
with req.get(url=url, stream=True) as response:
    image = PIL.Image.open(response.raw)
    image.load()
```
The actual loading of the data over the internet can happen when using `image.load()`. Thus all of this is wrapped in a retry.

Two unittests were added, which both mock the stream to only be 1024 bytes long, causing an `OSError` caught by the retry.

3. Tested the new `download_image()` function also with real urls from large images (5MB jpeg, 33MB tif). Download speeds were about 70MBit/s. 

## Open problem:
When turning off WiFi during the download of such big images, the download freezes (nothing happened for 30s. After turning WiFi on again, nothing happened (for 30 s). I tried without retry, so it was not in the return loop. Then, when hitting CTRL-C, the seek not supported error, just like in the issue, occurred:

```
  File "/Users/malteebnerlightly/Documents/GitHub/lightly/venv/lib/python3.9/site-packages/PIL/Image.py", line 2972, in open
    fp.seek(0)
io.UnsupportedOperation: seek

During handling of the above exception, another exception occurred:

  File "/Users/malteebnerlightly/Documents/GitHub/lightly/tests/UNMOCKED_end2end_tests/test_download_large_files.py", line 13, in <module>
    img = download_image(url_5MB)
  File "/Users/malteebnerlightly/Documents/GitHub/lightly/venv/lib/python3.9/site-packages/lightly/api/download.py", line 43, in download_image
    return PIL.Image.open(response.raw)
  File "/Users/malteebnerlightly/Documents/GitHub/lightly/venv/lib/python3.9/site-packages/PIL/Image.py", line 2974, in open
    fp = io.BytesIO(fp.read())
  File "/Users/malteebnerlightly/Documents/GitHub/lightly/venv/lib/python3.9/site-packages/urllib3/response.py", line 515, in read
    data = self._fp.read() if not fp_closed else b""
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/http/client.py", line 475, in read
    s = self._safe_read(self.length)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/http/client.py", line 625, in _safe_read
    chunk = self.fp.read(min(amt, MAXAMOUNT))
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/socket.py", line 704, in readinto
    return self._sock.recv_into(b)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ssl.py", line 1241, in recv_into
    return self.read(nbytes, buffer)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ssl.py", line 1099, in read
    return self._sslobj.read(len, buffer)
KeyboardInterrupt
```

Follow-Up issue to solve this: https://linear.app/lightly/issue/LIG-1113/add-timeout-to-retry-to-preven-frozen-docker

 ## Other
closes #LIG-1115 (Pip import of download failing if pyav is not installed)

### DIFF
--- a/lightly/api/download.py
+++ b/lightly/api/download.py
@@ -26,7 +26,7 @@ def _check_av_available() -> None:
         raise av
 
 def download_image(url: str, session: requests.Session = None) -> PIL.Image.Image:
-    """Downloads an image from an url.
+    """Downloads an image from a url.
 
     Args:
         url: 
@@ -38,16 +38,14 @@ def download_image(url: str, session: requests.Session = None) -> PIL.Image.Imag
         The downloaded image.
 
     """
-    def load_image(url, session):
-        req = requests if session is None else session
+    def load_image(url, req):
         with req.get(url=url, stream=True) as response:
             image = PIL.Image.open(response.raw)
             image.load()
         return image
 
-    image = load_image(url, session)
-
-    #image = utils.retry(load_image, url, session)
+    req = requests if session is None else session
+    image = utils.retry(load_image, url, req)
     return image
 
 
--- a/lightly/api/download.py
+++ b/lightly/api/download.py
@@ -26,7 +26,7 @@ def _check_av_available() -> None:
         raise av
 
 def download_image(url: str, session: requests.Session = None) -> PIL.Image.Image:
-    """Downloads an image from a url.
+    """Downloads an image from an url.
 
     Args:
         url: 
@@ -38,9 +38,17 @@ def download_image(url: str, session: requests.Session = None) -> PIL.Image.Imag
         The downloaded image.
 
     """
-    req = requests if session is None else session
-    response = utils.retry(req.get, url=url, stream=True)
-    return PIL.Image.open(response.raw)
+    def load_image(url, session):
+        req = requests if session is None else session
+        with req.get(url=url, stream=True) as response:
+            image = PIL.Image.open(response.raw)
+            image.load()
+        return image
+
+    image = load_image(url, session)
+
+    #image = utils.retry(load_image, url, session)
+    return image
 
 
 def download_all_video_frames(
--- a/lightly/api/download.py
+++ b/lightly/api/download.py
@@ -48,80 +48,301 @@ def download_image(url: str, session: requests.Session = None) -> PIL.Image.Imag
     image = utils.retry(load_image, url, req)
     return image
 
+if not isinstance(av, ModuleNotFoundError):
 
-def download_all_video_frames(
-    url: str, 
-    timestamp: Union[int, None] = None,
-    as_pil_image: int = True, 
-    thread_type: av.codec.context.ThreadType = av.codec.context.ThreadType.AUTO,
-    video_channel: int = 0,
-) -> Iterable[Union[PIL.Image.Image, av.VideoFrame]]:
-    """Lazily retrieves all frames from a video stored at the given url.
+    def download_all_video_frames(
+        url: str,
+        timestamp: Union[int, None] = None,
+        as_pil_image: int = True,
+        thread_type: av.codec.context.ThreadType = av.codec.context.ThreadType.AUTO,
+        video_channel: int = 0,
+    ) -> Iterable[Union[PIL.Image.Image, av.VideoFrame]]:
+        """Lazily retrieves all frames from a video stored at the given url.
 
-    Args:
-        url: 
-            The url where video is downloaded from.
-        timestamp:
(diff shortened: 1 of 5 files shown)


lightly/api/utils.py
```diff
@@ -45,17 +45,20 @@ def retry(func, *args, **kwargs):
...
             time.sleep(backoff)
             backoff = 2 * backoff if backoff < max_backoff else backoff
-        
-    # max retries exceeded
-    raise RuntimeError('The connection to the server timed out.')
+            current_retries += 1
+
+            # max retries exceeded
+            if current_retries >= max_retries:
+                raise RuntimeError from e
```
philippmwirth, line 61: I would keep some message in the `RuntimeError` e.g.
```python
raise RuntimeError('Maximum retries exceeded!') from e
```

tests/UNMOCKED_end2end_tests/test_download_large_files.py
```diff
@@ -0,0 +1,16 @@
...
+lightly.api.utils.RETRY_MAX_RETRIES = 1
+
+from lightly.api.download import download_image
+
+url_33MB = "https://cdn.eso.org/images/original/potw1130a.tif"
+url_5MB = "https://cdn.eso.org/images/large/potw1130a.jpg"
+url_1_5MB = "https://cdn.eso.org/images/publicationjpg/potw1130a.jpg"
+
+start = time.time()
+img = download_image(url_5MB)
```
philippmwirth, line 13: did this test fail before your changes?

philippmwirth (commented):
Looks great!

tests/UNMOCKED_end2end_tests/test_download_large_files.py
MalteEbner: No, it did not. However, for some implementations I tried, the unittests worked and the actual download from a url failed. Thus I added this test.

lightly/api/utils.py
MalteEbner: I changed it to
`raise RuntimeError(f'Maximum retries exceeded! Original exception: {type(e)}: {str(e)}') from e`

tests/UNMOCKED_end2end_tests/test_download_large_files.py
MalteEbner: Currently, the unmocked tests fail because of https://lightly-ai.slack.com/archives/C014FJT529M/p1652273127740529

philippmwirth: Can you fix this in this PR or is it a bigger problem?

MalteEbner: I am trying, the fix is the lates commit, the unmocked test is here: https://github.com/lightly-ai/lightly/actions/runs/2307560569

lightly/api/utils.py
MalteEbner: No, otherwise the original exception is not printed if you do `str(exception)`, like we do in the docker. It is only printed in the full print modes that also print the `exception.__cause__`.

lightly/api/download.py
```diff
@@ -38,84 +38,311 @@ def download_image(url: str, session: requests.Session = None) -> PIL.Image.Imag
...
+            image.load()
+        return image
+
     req = requests if session is None else session
-    response = utils.retry(req.get, url=url, stream=True)
-    return PIL.Image.open(response.raw)
+    image = utils.retry(load_image, url, req)
+    return image
 
+if not isinstance(av, ModuleNotFoundError):
```
philippmwirth, line 51: This is not the ideal way of handling this problem because now if a user doesn't have `av` installed but wants to use these functions the error message is very confusing. If you look at line `24` we have a function `_check_av_available` which raises a well-formatted and informative error if `av` is not available. Also, the error you've experienced comes from typehints only so my suggested fix is:
- Change all typehints which use `av.XYZ` to their string counterpart (i.e. `"av.XYZ"`)
- Add a call to `_check_av_available()` to each function using `av`.

philippmwirth (approved):

Issue #782, opened 2022-04-26
DCL: Decoupled Contrastive Learning


guarin:
## DCL: Decoupled Contrastive Learning

13.10.2021
https://arxiv.org/abs/2110.06848

Introduces a new contrastive loss function (DCL loss). Compatible with architectures such as SimCLR but trains much faster. Works with small batch sizes.

Estimated effort to implement in Lightly: Very low
* Add DCL Loss

IgorSusmelj:
There is a reference implementation here: https://github.com/raminnakhli/Decoupled-Contrastive-Learning/blob/main/loss/dcl.py

I'd suggest to keep this loss as a separate instead of trying to merge it with the NTX-Ent loss we already have.

PR #793, opened 2022-05-11
Guarin lig 1055 dcl decoupled contrastive learning

guarin:
* Add DCL loss
* Add DCLW loss
* Add DCL & DCLW examples docs
* Add lightly/utils/dist.py
* Refactor NTXentLoss

I got 84.1% with SimCLR + DCL, compared to 82.2% with normal SimCLR on the cifar10 benchmark. I am currently running the benchmark for DCLW and SimCLR (for comparison and to verify the changes). 

There is one outstanding question regarding the weighting function in DCLW. I think the output tensors passed to the weighting function should be detached but the paper does not mention it. In my opinion the weighting function should just scale the loss and not be backpropagated. I'll run a test with and without detaching to see what happens.

Closes: #782


philippmwirth (changes requested):
I only found one typo, other than that this looks great!

philippmwirth:
> There is one outstanding question regarding the weighting function in DCLW. I think the output tensors passed to the weighting function should be detached but the paper does not mention it. In my opinion the weighting function should just scale the loss and not be backpropagated. I'll run a test with and without detaching to see what happens.

Curious to see the results 👍

lightly/loss/dcl_loss.py
```diff
@@ -0,0 +1,219 @@
...
+            Shape: (batch_size, embedding_size)
+        sigma:
+            Similarities are scaled by inverse sigma.
+    Returns:
+        A tensor with shape (batch_size,) where each entry is the weight for one
+        of the input images.
+    
+    """
+    similarity = torch.einsum('nm,nm->n', out0, out1) / sigma
+    return 2 - out0.shape[0] * torch.nn.functional.softmax(similarity, dim=0)
```
IgorSusmelj, line 37: You can `import torch.nn as nn` as it's a common PyTorch convention. That might shorten some of the lines in the file. (not a must but more a suggestion)

lightly/loss/dcl_loss.py
```diff
@@ -0,0 +1,219 @@
...
+        >>> out1 = model(t1)
+        >>>
+        >>> # calculate loss
+        >>> loss = loss_fn(out0, out1)
+        
+    """
+    def __init__(
+        self,
+        temperature: float = 0.1,
+        weight_fn: Optional[Callable[[Tensor, Tensor], Tensor]] = None,
```
IgorSusmelj, line 79: If we add such "complex" type hints I'd expect somewhere examples/ explanations on what this means. Ideally, above where you describe the attribute.

lightly/loss/dcl_loss.py
```diff
@@ -0,0 +1,219 @@
...
+        Args:
+            out0:
+                Output projections of the first set of transformed images.
+                Shape: (batch_size, embedding_size)
+            out1:
+                Output projections of the second set of transformed images.
+                Shape: (batch_size, embedding_size)
+            out0_all:
+                Output projections of the first set of transformed images from
+                all processes.
```
IgorSusmelj, line 135: Add/ mention that "all processes" is for distributed processing (e.g. from all GPUs)

IgorSusmelj (changes requested):
This looks very good. I added a few suggestions. The only thing I'd def. do is explain some of the comments more and write out the code for the benchmarks.

IgorSusmelj:
And please don't forget to update readme, docs etc.

lightly/loss/dcl_loss.py
```diff
@@ -0,0 +1,219 @@
...
+                Shape: (batch_size, embedding_size)
+            out1:
+                Output projections of the second set of transformed images.
+                Shape: (batch_size, embedding_size)
+        
+        Returns:
+            Mean loss over the mini-batch.
+        """
+        # normalize the output to length 1
+        out0 = out0_all = torch.nn.functional.normalize(out0, dim=1)
```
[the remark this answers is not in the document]
guarin: > Also, what would be the reason to have self.gather_distributed == False?

You can save some communication overhead between the gpus. And it also does not break training if you forget by accident to use a [distributed sampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.distributed.DistributedSampler) when training with multiple gpus.

guarin:
I updated the pr with the suggested changes. I also added full example code for dcl to the docs, this should make it more consistent with the other examples.

[DCL & DCLW — lightly 1.2.14 documentation.pdf](https://github.com/lightly-ai/lightly/files/8685238/DCL.DCLW.lightly.1.2.14.documentation.pdf)

I am still running the benchmark with DCLW and detached weight calculation. Will add the values later.

guarin:
Added the benchmark results, runtime is somehow slower than for previous runs on the large gpu.

IgorSusmelj (approved):
Looks great!

philippmwirth (approved):

PR #796, opened 2022-05-12
Add a helper to create custom metadata config

philippmwirth:
# Add a helper to create custom metadata config

### DIFF
--- a/lightly/api/api_workflow_upload_metadata.py
+++ b/lightly/api/api_workflow_upload_metadata.py
@@ -8,8 +8,9 @@ from tqdm import tqdm
 
 from lightly.api.utils import retry
 from lightly.cli._helpers import print_as_warning
-from lightly.openapi_generated.swagger_client.models.sample_update_request import \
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_upload_metadata.py
```diff
@@ -214,3 +215,43 @@ def upload_sample_metadata(upload_request):
...
+            >>> )
+        
+        
+        """
+        config_set_request = ConfigurationSetRequest(name=name, configs=configs)
+        resp = self._metadata_configurations_api.create_meta_data_configuration(
+            body=config_set_request,
+            dataset_id=self.dataset_id,
+        )
+        return resp
```
MalteEbner, line 256: I don't see a reason to return the response.

MalteEbner (approved):
This looks almost perfect to me, good implementation, nice typehints and example.

Just one minor point: I would not return the response, it is not needed by the user.

PR #798, opened 2022-05-13
Guarin lig 919 add default values to model heads

guarin:
* Add default values from papers to model heads
* Add note that model examples do not follow reference settings
* Add BYOLPredictionHead
* Fix some wrong values in imagenette benchmark

### DIFF
--- a/lightly/models/modules/heads.py
+++ b/lightly/models/modules/heads.py
@@ -70,9 +70,9 @@ class BarlowTwinsProjectionHead(ProjectionHead):
     """
 
     def __init__(self,
-                 input_dim: int,
-                 hidden_dim: int,
-                 output_dim: int):
+                 input_dim: int = 2048,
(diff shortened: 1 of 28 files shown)


lightly/models/modules/heads.py
```diff
@@ -152,9 +174,9 @@ class NNCLRPredictionHead(ProjectionHead):
 
     """
     def __init__(self,
-                 input_dim: int,
-                 hidden_dim: int,
-                 output_dim: int):
+                 input_dim: int = 256,
+                 hidden_dim: int = 4095,
```
philippmwirth, line 178: I guess this is a typo and it should be `4096`

philippmwirth (changes requested):
One typo otherwise all the numbers seem to be correct 👍

philippmwirth (approved):

PR #807, opened 2022-05-18
lig-891 cd for docs

shaundaley39:
- in active learning tutorials (platform 3 and 4), now shows how to remove created tags
- the real motivation for this is to allow for (sufficiently) idempotent docs builds
- when a tutorial is built creating tags, tags are also then deleted during the build, avoiding naming collisions on the next build

Screenshot from 2022-05-18 10-06-35
Screenshot from 2022-05-18 10-06-49


philippmwirth (approved):
Looks good 👍

PR #808, opened 2022-05-19
Make retry function and request parameters configurable

guarin:


lightly/api/download.py
```diff
@@ -154,7 +170,7 @@ def video_frame_count(
             decoded.
 
         """
-        with av.open(url) as container:
+        with retry_fn(av.open, url) as container:
```
MalteEbner, line 173: Here I would retry the whole av.open block instead of just the open itself. Alternatively, do not retry at all. The actual frames are only loaded when the container is decoded. Retrying only the open, but not the frame loading does not make sense.

lightly/api/download.py
```diff
@@ -25,27 +25,39 @@ def _check_av_available() -> None:
...
+        request_kwargs:
+            Additional parameters passed to requests.get().
 
     Returns:
         The downloaded image.
 
     """
-    def load_image(url, req):
-        with req.get(url=url, stream=True) as response:
+    request_kwargs = request_kwargs or dict(stream=True, timeout=10)
```
MalteEbner, line 50: Here I would rather use `request_kwargs.set_default()`. Otherwise, there will be no `stream=True` if` {"timeout": 2}` is passed.

MalteEbner (commented):
The implementation looks good to me.

What I am missing are some unittests that pass the `retry_fn`, `request_kwargs` and `video_exceptions_to_catch`.

lightly/api/download.py
guarin: Ah, I maybe should have explained this better. In general we always use two nested retries when reading images/videos. The inner retry handles the opening part of the file/video. The outer retry is responsible for handling any errors that happen while the file is downloaded. For example if you try to load a file from an invalid url the inner retry will fail. If you have internet connection problems while loading a huge video file then it will probably fail in the outer retry.

The reason why there is not only an outer retry is two-fold. First, some libraries like `requests` already partially handle retries themselves, so you get automatically an inner retry. Second, the two retries handle different error cases and they should probably not interfere with each other. Note that if the inner retry fails because it reaches the maximum number of retries then the outer retry will NOT retry again.

MalteEbner: Thanks for explaining this well :)

MalteEbner (approved):
Great work!

PR #809, opened 2022-05-19
Malte lig 1014 improve video indexing speed for local datasets

MalteEbner:
## Description
- remove code trying to estimate the number of frames differently with pyav because it is buggy and deprecated. See https://lightly-ai.slack.com/archives/C014FJT529M/p1651762322215329?thread_ts=1651761912.070349&cid=C014FJT529M
- make offset and timestamp increasing calculation a bit faster using numpy and pre-allocated arrays.
- use multithreading for counting the frames (implemented with a torch dataloader and dataset). Once we switch to torch >= 1.11, we could use datapipes instead: https://github.com/pytorch/pytorch/tree/master/torch/utils/data/datapipes

## Tests:
with dataset with 2000 videos, 1.5GB
on a  6-core MBP

with 0 workers: 335s
with 4 workers: 94s
with 6 workers: 73s
with 12 workers: 76s

## Limitations
I did not have a look at improving the video indexing speed for `IterableDatasets`. It is already parallelized here: https://github.com/lightly-ai/lightly/blob/ac09efa69c183a0a98447350d2f30307956e1dcd/lightly/api/download.py#L294-L295 
This should be handled in a separate issue in my eyes.

### DIFF
--- a/lightly/data/_video.py
+++ b/lightly/data/_video.py
@@ -10,6 +10,7 @@ import threading
 import weakref
 import warnings
 
+import numpy as np
 from PIL import Image
 
 import torch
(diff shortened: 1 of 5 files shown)


lightly/data/_video.py
```diff
@@ -4,16 +4,19 @@
 # All Rights Reserved
 
 import os
+from multiprocessing import freeze_support
```
guarin, line 7: I don't think this import is used.

guarin (approved):
Looks good :) I think you imported something by accident that you don't use in the code.

I think we could use a `ProcessPoolExecutor` instead of a `ThreadPoolExecutor` for iterable dataset indexing. We used `ThreadPoolExecutor` in the beginning because we just read the `stream.frames` attribute to get the total number of frames. But as this can return wrong/no values we switched to reading all the frames. This should be faster when using multiprocessing instead of multithreading.

PR #817, opened 2022-06-01
Add collapse detection helper

philippmwirth:
# Add collapse detection helper

As described [here](https://ar5iv.labs.arxiv.org/html/2011.10566#S4.SS1) this can be used to monitor the level of collapse of the output of a self-supervised model.


guarin (approved):
Nice!

PR #818, opened 2022-06-02
plot positive and negative example pairs

philippmwirth:
# plot positive and negative example pairs

Adds a helper to plot augmented example images and their originals for the different collate functions implemented by `lightly`:

Example SimCLR:
hello

Example DINO (with real pictures)
hello


guarin (approved):
Looks great! Just two small suggestions :)

PR #835, opened 2022-06-10
Make sure we use default version for hydra config

IgorSusmelj:
## Changes

- There were warning messages due to not specifying the hydra config version and this PR resolves that
- Has been tested with different hydra versions and it seems to work fine :)
- Note that I also had to fix our CI pipeline here:
  - Clean cache of github actions
  - Make sure we don't use the latest AV package (seems broken). We now use `8.x` in our requirements.txt


lightly/cli/_helpers.py
```diff
@@ -51,6 +54,18 @@ def fix_input_path(path):
     return path
 
 
+def fix_hydra_arguments(config: str = 'config'):
```
philippmwirth, line 57: This should have two arguments: `config_path` and `config_name`. Also, I think `adapt_hydra_arguments` would probably be the better name but that's just a small thing.

lightly/cli/_helpers.py
```diff
@@ -51,6 +54,18 @@ def fix_input_path(path):
...
+    """Helper to make hydra arugments adaptive to installed hydra version
+    
+    Hydra introduced the `version_base` argument in version 1.2.0
+    We use this helper to provide backwards compatibility to older hydra verisons.    
+    """
+    if version_compare(hydra.__version__, '1.1.2') > 0:
```
philippmwirth, line 63: This `version_compare` is only safe if the version is of the format `X.Y.Z` and this is fine for our own version checks atm. However, we have no control over hydra versions (they could release `1.2.3beta` for example and I don't know how the version check would handle this. So my suggestion would be to wrap the whole thing in a try-catch and return some save default value.

lightly/utils/version_compare.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Examples:
+
+        >>> # compare two versions
+        >>> version_compare('1.2.0', '1.1.2')
+        >>> 1
+    """
+    v0 = [int(n) for n in v0.split('.')][::-1]
+    v1 = [int(n) for n in v1.split('.')][::-1]
+    assert len(v0) == 3
```
philippmwirth, line 21: I'd replace this by a `ValueError` (asserts for external checks is bad practice) and then catch the value error in the `fix_hydra_arguments` 🙂

philippmwirth (changes requested):

lightly/cli/_helpers.py
IgorSusmelj: I can expose both as arguments. Could you elaborate on where/ why we use different `config_path` and `config_name`? I quickly searched through the code and haven't seen any other use. 

I went for fix and not adapt or update because that's what it is. It's a temporary fix for version incompatibility. Similar to what we did with other fixes from PL or paths in the codebase. Just wanted to be consistent. At one point we hopefully can get rid of some of the rather manual fixes/ patches :D

philippmwirth: We don't use separate arguments yet but I would just expose both because bit's much cleaner. The `config_path` refers to the name of the directory where the config with `config_name` is stored so they actually mean different things. In our case it's just `config/config.yml`.

philippmwirth (approved):

PR #836, opened 2022-06-13
Fix bug in try except block

IgorSusmelj:
## Changes

- fixes the wrong fix from my previous PR
- I did not consider how try/ except in Python works when returning values and using the `finally` statement (see https://docs.python.org/3/reference/compound_stmts.html#the-try-statement)

This code has been tested against several `hydra-core` versions manually and it works now as expected :)

### DIFF
--- a/lightly/cli/_helpers.py
+++ b/lightly/cli/_helpers.py
@@ -63,7 +63,9 @@ def fix_hydra_arguments(config_path: str = 'config', config_name: str = 'config'
     try:
         if version_compare(hydra.__version__, '1.1.2') > 0:
             return {'config_path': config_path, 'config_name': config_name, 'version_base': '1.1'}
-    finally:
+        else:
+            return {'config_path': config_path, 'config_name': config_name} 
+    except:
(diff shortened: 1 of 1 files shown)


lightly/cli/_helpers.py
```diff
@@ -63,7 +63,9 @@ def fix_hydra_arguments(config_path: str = 'config', config_name: str = 'config'
     try:
         if version_compare(hydra.__version__, '1.1.2') > 0:
             return {'config_path': config_path, 'config_name': config_name, 'version_base': '1.1'}
-    finally:
+        else:
+            return {'config_path': config_path, 'config_name': config_name} 
+    except:
```
michal-lightly, line 68: Let's make this function cleaner by removing the returns from the try-catch block.

michal-lightly (changes requested):
Please rewrite this such that there is no `return` in the try-catch block to prevent future problems :)

michal-lightly (approved):
chefskiss

PR #838, opened 2022-06-14
Add docker config issues to faq

IgorSusmelj:
## Changes

- adds further instructions to FAQ in docs to overcome issues when pulling docker images with existing docker configs

### DIFF
--- a/docs/source/docker/known_issues_faq.rst
+++ b/docs/source/docker/known_issues_faq.rst
@@ -155,3 +155,40 @@ This can be avoided by setting your `token` via the `LIGHTLY_TOKEN` environment 
         -v {OUTPUT_DIR}:/home/shared_dir \
         lightly/worker:latest \
         ...
+
+
+.. _rst-docker-known-issues-faq-pulling-docker:
+
+No permission to pull the docker image
+--------------------------------------
+
+Please make sure the authentication succeeded as described in the 
+:ref:`ref-docker-download-and-install`  guide.
+
+If you still can't pull the docker image it might be that the docker config
+is causing the problem.
+
+You can check the config using the following command:
+
+.. code-block:: console
+
+    cat ~/.docker/config.json 
+
+You should see a section with the key for authentication. If you also see
+a section about the `credHelpers` they might overrule the authentication.
+
+In this case, you can delete the config and run the authentication again.
+
+Or you can work with two configs. We recommend creating a dedicated folder
+for the Lightly docker config.
+
+.. code-block:: console
+
+    mkdir -p ~/.docker_lightly/
+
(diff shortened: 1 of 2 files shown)


docs/source/docker/known_issues_faq.rst
```diff
@@ -155,3 +155,40 @@ This can be avoided by setting your `token` via the `LIGHTLY_TOKEN` environment
...
         ...
+
+
+.. _rst-docker-known-issues-faq-pulling-docker:
+
+No permission to pull the docker image
+--------------------------------------
+
+Please make sure the authentication succeeded as described in the 
+:ref:`ref-docker-download-and-install`  guide.
```
guarin, line 166: double whitespace between `:ref:'ref-docker-download-and-install'` and `guide`

docs/source/docker/known_issues_faq.rst
```diff
@@ -155,3 +155,40 @@ This can be avoided by setting your `token` via the `LIGHTLY_TOKEN` environment
...
+You can check the config using the following command:
+
+.. code-block:: console
+
+    cat ~/.docker/config.json 
+
+You should see a section with the key for authentication. If you also see
+a section about the `credHelpers` they might overrule the authentication.
+
+In this case, you can delete the config and run the authentication again.
```
guarin, line 180: I am a bit confused here. Do I understand correctly that if there is a `credHelpers` section in the config.json file one should delete the configuration and run authentication again?

What would be the authentication command without the config.json?

guarin (commented):
Looks good :) I had just a small question about the faq

docs/source/docker/known_issues_faq.rst
IgorSusmelj: No, I'll rephrase that then. Thanks for pointing it out. The credHelper overrides the source for the authentication key. If there is a credHelper section the user might not use the right credential to pull the image. So one solution is to drop the whole config and create a new one from scratch. Whenever you run docker login it should also create a config if there is none yet

guarin (approved):

docs/source/docker/known_issues_faq.rst
```diff
@@ -177,18 +177,27 @@ You can check the config using the following command:
...
+The Lightly docker images are hosted in the European location. Therefore,
+it's important that pulling from the `eu.gcr.io` domain is using 
+the provided credentials.
 
-.. code-block:: console
 
-    mkdir -p ~/.docker_lightly/
+There are two ways to solve the problem:
+
+- You can delete the config and run the authentication again.
```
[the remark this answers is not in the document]
IgorSusmelj: Let me add this for clarification :)

PR #839, opened 2022-06-14
Guarin lig 1220 masked autoencoder examples

guarin:


docs/source/examples/mae.rst
```diff
@@ -0,0 +1,55 @@
...
+
+MAE
+===
+
+Example implementation of the Masked Autoencoder (MAE) architecture. MAE is a
+transformer model based on the `Vision Transformer (ViT) <https://arxiv.org/abs/2010.11929>`_ 
+architecture. It learns image representations by predicting pixel values for
+masked patches on the input images. The network is split into an encoder and
+decoder. The encoder generates the image representation and the decoder predicts
+the pixel values from the representation. MAE outperforms other self-supervised
```
IgorSusmelj, line 11: Too much text. I'd strip the part starting with `MAE outperforms other...` as this part will be outdated soon anyway. I think a quick description of the model is fine but then we should make it consistently also for other models.

IgorSusmelj (changes requested):
Great work!

Two things I'd change:
- reduce the text describing the model. We're not building another model zoo with detailed descriptions. And this would not be aligned with the other models we have.
- I'd mention any limitations here. There is a `from_vit` part to build a backbone. So I can't use any transformer? Any limitation regarding input image size? Same limitations as e.g. DINO? We should point that also better out in DINO as transformer models are bit more restricted in that regard.

docs/source/examples/mae.rst
guarin: Removed it :)

IgorSusmelj (approved):

PR #845, opened 2022-06-16
fix broken code snippets in `lightly_at_a_glance` documentation

jwuphysics:
I made some simple updates to ensure that the code on this page will run correctly!

1. Replace `nn` with `torch.nn` since the former isn't imported
2. Include the `super().__init__()` line that is needed to make inherited classes initialize.
3. Add necessary parameter `lr` when initializing the `torch.optim.SGD` optimizer in the `SimCLR` class.


MalteEbner (approved):
Hey @jwuphysics 
thanks a lot for taking the time of proposing changes in the documentation! We are very glad to have you contributing to Lightly.
I checked all you changes, they are correct, thus this is ready to be merged :)

PR #850, opened 2022-06-23
Add PIRL

shikharmn:
- `collate.py`: Added new class `PIRLCollateFunction`, which inherits from `nn.Module`.
-  `jigsaw.py`: Added shuffling of the crops within jigsaw augmentation.
-  `test_data_collate.py`: Added three tests for `PIRLCollateFunction`.

This collate function directly inherits from `nn.Module` because PIRL requires data differently, compared to all other collate function classes that inherit from `BaseCollateFunction`. This collate function gives us two batches, one with the original image and another with 9 crops of size `input_size // n_grid`. As per #396.


guarin (approved):
This pr looks great! Thanks a lot for the contribution @shikharmn 💯

PR #863, opened 2022-07-06
Add timeout for all api requests

guarin:
Changes:

* Add default timeout for all api requests

The timeout is added by patching the request function of the api client generated by swagger. This is sadly necessary because the client uses no timeout if it is not explicitly passed to the request method.

### DIFF
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -43,7 +43,6 @@ from lightly.openapi_generated.swagger_client.models.dataset_data import \
     DatasetData
 from lightly.utils.reordering import sort_items_by_keys
 
-
 class ApiWorkflowClient(_UploadEmbeddingsMixin,
                         _SelectionMixin,
                         _UploadDatasetMixin,
@@ -89,6 +88,7 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin,
         api_client = ApiClient(configuration=configuration)
         self.user_agent = f"Lightly/{__version__}/python ({platform.platform()})"
         self.api_client = api_client
+        self.set_request_timeout(DEFAULT_API_TIMEOUT)
 
         self.token = token
         if dataset_id is not None:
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_client.py
```diff
@@ -220,3 +220,49 @@ def upload_file_with_signed_url(
...
+        """Sets a default timeout for all api requests.
+
+        Args:
+            timeout:
+                Timeout in seconds. Is either a single total_timeout value or a
+                (connect_timeout, read_timeout) tuple. 
+                See https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html?highlight=timeout#urllib3.util.Timeout
+                for details on the different values.
+        """
+        set_api_client_request_timeout(client=self.api_client, timeout=timeout)
```
japrescott, line 234: maybe add a sanity check if `self.api_client` is actually available? (otherwise its strongly dependent on the order of calling these fn)

lightly/api/api_workflow_client.py
```diff
@@ -220,3 +220,49 @@ def upload_file_with_signed_url(
...
+        timeout:
+            Timeout in seconds. Is either a single total_timeout value or a
+            (connect_timeout, read_timeout) tuple. 
+            See https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html?highlight=timeout#urllib3.util.Timeout
+            for details on the different values.
+
+    """
+    request_fn = client.rest_client.request
+
+    def new_request_fn(*args, **kwargs):
```
japrescott, line 262: HACKERMAN

japrescott (approved):
very nicely done!

lightly/api/api_workflow_client.py
guarin: The `self.api_client` attribute should always exist as it is created in the `__init__` method. The only error that can happen is if someone changes the order of the function calls in the init itself but this is covered by unit tests.

PR #879, opened 2022-07-27
allow parallel downloading and some blingbling

japrescott:
closes lig-1294
- speedup `lightly-download` by using workers to download files in parallel


philippmwirth (approved):
Looks like a bit of feature creep happened here ;)

japrescott:
@philippmwirth I would never do something like feature creep 👼

PR #880, opened 2022-07-28
Update openapi generated client

philippmwirth:


lightly/openapi_generated/.swagger-codegen/VERSION
```diff
@@ -1 +1 @@
-3.0.34
\ No newline at end of file
+3.0.26
```
japrescott, line 1: que pasa? maybe you want to update your swagger-codegen thingy locally

japrescott (changes requested):

lightly/openapi_generated/.swagger-codegen/VERSION
philippmwirth: Well the version in the script to generate on `ubuntu` is hardcoded to `3.0.26`. I can try to upgrade it but that's a problem of our setup.

philippmwirth: that was easy 👍

japrescott (approved):

PR #882, opened 2022-07-28
closes lig-1071 lightly test write file not removed

shaundaley39:
- the most direct way of ensuring that lightly-test-write-file.txt is removed is to include delete permissions in gcloud and azure
- this also makes IAM user instructions consistent across aws, gcloud and azure

The background investigation is covered in https://linear.app/lightly/issue/LIG-1071/lightly-test-write-file-not-removed

Screenshot from 2022-07-28 23-44-28
Screenshot from 2022-07-28 23-45-09


MalteEbner (approved):
This looks good to me.

PR #885, opened 2022-07-29
step2: lig-1541: use getSamplesPartialByDataset instead of fetching everything: slightly bypasses the 32MB restriction

japrescott:
closes lig-1541
- use getSamplesPartialByDataset instead of fetching everything: slightly bypasses the 32MB restriction


lightly/api/api_workflow_upload_metadata.py
```diff
@@ -161,8 +162,9 @@ def upload_custom_metadata(self,
         }
 
         samples = retry(
-            self._samples_api.get_samples_by_dataset_id,
-            dataset_id=self.dataset_id
+            self._samples_api.get_samples_partial_by_dataset_id,
```
philippmwirth, line 165: is the sample id also returned here because it's needed two lines below?

philippmwirth (commented):
One question

lightly/api/api_workflow_upload_metadata.py
japrescott: yes, id and filename is returned

philippmwirth (approved):

PR #886, opened 2022-08-03
Test python 3.10

michal-lightly:


guarin:
@michal-lightly can we close this PR?

PR #895, opened 2022-08-09
Guarin lig 1572 implement msn paper

### LINKED ISSUE #878: MSN: Masked Siamese Networks for Label-Efficient Learning
Hello, I would like to share this new interesting SSL method proposed by Meta that achieves SOTA results and shows interesting generalization properties.

It is based on ViT and I think it could be worth to integrate it in this codebase. Also, the official implementation is available :)

[MSN Paper](https://arxiv.org/abs/2204.07141), [MSN GIthub code](https://github.com/facebookresearch/msn)

guarin:
TODO:

- [x] projection head
- [x] collate function
- [x] loss
- [x] distributed
- [x] examples
- [x] benchmark model
- [x] benchmark results
- [x] docs
- [x] add to api docs
- [x] update readme

Closes #878 

Last benchmark results:
```
With image size 224 (and 96 for local views):
------------------------------------------------------------------------------------------
| Model         | Batch Size | Epochs |  KNN Test Accuracy |       Time | Peak GPU Usage |
------------------------------------------------------------------------------------------
| MSN           |        128 |    200 |              0.749 |  180.2 Min |     21.6 GByte |
| MSN           |        256 |    200 |              0.743 |  187.9 Min |     42.5 GByte |
| MSN           |        256 |    800 |              0.820 |  748.9 Min |     41.7 GByte |
------------------------------------------------------------------------------------------

With image size 128 (and 64 for local views):
------------------------------------------------------------------------------------------
| Model         | Batch Size | Epochs |  KNN Test Accuracy |       Time | Peak GPU Usage |
------------------------------------------------------------------------------------------
| MSN           |        256 |    200 |              0.741 |   92.7 Min |     16.3 GByte |
| MSN           |        256 |    800 |              0.834 |  376.1 Min |     16.3 GByte |
------------------------------------------------------------------------------------------
```

### DIFF
--- a/lightly/loss/msn_loss.py
+++ b/lightly/loss/msn_loss.py
@@ -1,0 +1,164 @@
+import math
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+
+def prototype_probabilities(
+    queries: torch.Tensor,
+    prototypes: torch.Tensor,
+    temperature: float,
+) -> torch.Tensor:
+    """Returns probability for each query to belong to each prototype.
+    
+    Args:
+        queries:
+            Tensor with shape (batch_size, dim)
+        prototypes:
+            Tensor with shape (num_prototypes, dim)
+        temperature:
+            Inverse scaling factor for the similarity.
+
+    Returns:
+        Probability tensor with shape (batch_size, num_prototypes) which sums to 1 along
+        the num_prototypes dimension.
+
+    """
+    return F.softmax(torch.matmul(queries, prototypes.T) / temperature, dim=1)
+
+def sharpen(probabilities: torch.Tensor, temperature: float) -> torch.Tensor:
+    """Sharpens the probabilities with the given temperature.
+    
+    Args:
+        probabilities:
+            Tensor with shape (batch_size, dim)
+        temperature:
+            Temperature in (0, 1]. Lower temperature results in stronger sharpening (
+            output probabilities are less uniform).
+    Returns:
+        Probabilities tensor with shape (batch_size, dim).
+
+    """
+    probabilities = probabilities ** (1.0 / temperature)
+    probabilities /= torch.sum(probabilities, dim=1, keepdim=True)
+    return probabilities
+
+@torch.no_grad()
+def sinkhorn(
+    probabilities: torch.Tensor, 
+    iterations: int = 3,
+) -> torch.Tensor:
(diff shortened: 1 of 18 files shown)


IgorSusmelj:
On which dataset are the benchmarks? And which GPU? The usage seems waaay off. If you want to add the benchmark results, please make sure you use the same config (batch size, epochs etc.) as we use by default such that all the benchmarks can be run together and we should get the same result :)

guarin:
> On which dataset are the benchmarks? And which GPU? The GPU memory usage seems waaay off. If you want to add the benchmark results, please make sure you use the same config (batch size, epochs etc.) as we use by default such that all the benchmarks can be run together and we should get the same result :)

Good point :) I ran the benchmarks with the augmentation settings from the paper (image size = 224). In total there are a couple reasons why GPU usage is higher compared to other models:

* ViT-S backbone instead of Res18
* Many views in single forward pass (12 views)
* Larger image size than for other benchmarks  (224 instead of 128)

I ran an additional benchmark with a smaller image size (128) and got the following results:

```
------------------------------------------------------------------------------------------
| Model         | Batch Size | Epochs |  KNN Test Accuracy |       Time | Peak GPU Usage |
------------------------------------------------------------------------------------------
| MSN           |        256 |    200 |              0.741 |   92.7 Min |     16.3 GByte |
------------------------------------------------------------------------------------------
```

I already updated the results in the benchmark file but not yet the docs. I am currently running the 800 epoch benchmark with smaller image size as well. Will update all the benchmark results once this one is done.

lightly/loss/msn_loss.py
```diff
@@ -0,0 +1,203 @@
...
+    Returns:
+        Probabilities tensor with shape (batch_size, dim).
+
+    """
+    probabilities = probabilities ** (1.0 / temperature)
+    probabilities /= torch.sum(probabilities, dim=1, keepdim=True)
+    return probabilities
+
+@torch.no_grad()
+def sinkhorn(
```
philippmwirth, line 49: Can we not re-use the sinkhorn vom swav? E.g. move it to utils

guarin: It is slightly different :(

philippmwirth (approved):
amazing work, let's get this merged :)

guarin:
I'll run the 800 epoch benchmark, update the benchmark docs and then merge it. Should be done by tomorrow.

PR #897, opened 2022-08-11
closes lig-1485 Lots of error messages when uploading data to datasource

shaundaley39:
403 permissions errors occur when uploading to Google Cloud Storage datasources, where the access role does not have either (`storage.objects.create` *and* `storage.objects.update`) *or* (`storage.objects.create` *and `storage.objects.delete`) permissions.

Here, we update the user documentation so that users (should!) have their datasource correctly configured. Before: 
https://docs.lightly.ai/getting_started/dataset_creation/dataset_creation_gcloud_bucket.html

After:
Screenshot from 2022-08-11 14-33-07
Screenshot from 2022-08-11 14-33-13

Note: both `delete` and `update` are stated as expected permissions to simplify user setup.


japrescott (changes requested):
?! I am confused. What about the update permissions? and whatabout the images in the docs?

japrescott (approved):

PR #898, opened 2022-08-12
Add instructions on how to re-use a checkpoint

philippmwirth:
# Add instructions on how to re-use a checkpoint

Straight-forward PR adding the information about how to re-use a checkpoint (tested it personally first).

Screenshot:
image

### DIFF
--- a/docs/source/docker/getting_started/first_steps.rst
+++ b/docs/source/docker/getting_started/first_steps.rst
@@ -362,6 +362,43 @@ you might want to change:
     )
 
 
+**Checkpoints** from your training process will be stored in the output directory.
+You can use such a checkpoint in future worker runs by copying the checkpoint to
+a `shared directory` and then passing the checkpoint filename to the container.
+
+.. code-block:: console
+    :emphasize-lines: 3
+    :caption: Starting the worker with a `shared directory`
+
+    docker run --gpus all --rm -it \
+        -v {OUTPUT_DIR}:/home/output_dir \
+        -v {SHARED_DIR}:/home/shared_dir \
+        lightly/worker:latest \
+        token=MY_AWESOME_TOKEN \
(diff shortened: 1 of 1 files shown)


docs/source/docker/getting_started/first_steps.rst
```diff
@@ -362,6 +362,43 @@ you might want to change:
     )
 
 
+**Checkpoints** from your training process will be stored in the output directory.
```
IgorSusmelj, line 365: Let's add an example here: 

Just a suggestion:

The output directory is defined in the docker run command as shown in #volume-mapping. 
If your docker run command looks like this:
```console
docker run --gpus all --rm -it \
    -v /home/mymachine/lightly-output/:/home/output_dir \
    lightly/worker:latest \
    token=MY_AWESOME_TOKEN \
    worker.worker_id=MY_WORKER_ID
```
The checkpoint path would be something like this: `/home/mymachine/lightly-output/my_checkpoint_file.ckpt`

docs/source/docker/getting_started/first_steps.rst
```diff
@@ -362,6 +362,43 @@ you might want to change:
...
+You can use such a checkpoint in future worker runs by copying the checkpoint to
+a `shared directory` and then passing the checkpoint filename to the container.
```
IgorSusmelj, line 367: Also here I'd give an example.

For example, you can create a new folder `/home/mymachine/lightly-shared` and copy the checkpoint (`my_checkpoint_file.ckpt`) into this folder. Then make sure to set `{SHARED_DIR}` to `/home/mymachine/lightly-shared`.

IgorSusmelj (changes requested):
Two minor changes. Otherwise, this looks awesome!

IgorSusmelj (approved):
LGTM

PR #903, opened 2022-08-17
If we use reredirects for moving pages to readme.com, this is how …

shaundaley39:
Do not merge.

This is for illustration and reproduction only.

If we want to use sphinx-reredirects for moving pages, this is how we would do it.


guarin:
@shaundaley39 can we close this PR?

PR #905, opened 2022-08-18
Make Lightly compatible with Pytorch Lightning 1.7

michal-lightly:
### What changed

Make Lightly run in an environment with Pytorch lightning 1.7.

Pytorch Lightning 1.7 introduces a breaking change, the parameter `weights_summary` can not be passed to `pytorch_lightning.Trainer` anymore. Instead, a `ModelSummary` class must be instantiated and passed to the `Trainer` as a callback.
* [Pytorch lightning PR](https://github.com/Lightning-AI/lightning/pull/13070/files) with the change

New field `summary_callback` is introduced in config file. The old field `trainer.weights_summary` is deprecated and will be removed in the future.

Other changes:
* Callbacks are now created inside `BaseEmbedding.train_embedding()`. Callback configs are passed in.
* New file `callbacks.py` created to instantiate trainer callbacks
* Old file `_callbacks.py` contained dead code and was removed
* Explicitly specify lower version bound for `pytorch_lightning` and `torchvision` requirements. Lower versions were not supported even before this change.

### How to modify your old config

Instead of `trainer.weights_summary: "full"` specify:
```
trainer.enable_model_summary: True
summary_callback.max_depth: -1
```

Instead of `trainer.weights_summary: "top"` specify:
```
trainer.enable_model_summary: True
summary_callback.max_depth: 1
```

### How is it tested
* CI passes
* Tested in a local environment with Pytorch Lightning versions 1.5, 1.6 and 1.7


tests/embedding/test_embedding.py
```diff
@@ -66,7 +66,7 @@ def test_embed_correct_order(self):
             device=device,
         )
 
-        np.testing.assert_allclose(embeddings_1_worker, embeddings_4_worker, rtol=1e-5)
+        np.testing.assert_allclose(embeddings_1_worker, embeddings_4_worker, rtol=2e-5)
```
michal-lightly, line 69: Fix of a flaky test.

lightly/embedding/callbacks.py
```diff
@@ -0,0 +1,92 @@
...
+            Where to save the checkpoint.
+
+    """
+    return ModelCheckpoint(
+        dirpath=os.getcwd() if dirpath is None else dirpath,
+        filename='lightly_epoch_{epoch:d}',
+        save_last=save_last,
+        save_top_k=save_top_k,
+        monitor=monitor,
+        auto_insert_metric_name=False)
```
michal-lightly, line 34: Moved from `_base.py`.

michal-lightly (commented):
Note: The structure of the touched codebase is not great (tangled dependencies, questionable tests). I attempted to make the code better but I had to limit the effort not to spend too much time on it.

One thing that would be nice to test is to view the model summary at least manually. Does anyone know how to? (Maybe @philippmwirth )

guarin (approved):
Looks great and awesome that you made the code much cleaner!

I just had a small suggestion but nothing crucial, already approved for merging :)

lightly/embedding/callbacks.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    # TODO: Drop support for the "weights_summary" argument.
+    weights_summary = trainer_config.get("weights_summary", None)
+    if weights_summary not in [None, "None"]:
+        summary_callback = _create_summary_callback_deprecated(weights_summary)
+    else:
+        summary_callback = _create_summary_callback(**summary_callback_config)
+
+    if "weights_summary" in trainer_config:
+        del trainer_config["weights_summary"]
```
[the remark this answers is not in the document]
michal-lightly: This is a great suggestion!

requirements/base.txt
```diff
@@ -1,14 +1,14 @@
...
+lightly_utils~=0.0.0
 numpy>=1.18.1
-pytorch_lightning>=1.0.4,<1.7  # TODO: Hotfix, remove upper bound
+python_dateutil>=2.5.3
+pytorch_lightning>=1.5.0
 requests>=2.23.0
-torchvision
+setuptools>=21.0.0
+six>=1.10
+torch>=1.6.0
```
IgorSusmelj, line 11: This would be a new dependency, no?

requirements/base.txt
```diff
@@ -1,14 +1,14 @@
 # the default package dependencies
-
+certifi>=14.05.14
 hydra-core>=1.0.0
+lightly_utils~=0.0.0
 numpy>=1.18.1
-pytorch_lightning>=1.0.4,<1.7  # TODO: Hotfix, remove upper bound
+python_dateutil>=2.5.3
+pytorch_lightning>=1.5.0
```
IgorSusmelj, line 7: And this might restrict some users?

requirements/base.txt
michal-lightly: It is not, we installed it as a second order dependency of `torchvision`.

requirements/base.txt
michal-lightly: I verified that 1.5.0 was before the lowest `pytorch_lightning` version that lightly would run with. So this line just states that explicitly.

IgorSusmelj: Hmm, that's weird. Let's double check this before we merge. I'm not a big fan of adding strict dependencies except if it's really required. 
We have some projects here that use much older versions: https://github.com/lightly-ai/lightly/network/dependents?package_id=UGFja2FnZS0xNDQ3MzY0NTYy

E.g. this one here uses lightly 1.2.27 with pytorch lightning 1.2.7 (https://github.com/huaxiuyao/Wild-Time/blob/main/requirements.txt)

michal-lightly: I did more investigation.

I managed to create an environment where I can install pytorch lightning 1.4. There are three tests that are failing (master branch):
```
FAILED tests/cli/test_cli_magic.py::TestCLIMagic::test_magic_without_upload_with_trainer - TypeError: __init__() got an unexpected keyword...
FAILED tests/cli/test_cli_train.py::TestCLITrain::test_train_weights_summary - TypeError: __init__() got an unexpected keyword argument 's...
FAILED tests/core/test_Core.py::TestCore::test_train_and_embed - TypeError: __init__() got an unexpected keyword argument 'strategy'
```

The error is:
```
lightly/embedding/_base.py:88: in train_embedding
    trainer = pl.Trainer(**kwargs, callbacks=[self.checkpoint_callback])
...
>       return fn(self, **kwargs)
E       TypeError: __init__() got an unexpected keyword argument 'strategy'
```

michal-lightly: @IgorSusmelj I see no great options here.
* We cannot support all the versions of pytorch_lightning because they make breaking changes too fast. Currently, cli does not work with versions before 1.5. We don't specify this lower bound in requirements.txt, so someone might install the package and it will not work.
* By specifying the lower bound, projects that use lightly with an older pytorch lightning version (which means that they don't use the broken part) will need to do dependency resolution.

I do think that the better option here is to specify the lower bound, even though some people will need to do dependency resolution.
* We are still supporting older versions (1.5+), not just the newest one (1.7)
* It is always possible to override and call `pip install` with an older version after lightly is installed
* I checked the list, there are just three that use lightly with and older pl version. I tried to install one locally (the one you mentioned). The dependencies though could not be resolved (I tried several python versions, ran `pip install -r requirements.txt`). Also they say it is a WIP.
* We have to draw a line somewhere. It seems to me that 1.5 is a reasonable version.

michal-lightly: Discussed with Igor in person.

Conclusion:
* We don't include the lower bound
* Create a [follow-up issue](https://github.com/lightly-ai/lightly/issues/912)

PR #906, opened 2022-08-18
Update shm in docs to higher default value

IgorSusmelj:
## Changes

- add default shared memory of 1GB to all docker run commands
- add a small explanation of the docker run command in the first steps section

### DIFF
--- a/docs/source/docker/advanced/active_learning.rst
+++ b/docs/source/docker/advanced/active_learning.rst
@@ -55,7 +55,7 @@ We can start a Lightly Worker using the following command
 
 .. code-block:: console
 
-  docker run --rm --gpus all -it \
+  docker run --shm-size="1024m" --rm --gpus all -it \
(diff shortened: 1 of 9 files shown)


docs/source/docker/advanced/active_learning.rst
```diff
@@ -55,7 +55,7 @@ We can start a Lightly Worker using the following command
 
 .. code-block:: console
 
-  docker run --rm --gpus all -it \
+  docker run --shm-size="1024m" --rm --gpus all -it \
```
guarin, line 58: Just FYI you can also use gigabytes as units: `--shm-size="1g"`

IgorSusmelj: I know. But I thought it's more handy to use m (for megabytes) as users can easily adapt to 512, 1500 etc.

PR #913, opened 2022-08-25
Add selection config conversion

guarin:
Adds automatic conversion from selection config as a dict to a SelectionConfig class. This allows to verify the selection config before it is sent to the API and works with snake_case entries.

### DIFF
--- a/lightly/api/api_workflow_compute_worker.py
+++ b/lightly/api/api_workflow_compute_worker.py
@@ -1,3 +1,4 @@
+import copy
 from typing import Any, Dict, List, Optional, Union
 
 from lightly.openapi_generated.swagger_client import SelectionConfig
@@ -25,6 +26,7 @@ from lightly.openapi_generated.swagger_client.models.docker_worker_config import
 from lightly.openapi_generated.swagger_client.models.docker_worker_config_create_request import (
     DockerWorkerConfigCreateRequest,
 )
+from lightly.openapi_generated.swagger_client import SelectionConfig, SelectionConfigEntryInput, SelectionConfigEntryStrategy, SelectionConfigEntry
 
 
 class _ComputeWorkerMixin:
@@ -85,11 +87,15 @@ class _ComputeWorkerMixin:
             The id of the created config.
 
         """
+        if isinstance(selection_config, dict):
+            selection = _selection_config_from_dict(cfg=selection_config)
+        else:
+            selection = selection_config
         config = DockerWorkerConfig(
             worker_type=DockerWorkerType.FULL,
             docker=worker_config,
             lightly=lightly_config,
-            selection=selection_config,
+            selection=selection,
         )
         request = DockerWorkerConfigCreateRequest(config)
         response = self._compute_worker_api.create_docker_worker_config(request)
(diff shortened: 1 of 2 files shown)


MalteEbner (commented):
Great!

Did you also test this once over the API? Just to make sure that also works well.

tests/api_workflow/test_api_workflow_compute_worker.py
```diff
@@ -114,4 +123,85 @@ def test_selection_config(self):
         )
         config = DockerWorkerConfig(worker_type=DockerWorkerType.FULL, selection=selection_config)
 
-        config_api = self._check_if_openapi_generated_obj_is_valid(config)
\ No newline at end of file
+        config_api = self._check_if_openapi_generated_obj_is_valid(config)
+
+def test___selection_config_from_dict() -> None:
```
michal-lightly, line 128: Minor: I would suggest to ignore leading underscores when naming tests. But in any case, Three underscores are too excessive 🙂

lightly/api/api_workflow_compute_worker.py
```diff
@@ -85,11 +87,15 @@ def create_compute_worker_config(
             The id of the created config.
 
         """
+        if isinstance(selection_config, dict):
+            selection = _selection_config_from_dict(cfg=selection_config)
+        else:
+            selection = selection_config
```
michal-lightly, line 93: I would really consider to not support both versions, just choose one.

If we go with `SelectionConfig` let's make `_selection_config_from_dict` a public function and use it in examples.

michal-lightly (approved):
The tests look great!

tests/api_workflow/test_api_workflow_compute_worker.py
```diff
@@ -114,4 +123,85 @@ def test_selection_config(self):
...
+        ],
+    }
+    with pytest.raises(TypeError, match="got an unexpected keyword argument 'invalid-key'"):
+        api_workflow_compute_worker._selection_config_from_dict(cfg)
+
+
+def test___selection_config_from_dict__typo() -> None:
+    cfg = {"nSamples": 10}
+    with pytest.raises(TypeError, match="got an unexpected keyword argument 'nSamples'"):
+        api_workflow_compute_worker._selection_config_from_dict(cfg)
```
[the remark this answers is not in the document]
guarin: Added the extra tests

tests/api_workflow/test_api_workflow_compute_worker.py
guarin: Ah yes, my bad :) I actually like double underscore for private functions because it avoids potential naming conflicts if we have `fun` and `_fun` in the same file.

lightly/api/api_workflow_compute_worker.py
MalteEbner: I would allow both for the moment.

tests/api_workflow/test_api_workflow_compute_worker.py
michal-lightly: You should actually never do that anyway:
1) to avoid confusion
2) to easily switch private/public indication

(I know I did it in a previous PR, I should not have done it :))

But I'm fine with the double underscore convention too.

PR #900, opened 2022-08-15
Malte lig 1262 advanced selection external docs

MalteEbner:
## Description

- adds docs for selection using the new config
- adapts existing docs (getting started, docker active learning)
- already included: DIVERSIFY->DIVERSITY

### Tabs vs. Dropdowns
Here tabs are used for the SelectionStrategy and Dropdowns for the configuration examples.
image

PR #914, opened 2022-08-25
Add proportion_samples to generated api client

guarin:
Adds api generated code for `proportion_samples` option.

## TODO
* ~~Wait until https://github.com/lightly-ai/lightly/pull/913 is merged and rebase~~
* Add docs --> This should actually be done in https://github.com/lightly-ai/lightly/pull/900

### DIFF
--- a/tests/api_workflow/test_api_workflow_compute_worker.py
+++ b/tests/api_workflow/test_api_workflow_compute_worker.py
@@ -128,6 +128,7 @@ class TestApiWorkflowComputeWorker(MockedApiWorkflowSetup):
 def test__selection_config_from_dict() -> None:
     cfg = {
         "n_samples": 10,
+        "proportion_samples": 0.1,
         "strategies": [
             {
                 "input": {"type": "EMBEDDINGS", "dataset_id": "some-dataset-id", "tag_name": "some-tag-name"},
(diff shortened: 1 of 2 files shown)


tests/api_workflow/test_api_workflow_compute_worker.py
```diff
@@ -163,32 +165,20 @@ def test__selection_config_from_dict() -> None:
...
-def test__selection_config_from_dict__missing_n_samples() -> None:
-    cfg = {
-        "strategies": [
-            {"input": {"type": "EMBEDDINGS"}, "strategy": {"type": "DIVERSIFY"}},
-        ]
-    }
-    with pytest.raises(ValueError, match="Invalid value for `n_samples`, must not be `None`"):
-        api_workflow_compute_worker._selection_config_from_dict(cfg)
-
-
```
guarin, line 175: This test no longer makes sense as `n_samples` is now optional

MalteEbner (approved):
LGTM

Perhaps you have already though about it, but could we also have the move the`_verify_stopping_condition_set()` method here and call it before sending the SelectionConfig?

tests/api_workflow/test_api_workflow_compute_worker.py
MalteEbner: Could we two tests instead that checks that both `n_samples` and `proportion_samples` can be set?

guarin:
> Perhaps you have already though about it, but could we also have the move the_verify_stopping_condition_set() method here and call it before sending the SelectionConfig, e.g. after calling _selection_config_from_dict()?

I didn't think about it :D But if we do this then the config is only verified if it is set by the pip client. It would not verified if it is set by the API or the webapp.

tests/api_workflow/test_api_workflow_compute_worker.py
guarin: Oh yes, great catch!

guarin: It's actually verified by the `test__selection_config_from_dict` test.

PR #915, opened 2022-08-26
Fix weights summary deletion

guarin:


lightly/embedding/_base.py
```diff
@@ -101,9 +104,10 @@ def train_embedding(
 
         # Remove weights_summary from trainer_config now that the summary callback
         # has been created. TODO: Drop support for the "weights_summary" argument.
-        trainer_config_copy = trainer_config.copy()
+        trainer_config_copy = copy.deepcopy(trainer_config)
         if "weights_summary" in trainer_config_copy:
-            del trainer_config_copy["weights_summary"]
+            with omegaconf.open_dict(trainer_config_copy):
+                del trainer_config_copy["weights_summary"]
```
guarin, line 110: Required change because Lightly Worker has a frozen config which does not allow deleting keys.

michal-lightly (approved):
LGTM!

Please write also a line in the PR summary about what has changed and why.

PR #921, opened 2022-09-06
Igor lig 1609 allow users to share datasets from code

IgorSusmelj:
## Changes

- users can now share datasets using the Python API
- add `share_dataset` method to share datasets with other users
- add `get_shared_users` method to get list of users that have access to a dataset
- add unit tests

I tested the whole setup also manually using a code similar to this:

Example
```Python
import lightly

api_client = lightly.api.ApiWorkflowClient(token='MY_TOKEN')

# share a dataset from Python client api
api_client.share_dataset(dataset_id="DATASET_ID", user_emails=['someone@something.com'])

# get list of shared users for a dataset
api_client.get_shared_users(dataset_id='630f7f90f17b7808bfebec51')
# Out: ['someone@something.com']
```

### DIFF
--- a/docs/source/docker/getting_started/first_steps.rst
+++ b/docs/source/docker/getting_started/first_steps.rst
@@ -530,6 +530,52 @@ It is also possible to directly download the actual files themselves as follows:
     )
 
 
+Sharing Datasets
+----------------
+
+Once a dataset has been created we can also make it accessible to other users by
+sharing it. Sharing works through e-mail addresses. 
+
+.. code-block:: python
+    :caption: Share a dataset
+
+    # we first need to have an api client (create a new or use an existing one)
+    client = ApiWorkflowClient(token="MY_AWESOME_TOKEN")
+
+    # share a dataset with an user
+    client.share_dataset_only_with(dataset_id="MY_DATASET_ID", user_emails=["user@something.com"])
+
+    # share dataset with a user while keep sharing it with previous users
+    user_emails = client.get_shared_users(dataset_id="MY_DATASET_ID")
+    user_emails.extend(["additional_user2@something.com"])
+    client.share_dataset_only_with(dataset_id="MY_DATASET_ID", user_emails=user_emails)
+
+    # revoke access to all users
+    client.share_dataset_only_with(dataset_id="MY_DATASET_ID", user_emails=[])
+
+
+If you want to get a list of users that have access to a given dataset we can do 
+this using the following code: 
+
+.. code-block:: python
+    :caption: Share a dataset
+
+    # we first need to have an api client (create a new or use an existing one)
+    client = ApiWorkflowClient(token="MY_AWESOME_TOKEN")
+
+    # get a list of users that have access to a given dataset
+    client.get_shared_users(dataset_id="MY_DATASET_ID")
+    # ["user@something.com"]
+
+You find more details about the individual API requests in the docs for the
+Python client: :py:class:`lightly.api.api_workflow_client`
+
+
+.. note::
+
+    You can share a dataset immediately after creating the dataset. You don't have
+    to wait for a Lightly Worker run to complete!
+
 
 Reporting
 ---------
--- a/docs/source/docker/getting_started/first_steps.rst
+++ b/docs/source/docker/getting_started/first_steps.rst
@@ -568,9 +568,6 @@ this using the following code:
     print(users)
     # ["user@something.com"]
 
-You find more details about the individual API requests in the docs for the
-Python client: :py:class:`lightly.api.api_workflow_client`
-
 
 .. note::
 
--- a/docs/source/docker/getting_started/first_steps.rst
+++ b/docs/source/docker/getting_started/first_steps.rst
@@ -542,12 +542,12 @@ sharing it. Sharing works through e-mail addresses.
     # we first need to have an api client (create a new or use an existing one)
     client = ApiWorkflowClient(token="MY_AWESOME_TOKEN")
 
-    # share a dataset with an user
+    # share a dataset with a user
     client.share_dataset_only_with(dataset_id="MY_DATASET_ID", user_emails=["user@something.com"])
 
     # share dataset with a user while keep sharing it with previous users
     user_emails = client.get_shared_users(dataset_id="MY_DATASET_ID")
-    user_emails.extend(["additional_user2@something.com"])
+    user_emails.append("additional_user2@something.com")
     client.share_dataset_only_with(dataset_id="MY_DATASET_ID", user_emails=user_emails)
 
     # revoke access to all users
@@ -565,6 +565,7 @@ this using the following code:
 
     # get a list of users that have access to a given dataset
     client.get_shared_users(dataset_id="MY_DATASET_ID")
+    print(users)
     # ["user@something.com"]
 
 You find more details about the individual API requests in the docs for the
--- a/lightly/api/api_workflow_collaboration.py
+++ b/lightly/api/api_workflow_collaboration.py
@@ -28,14 +28,14 @@ class _CollaborationMixin:
             List of email addresses of users to grant write permission
 
         Examples:
-          >>> # share a dataset with an user
+          >>> # share a dataset with a user
           >>> client = ApiWorkflowClient(token="MY_AWESOME_TOKEN")
           >>> client.share_dataset_only_with(dataset_id="MY_DATASET_ID", user_emails=["user@something.com"])
           >>>
           >>> # share dataset with a user while keep sharing it with previous users
           >>> client = ApiWorkflowClient(token="MY_AWESOME_TOKEN")
           >>> user_emails = client.get_shared_users(dataset_id="MY_DATASET_ID")
-          >>> user_emails.extend(["additional_user2@something.com"])
+          >>> user_emails.append("additional_user2@something.com")
           >>> client.share_dataset_only_with(dataset_id="MY_DATASET_ID", user_emails=user_emails)
           >>>
           >>> # revoke access to all users
--- a/lightly/api/api_workflow_collaboration.py
+++ b/lightly/api/api_workflow_collaboration.py
@@ -1,0 +1,76 @@
+from os import access
+from typing import List
+
+from lightly.openapi_generated.swagger_client.models.create_entity_response import CreateEntityResponse
+from lightly.openapi_generated.swagger_client.models.dataset_create_request import DatasetCreateRequest
+
+from lightly.openapi_generated.swagger_client.models.shared_access_config_create_request import SharedAccessConfigCreateRequest
+from lightly.openapi_generated.swagger_client.models.shared_access_config_data import SharedAccessConfigData
+from lightly.openapi_generated.swagger_client.models.shared_access_type import SharedAccessType
+
+from lightly.openapi_generated.swagger_client.models.dataset_data import DatasetData
+from lightly.openapi_generated.swagger_client.rest import ApiException
+from lightly.openapi_generated.swagger_client.models.dataset_type import DatasetType
+
+class _CollaborationMixin:
+
+    def share_dataset(self, dataset_id: str, user_emails: List[str]):
+        """Shares dataset with a list of users
+
+        This method overwrites the list of users that have had access to the dataset
+        before. If you want to add someone new to the list make sure you get the
+        list of users with access beforehand and add them as well.
+
+        Args:
+          dataset_id:
+            Identifier of dataset
+          user_emails:  
+            List of email addresses of users to grant write permission
+
+        Examples:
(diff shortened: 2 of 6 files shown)


lightly/api/api_workflow_collaboration.py
```diff
@@ -0,0 +1,76 @@
...
+          >>>
+          >>> # revoke access to all users
+          >>> client = ApiWorkflowClient(token="MY_AWESOME_TOKEN")
+          >>> user_emails = client.share_dataset(dataset_id="MY_DATASET_ID", user_emails=[])
+        """
+        body = SharedAccessConfigCreateRequest(access_type=SharedAccessType.WRITE, users=user_emails)
+        self._collaboration_api.create_or_update_shared_access_config_by_dataset_id(body=body, dataset_id=dataset_id)
+
+
+    def get_shared_users(self, dataset_id: str) -> List[str]:
```
japrescott, line 49: for future safety, allow one to possible to pass access_type but default to SharedAccessType.WRITE

lightly/api/api_workflow_collaboration.py
```diff
@@ -0,0 +1,76 @@
...
+from lightly.openapi_generated.swagger_client.models.shared_access_config_data import SharedAccessConfigData
+from lightly.openapi_generated.swagger_client.models.shared_access_type import SharedAccessType
+
+from lightly.openapi_generated.swagger_client.models.dataset_data import DatasetData
+from lightly.openapi_generated.swagger_client.rest import ApiException
+from lightly.openapi_generated.swagger_client.models.dataset_type import DatasetType
+
+class _CollaborationMixin:
+
+    def share_dataset(self, dataset_id: str, user_emails: List[str]):
```
japrescott, line 17: allow to pass in access_type but default to SharedAccessType.WRITE

japrescott, line 17: the naming of the function feels like one could "add to" the list. Maybe rename this one to `share_dataset_only_with` and add another fn which one could add more users (first get the old stuff and then concat) like `share_dataset` or `share_dataset_with`.

tests/api_workflow/test_api_workflow_collaboration.py
```diff
@@ -0,0 +1,17 @@
...
+class TestApiWorkflowDatasets(MockedApiWorkflowSetup):
+
+    def setUp(self) -> None:
+        self.api_workflow_client = MockedApiWorkflowClient(token="token_xyz")
+
+    def test_share_empty_dataset(self):
+        self.api_workflow_client.share_dataset(dataset_id="some-dataset-id", user_emails=[])
+
+    def test_share_dataset(self):
+        self.api_workflow_client.share_dataset(dataset_id="some-dataset-id", user_emails=["someone@something.com"])
```
japrescott, line 13: where is the validation?

tests/api_workflow/test_api_workflow_collaboration.py
```diff
@@ -0,0 +1,17 @@
+from tests.api_workflow.mocked_api_workflow_client import MockedApiWorkflowSetup, MockedApiWorkflowClient
+
+
+class TestApiWorkflowDatasets(MockedApiWorkflowSetup):
+
+    def setUp(self) -> None:
+        self.api_workflow_client = MockedApiWorkflowClient(token="token_xyz")
+
+    def test_share_empty_dataset(self):
+        self.api_workflow_client.share_dataset(dataset_id="some-dataset-id", user_emails=[])
```
japrescott, line 10: where is the validation?

tests/api_workflow/test_api_workflow_collaboration.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def test_share_dataset(self):
+        self.api_workflow_client.share_dataset(dataset_id="some-dataset-id", user_emails=["someone@something.com"])
+
+    def test_get_shared_users(self):
+        user_emails = self.api_workflow_client.get_shared_users(dataset_id="some-dataset-id")
```
japrescott, line 16: I think you should mock the return type here and not centrally

japrescott (dismissed):
looks great! Would love to have a "add these users to the shared list functionality" rather than always overriding!

lightly/api/api_workflow_collaboration.py
IgorSusmelj: I thought about that as well but then decided against it. I'd rather update the API specs as there is no use case for now to differentiate between read and write permission.

lightly/api/api_workflow_collaboration.py
IgorSusmelj: Same here. I'll create a follow up issue to discuss this.

IgorSusmelj: I would not add two different methods and just make it clear to the user what this method does. If we add a method to add individual users we should also add one to remove individual users.
I thought `share_dataset` is like sharing something once (so you would add everyone that should get access). But I see your point. I'll go with `share_dataset_only_with`

tests/api_workflow/test_api_workflow_collaboration.py
IgorSusmelj: Could you please explain what you mean with validation? I just followed the example tests we have for the dataset creation (see https://github.com/lightly-ai/lightly/blob/85e21e6008a7683ead1f6f7aaf6977412b9d7ddd/tests/api_workflow/test_api_workflow_datasets.py#L8)

lightly/api/api_workflow_collaboration.py
japrescott: changing the API specs is a breaking change. Then lets just pseudo default to write perm

lightly/api/api_workflow_collaboration.py
japrescott: it would just make this directly future proof without needing to touch the pip package.

tests/api_workflow/test_api_workflow_collaboration.py
japrescott: I might not be the best to explain this :D but as this endpoint is mocked we are essentially testing what we mocked and nothing else?
What I normally would do on the API is "share it", "get shared back", "check if shared is what i wanted to share it with", "share it with something else", "check if its changed".
Maybe you can mock the endpoint from the test itself and check that the API function was called with certain params but also that seems a bit ridiculous to do. @MalteEbner help :D

lightly/api/api_workflow_collaboration.py
japrescott: okeydokey

tests/api_workflow/test_api_workflow_collaboration.py
MalteEbner: @japrescott I assume you mean something like this:
https://gist.github.com/hectorcanto/ecc42cf5d5ad4ea309f51d21337d537c

Asking @michal-lightly how to do this best.

michal-lightly: I agree with Jeremy. Basically the test should do what is explained in the docstring (`lightly/api/api_workflow_collaboration.py:30`).

Regarding mocking, I don't know how APIWorkflowClient is mocked, so I can say just general things: Test the functionality that you added. If it has some expensive call that is mocked, you can check that the call has been made with the right parameters.

lightly/api/api_workflow_collaboration.py
IgorSusmelj: pseudo default --> what is in the PR or adding the default argument to use WRITE permission?

tests/api_workflow/test_api_workflow_collaboration.py
IgorSusmelj: We discussed this and Malte will organize a session on how to do the mocking properly.

docs/source/docker/getting_started/first_steps.rst
```diff
@@ -530,6 +530,52 @@ It is also possible to directly download the actual files themselves as follows:
...
+Once a dataset has been created we can also make it accessible to other users by
+sharing it. Sharing works through e-mail addresses. 
+
+.. code-block:: python
+    :caption: Share a dataset
+
+    # we first need to have an api client (create a new or use an existing one)
+    client = ApiWorkflowClient(token="MY_AWESOME_TOKEN")
+
+    # share a dataset with an user
```
MalteEbner, line 545: a user

docs/source/docker/getting_started/first_steps.rst
```diff
@@ -530,6 +530,52 @@ It is also possible to directly download the actual files themselves as follows:
...
+    client.share_dataset_only_with(dataset_id="MY_DATASET_ID", user_emails=["user@something.com"])
+
+    # share dataset with a user while keep sharing it with previous users
+    user_emails = client.get_shared_users(dataset_id="MY_DATASET_ID")
+    user_emails.extend(["additional_user2@something.com"])
```
MalteEbner, line 550: Easier code:
`user_emails.append("additional_user2@something.com")`

docs/source/docker/getting_started/first_steps.rst
```diff
@@ -530,6 +530,52 @@ It is also possible to directly download the actual files themselves as follows:
...
+this using the following code: 
+
+.. code-block:: python
+    :caption: Share a dataset
+
+    # we first need to have an api client (create a new or use an existing one)
+    client = ApiWorkflowClient(token="MY_AWESOME_TOKEN")
+
+    # get a list of users that have access to a given dataset
+    client.get_shared_users(dataset_id="MY_DATASET_ID")
```
MalteEbner, line 567: I would add an explicit print:
```python
users = client.get_shared_users(dataset_id="MY_DATASET_ID")
print(users)
```

docs/source/docker/getting_started/first_steps.rst
```diff
@@ -530,6 +530,52 @@ It is also possible to directly download the actual files themselves as follows:
...
+    # ["user@something.com"]
+
+You find more details about the individual API requests in the docs for the
+Python client: :py:class:`lightly.api.api_workflow_client`
```
MalteEbner, line 571: I am not sure why this is necessary. Do you want to linkt to all API requests in general?

Alternatively you can also link to specific methods:
`:py:meth:`get_shared_users <lightly.api.api_workflow_client.get_shared_users>``

lightly/api/api_workflow_datasets.py
```diff
@@ -1,7 +1,13 @@
+from os import access
 from typing import List
 
 from lightly.openapi_generated.swagger_client.models.create_entity_response import CreateEntityResponse
 from lightly.openapi_generated.swagger_client.models.dataset_create_request import DatasetCreateRequest
+
+from lightly.openapi_generated.swagger_client.models.shared_access_config_create_request import SharedAccessConfigCreateRequest
+from lightly.openapi_generated.swagger_client.models.shared_access_config_data import SharedAccessConfigData
+from lightly.openapi_generated.swagger_client.models.shared_access_type import SharedAccessType
+
```
MalteEbner, line 10: Why are these additional imports necessary?

lightly/api/api_workflow_collaboration.py
```diff
@@ -0,0 +1,76 @@
...
+from lightly.openapi_generated.swagger_client.models.create_entity_response import CreateEntityResponse
+from lightly.openapi_generated.swagger_client.models.dataset_create_request import DatasetCreateRequest
+
+from lightly.openapi_generated.swagger_client.models.shared_access_config_create_request import SharedAccessConfigCreateRequest
+from lightly.openapi_generated.swagger_client.models.shared_access_config_data import SharedAccessConfigData
+from lightly.openapi_generated.swagger_client.models.shared_access_type import SharedAccessType
+
+from lightly.openapi_generated.swagger_client.models.dataset_data import DatasetData
+from lightly.openapi_generated.swagger_client.rest import ApiException
+from lightly.openapi_generated.swagger_client.models.dataset_type import DatasetType
```
MalteEbner, line 13: Are all these imports necessary?

lightly/api/api_workflow_datasets.py
MalteEbner: I assume that they are a leftover from moving the code to a new file.

MalteEbner (commented):
This looks good to me! Some minor improvements are left, then this can be merged.

docs/source/docker/getting_started/first_steps.rst
IgorSusmelj: Yes, you're right. Let's drop it.

lightly/api/api_workflow_datasets.py
IgorSusmelj: Oh, my bad. Good catch!

MalteEbner (approved):
LGTM^2

PR #922, opened 2022-09-06
Jeremy lig 1745 ensure to retry all api calls and use

japrescott:
closes lig-1745
- new filename2readUrls mapping function `export_read_url_mapping_by_tag_name`
- paginate all export endpoints
- ~retry all endpoints~


michal-lightly:
Can you please separate the retry change from the rest (two PRs)? It will be much easier to review.

lightly/api/api_workflow_download_dataset.py
```diff
@@ -299,4 +305,53 @@ def export_filenames_by_tag_name(
...
         tag = self.get_tag_by_name(tag_name)
-        return self.export_filenames_by_tag_id(tag.id)
\ No newline at end of file
+        return self.export_filenames_by_tag_id(tag.id)    
+
+
+    def export_filenames_and_read_urls_by_tag_id(
+        self,
+        tag_id: str,
+    ) -> List[Dict]:
```
MalteEbner, line 314: This should be a `List[ReadURLMapping]`. Please be aware that a `ReadURLMapping` is a class, not a dict, despite the name.

lightly/api/api_workflow_download_dataset.py
```diff
@@ -299,4 +305,53 @@ def export_filenames_by_tag_name(
...
+            page_size=10000,
+            dataset_id=self.dataset_id,
+            tag_id=tag_id
+        )
+        return mappings
+
+    def export_filenames_and_read_urls_by_tag_name(
+        self,
+        tag_name: str,
+    ) -> List[Dict]:
```
MalteEbner, line 336: This should be a List[ReadURLMapping]. Please be aware that a `ReadURLMapping` is a class, not a dict, despite the name.

tests/api_workflow/mocked_api_workflow_client.py
```diff
@@ -92,13 +92,14 @@ def __init__(self, api_client):
                 is_processed=True,
```
MalteEbner, line 92: How did the autoformatting of this file happen? If you autoformat it, please use black.

MalteEbner (changes requested):
I like the general approach.

However, please be aware that the ReadUrlMapping is a class, not a dict. Thus the name is very misleading.

lightly/api/api_workflow_download_dataset.py
```diff
@@ -269,9 +274,10 @@ def export_filenames_by_tag_id(
             A list of the samples filenames within a certain tag.
 
         """
-        filenames = self._tags_api.export_tag_to_basic_filenames(
-            self.dataset_id,
-            tag_id,
+        filenames = retry(
```
MalteEbner, line 277: That retry is not necessary.

lightly/api/utils.py
```diff
@@ -64,6 +65,35 @@ def retry(func, *args, **kwargs):
                     f'Maximum retries exceeded! Original exception: {type(e)}: {str(e)}') from e
 
 
+
+def paginate_endpoint(fn, page_size=5000, *args, **kwargs) -> List:
```
MalteEbner, line 69: Great to have a general helper function or this.

MalteEbner (commented):
There is one retry not removed.

lightly/api/api_workflow_download_dataset.py
japrescott: why shouldn't it be necessary?

MalteEbner: I would remove the retry for two reasons:
- a) Consistency with all the other endpoints.
- b) Some of the endpoints are already retried by the docker. Thus double retries can happen, which can make the solution quickly unresponsive.

japrescott: I don't think "making it consistent" is the correct approach when the consistency is just plain wrong/stupid. Retrying things is good as the http request can fail for multiple reasons of which we can't influence many.
I don't see how a double retry can happen unless someone wraps an api_workflow_client function in a retry. Looking at the docker code, I don't see any place where that happens.

I don't know what the pythoneers plan is to address this issue or why you all are so totally against retrying stuff but just letting processes die after a couple of hours 'just because' is not a solution.

So you can choose
1. I will leave it this way
2. I will remove it as you want and also remove any retry logic of the pagination helper

MalteEbner: @japrescott You have convinced me, let's keep the retry logic here.

tests/api_workflow/mocked_api_workflow_client.py
japrescott: formatted it with black again

tests/api_workflow/mocked_api_workflow_client.py
```diff
@@ -214,14 +372,81 @@ def delete_tag_by_tag_id(self, dataset_id, tag_id, **kwargs):
...
+    def export_tag_to_basic_filenames_and_read_urls(
+        self, dataset_id: str, tag_id: str, **kwargs
+    ) -> List[FilenameAndReadUrl]:
+        if kwargs["page_offset"] and kwargs["page_offset"] > 0:
+            return []
+        return [
+            {
+                "fileName": "export-basic-test-sample-0.png",
+                "readUrl": "https://storage.googleapis.com/somwhere/export-basic-test-sample-0.png?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=CENSORED",
+            }
```
MalteEbner, line 446: Please don't return a `List[Dict]`, but a `List[FilenameAndReadUrl]` instead

tests/api_workflow/test_api_workflow_download_dataset.py
```diff
@@ -52,6 +52,10 @@ def test_export_label_studio_tasks_by_tag_name(self):
         self.assertIsNotNone(tasks)
         self.assertTrue(all(isinstance(task, dict) for task in tasks))
 
+    def test_export_tag_to_basic_filenames_and_read_urls(self):
+        mappings = self.api_workflow_client.export_filenames_and_read_urls_by_tag_name('initial-tag')
```
MalteEbner, line 56: Rename mappings to `"filenames_and_read_urls"`

tests/api_workflow/test_api_workflow_download_dataset.py
```diff
@@ -52,6 +52,10 @@ def test_export_label_studio_tasks_by_tag_name(self):
...
+        self.assertIsNotNone(mappings)
+        self.assertTrue(all(isinstance(mapping, dict) for mapping in mappings))
```
MalteEbner, line 58: Rather assert that it `isinstance(..., FilenameAndReadURL)` for every element in `filenames_and_read_urls`

tests/api_workflow/mocked_api_workflow_client.py
japrescott: can you please share code? I don't follow what you mean

MalteEbner, line 446:
```suggestion
            FilenameAndReadUrl(
            file_name="export-basic-test-sample-0.png", 
            read_url="https://storage.googleapis.com/somwhere/export-basic-test-sample-0.png?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=CENSORED",
            )
```

japrescott: thanks malte!

MalteEbner (approved):
This is great! Thanks that you also fixed other export functions.

PR #923, opened 2022-09-06
Expose selection_config_from_dict function

michal-lightly:
### What changed

Renames `_selection_config_from_dict` to `selection_config_from_dict` so that it can be used externally.

### How is it tested

CI tests pass


MalteEbner (approved):
I assume you used STRG-R, thus this should be fine.

PR #927, opened 2022-09-15
Remove model_output from Scorer

michal-lightly:
### What changed and why

Minor refactor - remove `model_output` attribute from `Scorer` class.

Scorer acts as a protocol class. Different implementations use `model_output` differently, in fact, each Scorer expects a different type. No function that gets a Scorer on input accessed `model_output`.

Also made `Scorer` an abstract class and discovered that `class_names()` method was not implemented for `ScorerSemanticSegmentation`.

### How is it tested

- Unit tests pass.
- Added a unit test for `ScorerSemanticSegmentation.class_names()`


MalteEbner (approved):
LGTM

PR #931, opened 2022-09-20
igor-lig-1671-similarity-search-docs

IgorSusmelj:
## Changes

- Add docs for similarity search
  - Docs added to the selection config
    - How to setup input using embeddings from another dataset
    - How to use similarity strategy
    - Added an example


guarin (commented):
Looks great! I think this is sufficient for now. We can always add a dedicated page explaining it more in detail. I would add one more sentence to explain more in detail what actually happens.

Another thing that is maybe not optimal is that there is no "Similarity Search" page in the Lightly Worker -> Advanced part of the docs. So it might be harder for users to discover that we support similarity search.

guarin (approved):
🚀

PR #932, opened 2022-09-21
Make create_dataset throw an error on existing datasets

guarin:
Changes:
* Add `dataset_name_exists` method
* Make `create_dataset` throw a ValueError if a dataset with the given name exists

Tested the code with:
* Add unit test
* Test manually

Note that this is a potentially breaking change for users and we should make sure to point this out in the release notes.

### DIFF
--- a/lightly/api/api_workflow_datasets.py
+++ b/lightly/api/api_workflow_datasets.py
@@ -83,14 +83,15 @@ class _DatasetsMixin:
     def create_dataset(self, dataset_name: str, dataset_type: Optional[str] = None):
         """Creates a dataset on the Lightly Platform..
 
-        Raises a ValueError if a dataset with the given name already exists.
-
         Args:
             dataset_name:
                 The name of the dataset to be created.
             dataset_type:
                 The type of the dataset. We recommend to use the API provided constants
                 `DatasetType.IMAGES` and `DatasetType.VIDEOS`.
+
+        Raises:
+            ValueError: If a dataset with dataset_name already exists.
 
         Examples:
             >>> from lightly.api import ApiWorkflowClient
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_datasets.py
```diff
@@ -100,13 +103,23 @@ def create_dataset(self, dataset_name: str, dataset_type: str = None):
             >>> client.create_dataset('your-dataset-name', dataset_type=DatasetType.VIDEOS)
         """
```
IgorSusmelj, line 104: Can you add `Raises: ` (following the style guide) here?

lightly/api/api_workflow_datasets.py
```diff
@@ -100,13 +104,23 @@ def create_dataset(self, dataset_name: str, dataset_type: str = None):
             >>> client.create_dataset('your-dataset-name', dataset_type=DatasetType.VIDEOS)
```
IgorSusmelj, line 104: Since we refactor this maybe also change this here to `dataset_id = client.create_dataset('your-dataset-name', dataset_type=DatasetType.VIDEOS)`

guarin: `create_dataset` does not return the new dataset id. The dataset id is set on the client:
```
client.create_dataset('your-dataset-name', dataset_type=DatasetType.VIDEOS)
dataset_id = client.dataset_id
```

guarin: I'll add some more explanations about this to the docstring.

IgorSusmelj: You're right! My fault

IgorSusmelj (approved):
Awesome!

PR #943, opened 2022-09-29
Add scores argument to ObjectDetectionOutput

guarin:
### What has changed and why?

Added an optional `scores` argument to the `ObjectDetectionOutput.__init__` method. The argument should be used when `scores` and `class_probabilities` are known when an instance of the class is created. This is different from the `ObjectDetectionOutput.from_scores` method which assumes that `class_probabilities` are unknown.

I decided to add it as an optional argument to not break backwards compatibility.

### How has it been tested?

Added unit tests.


michal-lightly (approved):
LGTM, thanks!

Issue #944, opened 2022-09-29
Add multiprototype-heads for SWAV


Atharva-Phatak:
I took a look at the implementation of SWAV lightly and the one in FAIR's repo (https://github.com/facebookresearch/swav/blob/main/src/resnet50.py). Lightly uses a single ```nn. Linear``` for prototypes but FAIR implementation has multiple prototype heads ? Any reason why lightly uses a single linear layer?

If not can we make it a Multi prototype head with multiple ```nn. Linear``` ? 

If so I can make a quick PR for these changes.

guarin:
Hi, yes you are completely right, we only added single prototype head. We added the single head because this is the default setting in SwAV. 

Adding a multiprototype head seems like a perfect addition to Lightly and a PR would be more than welcome :)

Atharva-Phatak:
Sure I will create a PR asap :)

PR #946, opened 2022-09-30
added multi-prototype head for SWav

Atharva-Phatak:
This PR implements a multi-prototype head for SWav as discussed in #944. 

Looking forward to feedback.

### DIFF
--- a/lightly/models/modules/heads.py
+++ b/lightly/models/modules/heads.py
@@ -3,7 +3,7 @@
 # Copyright (c) 2021. Lightly AG and its affiliates.
 # All Rights Reserved
 
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
@@ -370,11 +370,13 @@ class SwaVPrototypes(nn.Module):
 
     """
     def __init__(self,
-                input_dim: int,
-                n_prototypes: List[int]):
+                input_dim: int = 128,
+                n_prototypes: Union[List[int], int] = 3000):
         super(SwaVPrototypes, self).__init__()
-        self.number_of_heads = len(n_prototypes)
-        for layerNum, k in enumerate(self.number_of_heads):
+        self.number_of_heads = len(n_prototypes) if isinstance(n_prototypes, list) else 1
+        #Default to a list of 1 if n_prototypes is an int.
+        self.n_prototypes = n_prototypes if isinstance(n_prototypes, list) else [n_prototypes]
+        for layerNum, k in enumerate(n_prototypes):
             self.add_module("prototypes" + str(layerNum), nn.Linear(input_dim, k, bias = False))
 
     def forward(self, x):
--- a/lightly/models/modules/heads.py
+++ b/lightly/models/modules/heads.py
@@ -373,17 +373,16 @@ class SwaVPrototypes(nn.Module):
                 input_dim: int = 128,
                 n_prototypes: Union[List[int], int] = 3000):
         super(SwaVPrototypes, self).__init__()
-        self.number_of_heads = len(n_prototypes) if isinstance(n_prototypes, list) else 1
         #Default to a list of 1 if n_prototypes is an int.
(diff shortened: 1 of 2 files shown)


Atharva-Phatak:
Perfect the tests are working. Looking forward to the feedback and code reviews

lightly/models/modules/heads.py
```diff
@@ -370,14 +370,20 @@ class SwaVPrototypes(ProjectionHead):
...
+        #Default to a list of 1 if n_prototypes is an int.
+        self.n_prototypes = n_prototypes if isinstance(n_prototypes, list) else [n_prototypes]
+        for layerNum, k in enumerate(self.n_prototypes):
+            self.add_module("prototypes" + str(layerNum), nn.Linear(input_dim, k, bias = False))
+
+    def forward(self, x):
+        out = []
+        for head in range(self.number_of_heads):
+            out.append(getattr(self, "prototypes" + str(head))(x))
+        return out
```
guarin, line 386: This changes the behaviour of  the `SwaVPrototypes` head and might break existing code. Before this change it returned a single tensor and now it returns a list of tensors.
 
I would propose to keep the old behaviour if users specify `n_prototypes` with a single number. If `n_prototypes` is a list, the forward function should also return a list. This could be implemented by adding a `self.single_head = isinstance(n_prototypes, int)` line in the `__init__` method and checking the property in the forward function.

lightly/models/modules/heads.py
```diff
@@ -370,14 +370,20 @@ class SwaVPrototypes(ProjectionHead):
...
     """
     def __init__(self,
-                 input_dim: int = 128,
-                 n_prototypes: int = 3000):
-        super(SwaVPrototypes, self).__init__([])
-        self.layers = nn.Linear(input_dim, n_prototypes, bias=False)
-
-    def normalize(self):
-        """Normalizes the prototypes so that they are on the unit sphere."""
-        utils.normalize_weight(self.layers.weight)
```
guarin, line 380: I don't think this should be deleted. Instead we should loop over all heads and normalise them.

guarin (changes requested):
Thanks a lot for the PR 💯 

The code looks great. I added some comments and suggestions, mostly regarding backwards compatibility.

lightly/models/modules/heads.py
```diff
@@ -370,14 +370,20 @@ class SwaVPrototypes(ProjectionHead):
...
+                input_dim: int = 128,
+                n_prototypes: Union[List[int], int] = 3000):
+        super(SwaVPrototypes, self).__init__()
+        self.number_of_heads = len(n_prototypes) if isinstance(n_prototypes, list) else 1
+        #Default to a list of 1 if n_prototypes is an int.
+        self.n_prototypes = n_prototypes if isinstance(n_prototypes, list) else [n_prototypes]
+        for layerNum, k in enumerate(self.n_prototypes):
+            self.add_module("prototypes" + str(layerNum), nn.Linear(input_dim, k, bias = False))
```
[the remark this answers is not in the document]
guarin: I think we should use the builtin functionality provided by pytorch to register new modules.

lightly/models/modules/heads.py
Atharva-Phatak: Hey @guarin. So just to clarify, the output should be a tensor if n_prototypes is an int else, return a list. I think this should be easy to implement. Am I correct ?

guarin: Thanks for implementing the feature :)

lightly/models/modules/heads.py
guarin: I think this is the only thing that is missing then we should be good to merge :)

guarin (commented):
Thanks a lot for updating the PR. The only thing that is missing is the normalisation, then it should be ready to merge :)

Atharva-Phatak:
Added the normalisation.

guarin (approved):
Great, thanks a lot 💯

PR #948, opened 2022-10-04
Fix typehints for export functions

philippmwirth:
# Fix typehints for export functions

The typehints for the export functions are inaccurate because swagger codegen can't handle arrays of types correctly.

### DIFF
--- a/lightly/api/api_workflow_download_dataset.py
+++ b/lightly/api/api_workflow_download_dataset.py
@@ -157,7 +157,7 @@ class _DownloadDatasetMixin:
     def export_label_studio_tasks_by_tag_id(
         self,
         tag_id: str,
-    ) -> List[LabelStudioTask]:
+    ) -> List[Dict]:
         """Exports samples in a format compatible with Label Studio.
 
         The format is documented here:
@@ -182,7 +182,7 @@ class _DownloadDatasetMixin:
     def export_label_studio_tasks_by_tag_name(
         self,
         tag_name: str,
-    ) -> List[LabelStudioTask]:
+    ) -> List[Dict]:
         """Exports samples in a format compatible with Label Studio.
 
         The format is documented here:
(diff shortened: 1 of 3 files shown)


lightly/api/api_workflow_download_dataset.py
```diff
@@ -211,7 +211,7 @@ def export_label_studio_tasks_by_tag_name(
     def export_label_box_data_rows_by_tag_id(
         self,
         tag_id: str,
-    ) -> List[LabelBoxDataRow]:
+    ) -> List[Dict]:
```
japrescott, line 214: @MalteEbner made me add this all over the place.. if you really need to remove these, you'll also need to remove them from the tests and mocked stuff

japrescott (changes requested):

michal-lightly:
I have a few questions (but maybe someone has more context and can approve).
1. What is the error/problem caused by the incorrect types?
2. `LabelBoxDataRow` seems more concrete than `Dict`. Why is Dict the proper type?

It would also be nice to specify the Dict type (Dict[str, Any]?)

philippmwirth:
> I have a few questions (but maybe someone has more context and can approve).
> 
>     1. What is the error/problem caused by the incorrect types?
> 
>     2. `LabelBoxDataRow` seems more concrete than `Dict`. Why is Dict the proper type?
> 
> 
> It would also be nice to specify the Dict type (Dict[str, Any]?)

There's no error caused by the incorrect types other than that the information provided is faulty because the functions actually return dictionaries. This should also answer your second question: Dict is the proper type because that's what the function actually returns. An alternative solution would be to define the types `LabelBoxDataRows` as `Dict[str, Any]` but then there's ambiguity because `LabelBoxDataRows`  actually exists as a type (it's just not used).

japrescott (approved):

michal-lightly (approved):
Lightly PIP package is not type-checked anyway.

PR #949, opened 2022-10-04
Igor lig 1800 improve error messages about worker

IgorSusmelj:
## Changes

- Better error messages when monitoring Lightly Worker
- Add info that there is a `logs.txt` file with useful debugging info to the docs


japrescott (approved):
awesome stuff!

PR #953, opened 2022-10-11
Allow self signed certificates

philippmwirth:
# Allow self signed certificates

- Refactor the way a `Configuration` object is created to make it easily reusable.
- Allow passing of self-signed certificate bundles to the `ApiWorkflowClient`.

### DIFF
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -293,7 +293,7 @@ def set_api_client_request_timeout(
 
 
 def get_api_client_configuration(
-    token: Optional[str]
+    token: Optional[str] = None,
 ) -> Configuration:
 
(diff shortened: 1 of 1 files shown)


lightly/api/api_workflow_client.py
```diff
@@ -294,3 +290,25 @@ def new_request_fn(*args, **kwargs):
...
+    if token is None:
+        token = getenv("LIGHTLY_TOKEN", None)
+    if token is None:
+        raise ValueError(
+            "Either provide a 'token' argument or export a LIGHTLY_TOKEN environment variable"
+        )
+
+    configuration = Configuration()
+    configuration.api_key = {"token": token}
+    configuration.ssl_ca_cert = ssl_ca_cert
```
japrescott, line 311: I assume `None` is the default val of `ssl_ca_cert`, right?

japrescott (approved):
very well done señor

PR #954, opened 2022-10-11
Guarin lig 1719 Add subdir info to prediction and metadata docs

guarin:
* Change examples to use subdirectories by default

[Add Predictions to a Datasource — lightly 1.2.31 documentation.pdf](https://github.com/lightly-ai/lightly/files/9754418/Add.Predictions.to.a.Datasource.lightly.1.2.31.documentation.pdf)
[Add Metadata to a Datasource — lightly 1.2.31 documentation.pdf](https://github.com/lightly-ai/lightly/files/9754420/Add.Metadata.to.a.Datasource.lightly.1.2.31.documentation.pdf)

### DIFF
--- a/docs/source/docker/advanced/datasource_predictions.rst
+++ b/docs/source/docker/advanced/datasource_predictions.rst
@@ -29,24 +29,30 @@ of your input and output bucket will look like this:
 .. code-block:: bash
 
     input/bucket/
-        + image_1.png
-        + image_2.png
-        + ...
-        + image_N.png
+        + image_0.png
+        + subdir/
+            + image_1.png
+            + image_2.png
+            + ...
+            + image_N.png
   
     output/bucket/
         + .lightly/predictions/
             + tasks.json
             + task_1/
                  + schema.json
-                 + image_1.json
-                 ...
-                 + image_N.json
+                 + image_0.json
+                 + subdir/
+                    + image_1.json
+                    ...
+                    + image_N.json
             + task_2/
                  + schema.json
-                 + image_1.json
-                 ...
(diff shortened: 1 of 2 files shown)


docs/source/docker/advanced/datasource_predictions.rst
```diff
@@ -169,14 +175,14 @@ it's necessary to follow the naming convention:
 
 .. code-block:: bash
 
-    # filename of the prediction for image FILENAME.EXT
-    .lightly/predictions/${TASK_NAME}/${FILENAME}.json
+    # filename of the prediction for image input/bucket/SUBDIR/FILENAME.EXT
+    .lightly/predictions/${TASK_NAME}/${SUBDIR}/${FILENAME}.json
```
IgorSusmelj, line 179: I'm not sure about the `${SUBDIR}` notion. It might raise the question if multiple subdirs are also allowed or not (hierarchy). I think with using subdirs in the examples this is already much better. And I'd not add the `${SUBDIR}`.

IgorSusmelj (approved):
The changes look good. Thanks! It now almost appears a bit too strict (as if you must use a subdir and then the question pops up whether you can have hierarchies or not). But maybe that's just my perception. I made a small suggestion for a change (remove all the `${SUBDIR}`). But you can also keep them if you think it makes sense :)

docs/source/docker/advanced/datasource_predictions.rst
guarin: I agree, the `$SUBDIR` is a bit confusing and the other examples should be enough. Thanks for pointing it out!

PR #957, opened 2022-10-12
Update datapool docs

shaundaley39:
- separate tabs for AWS, GCS and Azure datasources
- handling case where dataset already exists - given client.create_dataset functionality has changed

Before:
Screenshot from 2022-10-12 15-30-22

After, we have two separate sections:
- first, dataset creation and datasource configuration (split across 3 tabs: AWS S3, GCS, Azure)
- second, we fetch the dataset by name and run a selection task against it

Create dataset (3 tabs):
Screenshot from 2022-10-12 16-35-39
Screenshot from 2022-10-12 16-35-44
Screenshot from 2022-10-12 16-35-50
Fetch dataset, run task (with stopping condition of 0.1):
Screenshot from 2022-10-17 11-46-34
Screenshot from 2022-10-17 11-46-49
Then with stopping condition of 0.2:
Screenshot from 2022-10-17 11-47-02
Screenshot from 2022-10-17 11-47-09

If this is approved, there are a bunch of other docs we would probably want to update in the same way, listed here:
https://linear.app/lightly/issue/LIG-1857#comment-0bf7e8ef


docs/source/docker/advanced/datapool.rst
```diff
@@ -67,6 +67,27 @@ first time.
     |-- passageway1-c1.avi
     `-- terrace1-c0.avi
 
+Let's create a Lightly dataset which uses that bucket (choose your tab - S3, GCS or Azure):
+
+.. tab-set::
+  .. tab-item:: AWS S3 Datasource
```
guarin, line 73: Could you use `.. tabs::` and `... tab::` instead of `.. tab-set::` and `.. tab-item::` just to be consistent with the rest of the docs.

docs/source/docker/advanced/datapool.rst
```diff
@@ -67,6 +67,27 @@ first time.
...
+
+  .. tab-item:: Azure Datasource
+
+    .. literalinclude:: ./code_examples/python_create_dataset_azure_example.py
+      :linenos:
+      :language: python
+
 Now we can run the following code to select a subset based on the 
 :code:`'stopping_condition_minimum_distance': 0.1` stopping condition. In a first,
 selection run we only select images with the specific minimum distance between
```
guarin, line 93: Maybe add a note below this paragraph (or the code snippet) that we use `client.set_dataset_id_by_name` instead of `client.create_dataset` to allow running the snippet multiple times on the same dataset. And that `create_dataset` will throw an error if a dataset with this name already exists.

You could also highlight the `client.set_dataset_id_by_name` line with:
```
.. code-block:: python
    :emphasize-lines: 8
```

guarin, line 93: The sentence on line 122 is a bit ambiguous now because it is no longer clear what "the script" is. Maybe rephrase the sentence to make it clear that the dataset creation should not be run again.

guarin (commented):
I really like the change! I added some comments/suggestions.

docs/source/docker/advanced/datapool.rst
guarin: Actually, I would maybe even copy/paste the script again and increase the `stopping_condition_minimum_distance` to `0.2` (highlight the change!) which should make it very clear what should be run again.

shaundaley39: done

docs/source/docker/advanced/datapool.rst
shaundaley39: done

docs/source/docker/advanced/datapool.rst
shaundaley39: nice suggestion! Done

guarin (approved):

PR #967, opened 2022-10-27
ensure set_embedding_id_to_latest sets the newest

japrescott:
closes lig-1971
- ensure `set_embedding_id_to_latest` actually sets to the latest

### DIFF
--- a/lightly/api/api_workflow_upload_embeddings.py
+++ b/lightly/api/api_workflow_upload_embeddings.py
@@ -41,7 +41,10 @@ class _UploadEmbeddingsMixin:
             self._embeddings_api.get_embeddings_by_dataset_id(
                 dataset_id=self.dataset_id
             )
-        self.embedding_id = embeddings_on_server[-1].id
+        if len(embeddings_on_server) == 0:
+            raise RuntimeError(f"There are no known embeddings for dataset_id {self.dataset_id}.")
+        # return first entry as the API returns newest first
+        self.embedding_id = embeddings_on_server[0].id
 
     def get_embedding_by_name(
             self, name: str, ignore_suffix: bool = True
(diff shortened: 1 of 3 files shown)


lightly/api/api_workflow_upload_embeddings.py
```diff
@@ -41,7 +41,10 @@ def set_embedding_id_to_latest(self):
             self._embeddings_api.get_embeddings_by_dataset_id(
                 dataset_id=self.dataset_id
             )
-        self.embedding_id = embeddings_on_server[-1].id
+        if len(embeddings_on_server) == 0:
+            raise RuntimeError(f"There are no known embeddings for dataset_id {self.dataset_id}.")
+        # return first entry as the API returns newest first
+        self.embedding_id = embeddings_on_server[0].id
```
IgorSusmelj, line 47: The entries we get from the API are sorted by what? (Not timestamp and latest entry is last?)

IgorSusmelj (approved):
LGTM. I assume you doubled checked what the API returns and in which order?

lightly/api/api_workflow_upload_embeddings.py
japrescott: `.const embeddings = await db.collection<IMongoEmbeddingSummary>('embeddingSummaries').find({ datasetId: toMongoId(datasetId) }).sort({ createdAt: -1 }).toArray();` thats how its called in the DB.

Don't know how to test it properly. Maybe you could upload multiple embeddings and verify that it always returns the first created embedding? Then we know this works as expected

MalteEbner: I would unittest this in the typescript tests an ensure it there.

MalteEbner (approved):
The change looks good to me, great that this error case now has a proper error message.

lightly/api/api_workflow_upload_embeddings.py
MalteEbner: Edit: I would unittest the order in typescript.

japrescott: verified

PR #971, opened 2022-11-02
Malte lig 1897 implement run worker label matching

MalteEbner:
## What has changed?
- generated from the new spec:
  - `get_docker_runs_scheduled_by_state_and_labels` with `labels`
  - `DockerRunScheduledCreateRequest` with `runs_on`
  - `DockerRunScheduledData` with `runs_on`
  - `DockerRunScheduledUpdateRequest`with `runs_on`
- `ApiWorkflowClient.schedule_compute_worker_run`: added `runs_on` argument
- `ApiWorkflowClient.register_compute_worker`: added `labels` argument
- mock the `urllib3.request.urlencode` to use `doseq=True`. Otherwise using arrays in the query fails. This is needed for `get_docker_runs_scheduled_by_state_and_labels`

## How was it tested? + corresponding lightly-core PR
see https://github.com/lightly-ai/lightly-core/pull/2037

### DIFF
--- a/lightly/api/__init__.py
+++ b/lightly/api/__init__.py
@@ -5,3 +5,10 @@
 
 from lightly.api.api_workflow_client import ApiWorkflowClient
 from lightly.api.api_workflow_compute_worker import ArtifactNotExist
+
+# Mock the urlencode to use doseq, so that arrays in the query are properly encoded
+from six.moves.urllib.parse import urlencode
+def doseq_urlencode(fields):
+    return urlencode(fields, doseq=True)
+import urllib3.request
+urllib3.request.urlencode = doseq_urlencode
--- a/lightly/api/__init__.py
+++ b/lightly/api/__init__.py
@@ -5,10 +5,9 @@
 
 from lightly.api.api_workflow_client import ApiWorkflowClient
 from lightly.api.api_workflow_compute_worker import ArtifactNotExist
+from lightly.api.patch_rest_client import patch_rest_client
 
-# Mock the urlencode to use doseq, so that arrays in the query are properly encoded
-from six.moves.urllib.parse import urlencode
-def doseq_urlencode(fields):
-    return urlencode(fields, doseq=True)
-import urllib3.request
-urllib3.request.urlencode = doseq_urlencode
+from lightly.openapi_generated.swagger_client.rest import RESTClientObject
+
+# Needed to handle list of arguments correctly
+patch_rest_client(RESTClientObject)
--- a/lightly/api/api_workflow_compute_worker.py
+++ b/lightly/api/api_workflow_compute_worker.py
@@ -68,7 +68,7 @@ class ComputeWorkerRunInfo:
 
 
 class _ComputeWorkerMixin:
-    def register_compute_worker(self, name: str = "Default", labels: List[str] = None) -> str:
+    def register_compute_worker(self, name: str = "Default", labels: Optional[List[str]] = None) -> str:
         """Registers a new compute worker.
 
         Args:
@@ -147,7 +147,7 @@ class _ComputeWorkerMixin:
         lightly_config: Optional[Dict[str, Any]] = None,
         selection_config: Optional[Union[Dict[str, Any], SelectionConfig]] = None,
         priority: str = DockerRunScheduledPriority.MID,
-        runs_on=None
+        runs_on: Optional[List[str]] = None
     ) -> str:
         """Schedules a run with the given configurations.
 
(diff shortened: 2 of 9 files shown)


lightly/api/__init__.py
```diff
@@ -5,3 +5,10 @@
 
 from lightly.api.api_workflow_client import ApiWorkflowClient
 from lightly.api.api_workflow_compute_worker import ArtifactNotExist
+
+# Mock the urlencode to use doseq, so that arrays in the query are properly encoded
+from six.moves.urllib.parse import urlencode
```
japrescott, line 10: will this not be overwritten when we generate the code again?

tests/api_workflow/test_api_workflow_compute_worker.py
```diff
@@ -368,6 +370,7 @@ def test_get_compute_worker_state_and_message_OPEN() -> None:
         state=DockerRunScheduledState.OPEN,
         created_at=0,
         last_modified_at=1,
+        runs_on=[6]
```
japrescott, line 373: I am going to put on my Malte hat and ask you to write tests :)

lightly/api/__init__.py
MalteEbner: This is not part of the generated code. The generated code is only under `lightly/openapi_generated`

lightly/api/__init__.py
```diff
@@ -5,3 +5,10 @@
...
+def doseq_urlencode(fields):
+    return urlencode(fields, doseq=True)
+import urllib3.request
+urllib3.request.urlencode = doseq_urlencode
```
guarin, line 14: This is pretty hacky as it changes urllib3 under the hood. A less intrusive alternative is to patch the requests method on the rest client:
```
from lightly.openapi_generated.swagger_client.rest import RESTClientObject

def patch_rest_client(rest_client):
    request = rest_client.request

    def request_patched(self, method, url, query_params=None, headers=None,
                    body=None, post_params=None, _preload_content=True,
                    _request_timeout=None):
        if query_params is not None:
            new_query_params = []
            for name, value in query_params:
                if isinstance(value, list):
                    new_query_params.extend([(name, val) for val in value])
                else:
                    new_query_params.append((name, value))
            query_params = new_query_params
        return request(self, method=method, url=url, query_params=query_params, headers=headers, body=body, post_params=post_params, _preload_content=_preload_content, _request_timeout=_request_timeout)

    rest_client.request = request_patched

patch_rest_client(RESTClientObject)
```

What it does is convert query parameters that look like this: `[('labels', ['A', 'B', 'C'])]` into a list of simple parameters: `[('labels', 'A'), ('labels', 'B'), ('labels', 'C')]` which is supported by swagger codegen.

tests/api_workflow/test_api_workflow_compute_worker.py
MalteEbner: Good point, I added them in the commit `added tests for schedule_compute_worker_run`

guarin (commented):
Looks already pretty good. Added some comments :)

lightly/api/__init__.py
MalteEbner: Thanks, good idea!
However, this still messes up with a library, but luckily only a generated one.

lightly/api/api_workflow_compute_worker.py
```diff
@@ -143,6 +147,7 @@ def schedule_compute_worker_run(
         lightly_config: Optional[Dict[str, Any]] = None,
         selection_config: Optional[Union[Dict[str, Any], SelectionConfig]] = None,
         priority: str = DockerRunScheduledPriority.MID,
+        runs_on=None
```
guarin, line 150: Type missing.

lightly/api/patch_rest_client.py
```diff
@@ -0,0 +1,20 @@
+from typing import Type
+
+
+def patch_rest_client(rest_client: Type):
```
guarin, line 4: Please add a short description what this does and why we need it.

lightly/api/api_workflow_compute_worker.py
```diff
@@ -68,19 +68,23 @@ def ended_successfully(self) -> bool:
 
 
 class _ComputeWorkerMixin:
-    def register_compute_worker(self, name: str = "Default") -> str:
+    def register_compute_worker(self, name: str = "Default", labels: List[str] = None) -> str:
```
guarin, line 71: Type should be `Optional[List[str]]`

guarin (commented):
Great! Just some minor things missing.

guarin (approved):
💯

PR #974, opened 2022-11-09
Use correct configuration everywhere

philippmwirth:
# Use correct configuration everywhere

This should solve the issues with the CA certificate.

### DIFF
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -294,6 +294,7 @@ def set_api_client_request_timeout(
 
 def get_api_client_configuration(
     token: Optional[str] = None,
+    raise_if_no_token_specified: bool = True,
 ) -> Configuration:
 
     host = getenv("LIGHTLY_SERVER_LOCATION", "https://api.lightly.ai")
@@ -301,7 +302,7 @@ def get_api_client_configuration(
 
     if token is None:
         token = getenv("LIGHTLY_TOKEN", None)
-    if token is None:
+    if token is None and raise_if_no_token_specified:
         raise ValueError(
             "Either provide a 'token' argument or export a LIGHTLY_TOKEN environment variable"
         )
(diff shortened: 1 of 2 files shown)


lightly/api/version_checking.py
```diff
@@ -52,7 +49,7 @@ def do_version_check(current_version: str):
 
 
 def get_versioning_api() -> VersioningApi:
-    configuration = Configuration()
+    configuration = api_workflow_client.get_api_client_configuration_no_token()
     configuration.host = getenv('LIGHTLY_SERVER_LOCATION', 'https://api.lightly.ai')
```
japrescott, line 53: this line is not needed anymore

lightly/api/api_workflow_client.py
```diff
@@ -311,4 +311,8 @@ def get_api_client_configuration(
     configuration.ssl_ca_cert = ssl_ca_cert
     configuration.host = host
 
-    return configuration
\ No newline at end of file
+    return configuration
+
+
+def get_api_client_configuration_no_token() -> Configuration:
+    return get_api_client_configuration(token="dummy_token")
```
japrescott, line 318: call this "no_token_specified" or something

lightly/api/version_checking.py
```diff
@@ -52,7 +49,7 @@ def do_version_check(current_version: str):
 
 
 def get_versioning_api() -> VersioningApi:
-    configuration = Configuration()
+    configuration = api_workflow_client.get_api_client_configuration_no_token()
```
japrescott, line 52: do we really not have the token here already? We should unless this code is called directly, no? Couldn't we rather pass a flag to not raise an exception in `get_api_client_configuration`?

japrescott (changes requested):

lightly/api/version_checking.py
philippmwirth: > do we really not have the token here already? 

No, we don't have a token.

> Couldn't we rather pass a flag to not raise an exception in get_api_client_configuration?

We could 👍

japrescott (approved):

Issue #779, opened 2022-04-26
VICReg: Variance-Invariance-Covariance Regularization for Self-Supervised Learning


guarin:
## VICReg: Variance-Invariance-Covariance Regularization for Self-Supervised Learning

21.05.2021
https://arxiv.org/abs/2105.04906
https://github.com/facebookresearch/vicreg

New model similar to [Barlow Twins](https://arxiv.org/abs/2103.03230) but combines three losses: invariance, variance, and covariance. Uses a fully symmetric model architecture and does not use feature normalization.

![Screenshot 2022-04-26 at 09 20 24](https://user-images.githubusercontent.com/43336610/165244231-030a7110-862a-487a-97be-a100bd1ee481.png)

Estimated effort to implement in Lightly: Low
* Add the three losses

guarin:
Closed by #989

PR #979, opened 2022-11-16
Add VICReg loss

b-albar:
It adds a VICReg loss that is described in this [paper](https://arxiv.org/abs/2105.04906) also as proposed in #779 
The implementation follows the official code, [here](https://github.com/facebookresearch/vicreg). In term of architecture, it's similar to BarlowTwins, so there is no need for a new projection head.

I would appreciate any feedback on this.

### DIFF
--- a/lightly/loss/vicreg_loss.py
+++ b/lightly/loss/vicreg_loss.py
@@ -1,0 +1,98 @@
+import torch
+import torch.distributed as dist
+import torch.nn.functional as F
+
+class VICRegLoss(torch.nn.Module):
+    """Implementation of the VICReg Loss from VICReg[0] paper.
+    This implementation follows the code published by the authors. [1]
+
+    [0] Bardes, A. et. al, 2022, VICReg... https://arxiv.org/abs/2105.04906
+    [1] https://github.com/facebookresearch/vicreg/
+        Examples:
+        >>> # initialize loss function
+        >>> loss_fn = VICRegLoss()
+        >>>
+        >>> # generate two random transforms of images
+        >>> t0 = transforms(images)
+        >>> t1 = transforms(images)
+        >>>
+        >>> # feed through model
+        >>> out0, out1 = model(t0, t1)
+        >>>
+        >>> # calculate loss
+        >>> loss = loss_fn(out0, out1)
+    """
+
+    def __init__(
+        self,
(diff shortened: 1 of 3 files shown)


philippmwirth:
Thanks for the contribution @b-albar! We'll take a look shortly 🙂

lightly/loss/vicreg_loss.py
```diff
@@ -0,0 +1,98 @@
...
+    def forward(self, z_a: torch.Tensor, z_b: torch.Tensor) -> torch.Tensor:
+
+        # invariance term of the loss
+        repr_loss = F.mse_loss(z_a, z_b)
+
+        # gather all batches
+        if self.gather_distributed and dist.is_initialized():
+            world_size = dist.get_world_size()
+            if world_size > 1:
+                dist.all_reduce(z_a)
```
philippmwirth, line 69: I believe this should be `all_gather` rather than `all_reduce`. What do you think @b-albar  and @guarin?

philippmwirth (commented):
Thanks a lot for the contribution! I think once we fix the distributed gathering we can merge it 🙂 

@b-albar please let us know if you want to do it yourself or if you want someone to take over the PR.

lightly/loss/vicreg_loss.py
b-albar: Yes I agree it should be all_gather. But in this case, I think it should use the GatherLayer in lightly.utils.dist to avoid breaking the gradient flow during backprop. I'll fix the PR.

guarin (approved):

guarin (changes requested):
Hi, thanks a lot for the PR and for fixing the gathering operation!

We had some issues with the CI tests not running through because setuptools was missing. This has now been fixed on the master branch. I just tried your code with the updates from the master branch but now the vicreg tests fail because `nan` is returned.

New branch I made: https://github.com/lightly-ai/lightly/pulls
Failing tests: https://github.com/lightly-ai/lightly/actions/runs/3530335397/jobs/5922217442

Do you know what the issue for the nan could be? And could you also merge our master branch back into your branch to apply the latest changes?

guarin:
Hi @b-albar I managed to fix the bug in the tests (batch size must be at least 2) and merged the changes in #989. Thanks a ton for your contribution!

PR #985, opened 2022-11-22
Malte lig 1302 add prediction upload helpers

MalteEbner:
## Description

Added two new endpoints:
- `create_or_update_prediction_task_schema`
- `create_or_update_predictions`

The latter takes a mapping from filenames to their predictions as input and uploads the predictions for all samples.

## Open questions

To create the prediction singletons properly, I created the file `prediction_singletons.py` and already added the `PredictionSingletonClassificationRepr` to it. They allow to create the Singletons properly, as the openapi-generated singletons are useless, because the API specs use Inheritance and `OneOf`.

Should I follow this and also create the singletons for the other task types?

## Tests

- unittests
- direct test: Worked successfully, had 70 uploads per. See the docstring examples for how to run it.
```
Creating prediction schema
Uploading predictions
Uploading predictions: 100%|██████████| 356/356 [00:04<00:00, 78.84 predictions/s] 
```
## PROBLEM with UNITTEST `assert_has_calls`

Somehow `test_create_or_update_predictions` always fails for me with 
```
AssertionError: 'create_or_update_prediction' does not contain all of (call('', (<MagicMock spec='ApiWorkflowClient' id='140549846505408'>, 'sample_id_0', [<lightly.api.prediction_singletons.PredictionSingletonClassificationRepr object at 0x7fd44d87dbe0>], 1669297116), {}), call('', (<MagicMock spec='ApiWorkflowClient' id='140549846505408'>, 'sample_id_1', [<lightly.api.prediction_singletons.PredictionSingletonClassificationRepr object at 0x7fd44fac90d0>], 1669297116), {}), call('', (<MagicMock spec='ApiWorkflowClient' id='140549846505408'>, 'sample_id_2', [<lightly.api.prediction_singletons.PredictionSingletonClassificationRepr object at 0x7fd44fac9040>], 1669297116), {}), call('', (<MagicMock spec='ApiWorkflowClient' id='140549846505408'>, 'sample_id_3', [<lightly.api.prediction_singletons.PredictionSingletonClassificationRepr object at 0x7fd44fac9130>], 1669297116), {})) in its call list, found [TypeError("missing a required argument: 'self'"), TypeError("missing a required argument: 'self'"), TypeError("missing a required argument: 'self'"), TypeError("missing a required argument: 'self'")] instead
```


lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,155 @@
...
+        schema: PredictionTaskSchema,
+        prediction_uuid_timestamp: int,
+    ) -> None:
+        """Creates or updates the prediction task schema
+
+        Args:
+            schema:
+                The prediction task schema.
+            prediction_uuid_timestamp:
+                The uuid timestamp in nanoseconds since 1970. Get it via 'int(time.time())'.
```
japrescott, line 30: its important to know that this is the unique identifier (that is also a timestamp) and not just a random timestamp. If its nanoseconds, milliseconds or seconds is very irrelevant. The most important aspect is that this the `key` for everything happening around the schemas and the singletons.

lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,155 @@
...
+        self._predictions_api.create_or_update_prediction_task_schema_by_dataset_id(
+            body=schema,
+            dataset_id=self.dataset_id,
+            prediction_uuid_timestamp=prediction_uuid_timestamp,
+        )
+
+
+    def create_or_update_predictions(
+        self,
+        filename_to_prediction_singletons: Union[
```
japrescott, line 69: AFAIK this should be a `Dict[str, List[Union[PredictionSingletonClassificationRepr, PredictionSingletonObjectDetection, PredictionSingletonInstanceSegmentation, PredictionSingletonKeypointDetection]]`, no? We have singletons for each file and these singletons can be mixed if multiple tasks are done, no?

lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,155 @@
...
+          >>> client.create_or_update_predictions(
+          >>>     filename_to_prediction_singletons_dummy,
+          >>>     prediction_uuid_timestamp=timestamp,
+          >>>     progress_bar=tqdm(desc="Uploading predictions", total=len(filenames), unit=" predictions")
+          >>> )
+
+
+        """
+
+        samples = retry(
```
japrescott, line 126: I see as this is a pip function and thus it makes sense to check this for sanity reasons. However I am not a big fan of querying the server again when we should already have this info. Maybe we can allow to skip this (when using from lightlyworker) somehow? Or we pass the dictionary instead of `filename<->singleton[]` as `sampleId<->singleton[]` instead?

lightly/api/api_workflow_predictions.py
MalteEbner: I changed the docstring accordingly.

lightly/api/api_workflow_predictions.py
MalteEbner: Yes, I changed that.

lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,152 @@
...
+        self._predictions_api.create_or_update_prediction_task_schema_by_dataset_id(
+            body=schema,
+            dataset_id=self.dataset_id,
+            prediction_uuid_timestamp=prediction_uuid_timestamp,
+        )
+
+
+    def create_or_update_predictions(
+        self,
+        filename_to_prediction_singletons: Dict[str, List[Union[PredictionSingletonClassificationRepr]]],
```
michal-lightly, line 70: There is just one thing in the union because this is a work in progress, right?

lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,152 @@
...
+        prediction_uuid_timestamp: int,
+        progress_bar: Optional[tqdm.tqdm] = None,
+        max_workers: int = 8
+    ) -> None:
+        """Creates or updates the predictions for specific samples
+
+        Args:
+            filename_to_prediction_singletons
+                A mapping from the filename of the sample to its corresponding prediction singletons.
+                All singletons must be of the same type.
```
michal-lightly, line 80: Why is there this limitation?

lightly/api/api_workflow_predictions.py
michal-lightly: Probably @guarin is better to answer because he knows what data we have available in worker. Jeremy is on point, the worker should already know a sampleId. (I have not heard of sampleId though before, do we use it in worker?)

lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,152 @@
...
+            self._samples_api.get_samples_partial_by_dataset_id,
+            dataset_id=self.dataset_id,
+            mode=SamplePartialMode.FILENAMES,
+        )
+        api_filename_to_sample_id = {sample.file_name: sample.id for sample in samples}
+
+        if any(filename_to_upload not in api_filename_to_sample_id for filename_to_upload in filename_to_prediction_singletons.keys()):
+            raise ValueError(f"Some filenames to upload are not found as samples on the server.")
+
+        def upload_prediction(filename_and_predictions_tuple) -> bool:
```
michal-lightly, line 133: Nit: Type would be nice

lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,152 @@
...
+            (filename, predictions) = filename_and_predictions_tuple
+            sample_id = api_filename_to_sample_id[filename]
+            prediction_singletons_for_sending = [vars(singleton) for singleton in predictions]
+            retry(
+                self._predictions_api.create_or_update_prediction_by_sample_id,
+                body=prediction_singletons_for_sending,
+                dataset_id=self.dataset_id,
+                sample_id=sample_id,
+                prediction_uuid_timestamp=prediction_uuid_timestamp,
+            )
```
michal-lightly, line 143: Minor, optional: This function can be moved outside of the enclosing function. It will need an extra argument and then needs to be passed in as a partial function:
```py
def _upload_prediction(prediction_uuid_timestamp, filename_and_predictions_tuple) -> bool:
    ...

def create_or_update_predictions(...):
    ...

    upload_prediction_partial = functools.partial(_upload_prediction, prediction_uuid_timestamp)
    for _ in executor.map(upload_prediction_partial, filename_to_prediction_singletons.items()):
        ...
```

lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # handle the case where len(filename_to_sample_id) < max_workers
+        max_workers = min(len(api_filename_to_sample_id), max_workers)
+        max_workers = max(max_workers, 1)
+
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
+            for _ in executor.map(upload_prediction, filename_to_prediction_singletons.items()):
+                if progress_bar is not None:
+                    progress_bar.update(1)
```
michal-lightly, line 152: Question: I would expect `filename_to_prediction_singletons.items()` just wrapped in `tqdm`. Why do we accept `progress_bar` as a parameter? To customise the message?

lightly/api/prediction_singletons.py
```diff
@@ -0,0 +1,24 @@
...
+from lightly.openapi_generated.swagger_client import TaskType
+
+
+class PredictionSingletonRepr(ABC):
+    type: str
+    taskName: str
+    cropDatasetId: Optional[str] = None
+    cropSampleId: Optional[str] = None
+    categoryId: int
+    score: float
```
michal-lightly, line 13: Is the reason not to use dataclasses that we support python 3.6?

michal-lightly: I would recommend using instance variables instead of class variables; that way we can prevent sharing the instances. Without dataclasses there is a little boilerplate code:
```py
class PredictionSingletonRepr(ABC):
    def __init__(...)
        self.type = type
        self.taskName = taskName
        self.categoryId = categoryId
        self.score = score

class PredictionSingletonClassificationRepr(PredictionSingletonRepr):
    def __init__(...):
        super().__init__(...)
        self.probabilities = probabilities
```

lightly/api/prediction_singletons.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class PredictionSingletonClassificationRepr(PredictionSingletonRepr):
```
michal-lightly, line 16: Nit: The name is a mouthful :) What does "Repr" mean?

lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,152 @@
...
+        prediction_uuid_timestamp: int,
+    ) -> None:
+        """Creates or updates the prediction task schema
+
+        Args:
+            schema:
+                The prediction task schema.
+            prediction_uuid_timestamp:
+                This timestamp is used as a key to distinguish different predictions for the same sample.
+                Get it e.g. via 'int(time.time())'
```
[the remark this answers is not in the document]
MalteEbner: This is the parameter as defined in the API. See https://github.com/lightly-ai/lightly/pull/985#discussion_r1029522176
If you want to challenge the current design of it in the API, I would do that outside of this PR.

lightly/api/api_workflow_predictions.py
MalteEbner: Yes, so that we can use `logging_tqdm` instead of `tqdm` when calling it from the worker.

michal-lightly:
@MalteEbner I'd like to check what is the intended use:

* First call `create_or_update_prediction_task_schema` for each prediction task, this essentially defines what tasks are available
* Then call `create_or_update_predictions`

Is that right?

What about schema validation? How is that handled?

lightly/api/prediction_singletons.py
MalteEbner: > Is the reason not to use dataclasses that we support python 3.6?

That and we need custom initialisers.

lightly/api/api_workflow_predictions.py
MalteEbner: It will need even more extra arguments: the `self` and the `api_filename_to_sample_id`. Thus I think that extracting it makes the code more ugly.

lightly/api/api_workflow_predictions.py
MalteEbner: We also use it elsewhere, see e.g. https://github.com/lightly-ai/lightly/blob/master/lightly/api/api_workflow_datasources.py#L108

lightly/api/api_workflow_predictions.py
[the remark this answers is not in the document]
michal-lightly: Sound good to have it out of scope of this PR (then we will need to update two places though).

@japrescott Would you know the answer to my question?

lightly/api/prediction_singletons.py
```diff
@@ -0,0 +1,45 @@
...
+        cropSampleId: Optional[str] = None,
+    ):
+        self.type = type
+        self.taskName = taskName
+        self.categoryId = categoryId
+        self.score = score
+        if cropDatasetId is not None:
+            self.cropDatasetId = cropDatasetId
+        if cropSampleId is not None:
+            self.cropSampleId = cropSampleId
```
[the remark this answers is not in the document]
MalteEbner:
```
HTTP response body: {
    "code": "MALFORMED_REQUEST",
    "error": "[0].cropDatasetId must be a `string` type, but the final value was: `null`.\n If \"null\" is intended as an empty value be sure to mark the schema as `.nullable()`",
    "requestId": "8d2a17281893ec40a032763d2e89b67f"
}
```

MalteEbner:
> @MalteEbner I'd like to check what is the intended use:
> 
> * First call `create_or_update_prediction_task_schema` for each prediction task, this essentially defines what tasks are available
> * Then call `create_or_update_predictions`
> 
> Is that right?
> 
> What about schema validation? How is that handled?

See https://linear.app/lightly/issue/LIG-1303/upload-predictions-to-api

lightly/api/api_workflow_predictions.py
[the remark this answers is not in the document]
MalteEbner: See https://linear.app/lightly/issue/LIG-1303/upload-predictions-to-api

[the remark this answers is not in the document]
japrescott: @michal-lightly I see your point that its "counter intuitive". The reason its a mix is because I didn't want to create more endpoints for versioning etc. before we don't even know if the basic concept works :)

A customer has predictions at time X. Then uses AL, labels them, retrains models, gets new predictions. Then uses Lightly again at time Y. The idea was that the customer can see the difference between time X and time Y in the UI.

"Correct" would have been that the customer could define a name and we would then reference that entity which would have a timestamp attached for sorting purposes. I decided for the "less complicated" way but maybe that is not ideal.

> What if we want to upload two different prediction schemas and it happens to be in the same second?

A schema is unique by `datasetId+uuidtimestamp+name` so this does not matter.

> Why not just use an incremental uuid

Sure, that is up to you. We can also roll with `0` for the moment and always overwrite the schemas. Or we can always set it to the current timestamp so there is a "time element" associated to the value.

lightly/api/prediction_singletons.py
```diff
@@ -0,0 +1,101 @@
...
+            score=score,
+        )
+        self.probabilities = probabilities
+
+
+class PredictionSingletonObjectDetectionRepr(PredictionSingletonRepr):
+    def __init__(
+        self,
+        taskName: str,
+        cropDatasetId: str,
```
japrescott, line 49: why are the argumentst in a different order order compared to the other classes?

lightly/api/prediction_singletons.py
```diff
@@ -0,0 +1,101 @@
...
+        self.probabilities = probabilities
+
+
+class PredictionSingletonInstanceSegmentationRepr(PredictionSingletonRepr):
+    def __init__(
+        self,
+        taskName: str,
+        categoryId: int,
+        score: float,
+        segmentation: str,
```
japrescott, line 74: why can't I pass cropDatasetId or cropSampleId?

lightly/api/prediction_singletons.py
```diff
@@ -0,0 +1,101 @@
...
+    ):
+        super().__init__(
+            type=TaskType.INSTANCE_SEGMENTATION,
+            taskName=taskName,
+            categoryId=categoryId,
+            score=score,
```
japrescott, line 80: why can't I pass cropDatasetId or cropSampleId?

japrescott (changes requested):

lightly/api/prediction_singletons.py
MalteEbner: I fixed that.

lightly/api/prediction_singletons.py
MalteEbner: I had thought that the instance segmentation does not have those values. I passed it now.

lightly/api/prediction_singletons.py
MalteEbner: See https://github.com/lightly-ai/lightly/pull/985#discussion_r1032394710

lightly/api/api_workflow_predictions.py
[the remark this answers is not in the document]
michal-lightly: Jeremy and I had a discussion in which Jeremy explained the meaning of the field. It is worth knowing that:
* This interface was designed 9 months ago, and is imperfect
* Changing it now (even a rename) would not be trivial

We decided to keep the API as is, but give it a very good documentation.

The `prediction_uuid_timestamp` should be properly named `predicionVersionId`. It is any id specified by the user. If the user does not care about keeping history, they can simply always pass the same number.

Note that originally the endpoint was not meant for customer use. The idea was to display a timestamp in the UI and the field was (a bit hackily) meant to double as the timestamp.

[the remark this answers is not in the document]
michal-lightly: For the ApiWorkflowClient function, I propose to change the parameters as follows:
* Call the param `predicion_version_id` and make it optional
* If not specified, pass always `-1` to the api

Pros:
* Clearer to understand
* Easier to use
Cons:
* Not 1:1 correspondence to the API call
* The version id is an int, but str would be more natural
* Special value -1

MalteEbner:
> `cropDatasetId` and `cropSampleId` are available on all Singletons.

Yes, but they don't exist if we don't have a bounding box. Thus they don't exist for `keypoint detection` and `classification`. 

> And they are also optional so they should be towards the end of the argument list by convention.

I used the order of first the ones in `PredictionSingletonRepr`, then the ones in the child class. In my eyes, they are required for the ObjectDetection and Instance Segmentation, but must be None for the Keypoint detection and classification. I know that this is different from the API specs. The more I look at them, the more I think that they should be refactored. E.g. the `cropSampleId` should not be on the `PredictionSingletonBase`, but on the `PredictionSingletonObjectDetection`. 

@philippmwirth As you reviewed the code, could you look again at the API specs and make sure they still mostly make sense? In my eyes, a rework could be worth it. https://github.com/lightly-ai/lightly-core/blob/9afb2b4ff4d80ee6db807ca56649b16655393be8/openapi/spec/components/Prediction.yml#L30

lightly/api/api_workflow_predictions.py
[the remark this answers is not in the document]
MalteEbner: I agree and will change it.

michal-lightly (commented):
Looks good to me! The only thing is the meaning of uuid_timestamp for `schema`, we need an input from @japrescott

lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,169 @@
...
+        schema: PredictionTaskSchema,
+        prediction_version_id: int = -1,
+    ) -> None:
+        """Creates or updates the prediction task schema
+
+        Args:
+            schema:
+                The prediction task schema.
+            prediction_version_id:
+                And id to distinguish different predictions for the same sample.
```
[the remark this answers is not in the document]
japrescott: Its just how we pull together the schemas with the predictions. Aka it could be that suddenly there are more categoryIds with newer predictions.

I would rewrite it slightly;
```
 A numerical id (e.g timestamp) to distinguish different predictions of different model versions. Use the same id if you don't require versioning or if you wish to overwrite the previous schema. Optional, set to -1 by default.
```

japrescott (approved):

lightly/api/api_workflow_predictions.py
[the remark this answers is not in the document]
michal-lightly: One thing I don't understand: How do we know which schema belong to a prediction?

lightly/api/api_workflow_predictions.py
```diff
@@ -0,0 +1,169 @@
...
+
+        Args:
+            sample_id_to_prediction_singletons
+                A mapping from the sample_id of the sample to its corresponding prediction singletons.
+                The singletons can be from different tasks and different types.
+
+            prediction_version_id:
+                 A numerical id (e.g timestamp) to distinguish different predictions of different model versions.
+                 Use the same id if you don't require versioning or if you wish to overwrite the previous schema.
+                 This id must match th id of a prediction task schema.
```
michal-lightly, line 84:
```suggestion
                 This id must match the id of a prediction task schema.
```

michal-lightly (approved):
Let's merge this, thanks a lot!

PR #987, opened 2022-11-22
Fix failing github actions

guarin:
CI started failing about 24h ago due to a new release of setuptools (v65.6.0) which causes bugs in downstream packages: https://github.com/pypa/setuptools/issues/3693

Example of a failing action: https://github.com/lightly-ai/lightly/actions/runs/3524405267

## Changes
* Limited setuptools to v65.5.1 in requirements.txt
* Updated actions versions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,14 +22,14 @@ jobs:
 
     steps:
     - name: Checkout Code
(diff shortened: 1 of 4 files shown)


michal-lightly (approved):
LGTM!

PR #992, opened 2022-11-25
Niccolo lig 2064 vicreg

Ra1Nik:
Added VICReg example and updated the documentation. Tested working running the script and the rendered documentation print can be found here:
[lightly.loss.pdf](https://github.com/lightly-ai/lightly/files/10090100/lightly.loss.pdf)
[VICReg example.pdf](https://github.com/lightly-ai/lightly/files/10090103/VICReg.example.pdf)

### DIFF
--- a/docs/source/examples/vicreg.rst
+++ b/docs/source/examples/vicreg.rst
@@ -1,0 +1,50 @@
+.. _vicreg:
+
+ VICReg
+ ===
+
+ VICReg (Variance-Invariance-Covariance Regularization) is a method that explicitly
+ avoids the collapse problem with a simple regularization term on the variance of the embeddings along each dimension individually. VICReg combines the
+ variance term with a decorrelation mechanism based on redundancy reduction and covariance regularization. It inherits the model structure from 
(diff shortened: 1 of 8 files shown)


docs/source/examples/vicreg.rst
```diff
@@ -0,0 +1,50 @@
...
+        the system. The example can be run from the command line with::
+
+            python lightly/examples/pytorch_lightning_distributed/vicreg.py
+
+        The model differs in the following ways from the non-distributed
+        implementation:
+
+        - Distributed Data Parallel is enabled
+        - Distributed Sampling is used in the dataloader
+        - Distributed Sinkhorn is used in the loss calculation
```
guarin, line 45: Could you just drop this line? VICReg does not use sinkhorn.

guarin (changes requested):
Awesome! 

Could you remove the part about Sinkhorn as VICReg does not use it.
And I also noticed that the example of the VICReg loss is not properly rendered in the docs:
Screen Shot 2022-11-25 at 10 43 20

For the other losses it looks like this:
Screen Shot 2022-11-25 at 10 43 24

Could you also fix this in `lightly/loss/vicreg_loss.py`? I believe it is just a missing newline or something in the docstring.

guarin (approved):

PR #999, opened 2022-12-05
Update Lightly citations

utkuozbulak:
This bibtex returns the following error: **Too many commas in name 1 of "Igor Susmelj, Matthias Heller, Philipp Wirth, Jeremy Prescott, Malte Ebner et al." for entry lightly.**

This is because .bib entries in LaTeX expect author separation with "and", not comma. Comma is used when author names are written in (surname,name) format. Here is a simple fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,7 +334,7 @@ If you want to cite the framework feel free to use this:
 ```bibtex
 @article{susmelj2020lightly,
   title={Lightly},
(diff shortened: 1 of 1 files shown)


MalteEbner (approved):
Thanks for pointing this out!

This is indeed the recommended way: https://tex.stackexchange.com/a/36397

Issue #1006, opened 2022-12-13
Queue not implemented for SwaV


ibro45:
Hi, thanks for your great work!

From the [SwaV paper](https://arxiv.org/pdf/2006.09882.pdf):
<img width="1404" alt="image" src="https://user-images.githubusercontent.com/18015788/207389122-4adaffe6-eb36-481f-bcf6-92bbc2a7b11a.png">

[Lightly's `SwavLoss` ](https://github.com/lightly-ai/lightly/blob/7d3bc64ac3372c6e7ec8e24a8c56fb499209957f/lightly/loss/swav_loss.py#L65) does not implement this.

Here you can find [VISSL's implementation](https://github.com/facebookresearch/vissl/blob/346114aa1152aeadeb42d410951a95a28c53a207/vissl/losses/swav_loss.py#L191). It is quite messy though.

guarin:
Hi, thanks a lot for pointing this out! You are right that we did not implement the queue.

There is sadly no easy way to add the queue to the loss as the features in the queue have to be matched to the prototypes which are stored in the model. We would have to store the queue outside of the loss. I think something like this could work (modified version of our [SwaV example](https://docs.lightly.ai/self-supervised-learning/examples/swav.html)):

```
class SwaV(nn.Module):
    def __init__(self, backbone):
        super().__init__()
        self.backbone = backbone
        self.projection_head = SwaVProjectionHead(512, 512, 128)
        self.prototypes = SwaVPrototypes(128, n_prototypes=512)

    def forward(self, x):
        x = self.backbone(x).flatten(start_dim=1)
        x = self.projection_head(x)
        x = nn.functional.normalize(x, dim=1, p=2)
        p = self.prototypes(x)
        return p, x    # return prototypes assignments and features

...

# Create one queue for each high resolution view.
queues = [MemoryBankModule(size=3, 840 ) for _ in range(2)]
for queue in queues:
    queue.to(device)

for epoch in range(10):
    total_loss = 0
    for batch, _, _ in dataloader:
        model.prototypes.normalize()
        multi_crop_outputs = [model(x.to(device)) for x in batch]
        multi_crop_prototypes = [prototypes for prototypes, _ in multi_crop_outputs]
        high_resolution = multi_crop_prototypes[:2]
        low_resolution = multi_crop_prototypes[2:]

        # Get queue features
        queue_prototypes = []
        with torch.no_grad():
[...]

PR #1009, opened 2022-12-14
Fix Memory Bank

guarin:
### Changes

Add missing assignment after normalizing memory bank. This made the network collapse during training because the norm of the randomly initialized bank was too large. The bug was introduced in #993 (lightly v1.2.37).

### How was it tested?

Ran the [MoCo Tutorial](https://docs.lightly.ai/self-supervised-learning/tutorials/package/tutorial_moco_memory_bank.html) and verified that the loss decreases.

Closes #1007

PR #1010, opened 2022-12-16
Queue for SwaV

ibro45:
Aims to solve https://github.com/lightly-ai/lightly/issues/1006

Implements queue for SwaV with minimal code change in the training loop. Implemented as per details from the paper:

Rationale

image

A trick for small batch sizes - start using queue prototypes at a later epoch

image


lightly/loss/swav_loss.py
```diff
@@ -110,7 +110,8 @@ def subloss(self, z: torch.Tensor, q: torch.Tensor):
 
     def forward(self,
                 high_resolution_outputs: List[torch.Tensor],
-                low_resolution_outputs: List[torch.Tensor]):
+                low_resolution_outputs: List[torch.Tensor],
+                queue_outputs: List[torch.Tensor]):
```
guarin, line 114: Could you make the `queue_outputs` optional (default is `None`)?

lightly/models/modules/heads.py
```diff
@@ -371,7 +371,7 @@ class SwaVPrototypes(nn.Module):
     """Multihead Prototypes used for SwaV.
 
     Each output feature is assigned to a prototype, SwaV solves the swapped
-    predicition problem where the features of one augmentation are used to
+    prediction problem where the features of one augmentation are used to
```
guarin, line 374: Thanks for fixing this one :)

lightly/models/swav.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class SwaV(nn.Module):
+    def __init__(self, backbone, num_ftrs, out_dim, n_prototypes, queue_length=0):
+        super().__init__()
+        self.backbone = backbone
+        self.projection_head = SwaVProjectionHead(num_ftrs, num_ftrs, out_dim)
+        self.prototypes = SwaVPrototypes(out_dim, n_prototypes=n_prototypes)
+        # Queues are initialized in the first forward call
+        self.queues = []
```
guarin, line 15: I think you could register the queues in the beginning as a [ModuleList](https://pytorch.org/docs/stable/generated/torch.nn.ModuleList.html):

```
self.queues = ModuleList(modules=[MemoryBankModule(size=queue_length) for _ in range(n_queues)])
```

This would make sure that queues are moved to the correct device on `model.to(device)` calls. The drawback is that `n_queues` has to be passed as argument but I think that it is ok to assume that the number of queues is known from the beginning.

What do you think?

lightly/models/swav.py
```diff
@@ -0,0 +1,55 @@
+import torch
+from torch import nn
+
+from lightly.models.modules import SwaVProjectionHead, SwaVPrototypes
+from lightly.loss.memory_bank import MemoryBankModule
```
guarin, line 5: We decided a while ago to not add "full" models to `lightly/models` anymore. The reason was that users usually want to customize models in some way or another and would have to change the source code to do so. Instead we started writing example model implementations in `examples/` that are easy to copy paste and modify.

Could you move everything in this file to `swav_new.py` (maybe rename the file to `swav_queue.py` or similar)? Then it would perfectly follow our examples "philosophy".

lightly/models/swav.py
```diff
@@ -0,0 +1,55 @@
...
+        self.prototypes.normalize()
+
+        high_resolution_features = [self._subforward(x) for x in high_resolution]
+        low_resolution_features = [self._subforward(x) for x in low_resolution]
+        queue_features = self._get_queue_features(high_resolution_features)
+
+        high_resolution_prototypes = [self.prototypes(x) for x in high_resolution_features]
+        low_resolution_prototypes = [self.prototypes(x) for x in low_resolution_features]
+        queue_prototypes = [self.prototypes(x) for x in queue_features]
+
```
guarin, line 30: I believe this line should be in a `torch.no_grad()` block:
```
queue_prototypes = [self.prototypes(x) for x in queue_features]
```

The original repo does queue prototype assignment also with `torch.no_grad()`: https://github.com/facebookresearch/swav/blob/06b1b7cbaf6ba2a792300d79c7299db98b93b7f9/main_swav.py#L291-L301

guarin (commented):
Hi, thanks a ton for the lightning fast PR!

I had a look at the code and this looks very promising. Let me know if I can somehow help you with the memory bank issue. Are you aware that there was a very recent change to the memory bank in #1009 ?

ibro45:
Thanks so much for your feedback! And yes, I did fork from the last commit with the MemoryBank fix

guarin:
Awesome! Were you able to fix the issue with the memory bank? 
Let me know if the PR is ready for review again.

ibro45:
This was the issue with the memory bank. https://github.com/lightly-ai/lightly/blob/cf30f901b942eda7c426ad2b456b1f5d75504a91/examples/pytorch/swav_queue.py#L58

Is there a reason why the memory bank output dimensions are swapped?

guarin:
> Is there a reason why the memory bank output dimensions are swapped?

I am not aware of any particular reason. I guess we followed to original implementation where it is also swapped: https://github.com/facebookresearch/moco/blob/78b69cafae80bc74cd1a89ac3fb365dc20d157d3/moco/builder.py#L39

ibro45:
I see, then I'll remove those todo comments and leave as is. Otherwise it's ready for a review

guarin (approved):
This looks great, thanks a lot for the PR!

We'll try to run the code on one of our benchmarks :)

ibro45:
Awesome, thank you :)

guarin:
Hi @ibro45, I tested the queue implementation and it seems to work (accuracy >45% on Imagenet100 after 50 epochs with batch size 128 on a Tesla T4). The accuracy only seems to increase after epoch 15 once the queue is used. Before, the accuracy was always hovering around 1%. Maybe one could use the queue even earlier to speed up training, although the paper mentions that the queue should not be used too early.

To verify that the implementation matches the original we would have to train on full Imagenet but we don't have time to do that right now. I pushed the code I used here #1015 if you want to have a look (had to adapt it to pytorch lightning).

ibro45:
That's great to hear!

I think you're right, while 15 epochs before the queue were good for ImageNet, maybe a lower number would be more suitable for ImageNet100.

On another note, do you think it'd be possible to add the plots of losses and metrics of these models? A Weights&Biases project that includes all the models would be amazing for newbies like me to know how these models behave. For example, someone training a SwaV might get alarmed about the accuracy being that low before the queue kicks. Actually, that's what I'm struggling with right now, I'm trying to apply it to medical images and it's difficult to know what's normal and what needs more parameter tuning.

PR #1011, opened 2022-12-19
Philipp fix version check

philippmwirth:
# Philipp fix version check

The version check was broken because of circular imports that were swallowed by a try-catch. I resolved the circular import by moving the relevant function to `lightly.api.utils`.

### DIFF
--- a/lightly/__init__.py
+++ b/lightly/__init__.py
@@ -75,7 +75,7 @@ The framework is structured into the following modules:
 # All Rights Reserved
 
 __name__ = 'lightly'
-__version__ = '1.2.39'
+__version__ = '1.2.28'
 
 import os
 
@@ -123,12 +123,11 @@ else:
     from lightly import transforms
     from lightly import utils
     
-    from lightly.api.version_checking import do_version_check
-
     if os.getenv('LIGHTLY_DID_VERSION_CHECK', 'False') == 'False':
         os.environ['LIGHTLY_DID_VERSION_CHECK'] = 'True'
 
         try:
-            do_version_check(current_version=__version__)
+          from lightly.api.version_checking import do_version_check
+          do_version_check(current_version=__version__)
         except Exception as e:
             pass
--- a/lightly/__init__.py
+++ b/lightly/__init__.py
@@ -129,6 +129,6 @@ else:
         if current_process().name == 'MainProcess':
             from lightly.api.version_checking import is_latest_version
             try:
-              is_latest_version(current_version=__version__)
+                is_latest_version(current_version=__version__)
             except Exception as e:
                 pass
--- a/lightly/__init__.py
+++ b/lightly/__init__.py
@@ -75,7 +75,7 @@ The framework is structured into the following modules:
 # All Rights Reserved
 
 __name__ = 'lightly'
-__version__ = '1.2.28'
+__version__ = '1.2.39'
 
 import os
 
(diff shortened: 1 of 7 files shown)


lightly/__init__.py
```diff
@@ -123,12 +123,11 @@ def _is_prefetch_generator_available():
...
-    from lightly.api.version_checking import do_version_check
-
     if os.getenv('LIGHTLY_DID_VERSION_CHECK', 'False') == 'False':
         os.environ['LIGHTLY_DID_VERSION_CHECK'] = 'True'
 
         try:
-            do_version_check(current_version=__version__)
+          from lightly.api.version_checking import do_version_check
+          do_version_check(current_version=__version__)
         except Exception as e:
```
michal-lightly, line 132: Let's also fix the root of the problem and catch the specific exception we care about.

michal-lightly: Why did you move the import inside of the try-block?

lightly/api/api_workflow_client.py
```diff
@@ -120,9 +117,9 @@ def __init__(
     def check_version_compatibility(self):
         minimum_version = get_minimum_compatible_version()
         if version_compare(__version__, minimum_version) < 0:
-            raise ValueError(f"Incompatible Version of lightly pip package. "
-                             f"Please upgrade to at least version {minimum_version} "
-                             f"to be able to access the api and webapp")
+            raise ValueError(f"Incompatible version of lightly pip package. "
+                             f"Please upgrade to at least version '{minimum_version}' "
+                             f"to be able to access the api.")
```
michal-lightly, line 122: Minor: What do you think about changing this function to return bool and naming is something like `is_version_compatible`?

michal-lightly: Actually, for this PR the relevant function is `do_version_check` which is raising, but could return a bool instead.

Up to you if you want to change or keep.

lightly/api/version_checking.py
```diff
@@ -33,23 +31,13 @@ def do_version_check(current_version: str):
...
-
-            try:
-                if version_compare(current_version, latest_version) < 0:
-                    # local version is behind latest version
-                    pretty_print_latest_version(current_version, latest_version)
-            except ValueError:
-                pass
-                # error during version compare
-                # we just skip the version check
-
```
michal-lightly, line 48: What about this code block, why is it not necessary anymore?

michal-lightly (changes requested):
Let's please catch a concrete exception to prevent swallowing unwanted errors in the future.

lightly/api/api_workflow_client.py
philippmwirth: Yes, sounds good to me 👍

lightly/api/version_checking.py
philippmwirth: If we keep the block we will print a message whenever someone is not using the latest lightly version. This check has been broken for some time now (`current_version` was overridden by the latest version) so the message was never printed. I removed it now completely so that we're at least not printing the message on purpose.

lightly/__init__.py
michal-lightly: How can the import fail?

philippmwirth: I moved the import outside again 👍 my comment was about the type of exceptions we catch

tests/api/test_version_checking.py
```diff
@@ -21,6 +21,18 @@ def test_get_latest_version(self):
     def test_get_minimum_compatible_version(self):
         get_minimum_compatible_version()
 
+    def test_is_latest_version(self) -> None:
+        assert is_latest_version("1.2.8")
+        assert not is_latest_version("1.2.7")
+        assert not is_latest_version("1.1.8")
+        assert not is_latest_version("0.2.8")
+
+    def test_is_latest_version(self) -> None:
```
michal-lightly, line 30: Function redefinition

michal-lightly (approved):
LGTM!

PR #1012, opened 2022-12-20
Add download embeddings method

guarin:
### What has changed and why?

* Add `download_embeddings_csv` to `ApiWorkflowClient`
* Remove unused imports

I limited to method to download only the latest embedding file. This file should include all embeddings for all samples in the dataset. There is currently no way to get the embeddings from a specific run as the embeddings are not linked to runs.

### How has it been tested?

* Add unit tests
* Tested it also manually

### DIFF
--- a/lightly/api/api_workflow_download_dataset.py
+++ b/lightly/api/api_workflow_download_dataset.py
@@ -149,14 +149,15 @@ class _DownloadDatasetMixin:
             warnings.warn(msg)
 
     def download_embeddings_csv(self, output_path: str) -> None:
-        """Downloads the last created embedding from the dataset to the output path.
+        """Downloads the latest embeddings from the dataset and saves them to the output
+        path.
 
         Raises:
             RuntimeError:
-                If no embedding could be found for the dataset.
+                If no embeddings could be found for the dataset.
 
         """
-        last_embedding = self._get_last_default_embeddings_data()
+        last_embedding = self._get_latest_default_embeddings_data()
         if last_embedding is None:
             raise RuntimeError(
                 f"Could not find embedding for dataset with id '{self.dataset_id}'."
@@ -167,9 +168,9 @@ class _DownloadDatasetMixin:
         )
         download.download_and_write_file(url=read_url, output_path=output_path)
 
-    def _get_last_default_embeddings_data(self) -> Optional[DatasetEmbeddingData]:
-        """Returns the last created embedding with a default name from the dataset or
-        None if no such embedding exists.
+    def _get_latest_default_embeddings_data(self) -> Optional[DatasetEmbeddingData]:
+        """Returns the latest embeddings with a default name from the dataset or None
+        if no such embedding exists.
         """
         embeddings = self._embeddings_api.get_embeddings_by_dataset_id(
             dataset_id=self.dataset_id
--- a/lightly/api/api_workflow_download_dataset.py
+++ b/lightly/api/api_workflow_download_dataset.py
@@ -170,18 +170,6 @@ class _DownloadDatasetMixin:
             f"'{self.dataset_id}'."
         )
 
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_download_dataset.py
```diff
@@ -150,9 +148,38 @@ def lambda_(i):
             msg += 'Failed at image: {}'.format(results.index(False))
             warnings.warn(msg)
 
+    def download_embeddings_csv(self, output_path: str) -> None:
+        """Downloads the last created embedding from the dataset to the output path.
```
philippmwirth, line 152: The wording is a bit complicated, can we do "Downloads the latest embedding from the dataset and saves it to the output path"

lightly/api/api_workflow_download_dataset.py
```diff
@@ -150,9 +148,38 @@ def lambda_(i):
...
+                f"Could not find embedding for dataset with id '{self.dataset_id}'."
+            )
+        read_url = self._embeddings_api.get_embeddings_csv_read_url_by_id(
+            dataset_id=self.dataset_id,
+            embedding_id=last_embedding.id,
+        )
+        download.download_and_write_file(url=read_url, output_path=output_path)
 
+    def _get_last_default_embeddings_data(self) -> Optional[DatasetEmbeddingData]:
+        """Returns the last created embedding with a default name from the dataset or
```
philippmwirth, line 171: Just use latest instead of last created

lightly/api/api_workflow_download_dataset.py
```diff
@@ -150,9 +148,38 @@ def lambda_(i):
...
+            RuntimeError:
+                If no embedding could be found for the dataset.
 
+        """
+        last_embedding = self._get_last_default_embeddings_data()
+        if last_embedding is None:
+            raise RuntimeError(
+                f"Could not find embedding for dataset with id '{self.dataset_id}'."
+            )
+        read_url = self._embeddings_api.get_embeddings_csv_read_url_by_id(
```
philippmwirth, line 164: Can we two other functions to get an embedding by name and id? Then we can do

```python
    def dowload_embeddings_csv(self, output_path: str) -> None:
        last_embedding = self._get_last_default_embeddings_data()
        if last_embedding is None:
            raise RuntimeError(
                f"Could not find embedding for dataset with id '{self.dataset_id}'."
            )
        self.download_embeddings_csv_by_id(embeddings_id=last_embedding.id)
```

philippmwirth (changes requested):

lightly/api/api_workflow_download_dataset.py
```diff
@@ -150,8 +148,72 @@ def lambda_(i):
...
+
+        Raises:
+            ValueError:
+                If no embedding with this name exists.
 
+        """
+        for embedding_data in self.get_all_embedding_data():
+            if embedding_data.name == name:
+                return embedding_data
+        raise ValueError(
```
philippmwirth, line 168: Minor: I don't think the value error is the best choice but let's keep it for now

lightly/api/api_workflow_download_dataset.py
```diff
@@ -150,8 +148,72 @@ def lambda_(i):
...
+            f"There are no embeddings with name '{name}' for dataset with id "
+            f"'{self.dataset_id}'."
+        )
 
+    def _get_latest_default_embedding_data(self) -> Optional[DatasetEmbeddingData]:
```
philippmwirth, line 173: Let's make this take a list of embeddings and remove it from the api workflwo client
```python
def _get_latest_default_embedding_data(embeddings: list[DatasetEmbeddingData]) -> Optional[DatasetEmbeddingData]

philippmwirth (approved):
The PR looks good but I have two more comments. Feel free to ignore them.

philippmwirth (approved):

PR #1016, opened 2022-12-21
Add new TiCo Model.

Ra1Nik:
- Implement TiCo Loss and Projection Head
- Add Example
- Add Unit Test
- Update Docs

### DIFF
--- a/examples/pytorch/tico.py
+++ b/examples/pytorch/tico.py
@@ -1,0 +1,96 @@
+# Note: The model and training settings do not follow the reference settings
+# from the paper. The settings are chosen such that the example can easily be
+# run on a small dataset with a single GPU.
+
+import torch
+from torch import nn
+import torchvision
+import copy
+
+from lightly.data import LightlyDataset
+from lightly.data import SimCLRCollateFunction
+from lightly.loss.tico_loss import TiCoLoss
+from lightly.models.modules.heads import TiCoProjectionHead
+from lightly.models.utils import deactivate_requires_grad
+from lightly.models.utils import update_momentum
+from lightly.models.utils import schedule_momentum
+from torch.autograd import Variable
+
+class TiCo(nn.Module):
+    def __init__(self, backbone):
+        super().__init__()
+
+        self.backbone = backbone
+        self.projection_head = TiCoProjectionHead(512, 1024, 256)
+
+        self.backbone_momentum = copy.deepcopy(self.backbone)
+        self.projection_head_momentum = copy.deepcopy(self.projection_head)
+
+        deactivate_requires_grad(self.backbone_momentum)
+        deactivate_requires_grad(self.projection_head_momentum)
+
+    def forward(self, x):
+        y = self.backbone(x).flatten(start_dim=1)
+        z = self.projection_head(y)
+        return z
+
+    def forward_momentum(self, x):
+        y = self.backbone_momentum(x).flatten(start_dim=1)
+        z = self.projection_head_momentum(y)
+        z = z.detach()
+        return z
+
+resnet = torchvision.models.resnet18()
+backbone = nn.Sequential(*list(resnet.children())[:-1])
+model = TiCo(backbone)
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+model.to(device)
+
+cifar10 = torchvision.datasets.CIFAR10("/home/ubuntu/datasets/cifar10", download=True)
+dataset = LightlyDataset.from_torch_dataset(cifar10)
+# or create a dataset from a folder containing images or videos:
+# dataset = LightlyDataset("path/to/folder")
+
+collate_fn = SimCLRCollateFunction(input_size=32)
+
+dataloader = torch.utils.data.DataLoader(
+    dataset,
+    batch_size=8,
+    collate_fn=collate_fn,
+    shuffle=True,
+    drop_last=True,
+    num_workers=8,
+)
+
+criterion = TiCoLoss(
+    beta_param = 0.9,
+    ro_param = 20.0,
+)
+
+optimizer = torch.optim.SGD(model.parameters(), lr=0.06)
+
+C_prev = Variable(torch.zeros(256, 256), requires_grad=True).to(device)
+C_prev = C_prev.detach()
+
+epochs = 10
+
+print("Starting Training")
+for epoch in range(epochs):
+    total_loss = 0
+    momentum_val = schedule_momentum(epoch, epochs, m = 0.99)
+    for (x0, x1), _, _ in dataloader:
+        update_momentum(model.backbone, model.backbone_momentum, m=momentum_val)
+        update_momentum(model.projection_head, model.projection_head_momentum, m=momentum_val)
+        x0 = x0.to(device)
+        x1 = x1.to(device)
+        z0 = model(x0)
+        z1 = model.forward_momentum(x1)
+        loss, C = criterion(C_prev, z0, z1)
+        C_prev = C.detach()
+        total_loss += loss.detach()
+        loss.backward()
+        optimizer.step()
+        optimizer.zero_grad()
+    avg_loss = total_loss / len(dataloader)
+    print(f"epoch: {epoch:>02}, loss: {avg_loss:.5f}")
--- a/examples/pytorch/tico.py
+++ b/examples/pytorch/tico.py
@@ -56,22 +56,16 @@ collate_fn = SimCLRCollateFunction(input_size=32)
 
 dataloader = torch.utils.data.DataLoader(
     dataset,
-    batch_size=8,
+    batch_size=256,
     collate_fn=collate_fn,
     shuffle=True,
     drop_last=True,
     num_workers=8,
 )
 
-criterion = TiCoLoss(
-    beta_param = 0.9,
-    ro_param = 20.0,
-)
+criterion = TiCoLoss()
 
 optimizer = torch.optim.SGD(model.parameters(), lr=0.06)
-
-C_prev = Variable(torch.zeros(256, 256), requires_grad=True).to(device)
-C_prev = C_prev.detach()
 
 epochs = 10
 
@@ -86,8 +80,7 @@ for epoch in range(epochs):
         x1 = x1.to(device)
         z0 = model(x0)
         z1 = model.forward_momentum(x1)
-        loss, C = criterion(C_prev, z0, z1)
-        C_prev = C.detach()
+        loss = criterion(z0, z1)
         total_loss += loss.detach()
         loss.backward()
         optimizer.step()
--- a/examples/pytorch/tico.py
+++ b/examples/pytorch/tico.py
@@ -13,8 +13,8 @@ from lightly.loss.tico_loss import TiCoLoss
 from lightly.models.modules.heads import TiCoProjectionHead
 from lightly.models.utils import deactivate_requires_grad
 from lightly.models.utils import update_momentum
-from lightly.models.utils import cosine_decay_schedule
-from torch.autograd import Variable
+from lightly.models.utils import cosine_schedule
+
 
 class TiCo(nn.Module):
     def __init__(self, backbone):
@@ -47,7 +47,7 @@ model = TiCo(backbone)
 device = "cuda" if torch.cuda.is_available() else "cpu"
 model.to(device)
 
-cifar10 = torchvision.datasets.CIFAR10("/home/ubuntu/datasets/cifar10", download=True)
+cifar10 = torchvision.datasets.CIFAR10("datasets/cifar10", download=True)
 dataset = LightlyDataset.from_torch_dataset(cifar10)
 # or create a dataset from a folder containing images or videos:
 # dataset = LightlyDataset("path/to/folder")
@@ -72,7 +72,7 @@ epochs = 10
 print("Starting Training")
 for epoch in range(epochs):
     total_loss = 0
-    momentum_val = cosine_decay_schedule(epoch, epochs)
+    momentum_val = cosine_schedule(epoch, epochs, 0.996, 1)
     for (x0, x1), _, _ in dataloader:
         update_momentum(model.backbone, model.backbone_momentum, m=momentum_val)
         update_momentum(model.projection_head, model.projection_head_momentum, m=momentum_val)
--- a/examples/pytorch/tico.py
+++ b/examples/pytorch/tico.py
@@ -13,7 +13,7 @@ from lightly.loss.tico_loss import TiCoLoss
 from lightly.models.modules.heads import TiCoProjectionHead
 from lightly.models.utils import deactivate_requires_grad
 from lightly.models.utils import update_momentum
-from lightly.models.utils import schedule_momentum
+from lightly.models.utils import cosine_decay_schedule
 from torch.autograd import Variable
 
 class TiCo(nn.Module):
@@ -72,7 +72,7 @@ epochs = 10
 print("Starting Training")
 for epoch in range(epochs):
     total_loss = 0
-    momentum_val = schedule_momentum(epoch, epochs, m = 0.99)
+    momentum_val = cosine_decay_schedule(epoch, epochs)
     for (x0, x1), _, _ in dataloader:
         update_momentum(model.backbone, model.backbone_momentum, m=momentum_val)
         update_momentum(model.projection_head, model.projection_head_momentum, m=momentum_val)
--- a/examples/pytorch_lightning/tico.py
+++ b/examples/pytorch_lightning/tico.py
@@ -1,0 +1,86 @@
+import torch
+from torch import nn
+import torchvision
+import copy
+import pytorch_lightning as pl
+
+from lightly.data import LightlyDataset
+from lightly.data import SimCLRCollateFunction
+from lightly.loss.tico_loss import TiCoLoss
+from lightly.models.modules.heads import TiCoProjectionHead
+from lightly.models.utils import deactivate_requires_grad
+from lightly.models.utils import update_momentum
+from lightly.models.utils import schedule_momentum
+from torch.autograd import Variable
+
+
+
+class TiCo(pl.LightningModule):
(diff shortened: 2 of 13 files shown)


Ra1Nik:
Renderings of the docs:

[lightly.loss — lightly 1.2.40 documentation.pdf](https://github.com/lightly-ai/lightly/files/10276538/lightly.loss.lightly.1.2.40.documentation.pdf)
[TiCo — lightly 1.2.40 documentation.pdf](https://github.com/lightly-ai/lightly/files/10276539/TiCo.lightly.1.2.40.documentation.pdf)

examples/pytorch/tico.py
```diff
@@ -0,0 +1,96 @@
...
+
+criterion = TiCoLoss(
+    beta_param = 0.9,
+    ro_param = 20.0,
+)
+
+optimizer = torch.optim.SGD(model.parameters(), lr=0.06)
+
+C_prev = Variable(torch.zeros(256, 256), requires_grad=True).to(device)
+C_prev = C_prev.detach()
```
guarin, line 74: Can we put `C` into `TiCoLoss` instead of passing it around?

I would propose to use the following definition for the forward method in the loss:
```
def forward(self, z_a: torch.Tensor, z_b: torch.Tensor, update_covariance_matrix: bool = True) -> torch.Tensor:
# only update C if update_covariance_matrix == True
```

examples/pytorch/tico.py
```diff
@@ -0,0 +1,96 @@
...
+    batch_size=8,
+    collate_fn=collate_fn,
+    shuffle=True,
+    drop_last=True,
+    num_workers=8,
+)
+
+criterion = TiCoLoss(
+    beta_param = 0.9,
+    ro_param = 20.0,
```
guarin, line 68: No need to set `beta_param` and `ro_param` as they are already the default values.

lightly/loss/tico_loss.py
```diff
@@ -0,0 +1,80 @@
...
+
+from lightly.utils.dist import gather
+
+class TiCoLoss(torch.nn.Module):
+    """Implementation of the Tico Loss from Tico[0] paper.
+    This implementation takes inspiration from the code published 
+    by sayannag using Lightly. [1]
+
+    [0] Jiachen Zhu et. al, 2022, Tico... https://arxiv.org/abs/2206.10698
+    [1] https://github.com/sayannag/TiCo-pytorch
```
guarin, line 14: Could you format this docstring in the same format as the one for `DINOLoss`?

lightly/loss/tico_loss.py
```diff
@@ -0,0 +1,80 @@
...
+        >>>
+        >>> # calculate loss
+        >>> loss = loss_fn(out0, out1)
+    """
+
+    def __init__(
+        self,
+        beta_param: float = 0.9,
+        ro_param: float = 20.0,
+        gather_distributed : bool = False,
```
guarin, line 36: -> `gather_distributed: bool = False,`

(remove whitespace)

lightly/loss/tico_loss.py
```diff
@@ -0,0 +1,80 @@
...
+        >>> out0, out1 = model(t0, t1)
+        >>>
+        >>> # calculate loss
+        >>> loss = loss_fn(out0, out1)
+    """
+
+    def __init__(
+        self,
+        beta_param: float = 0.9,
+        ro_param: float = 20.0,
```
guarin, line 35: `beta_param` -> `beta`
`ro_param` -> `rho`

lightly/loss/tico_loss.py
```diff
@@ -0,0 +1,80 @@
...
+        gather_distributed : bool = False,
+    ):
+        """Lambda, mu and nu params configuration with default value like in [0]
+        Args:
```
guarin, line 39: Where do lambda, mu, and nu come from? I cannot find them in the paper.

Could you also add a newline before `Args:`?

lightly/loss/tico_loss.py
```diff
@@ -0,0 +1,80 @@
...
+        # gather all batches
+        if self.gather_distributed and dist.is_initialized():
+            world_size = dist.get_world_size()
+            if world_size > 1:
+                z_a = torch.cat(gather(z_a), dim=0)
+                z_b = torch.cat(gather(z_b), dim=0)
+
+        # normalize repr. along the batch dimension
+        z_a = z_a - z_a.mean(0) # NxD
+        z_b = z_b - z_b.mean(0) # NxD
```
guarin, line 69: Could not find this in the paper. Are you sure we have to do this?

lightly/loss/tico_loss.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # normalize image
+        z_a = torch.nn.functional.normalize(z_a, dim = -1)
+        z_b = torch.nn.functional.normalize(z_b, dim = -1)
```
guarin, line 73: I believe dim should be `1` instead of `-1`. Eventually, we have to flatten the representations first:
`z_a.flatten(start_dim=1)`.

lightly/loss/tico_loss.py
```diff
@@ -0,0 +1,80 @@
...
+        
+        # compute loss
+        B = torch.mm(z_a.T, z_a)/z_a.shape[0]        
+        C = self.beta_param * C + (1 - self.beta_param) * B
+        loss = 1 - (z_a * z_b).sum(dim=1).mean() + self.ro_param * (torch.mm(z_a, C) * z_a).sum(dim=1).mean()
```
guarin, line 78: Why `1 - (z_a, * z_b )...`? I can only find ` - (z_a * z_b) ...` in the paper.

lightly/models/utils.py
```diff
@@ -461,3 +462,31 @@ def random_token_mask(
...
+
+def schedule_momentum(
+    iteration: int, 
+    max_iter: int, 
+    m: float = 0.99
+) -> float:
+
+    """
+    Add sinusoidal noise to momentum value based on iteration number. Common
+    training tecnique in SSL
```
guarin, line 474: tecnique -> technique

Also, missing `.` at end of sentence 😉

lightly/models/utils.py
```diff
@@ -461,3 +462,31 @@ def random_token_mask(
...
+
+    Args:
+        iteration:
+            Size of the token batch for which to generate masks.
+            Should be (batch_size, sequence_length).
+        max_iter:
+            Percentage of tokens to mask.
+        m:
+            If False the class token is never masked. If True the class token
+            might be masked.
```
guarin, line 484: Why does this mention mask tokens? Could you update the docstrings?

tests/loss/test_TicoLoss.py
```diff
@@ -0,0 +1,46 @@
+import unittest
+import torch
+
+from lightly.loss.tico_loss import TiCoLoss
+torch.manual_seed(0)
+class TestVICRegLoss(unittest.TestCase):
```
guarin, line 6: `TestTiCoLoss`

tests/loss/test_TicoLoss.py
```diff
@@ -0,0 +1,46 @@
+import unittest
+import torch
+
+from lightly.loss.tico_loss import TiCoLoss
+torch.manual_seed(0)
```
guarin, line 5: Please add whitespace after imports and move `torch.manual_seed(0)` into each test method.

guarin (changes requested):
Awesome stuff!

Left a bunch of comments, will have a second look tomorrow :)

lightly/loss/tico_loss.py
Ra1Nik: Schermata 2022-12-22 alle 10 00 34

Equation (4)

guarin: Interesting, the pseudocode doesn't have the `1 -`:
Screen Shot 2022-12-22 at 10 07 12

Ra1Nik: Also the code from the implementation I followed had no 1-. But then I saw that the loss started being negative and spotted the problem.

guarin: I think you can drop the 1 as it is just a constant added to the loss which should not change training.

Ra1Nik: Yes, but isn't it good to always have a positive loss?

Ra1Nik: I would personally let it there as I think that it is not really affecting the computation

lightly/models/utils.py
```diff
@@ -471,20 +471,18 @@ def schedule_momentum(
...
-            Size of the token batch for which to generate masks.
-            Should be (batch_size, sequence_length).
+            Current epoch number.
         max_iter:
-            Percentage of tokens to mask.
+            Total number of epochs.
         m:
-            If False the class token is never masked. If True the class token
-            might be masked.
+            Starting momentum value.
```
guarin, line 482: I would propose to make this method more general and rename it to `cosine_schedule` with the following parameters:
```
step (instead of iteration)
max_steps (instead of max_iter)
start_value (instead of m)
end_value (new)
```
The method would then interpolate between `start_value` and `end_value` with a cosine decay.

lightly/loss/tico_loss.py
guarin: Did you verify that `z_a` has only 2 dimensions?

guarin (changes requested):
Thanks a lot for the changes :)

Made a second sweep and left some comments.

lightly/loss/tico_loss.py
Ra1Nik: Yes, doing a print. Should I also do it like an assertion?

lightly/loss/tico_loss.py
```diff
@@ -31,31 +47,39 @@ class TiCoLoss(torch.nn.Module):
 
     def __init__(
         self,
-        beta_param: float = 0.9,
-        ro_param: float = 20.0,
-        gather_distributed : bool = False,
+        beta: float = 0.9,
+        rho: float = 20.0,
+        C: torch.Tensor = Variable(torch.zeros(256, 256), requires_grad=True),
```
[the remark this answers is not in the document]
Ra1Nik: Actually, letting C as an argument might be useful because if someone wants to change the last layer dimension can overwrite C as Variable(torch.zeros(32, 32), requires_grad=True) for example.

lightly/models/utils.py
Ra1Nik: What would be the purpose of the end_value? I mean, if you put max steps, it is already an "end value" in the sense that you compute the end value with the step=max_steps

lightly/loss/tico_loss.py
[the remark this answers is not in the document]
guarin: My proposal below would do this automatically: https://github.com/lightly-ai/lightly/pull/1016#discussion_r1055286323

lightly/models/utils.py
guarin: Let's say you want cosine decay from 0.5 to 0.1 over 100 epochs. Then you would set:
```
max_steps = 100
start_value = 0.5
end_value = 0.1
```

lightly/models/utils.py
```diff
@@ -463,28 +463,33 @@ def random_token_mask(
...
+        end_value:
+            Target momentum value.
             
     Returns:
         New momentum value to be used with update_momentum.
 
     """
-    
-    momentum = m + (1 - m)*np.sin((np.pi / 2)* iteration / (max_iter - 1))
+    if (max_steps == 1):
```
guarin, line 491: Could you raise `ValueError` for the cases: `step < 0`, or `max_steps < 0`, or `step >= max_steps`.

lightly/models/utils.py
```diff
@@ -463,28 +463,33 @@ def random_token_mask(
 
     return idx_keep, idx_mask
 
-def schedule_momentum(
-    iteration: int, 
-    max_iter: int, 
-    m: float = 0.99
+def cosine_decay_schedule(
+    step: int,
```
guarin, line 467: Can we just call it `cosine_schedule`? Decay implies that the value decreases but it also supports increasing values.

lightly/models/utils.py
```diff
@@ -463,28 +463,33 @@ def random_token_mask(
...
 
-def schedule_momentum(
-    iteration: int, 
-    max_iter: int, 
-    m: float = 0.99
+def cosine_decay_schedule(
+    step: int, 
+    max_steps: int, 
+    start_value: float = 0.996,
+    end_value: float = 1
```
guarin, line 470: I wouldn't add default values here as the function is very general. It is, for example, also often used for learning rate schedules.

lightly/models/utils.py
```diff
@@ -463,28 +463,33 @@ def random_token_mask(
...
 ) -> float:
 
     """
-    Add sinusoidal noise to momentum value based on iteration number. Common
-    training technique in SSL.
+    Use cosine decay to gradually modify start_value to reach target end_value during training.
+    Introduced in BYOL (https://arxiv.org/pdf/2006.07733.pdf) to update the momentum.
```
guarin, line 475: No need to cite anything here. Cosine schedules are common and existed before BYOL :)

lightly/loss/tico_loss.py
```diff
@@ -92,13 +85,19 @@ def forward(self, z_a: torch.Tensor, z_b: torch.Tensor, update_covariance_matrix
         z_a = torch.nn.functional.normalize(z_a, dim = 1)
         z_b = torch.nn.functional.normalize(z_b, dim = 1)
         
+        # compute ausiliary matrix B
```
guarin, line 88: ausiliary -> auxiliary

guarin (changes requested):
Nice, I think we can merge this soon :)

examples/pytorch_lightning/tico.py
```diff
@@ -43,7 +42,7 @@ def forward_momentum(self, x):
 
     def training_step(self, batch, batch_idx):
         (x0, x1), _, _ = batch
-        momentum = cosine_decay_schedule(batch_idx, 10)
+        momentum = cosine_schedule(batch_idx, 10, 0.996, 1)
```
guarin, line 45: I believe we have to use current epoch instead of batch idx here:
```
momentum = cosine_schedule(self.current_epoch, 10, 0.996, 1)
```

examples/pytorch_lightning_distributed/tico.py
```diff
@@ -43,7 +42,7 @@ def forward_momentum(self, x):
 
     def training_step(self, batch, batch_idx):
         (x0, x1), _, _ = batch
-        momentum = cosine_decay_schedule(batch_idx, 10)
+        momentum = cosine_schedule(batch_idx, 10, 0.996, 1)
```
guarin, line 45: I guess this should also use self.current_epoch instead of batch_idx

lightly/models/utils.py
```diff
@@ -488,6 +487,10 @@ def cosine_decay_schedule(
         New momentum value to be used with update_momentum.
```
guarin, line 488: Could you remove the references to "momentum" in the docstring?

lightly/models/utils.py
```diff
@@ -488,6 +487,10 @@ def cosine_decay_schedule(
...
 
     """
+    assert step >= 0, "Current number can't be negative"
+    assert max_steps > 0, "Total epochs must be > 0"
+    assert step <= max_steps, f"Current step is bigger than total number of steps. Found {step} > {max_steps}."
```
guarin, line 492: I would prefer `ValueError` instead of `assert` here.

guarin, line 492: Are you sure that it is `step <= max_steps` and not `step < max_steps`? Because further down we have: `step / (max_steps - 1)`

guarin (changes requested):

lightly/models/utils.py
Ra1Nik: Yes, as I cut out the max_steps = 1 case with the if clause (and "step" starts from 0) and if I have step = max_steps = 1  is a case that I would like to keep

guarin: Yes but the calculation doesn't work if we have for example, step=max_steps=4. Then the formula does: `step / (max_steps - 1) = 4 / 3` which is `>1` and we want the schedule to only go to 1 no?

Ra1Nik: Oh yes sorry, step will never be = max step because as I wrote "step" starts from 0. Sorry

lightly/models/utils.py
guarin: There are still references about momentum in the docstring.

lightly/models/utils.py
```diff
@@ -484,12 +484,15 @@ def cosine_schedule(
...
     Returns:
-        New momentum value to be used with update_momentum.
+        Cosine decay value.
 
     """
-    assert step >= 0, "Current number can't be negative"
-    assert max_steps > 0, "Total epochs must be > 0"
-    assert step <= max_steps, f"Current step is bigger than total number of steps. Found {step} > {max_steps}."
+    if step < 0:
+        raise ValueError("Current epoch number can't be negative")
```
guarin, line 491: Should be "step" instead of "epoch"

guarin: Also in the other errors.

lightly/models/utils.py
```diff
@@ -484,12 +484,15 @@ def cosine_schedule(
...
+    if max_steps < 1:
+        raise ValueError("Total epochs number must be >= 1")
+    if step >= max_steps:
+        raise ValueError(f"Current step is bigger than total number of steps. Found {step} > {max_steps}.")
```
guarin, line 495: The check and the error message do not say the same `step >= max_steps` and `step > max_steps`. 

The error message might be more helpful if it said something along the lines: `The current step must be smaller than max_steps but found step {step} and max_steps {max_steps}.`

guarin (changes requested):

guarin (approved):
Awesome!

PR #1019, opened 2022-12-23
Use queue prototypes only when enough features are queued

ibro45:
As implemented here https://github.com/facebookresearch/swav/blob/06b1b7cbaf6ba2a792300d79c7299db98b93b7f9/main_swav.py#L296


philippmwirth (approved):
Looks good, thanks a lot!

PR #1021, opened 2022-12-29
Add new Model: VICRegL

Ra1Nik:
- Add example
- Add global loss from VICReg
- Add custom collate which passes locations other than images


lightly/transforms/random_crop_and_flip_with_grid.py
```diff
@@ -0,0 +1,147 @@
...
+        ).forward(img_and_location)
+
+        img_and_location = RandomHorizontalFlipReturnsIfFlip(self.hf_prob).forward(
+            img_and_location
+        )
+        transform = T.ToTensor()
+        image_tensor = transform(img_and_location.image)
+        image_tensor = T.Normalize(
+            mean=self.normalize["mean"], std=self.normalize["std"]
+        ).forward(image_tensor)
```
MalteEbner, line 112: Can you move all the created classes to be created in the init? This should make it faster.
In particular
- RandomResizedCropWithLocation
- RandomHorizontalFlipReturnsIfFlip
- T.ToTensor
- T.Normalize
The instance created of these classes are static. thus they can created in the init. Then you can reuse them in the forward pass.

Ra1Nik:
For the reviewer:
- the model is still very slow, suggestion on how to improve speed performance are welcomed
- I set a lower alpha than the one in the paper because the standard (global) VICReg loss is diverging with high alpha values during optimization. I guess it is due to the fact that the collate function is different from the one used in VICReg example.

lightly/data/collate.py
```diff
@@ -1010,6 +1011,112 @@ def __init__(
...
+            batch:
+                A list of tuples containing an image (as a PIL Image), a label (int), and a filename (str).
+
+        Returns:
+            A tuple of transformed images (as a 4-tuple of torch.Tensors), labels (as a list of ints), and filenames (as a list of strs).
+
+        """
+       
+        transforms_0 = [self.gridcreator0.forward(self.transforms[0](item[0])) for item in batch]
+        transforms_1 = [self.gridcreator1.forward(self.transforms[1](item[0])) for item in batch]
```
guarin, line 1102: Danger! This is super slow. The reason is that the transformations in `self.transforms` happen before the resize & crop. Image transformations are pretty expensive and we should run them always after resize&crop on the smaller images.

lightly/data/collate.py
```diff
@@ -1010,6 +1011,112 @@ def __init__(
...
+
+        """
+        Applies transforms to images in the input batch.
+
+        Args:
+            batch:
+                A list of tuples containing an image (as a PIL Image), a label (int), and a filename (str).
+
+        Returns:
+            A tuple of transformed images (as a 4-tuple of torch.Tensors), labels (as a list of ints), and filenames (as a list of strs).
```
guarin, line 1097: `as a list of ints` -> `as tensor of ints`

lightly/models/utils.py
```diff
@@ -461,3 +461,97 @@ def random_token_mask(
...
+
+    if num_matches is None or num_matches == -1:
+        num_matches = input_maps.size(1)
+
+    topk_values, topk_indices = distances.topk(k=1, largest=False)
+    topk_values = topk_values.squeeze(-1)
+    topk_indices = topk_indices.squeeze(-1)
+
+    sorted_values, sorted_values_indices = torch.sort(topk_values, dim=1)
+    sorted_indices, sorted_indices_indices = torch.sort(sorted_values_indices, dim=1)
```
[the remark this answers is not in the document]
Ra1Nik: redo the function from scratch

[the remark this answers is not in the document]
Ra1Nik: split In two functions

lightly/transforms/random_crop_and_flip_with_grid.py
```diff
@@ -1,44 +1,30 @@
...
+    left: float
     # The height of the crop.
-    h: float
+    height: float
     # The width of the crop.
-    w: float
+    width: float
     # The height of the original image.
-    H: float
+    HEIGHT: float
```
guarin, line 21: Please call this `image_height` instead. Having `height` and `HEIGHT` is a bit confusing. Same for width.

lightly/transforms/random_crop_and_flip_with_grid.py
```diff
@@ -47,29 +33,28 @@ class RandomResizedCropWithLocation(T.RandomResizedCrop):
...
-        location = Location(i=i, j=j, h=h, w=w, H=height, W=width)
-        imageandlocation = ImageAndLocation(
-            image=T.functional.resized_crop(
-                img.image, i, j, h, w, self.size, self.interpolation, antialias=self.antialias
-            ),
-            location=location,
-        )
-        return imageandlocation
+        top, left, height, width = self.get_params(img, self.scale, self.ratio)
+        width, height = T.functional.get_image_size(img)
```
guarin, line 46: Here you overwrite `width` and `height` which was set in the line above.

lightly/transforms/random_crop_and_flip_with_grid.py
```diff
@@ -79,22 +64,51 @@ class RandomHorizontalFlipReturnsIfFlip(T.RandomHorizontalFlip):
...
+            PIL Image or Tensor: Randomly flipped image 
+            Location: Location object with updated location.horizontal_flip parameter
         """
         
         if torch.rand(1) < self.p:
-            img.image = F.hflip(img.image)
-            img.location.flip = True
-        return img
+            img = F.hflip(img)
+            location.horizontal_flip = True
```
guarin, line 79: Technically this should be `location.horizontal_flip = not location.horizontal_flip` because you could have multiple flip transformations. But its fine if you leave it like that.

lightly/transforms/random_crop_and_flip_with_grid.py
```diff
@@ -79,22 +64,51 @@ class RandomHorizontalFlipReturnsIfFlip(T.RandomHorizontalFlip):
...
+            Location: Location object linked to the image
+        Returns:
+            PIL Image or Tensor: Randomly flipped image 
+            Location: Location object with updated location.vertical_flip parameter
+
+        """
+        
+        if torch.rand(1) < self.p:
+            img = F.hflip(img)
+            location.vertical_flip = True
```
guarin, line 108: `F.vflip`

lightly/transforms/random_crop_and_flip_with_grid.py
```diff
@@ -158,58 +167,36 @@ def forward(self, img: Image.Image) -> ImageTensorAndGrid:
 
         """
 
-        location = Location(i=0, j=0, h=0, w=0, H=img.height, W=img.width)
-        img_and_location = ImageAndLocation(image=img, location=location)
-        img_and_location = self.randomresizedcropwithlocation.forward(img_and_location)
```
guarin: Please update the docstring above.

lightly/transforms/random_crop_and_flip_with_grid.py
```diff
@@ -158,58 +167,36 @@ def forward(self, img: Image.Image) -> ImageTensorAndGrid:
...
-                    final_grid[k, self.N - 1 - l] = swap
-        return final_grid
+        cell_width = location.width / self.grid_size 
+        cell_height = location.height / self.grid_size
+        x = torch.linspace(location.left, location.left + location.width, self.grid_size) + (cell_width / 2)
+        y = torch.linspace(location.top, location.top + location.height, self.grid_size) + (cell_height / 2)
+        if location.horizontal_flip:
+            x = torch.flip(x, dims=[0])
+        if location.vertical_flip:
+            y = torch.flip(y, dims=[0])
```
guarin, line 200: Oh nice! Didn't know about `torch.flip`!

guarin (changes requested):
Awesome! Thanks a lot for the changes.

Could you change two additional things?
* Remove the `_a`, `_b`, `_0`, `_1` variable suffixes and use `_global` and `_local` instead. I think this could make the code easier to read.
* Add a test to check if the loss works.

lightly/transforms/random_crop_and_flip_with_grid.py
Ra1Nik: It is less efficient, but indexing like you did is not possible in torch, only in numpy

lightly/models/utils.py
```diff
@@ -518,40 +521,15 @@ def nearest_neighbors(
...
-
-    filtered_input_maps = _batched_index_select(input_maps, 1, indices_selected)
-    filtered_candidate_maps = _batched_index_select(
-        candidate_maps, 1, topk_indices_selected
-    )
+    topk_values, topk_indices = distances.topk(k=1, dim=2, largest=False) # [bsz, map_size_0, 1]
+    topk_values = topk_values.squeeze(-1) # [bsz, map_size_0]
+    topk_indices = topk_indices.squeeze(-1) # [bsz, map_size_0]
+    min_values_1, min_indices_2 = topk_values.topk(k=num_matches, dim=1, largest=False) # [bsz, num_matches]
+    filtered_input_maps = _batched_index_select(input_maps, 1, min_indices_2) # [bsz, num_matches, num_input_maps]
```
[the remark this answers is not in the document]
Ra1Nik: That's true, I guess I have to do a gather as for the candidate maps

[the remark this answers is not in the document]
Ra1Nik: After some thought, I think it is correct. topk_values has the same shape and indexing of input_maps. It is just a max on distances in the maps lets say second dimension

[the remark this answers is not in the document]
guarin: Hmm, maybe I don't understand something. Let's assume we have the following example:

```
topk_values = [0.1, 1.0, 0.2, 0.5]
topk_indices = [3, 1, 2, 0]
_, min_inidices = topk_values.topk(k=2)
```
Now we have `min_indices = [1, 3]`. But we cannot use this one the select the entries from the `input_maps` because `min_indices` is an index into `topk_values`. We know that that we want the entries with `topk_values=[1.0, 0.5]` which are at index `[1, 0]` in the `input_maps`. To get `[1, 0]` we have to do `topk_indices[min_indices]` no?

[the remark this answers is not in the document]
guarin: Aah I see now! Thanks a lot for the explanation! Could you add maybe a couple more comments on what is happening?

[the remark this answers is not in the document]
Ra1Nik: I'll try. But it is very difficult to visualize it and understand fully what's happening without a sketch 😂

guarin (changes requested):

lightly/models/utils.py
```diff
@@ -461,3 +461,82 @@ def random_token_mask(
...
+    
+    # Select num_matches neighbors pairs having the lowest distance value.
+    _, min_indices = topk_values.topk(k=num_matches, dim=1, largest=False) # [bsz, num_matches]
+    
+    # Create the filtered input map with num_matches lowest distance values.
+    filtered_input_maps = _batched_index_select(input_maps, 1, min_indices) # [bsz, num_matches, num_input_maps]
+    
+    # Create candidate maps in the same way as input maps, but using corrispondent candidate values
+    selected_candidate_maps = torch.gather(candidate_maps, 1, topk_indices.unsqueeze(-1).repeat(1,1, input_maps.shape[2])) # [bsz, map_size_0, num_input_maps]
+    filtered_candidate_maps = _batched_index_select(selected_candidate_maps, 1, min_indices) # [bsz, num_matches, num_input_maps]
```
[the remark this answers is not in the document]
Ra1Nik: The only thing I can think of is to create a function that calls lines 539 and 540, but it is less legible than it is now

lightly/models/utils.py
```diff
@@ -461,3 +461,82 @@ def random_token_mask(
...
+    """
+
+    for ii in range(1, len(input.shape)):
+        if ii != dim:
+            index = index.unsqueeze(ii)
+    expanse = list(input.shape)
+    expanse[0] = -1
+    expanse[dim] = -1
+    index = index.expand(expanse)
+    return torch.gather(input, dim, index)
```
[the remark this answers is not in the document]
Ra1Nik: The only thing I can do is to "overfit" to my case, so to hardcode the "expanse" list:

`expanse = [-1, -1, input.shape[2]
index = index.expand(expanse)
return torch.gather(input, dim, index)`

lightly/loss/vicregl_loss.py
```diff
@@ -76,118 +75,158 @@ def __init__(
...
+        
+        Args:
+        z_global: 
+            A tensor of global maps. It has size: [batch_size, global_image_height_crop, global_image_width_crop, global_feature_dimension]
+        z_local: 
+            A tensor of local maps. It has size: [batch_size, local_image_height_crop, local_image_width_crop, local_feature_dimension]
+        grid_global: 
+            A tensor of grids for the global maps. It has size: [batch_size, grid_size, grid_size, 2]
+        grid_local: 
+            A tensor of grids for the local maps. It has size: [batch_size, grid_size, grid_size, 2]
```
guarin, line 127: Should be indented by one more tab:

```
Args:
    z_global:
        ...

File: lightly/transforms/random_crop_and_flip_with_grid.py [hunk 15]
```diff
@@ -109,26 +97,26 @@ def forward(self, img: Image.Image, location:Location) -> Tuple[Image.Image, Loc
         return img, location
 
 class RandomResizedCropAndFlip(nn.Module):
-    """
-        A PyTorch module that applies random cropping and horizontal flipping to an image,
-        and returns the transformed image and a grid tensor used to map the image back to the
-        original image space in an NxN grid.
+    """Flip and crop and image and return grid
```
guarin, line 100: Could you make this a full sentence?

lightly/models/utils.py
```diff
@@ -462,81 +462,51 @@ def random_token_mask(
...
     input_maps: torch.Tensor,
     candidate_maps: torch.Tensor,
     distances: torch.Tensor, 
     num_matches: int
 ) -> Tuple[torch.Tensor, torch.Tensor]:
-
-    """Finds the nearest neighbors of the maps in input_maps in candidate_maps.
+    """Compute nearest neighbors
+    
+    Finds the nearest neighbors of the maps in input_maps in candidate_maps.
```
guarin, line 474:
```suggestion
    """Finds the nearest neighbors of the maps in input_maps in candidate_maps.
```

guarin (commented):
Awesome, just some small formatting left and then this is good to merge. Could you also run black again on the changed files?

lightly/loss/vicregl_loss.py
```diff
@@ -190,31 +190,28 @@ def forward(
     ) -> torch.Tensor:
         """Compute the overall loss between two sets of maps, using global loss and local loss.
 
-        it computes global loss using the VICReg loss module and z_global and z_local, and local loss .
+        It computes global loss using the VICReg loss module and z_global and z_local, and local loss .
```
guarin, line 193:
```suggestion
        Computes the global loss using VicRegLoss and local loss using nearest neighbor search.
```

lightly/loss/vicregl_loss.py
```diff
@@ -190,31 +190,28 @@ def forward(
...
+            z_local_local_features:
+                A tensor of local features for the local maps. It has size: [batch_size, global_image_height_crop, global_image_width_crop, global_feature_dimension]
+            grid_global:
+                A tensor of grids for the global maps. It has size: [batch_size, grid_size, grid_size, 2]
+            grid_local:
+                A tensor of grids for the local maps. It has size: [batch_size, grid_size, grid_size, 2]
 
         Returns:
-        A tensor of the overall loss between the two sets of maps. It has size: [batch_size]"
+            A tensor of the overall loss between the two sets of maps. It has size: [batch_size]"
```
guarin, line 211:
```suggestion
            A tensor of the overall loss between the two sets of maps. It has size: [batch_size].
```

guarin (approved):

PR #1029, opened 2023-01-06
Fix bug in SMoGProjectionHead()

DrIsDr:
Fix bug in SMoGProjectionHead().

The current implementation has the number of BatchNorm1d nodes hard-coded, so if the number of hidden or output layers is modified, it will cause a bug.


philippmwirth (approved):
Thanks a lot! Great catch 🙂

PR #1030, opened 2023-01-06
Add list_datasource_permissions function to the api client.

michal-lightly:
### What changed and why

Expose a function to check datasource permission. This was possible before only by using a private api.

### How was it tested

Manually in repl.

Generated docs:

Screenshot 2023-01-06 at 13 02 09

### DIFF
--- a/lightly/api/api_workflow_datasources.py
+++ b/lightly/api/api_workflow_datasources.py
@@ -3,7 +3,6 @@ from typing import Dict, List, Optional, Tuple, Union
 
 import tqdm
 import warnings
-import re
 
 from lightly.openapi_generated.swagger_client.models.datasource_config import (
     DatasourceConfig,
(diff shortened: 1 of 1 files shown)


michal-lightly:
Question: Is it possible to build and view the docs locally?

lightly/api/api_workflow_datasources.py
```diff
@@ -590,4 +587,30 @@ def get_custom_embedding_read_url(
         return self._datasources_api.get_custom_embedding_file_read_url_from_datasource_by_dataset_id(
             self.dataset_id,
             filename,
-        )
\ No newline at end of file
+        )
+
+    def list_datasource_permissions(
+        self,
```
michal-lightly, line 593: Is it OK that the function uses `client.dataset_id` or should it get the dataset id as the input?

guarin: Yes that is ok, most functions assume that `client.dataset_id` is used.

guarin (approved):

PR #1033, opened 2023-01-10
Niccolo lig 2388 fix vicreg collate function

Ra1Nik:
Add right collate function for VICReg and tests


guarin (approved):
Awesome, thanks a lot! 💯

PR #1041, opened 2023-01-19
Benchmarks on TiCo and VICRegL

Ra1Nik:
Add benchmarks to be run on the machine.

### DIFF
--- a/docs/source/getting_started/benchmarks.rst
+++ b/docs/source/getting_started/benchmarks.rst
@@ -56,7 +56,7 @@ The current benchmark contains the following models:
    "SwaV", 800, 256, 0.899, "554.7 Min", "6.6 GByte"
    "TiCo", 800, 256, 0.672, "321.1 Min", "4.0 GByte"
    "VICReg", 800, 256, 0.783, "316.7 Min", "5.7 GByte"
-   "VICRegL", 800, 256, 0.800, "414.8 Min", "8.6 GByte"
+   "VICRegL", 800, 256, 0.817, "302.0 Min", "4.4 GByte"
 
 (*): Different runtime and memory requirements due to different hardware settings
 and pytorch version. Runtime and memory requirements are comparable to SimCLR
--- a/docs/source/getting_started/benchmarks.rst
+++ b/docs/source/getting_started/benchmarks.rst
@@ -33,7 +33,9 @@ The current benchmark contains the following models:
 - :ref:`SimMiM <simmim>`
 - :ref:`SimSiam <simsiam>`
 - :ref:`SwAV <swav>`
+- :ref:`TiCo <tico>`
 - :ref:`VICReg <vicreg>`
+- :ref:`VICRegL <vicregl>`
 
 
 .. csv-table:: ImageNette benchmark results using kNN evaluation on the test set using 128x128 input resolution.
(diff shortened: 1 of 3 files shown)


docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -142,40 +147,42 @@
...
 
-# Single crop augmentation for MAE
+# Single crop augmentation for MAE
 mae_collate_fn = lightly.data.MAECollateFunction()
 
 # Multi crop augmentation for MSN
 msn_collate_fn = lightly.data.MSNCollateFunction(random_size=128, focal_size=64)
 
+# Collate function passing geometrical transformation for VICRegL
+vicregl_collate_fn = lightly.data.VICRegLCollateFunction()
```
guarin, line 157: Oh no, we usually use smaller image sizes than default for the imagenette benchmarks as the images are only 160px large. The global crop size is controlled by `input_size` variable set at the top of the file. And I would set local size to 64.

docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -1048,8 +1086,140 @@ def training_step(self, batch, batch_index):
...
+            return epoch / self.warmup_epochs
+        else:
+            return 0.5 * (
+                1.0
+                + math.cos(
+                    math.pi
+                    * (epoch - self.warmup_epochs)
+                    / (max_epochs - self.warmup_epochs)
+                )
+            )
```
guarin, line 1110: Why did you have to change the `VICRegModel` as well? And I saw there were also some changes to `SimMIM`.

guarin (changes requested):
Nice PR! Also thanks a lot for formatting :)

The code looks great! But I believe the models could have been trained on the wrong image size, or did you revert a change? Could you rerun them on the smaller image size if necessary?

docs/source/getting_started/benchmarks/imagenette_benchmark.py
Ra1Nik: It's only formatting and adding a comment

docs/source/getting_started/benchmarks.rst
```diff
@@ -56,7 +56,7 @@ The current benchmark contains the following models:
    "SwaV", 800, 256, 0.899, "554.7 Min", "6.6 GByte"
    "TiCo", 800, 256, 0.672, "321.1 Min", "4.0 GByte"
    "VICReg", 800, 256, 0.783, "316.7 Min", "5.7 GByte"
-   "VICRegL", 800, 256, 0.800, "414.8 Min", "8.6 GByte"
+   "VICRegL", 800, 256, 0.817, "302.0 Min", "4.4 GByte"
```
guarin, line 60: 😍

lightly/models/utils.py
```diff
@@ -498,7 +498,7 @@ def nearest_neighbors(
         They both have size: [batch_size, input_map_size, feature_dimension]
     """
 
-    if num_matches is None or num_matches == -1:
+    if num_matches is None or num_matches == -1 or num_matches > input_maps.size(1):
```
guarin, line 501: Would it make sense to raise an error if `num_matches > input_maps.size(1)`?

Ra1Nik: I thought about that, but to be honest, if something like that happens I would love not to get interrupted. Like the -1. If I put more matches, it makes sense to just take them all imho.

Ra1Nik: The `num_matches` parameter is useful only if it is less than `input_maps.size(1)`. Otherwise, I would just not interrupt stuff.

guarin (approved):

PR #1043, opened 2023-01-23
Guarin lig 2221 add tag info to runs pip

guarin:
## Changes

* Add `get_compute_worker_run_tags` endpoint

## How was it tested?

* Added unit tests
* Manual tests

### DIFF
--- a/tests/api_workflow/test_api_workflow_compute_worker.py
+++ b/tests/api_workflow/test_api_workflow_compute_worker.py
@@ -601,7 +601,7 @@ def test_get_compute_worker_run_tags__multiple_tags(mocker: MockerFixture) -> No
     tag_1 = TagData(
         id="tag-1",
         dataset_id="dataset-0",
-        prev_tag_id=None,
+        prev_tag_id="tag-0",
         bit_mask_data="0x1",
         name="tag-1",
         tot_size=0,
--- a/tests/api_workflow/test_api_workflow_compute_worker.py
+++ b/tests/api_workflow/test_api_workflow_compute_worker.py
@@ -26,16 +26,13 @@ from lightly.openapi_generated.swagger_client import (
     DockerWorkerConfig,
     DockerWorkerType,
     DockerRunArtifactData,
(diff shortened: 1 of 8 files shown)


tests/api_workflow/test_api_workflow_compute_worker.py
```diff
@@ -557,6 +554,82 @@ def test_download_compute_worker_run_artifacts(mocker: MockerFixture) -> None:
...
+        name="tag-0",
+        tot_size=0,
+        created_at=0,
+        changes=dict(),
+        run_id="run-0",
+    )
+    tag_1 = TagData(
+        id="tag-1",
+        dataset_id="dataset-0",
+        prev_tag_id=None,
```
japrescott, line 604: technically should be the id of tag_0

tests/api_workflow/test_api_workflow_compute_worker.py
```diff
@@ -557,6 +554,82 @@ def test_download_compute_worker_run_artifacts(mocker: MockerFixture) -> None:
...
+        tot_size=0,
+        created_at=1,
+        changes=dict(),
+        run_id="run-0",
+    )
+    # tag from a different dataset
+    tag_2 = TagData(
+        id="tag-2",
+        dataset_id="dataset-1",
+        prev_tag_id=None,
```
japrescott, line 616: id of tag_0 or tag_1

japrescott (approved):

tests/api_workflow/test_api_workflow_compute_worker.py
guarin: Ah yes, thanks!

tests/api_workflow/test_api_workflow_compute_worker.py
guarin: I think this one is correct because the `dataset_id` is different.

japrescott: ah true sorry, didn't see that

PR #1044, opened 2023-01-24
Update LightningModule import

guarin:
## Changes

* Import `LightningModule` from `pytorch_lightning` instead of `pytorch_lightning.core.lightning`
* Fix exit status of weekly dependency test action
* Move weekly dependency from Tuesday to Thursday

Old failing action with broken exit status: https://github.com/lightly-ai/lightly/actions/runs/3990939451/jobs/6845182937
New failing action with fixed exit status: https://github.com/lightly-ai/lightly/commit/0c147d32c1fd4736aff4fe214e42688a648ff2f5/checks
New passing action with fixed exit status and fixed imports: https://github.com/lightly-ai/lightly/actions/runs/3996313813

### DIFF
--- a/.github/workflows/weekly_dependency_test.yml
+++ b/.github/workflows/weekly_dependency_test.yml
@@ -7,7 +7,7 @@ name: Weekly Dependency Test
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 22 * * mon' # each Monday night, at 22:00 UTC
+    - cron: '0 22 * * THU' # each Thursday night, at 22:00 UTC
 
 jobs:
   test_fresh_install:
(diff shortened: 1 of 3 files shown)


.github/workflows/weekly_dependency_test.yml
```diff
@@ -7,7 +7,7 @@ name: Weekly Dependency Test
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 22 * * mon' # each Monday night, at 22:00 UTC
+    - cron: '0 22 * * THU' # each Thursday night, at 22:00 UTC
```
philippmwirth, line 10: Is there a reason you changed this to all caps?

philippmwirth (changes requested):
I think you forgot all the examples, benchmarks, and docs.

guarin:
> I think you forgot all the examples, benchmarks, and docs.

They already use the correct import :)

.github/workflows/weekly_dependency_test.yml
guarin: Caps seems to be the default and VSCode marks lowercase as wrong.

philippmwirth (approved):

PR #1045, opened 2023-01-25
Malte lig 1455 expose default lightly config

MalteEbner:
## Description

Adds a function to get the default lighty config. It is used in https://github.com/lightly-ai/lightly-core/pull/2292

## Tests

See new unittest


tests/cli/test_cli_get_lighty_config.py
```diff
@@ -0,0 +1,11 @@
+from lightly.cli.config.get_config import get_lightly_config
+
+
+def test_get_lightly_config() -> None:
+    conf = get_lightly_config()
+    # Assert some default values
+    assert conf.append == False
```
philippmwirth, line 7:
```suggestion
    assert not conf.append
```

philippmwirth (approved):

lightly/cli/config/get_config.py
```diff
@@ -0,0 +1,9 @@
+from pathlib import Path
```
michal-lightly, line 1: I'm not very familiar with `lightly` package layout, it looks a bit surprising that this is in the `cli` submodule. Briefly looking at the folder structure `utils` could be suitable. But I believe you know it better so I'll trust your judgement 🙂

michal-lightly (approved):

lightly/cli/config/get_config.py
MalteEbner: The `config.yaml` is in the config directory, thus I also put the function here.

PR #1046, opened 2023-01-26
Add SwaV queue PL example and docs

Ra1Nik:
Add Pytorch Lightning example and docs


Ra1Nik:
Rendered SwaV docs:
[SwaV — lightly 1.2.43 documentation.pdf](https://github.com/lightly-ai/lightly/files/10510468/SwaV.lightly.1.2.43.documentation.pdf)

examples/pytorch_lightning/swav_queue.py
```diff
@@ -0,0 +1,155 @@
...
+    def __init__(
+        self,
+        num_ftrs,
+        out_dim,
+        n_prototypes,
+        n_queues,
+        queue_length=0,
+        start_queue_at_epoch=0,
+        n_steps_frozen_prototypes=0,
+    ):
```
guarin, line 28: We usually hardcode the different parameters in the examples, see `examples/pytorch_lightning/swav.py`

examples/pytorch_lightning/swav_queue.py
```diff
@@ -0,0 +1,155 @@
...
+        self.prototypes = SwaVPrototypes(
+            out_dim, n_prototypes, n_steps_frozen_prototypes
+        )
+        self.queues = None
+        if n_queues > 0:
+            self.queues = [MemoryBankModule(size=queue_length) for _ in range(n_queues)]
+            self.queues = nn.ModuleList(self.queues)
+            self.queue_length = queue_length
+            self.num_features_queued = 0
+            self.start_queue_at_epoch = start_queue_at_epoch
```
guarin, line 42: I would ignore the case with no queues and just set it directly:

```suggestion
        self.start_queue_at_epoch = 15
        self.queues = nn.ModuleList([MemoryBankModule(size=3840) for _ in range(2)])
```

examples/pytorch_lightning/swav_queue.py
```diff
@@ -0,0 +1,155 @@
...
+        self.criterion = SwaVLoss()
+        self.step = 0
```
guarin, line 44: LightningModules have a `self.global_step` attribute that you can use instead. No need to count the steps yourself.

examples/pytorch_lightning/swav_queue.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def forward(self, high_resolution, low_resolution, epoch=None, step=None):
```
guarin, line 46: You can remove the `epoch` and `step` parameters and use `self.current_epoch` and `self.global_step` instead.

guarin, line 46: I would actually suggest to remove the `forward` method completely. The forward method is intended for evaluation and not for training. As we only show training we don't really need it. I would move all its contents to the `training_step` method.

examples/pytorch_lightning/swav_queue.py
```diff
@@ -0,0 +1,155 @@
...
+            _, features = self.queues[i](high_resolution_features[i], update=True)
+            # Queue features are in (num_ftrs X queue_length) shape, while the high res
+            # features are in (batch_size X num_ftrs). Swap the axes for interoperability.
+            features = torch.permute(features, (1, 0))
+            queue_features.append(features)
+
+        # Do not return queue prototypes if not enough features have been queued
+        self.num_features_queued += high_resolution_features[0].shape[0]
+        if self.num_features_queued < self.queue_length:
+            return None
```
guarin, line 91: Is this in the original code?

examples/pytorch_lightning/swav_queue.py
```diff
@@ -0,0 +1,155 @@
...
+        # Assign prototypes
+        queue_prototypes = [self.prototypes(x) for x in queue_features]
+        return queue_prototypes
+
+    def training_step(self, batch, batch_idx):
+        batch_swav, _, _ = batch
+        high_resolution, low_resolution = batch_swav[:2], batch_swav[2:]
+        high_resolution, low_resolution, queue = model(
+            high_resolution, low_resolution, self.current_epoch, self.step
+        )
```
guarin, line 113: You shouldn't call `model` here as it references the `model` object outside the class. You can call `self(high_resolution, ...)` instead.

guarin (changes requested):
Awesome! Some cleanup is needed to make it consistent with our other examples. I added some comments :)

examples/pytorch_lightning/swav_queue.py
Ra1Nik: Thanks, I didn't know. I imagined that there was something like that but I didn't manage to find it.

Ra1Nik:
In general, I translated everything from the example in Pytorch. I will also apply the suggestions to that example

docs/source/examples/swav.rst
```diff
@@ -3,7 +3,9 @@
 SwaV
 ====
 
-Example implementation of the SwaV architecture.
+Example implementation of the SwaV architecture. This model takes advantage of contrastive methods without requiring to compute pairwise comparisons. 
+Specifically, this method simultaneously clusters the data while enforcing consistency between cluster assignments produced for different augmentations of the same image,
+instead of comparing features directly as in contrastive learning. It can be trained with large and small batches and can scale to virtually unlimited amounts of data.
```
guarin, line 8:
```suggestion
instead of comparing features directly as in contrastive learning. It can be trained with large and small batch sizes.
```

examples/pytorch/swav_queue.py
```diff
@@ -13,21 +13,14 @@
...
+        self.prototypes = SwaVPrototypes(128, 512, 10)
 
-        self.queues = None
-        if n_queues > 0:
-            self.queues = [MemoryBankModule(size=queue_length) for _ in range(n_queues)]
-            self.queues = nn.ModuleList(self.queues)
-            self.queue_length = queue_length
-            self.num_features_queued = 0
-            self.start_queue_at_epoch = start_queue_at_epoch
+        self.start_queue_at_epoch = 15
```
guarin, line 22: Let's set this to 2 because the example has only 10 epochs.

examples/pytorch_lightning/swav_queue.py
guarin: I believe you forgot this one.

examples/pytorch_lightning/swav_queue.py
```diff
@@ -0,0 +1,120 @@
...
+class SwaV(pl.LightningModule):
+    def __init__(self):
+        super().__init__()
+        resnet = torchvision.models.resnet18()
+        self.backbone = nn.Sequential(*list(resnet.children())[:-1])
+        self.projection_head = SwaVProjectionHead(512, 512, 128)
+        self.prototypes = SwaVPrototypes(
+            128, 512, 10
+        )
+        self.start_queue_at_epoch = 15
```
guarin, line 27: Set to 2

examples/pytorch_lightning/swav_queue.py
```diff
@@ -0,0 +1,120 @@
...
+    def _subforward(self, input):
+        features = self.backbone(input).flatten(start_dim=1)
+        features = self.projection_head(features)
+        features = nn.functional.normalize(features, dim=1, p=2)
+        return features
+
+    @torch.no_grad()
+    def _get_queue_prototypes(self, high_resolution_features):
+        if self.queues is None:
+            return None
```
guarin, line 41: Drop

examples/pytorch_lightning/swav_queue.py
```diff
@@ -0,0 +1,120 @@
...
+            queue_features.append(features)
+
+        # If loss calculation with queue prototypes starts at a later epoch,
+        # just queue the features and return None instead of queue prototypes.
+        if self.start_queue_at_epoch > 0:
+            if self.current_epoch is None:
+                raise ValueError(
+                    "The epoch number must be passed to the `forward()` "
+                    "method if `start_queue_at_epoch` is greater than 0."
+                )
```
guarin, line 65: Drop

examples/pytorch/swav_queue.py
guarin: Please also remove the 
```
        if self.queues is None:
            return None
```

part in `_get_queue_prototypes `

examples/pytorch/swav_queue.py
```diff
@@ -13,21 +13,14 @@
...
+        self.queues = nn.ModuleList([MemoryBankModule(size=3840) for _ in range(2)])
 
     def forward(self, high_resolution, low_resolution, epoch=None, step=None):
         self.prototypes.normalize()
```
guarin, line 26: Does it make sense that `epoch` and `step` can be `None`? Or would it make more sense to always pass them?

examples/pytorch_lightning/swav_queue.py
```diff
@@ -0,0 +1,120 @@
...
+            if self.current_epoch < self.start_queue_at_epoch:
+                return None
+
+        # Assign prototypes
+        queue_prototypes = [self.prototypes(x) for x in queue_features]
+        return queue_prototypes
+
+    def training_step(self, batch, batch_idx):
```
guarin, line 73: Let's move the `training_step` and `configure_optimizers` methods just below the `__init__` method. It is best to put first the public methods and then the "private" methods starting with an underscore.

examples/pytorch/swav_queue.py
```diff
@@ -17,20 +17,24 @@ def __init__(self, backbone):
...
 
         high_resolution_features = [self._subforward(x) for x in high_resolution]
         low_resolution_features = [self._subforward(x) for x in low_resolution]
 
-        high_resolution_prototypes = [self.prototypes(x, step) for x in high_resolution_features]
-        low_resolution_prototypes = [self.prototypes(x, step) for x in low_resolution_features]
-        queue_prototypes = self._get_queue_prototypes(high_resolution_features, epoch)
+        high_resolution_prototypes = [
+            self.prototypes(x, step) for x in high_resolution_features
+        ]
```
guarin, line 33: I just checked the swav paper and they freeze the prototypes by epochs and not iterations. I guess we can just pass the epoch here instead of the step. Then we also don't have to count the steps in an extra variable.

examples/pytorch/swav_queue.py
```diff
@@ -17,20 +17,24 @@ def __init__(self, backbone):
         super().__init__()
         self.backbone = backbone
         self.projection_head = SwaVProjectionHead(512, 512, 128)
-        self.prototypes = SwaVPrototypes(128, 512, 10)
+        self.prototypes = SwaVPrototypes(128, 512, 5)
```
guarin, line 20: The paper freezes the prototypes for one epoch.

guarin (commented):
Nice! It would be great if we could also drop the step parameter.

examples/pytorch/swav_queue.py
Ra1Nik: That's true. I'll change that in both examples.

guarin (approved):

PR #1047, opened 2023-01-26
Create CosineWarmupScheduler class in benchmarks

Ra1Nik:
Add CosineWarmupScheduler class in benchmarks for imagenette


docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -237,6 +234,25 @@ def get_data_loaders(batch_size: int, model):
...
 
+class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
+    def __init__(self, optimizer, warmup_epochs, last_epoch=-1, verbose=False):
+        self.warmup_epochs = warmup_epochs
+        super().__init__(optimizer, self.scale_lr, last_epoch, verbose=False)
+
+    def scale_lr(self, epoch):
+        if epoch < self.warmup_epochs:
+            return epoch / self.warmup_epochs
+        else:
```
Ra1Nik, line 245: I could also do here:

`return 0.5*utils.cosine_schedule(epoch - self.warmup_epochs, max_epochs - self.warmup_epochs, 0, 1)`

But I think it hurts legibility

guarin: It is probably better to use the existing `cosine_schedule` method as it is well tested. Also I believe you don't need the `0.5` factor.

docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -237,6 +234,25 @@ def get_data_loaders(batch_size: int, model):
     return dataloader_train_ssl, dataloader_train_kNN, dataloader_test
 
 
+class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
```
guarin, line 237: As we use this class in many places it would maybe be better to move it into its own file. I would propose `lightly/utils/scheduler.py`. Could you also move the `cosine_schedule` method from `lightly/models/utils.py` there and update all the imports?

guarin (changes requested):
Thanks a lot for the PR! This will make things a lot cleaner. Added some comments :)

tests/utils/test_scheduler.py
```diff
@@ -0,0 +1,29 @@
+import unittest
+import torch
+from torch import nn
+
+from lightly.utils import cosine_schedule
+from lightly.utils import CosineWarmupScheduler
+
+
+class TestModelUtils(unittest.TestCase):
```
guarin, line 9:
```suggestion
class TestScheduler(unittest.TestCase):
```

tests/models/test_ModelUtils.py
```diff
@@ -222,73 +230,56 @@ def _test_random_token_mask_parameters(self, device):
...
         assert candidate_maps_filtered.shape == (3, 1, 2)
-        assert candidate_maps_filtered.equal(torch.tensor([
-            [[1, 1]],
-            [[3, 3]],
-            [[1, 1]]
-        ]))
-
-
-    def test_cosine_schedule(self):
-        momentum_0 = cosine_schedule(1, 10, 0.99, 1)
```
guarin, line 281: Great that you moved the tests!

guarin (commented):
Nice! Forgot something in the last review but we can merge this soon.

guarin (approved):
Awesome!

PR #1048, opened 2023-01-27
add user agent header with lightly version

MalteEbner:
## Description
- When using the ApiWorkflowClient, any API calls include the Lightly version and platform in the user agent of the header.
- Added unittest for this.

This does NOT finish the complete issue, but only the pip part of it. However, I need to merge it to then use it in the docker.

## How was it tested?
- New unittest.
- Ensuring through manual debugging, that the header is passed through all the connection functions:
imageimage

### DIFF
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -97,7 +97,7 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin,
 
         configuration = get_api_client_configuration(token=token)
         self.api_client = ApiClient(configuration=configuration)
-        self.user_agent = f"Lightly/{__version__}/python ({platform.platform()})"
+        self.api_client.user_agent = f"Lightly/{__version__}/python ({platform.platform()})"
         self.set_request_timeout(DEFAULT_API_TIMEOUT)
 
         self.token = configuration.api_key["token"]
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -97,7 +97,7 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin,
 
         configuration = get_api_client_configuration(token=token)
         self.api_client = ApiClient(configuration=configuration)
-        self.api_client.user_agent = f"Lightly/{__version__}/python/{platform.python_version()} (platform; {platform.platform()})"
+        self.api_client.user_agent = f"Lightly/{__version__} ({platform.system()}/{platform.release()}; {platform.platform()}; {platform.processor()};) python/{platform.python_version()}"
         self.set_request_timeout(DEFAULT_API_TIMEOUT)
 
         self.token = configuration.api_key["token"]
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -97,7 +97,7 @@ class ApiWorkflowClient(_UploadEmbeddingsMixin,
 
         configuration = get_api_client_configuration(token=token)
         self.api_client = ApiClient(configuration=configuration)
-        self.api_client.user_agent = f"Lightly/{__version__}/python ({platform.platform()})"
+        self.api_client.user_agent = f"Lightly/{__version__}/python/{platform.python_version()} (platform; {platform.platform()})"
         self.set_request_timeout(DEFAULT_API_TIMEOUT)
 
         self.token = configuration.api_key["token"]
--- a/tests/api_workflow/mocked_api_workflow_client.py
+++ b/tests/api_workflow/mocked_api_workflow_client.py
@@ -1037,7 +1037,6 @@ class MockedApiWorkflowClient(ApiWorkflowClient):
 
     def __init__(self, *args, **kwargs):
         lightly.api.version_checking.VersioningApi = MockedVersioningApi
-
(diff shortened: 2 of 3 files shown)


tests/api_workflow/mocked_api_workflow_client.py
```diff
@@ -1075,7 +1036,6 @@ class MockedApiWorkflowClient(ApiWorkflowClient):
     n_embedding_rows_on_server = N_FILES_ON_SERVER
 
     def __init__(self, *args, **kwargs):
-        lightly.api.api_workflow_client.ApiClient = MockedApiClient
```
MalteEbner, line 1078: This line was overwriting the `ApiClient` globally (for all tests), so that it was raising an error if you accidentally called it.
The global overwrite caused the new unittest to fail, thus I had to remove it. It was not needed anyway and only helped you when writing new unittests, as it improved the error message.

lightly/api/api_workflow_client.py
```diff
@@ -97,7 +97,7 @@ def __init__(
 
         configuration = get_api_client_configuration(token=token)
         self.api_client = ApiClient(configuration=configuration)
-        self.user_agent = f"Lightly/{__version__}/python ({platform.platform()})"
```
MalteEbner, line 100: This line is the only change needed.

japrescott, line 100: how does this allow one to pass the worker version together with the python version as the user agent when the pip package is used within the worker?

japrescott (changes requested):

lightly/api/api_workflow_client.py
MalteEbner: I don't want to expose the user agent publicly, as it should not be used publicly
In the worker, I would just set it via
```
client: ApiWorkflowClient
client.api_client.user_agent == f"Lightly/{__version__}/python ({platform.platform()}) ({worker_version})"
```

japrescott: Okey, thats a workable solution if its possible to import `__version__ ` from `lightly` 👍 

Maybe make the string
`f"Lightly/{__version__}/python (platform; {platform.platform()}) LightlyWorker/{worker_version}"`

Maybe you can also send some other information with the platform like the python version used? Wdyt? What could be valuable to know for us?

MalteEbner: Good idea to include the python version, then we can decide more easily in the future if we can deprecate support for an older version. I changed it to:
f"Lightly/{__version__}/python/{platform.python_version()} (platform; {platform.platform()})"

MalteEbner: > Okey, thats a workable solution if its possible to import __version__  from lightly

Yes, you can import it. I like your proposal with `LightlyWorker/{worker_version}`, but let's discuss it in the lightly-core PR.

tests/api_workflow/mocked_api_workflow_client.py
MalteEbner: Now that we write new unittests differently, we don't need it anymore.

lightly/api/api_workflow_client.py
```diff
@@ -97,7 +97,7 @@ def __init__(
...
+        self.api_client.user_agent = f"Lightly/{__version__}/python/{platform.python_version()} (platform; {platform.platform()})"
```
japrescott, line 100: Following the syntax of https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#syntax

it should rather be;
```
f"Lightly/{__version__} ({platform.system()}/{platform.release()}; {platform.platform()}; {platform.processor()};) python/{platform.python_version()}"
```

wdyt?

MalteEbner: Makes sense to me, it yields
`Lightly/1.2.43 (Darwin/22.2.0; macOS-10.16-x86_64-i386-64bit; i386;) python/3.9.7`

japrescott: nice

tests/api_workflow/test_api_workflow_client.py
```diff
@@ -79,3 +80,8 @@ def raise_connection_error(*args, **kwargs):
                     file=mock.Mock(),
                     signed_write_url='',
                 )
+
+def test_user_agent_header() -> None:
+    client = ApiWorkflowClient(token="")
```
philippmwirth, line 85: Is this mocked now?

MalteEbner: No, this test is completely unmocked.

philippmwirth: Ok, I just saw that it works anyways 👍 

Can you please change the test as follows:
```python

 # patch lightly.__version__, platform.system, platform.release, platform.platform, platform.processor and platform.python_version here

client = ApiWorkflowClient(token"")
assert client.api_client.user_agent == "Lightly/mock_version (mock_system/mock_release; mock_platform; mock_processor;) python/mock_python_version"

MalteEbner: I mocked everything.

philippmwirth (approved):

japrescott (approved):

PR #1049, opened 2023-01-27
Add unit test for frozen prototypes 

Ra1Nik:
Add unit test


tests/models/modules/test_SwaVFreezing.py
```diff
@@ -0,0 +1,33 @@
...
+    def test_swav_frozen_prototypes(self, device: str = "cpu", seed=0):
+        criterion = torch.nn.L1Loss()
+        linear_layer = torch.nn.Linear(8, 8, bias=False)
+        prototypes = SwaVPrototypes(8, 8, 2)
+        optimizer = torch.optim.SGD(prototypes.parameters(), lr=0.01)
+        for step in range(4):
+            torch.manual_seed(seed)
+            in_features = torch.rand(4, 8, device="cpu")
+            out_features = linear_layer(in_features)
+            target_features = torch.ones(4, 8, device="cpu")
```
guarin, line 17: I think the features only have to be defined once. Same for setting the seed.

```suggestion
        torch.manual_seed(seed)
        in_features = torch.rand(4, 8, device="cpu")
        target_features = torch.ones(4, 8, device="cpu")
        for step in range(4):
            out_features = linear_layer(in_features)
```

tests/models/modules/test_SwaVFreezing.py
```diff
@@ -0,0 +1,33 @@
...
+
+import torch
+
+from lightly.models.modules.heads import SwaVPrototypes
+
+class TestSwavFreezingPrototypes(unittest.TestCase):
+    def test_swav_frozen_prototypes(self, device: str = "cpu", seed=0):
+        criterion = torch.nn.L1Loss()
+        linear_layer = torch.nn.Linear(8, 8, bias=False)
+        prototypes = SwaVPrototypes(8, 8, 2)
```
guarin, line 11: Please use argument names so it is easier to know what the `8` and `2` belong to.

tests/models/modules/test_SwaVFreezing.py
```diff
@@ -0,0 +1,33 @@
...
+            loss = criterion(out_features, target_features)
+            loss.backward()
+            optimizer.step()
+            optimizer.zero_grad()
+            if step == 0:
+                loss0 = loss
+            if step <= 2:
+                self.assertEqual(loss, loss0)
+            if step > 2:
+                self.assertNotEqual(loss, loss0)
```
guarin, line 28: Neat :)

tests/models/modules/test_SwaVFreezing.py
```diff
@@ -0,0 +1,33 @@
+import unittest
+
+import torch
+
+from lightly.models.modules.heads import SwaVPrototypes
+
+class TestSwavFreezingPrototypes(unittest.TestCase):
+    def test_swav_frozen_prototypes(self, device: str = "cpu", seed=0):
```
guarin, line 8: Could you move the test also to `test_ProjectionHeads`? Tests for functions from the same file should also always be in the same file.

tests/models/test_ProjectionHeads.py
```diff
@@ -92,6 +92,22 @@ def test_swav_prototypes(self, device: str = 'cpu', seed=0):
                             y = prototypes(x)
                         self.assertEqual(y.shape[0], batch_size)
                         self.assertEqual(y.shape[1], n_prototypes)
+
+    def test_swav_frozen_prototypes(self, device: str = 'cpu', seed=0):
```
guarin, line 96: No need for this test, the `test_swav_prototypes` covers this already.

guarin (changes requested):
Great stuff! I think you can even drop one of the tests.

guarin (approved):

PR #1050, opened 2023-01-30
Add Swav Queue benchmarks on Imagenette

Ra1Nik:
Add benchmark code for benchmarking swav on Imagenette.

### DIFF
--- a/docs/source/getting_started/benchmarks.rst
+++ b/docs/source/getting_started/benchmarks.rst
@@ -33,6 +33,7 @@ The current benchmark contains the following models:
 - :ref:`SimMiM <simmim>`
 - :ref:`SimSiam <simsiam>`
 - :ref:`SwAV <swav>`
+- :ref:`SwAV Queue <swav_queue>`
 - :ref:`TiCo <tico>`
 - :ref:`VICReg <vicreg>`
 - :ref:`VICRegL <vicregl>`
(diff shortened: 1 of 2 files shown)


docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -1158,6 +1158,88 @@ def configure_optimizers(self):
...
+    def training_step(self, batch, batch_idx):
+        batch_swav, _, _ = batch
+        high_resolution, low_resolution = batch_swav[:2], batch_swav[2:]
+        self.prototypes.normalize()
+
+        high_resolution_features = [self._subforward(x) for x in high_resolution]
+        low_resolution_features = [self._subforward(x) for x in low_resolution]
+
+        high_resolution_prototypes = [
+            self.prototypes(x, self.global_step) for x in high_resolution_features
```
guarin, line 1187: I believe this should be `self.current_epoch` instead of `self.global_step`.

docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -1158,6 +1158,88 @@ def configure_optimizers(self):
...
+        # Assign prototypes
+        queue_prototypes = [self.prototypes(x, self.current_epoch) for x in queue_features]
+        return queue_prototypes
+
+    def configure_optimizers(self):
+        optim = torch.optim.SGD(
+            self.parameters(),
+            lr=4.8 * lr_factor,
+            weight_decay=1e-6,
+        )
```
guarin, line 1239: SwAV uses Adam optimizer. See the `SwAVModel` above :)

guarin (changes requested):

docs/source/getting_started/benchmarks.rst
```diff
@@ -54,6 +55,7 @@ The current benchmark contains the following models:
    "SimMiM", 800, 256, 0.355, "397.8 Min", "10.5 GByte"
    "SimSiam", 800, 256, 0.852, "316.0 Min", "3.9 GByte"
    "SwaV", 800, 256, 0.899, "554.7 Min", "6.6 GByte"
+   "SwaV Queue", 800, 256, 0.894, "273.7 Min", "6.4 GByte"
```
guarin, line 58: How come it is so much faster than the normal SwaV model? 273 min seems super fast compared to the other models.

guarin: And do these results already include the gaussian blur fix?

guarin (approved):
Beautiful 😍

docs/source/getting_started/benchmarks.rst
Ra1Nik: It includes the blur fix, that's why it is so high, and I was also surprised by the timing.

Issue #1051, opened 2023-01-30
Issue in Lightly Gaussian Blur Augmentation


AurelienGauffre:
Hi,
Thank you for the amazing work on this repo. I have identified what might be an issue with the implementation of Lightly's Gaussian Blur that is negatively impacting the performance of most of the Lightly models. The source of the problem appears to be a confusion around the interpretation of the "radius" parameter in PIL's Gaussian Blur filter. Simply removing the blur augmentation currently increases significantly performances on a few Lightly models I have tested. With a quick fix of this blur augmentation, all the models I have tested seem to perform even better (cf. table below), except for SwaV since I have noticed that blur augmentation [has been disabled](https://github.com/lightly-ai/lightly/blob/5f7613b73c59a0418b449f0d7c3e1c7e2262471c/lightly/data/collate.py#L354). This could also explain why SwaV outperforms some models by far in current [Lightly Imagenette Benchmark](https://docs.lightly.ai/self-supervised-learning/getting_started/benchmarks.html) .

## 1. What should be done according to SimClr

According to its documentation, the [ImageCollateFunction()](https://github.com/lightly-ai/lightly/blob/5f7613b73c59a0418b449f0d7c3e1c7e2262471c/lightly/data/collate.py#L90) class from Lightly is designed to replicate the augmentation techniques outlined in the original [SimCLR](https://arxiv.org/abs/2002.05709) paper. Concerning the Gaussian Blur augmentation parameters, the paper describes in Appendix A. :

"We blur the image 50% of the time using a Gaussian kernel. We randomly sample σ ∈ [0.1, 2.0], and the kernel size is set to be 10% of the image height/width." Indeed, a proper Gaussian Blur filter requires two parameters to be defined: the size of the convolution kernel (or, equivalently, the radius, which is half the size of the kernel) and the standard deviation (σ) of this Gaussian kernel. 

## 2. How it is currently done in Lightly VS elsewhere

The Gaussian blur PIL filter used in Lightly only allows for the input of one parameter, labeled as "radius". This lead to potential confusion, as it is unclear how this parameter relates to kernel size and σ. This question has already been raised several times including on [Torchvision github](https://github.com/pytorch/vision/issues/5194) or on [Stackoverflow](https://stackoverflow.com/questions/62968174/for-pil-imagefilter-gaussianblur-how-what-kernel-is-used-and-does-the-radius-par), without very clear answers though, since PIL approximates Gaussian Kernel filter with box linear filter.

[...]

PR #1052, opened 2023-01-30
Fix Gaussian Blur 

AurelienGauffre:
Fix the Gaussian Blur as detailed in #1051. This is just a quick fix example, feel free to reject this PR and write your own clean implementation, although I would love to help if I can. 

In particular, this is probably a classical problem, but I am not sure how to handle cleanly the old parameters of [GaussianBlur](https://github.com/lightly-ai/lightly/blob/93db513650a1dcc5a9d791c99519ed77269ead11/lightly/transforms/gaussian_blur.py#L10) filter as scale or kernel_size that we don't need anymore with this fix : either one can delete these two parameters, but it does not ensure compatibility with older version of Lightly (if a user happened to use the Gaussian Blur function, this will lead to an issue), or one can leave these two unused parameters as is with default values, resulting in a not very clean implementation having useless parameters (current state of the PR). 

One important question would be about using the Gaussian Blur either from PIL or from TorchVision.

### DIFF
--- a/lightly/data/collate.py
+++ b/lightly/data/collate.py
@@ -505,7 +505,7 @@ class SwaVCollateFunction(MultiCropCollateFunction):
         cj_prob: float = 0.8,
         cj_strength: float = 0.8,
         random_gray_scale: float = 0.2,
-        gaussian_blur: float = 0.0,
+        gaussian_blur: float = .5,
         kernel_size: float = 1.0,
         normalize: dict = imagenet_normalize,
     ):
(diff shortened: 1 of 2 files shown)


Ra1Nik:
Hi, awesome PR, really great finding! I launched the benchmarks running your code. I will have the results tomorrow. I will be back to you as soon as the computation is over!

Ra1Nik:
I tested the code, and the results are great! Thanks a lot for the finding. You can merge the PR. 

We'll stick to PIL images, so we don't need torchvision v0.8.0 as a requirement for lightly. As the next step, we will create an issue to add a deprecation warning, so we will follow your approach. Thanks again!

guarin (approved):
Thanks a lot!

PR #1061, opened 2023-02-10
Malte lig 2469 pass datasetcreator when creating a

MalteEbner:
Pip part of the issue, the docker part will be based on this.

## Description
Added the DatasetCreator argument to
- ApiWorkflowClient.create_dataset
- ApiWorkflowClient.create_new_dataset_with_unique_name

And added unittests for the new argument.

## Other
- generated new openapi specs
- autoformatted the two changed files with black

PR #1062, opened 2023-02-14
set different dataset_creator in upload_cli

MalteEbner:
Follow-up for the missing part in #1061

### DIFF
--- a/lightly/cli/upload_cli.py
+++ b/lightly/cli/upload_cli.py
@@ -29,6 +29,8 @@ from lightly.cli._helpers import fix_hydra_arguments
 
 from lightly.api.api_workflow_client import ApiWorkflowClient
 from lightly.data import LightlyDataset
+from lightly.openapi_generated.swagger_client import DatasetCreator
+
 SUCCESS_RETURN_VALUE = "Success"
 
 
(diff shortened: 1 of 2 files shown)


guarin (changes requested):
Could you also add it in the download cli?

guarin (approved):

lightly/cli/upload_cli.py
```diff
@@ -74,10 +76,10 @@ def _upload_cli(cfg, is_cli_call=True) -> Union[str, None]:
             cli_api_args_wrong = True
         else:
             api_workflow_client = \
-                ApiWorkflowClient(token=token, dataset_id=dataset_id)
+                ApiWorkflowClient(token=token, dataset_id=dataset_id, dataset_creator=DatasetCreator.USER_PIP_LIGHTLY_MAGIC)
```
japrescott, line 79: could this logic also be applied for the tags which are created by lightly-magic (and other stuff lightly-magic does?)

MalteEbner: Two points I checked:
- Yes, if you upsize using the CLI, the tag creator will be the `USER_PIP`.
- Yes, you could set a different tag creator here. However, I am not sure if we already have one.

In any case, this belongs to a follow-up issue. However, I am not sure about its importance.

MalteEbner:
Only codecov was failing, thus I merged.

lightly/cli/upload_cli.py
japrescott: yes its important as we want the pip users separate in our stats. Will follow up with an issue

PR #1064, opened 2023-02-15
Add set_obs_config

philippmwirth:
# Add set_obs_config

* Add `ApiWorkflowClient.set_obs_config`.
* I tested it manually by configuring a dataset and checking if the access rights are correct.


guarin (approved):
LGTM!

PR #1065, opened 2023-02-16
Fix API client imports part 1 (breaking changes)

philippmwirth:
Part one of a four-step plan to remove the dependency of `lightly.api` on `torch` and `torchvision`.
1. Drop all imports in `lightly.utils.__init__.py` (this PR).
    - Some utils depend on torch and others don't. We don't want to import all of `utils` when accessing only a subset.
    - Change some of the import specs:
    ```python
    # no longer possible!
    from lightly.utils import save_embeddings
    from lightly.utils import save_custom_metadata
    from lightly.utils import version_check
    ... # these are just examples, there are more
  
    # use this instead
    from lightly.utils.io import save_embeddings
    from lightly.utils.io import save_custom_metadata
    from lightly.utils.version_check import version_check
    ...
    ```
2. Drop all imports in `lightly/__init__.py` (will require more changes to how the imports work).
3. Remove references to `LightlyDataset` in the `ApiWorkflowClient`.
4. Drop direct dependency on torch.


docs/source/tutorials_source/platform/tutorial_aquarium_custom_metadata.py
```diff
@@ -94,7 +94,7 @@
 .. code-block:: python
 
     import json
-    from lightly.utils import save_custom_metadata
+    from lightly.utils,io import save_custom_metadata
```
guarin, line 97:
```suggestion
    from lightly.utils.io import save_custom_metadata
```

tests/imports/test_from_imports.py
```diff
@@ -66,16 +66,6 @@ def test_from_imports(self):
         from lightly.transforms import RandomRotate
         from lightly.transforms.rotation import RandomRotate
 
-        # utils imports
-        from lightly.utils import save_embeddings
-        from lightly.utils.io import save_embeddings
-        from lightly.utils import load_embeddings
```
guarin, line 72: I would drop remove the `test_from_imports.py` file completely.

tests/imports/test_seminested_imports.py
```diff
@@ -75,14 +75,3 @@ def test_seminested_imports(self):
         transforms.gaussian_blur.GaussianBlur
         transforms.RandomRotate
         transforms.rotation.RandomRotate
-
-        # utils imports
-        from lightly import utils
-        utils.save_embeddings
-        utils.io.save_embeddings
```
guarin, line 82: I would drop the `test_seminested_imports.py` file completely.

guarin (approved):
LGTM

PR #1066, opened 2023-02-17
Fix API client imports part 2 (breaking changes)

philippmwirth:
Part two of a four-step plan to remove the dependency of `lightly.api` on `torch` and `torchvision`.
1. Drop all imports in `lightly.utils.__init__.py` (#1065).
2. Drop all imports in `lightly/__init__.py` (will require more changes to how the imports work). (This PR)
3. Remove references to `LightlyDataset` in the `ApiWorkflowClient`.
4. Drop direct dependency on torch.

Note: The failing unit tests are just a codecov upload problem.

PR #1067, opened 2023-02-17
Move upload dataset out of ApiWorkflowClient

philippmwirth:
Part three of a four-step plan to remove the dependency of `lightly.api` on `torch` and `torchvision`.
1. Drop all imports in `lightly.utils.__init__.py` (#1065).
2. Drop all imports in `lightly/__init__.py` (will require more changes to how the imports work). (#1066)
3. Remove references to `LightlyDataset` in the `ApiWorkflowClient` (this PR).
4. Drop direct dependency on torch.

Note: The failing unit tests are just a codecov upload problem.


lightly/api/api_workflow_upload_dataset.py
```diff
@@ -35,13 +35,26 @@
...
 
 
+try:
+    from lightly.data import LightlyDataset
+    LightlyDatasetType = LightlyDataset
+    _lightly_dataset_available = True
+except (
+    ImportError,
+    OSError,
+    ModuleNotFoundError
```
guarin, line 45: Could you add a comment in which scenarios these errors can happen?

lightly/api/api_workflow_upload_dataset.py
```diff
@@ -35,13 +35,26 @@
...
+
+
 class _UploadDatasetMixin:
     """Mixin to upload datasets to the Lightly Api.
 
     """
 
     def upload_dataset(self,
-                       input: Union[str, LightlyDataset],
+                       input: Union[str, LightlyDatasetType],
```
guarin, line 57: Can we not set it to `input: Union[str, "LightlyDataset"]`? Then the docs will be nicer as users don't have to figure out what `LightlyDatasetType` is.

philippmwirth: You are right, it's nicer without.

lightly/api/api_workflow_upload_dataset.py
```diff
@@ -35,15 +35,25 @@
 from lightly.openapi_generated.swagger_client.rest import ApiException
 
 
+try:
+    from lightly.data import LightlyDataset
+    _lightly_dataset_available = True
+except (
+    OSError,              # Different CUDA versions for torch and torchvision
+    ModuleNotFoundError,  # No installation of torch or torchvision
```
guarin, line 43: Why could you remove the importerror?

philippmwirth: Ah I removed the wrong one, good catch! `ModuleNotFoundError` is a subclass of `ImportError`. I will change it.

guarin: Can we also update the exception catching in `lightly/__init__.py`?

```
    try:
        import torchvision.models.vision_transformer
        _torchvision_vit_available = True
    except (
        ImportError,          # Old version of torchvision
        OSError,              # Different CUDA versions for torch and torchvision
        ModuleNotFoundError,  # No installation of torchvision
    ):
```

guarin (approved):
LGTM! I left one last comment, feel free to address/resolve.

PR #1068, opened 2023-02-17
Add SimCLR implementation with dataset transforms

Ra1Nik:


lightly/transforms/simclr_transform.py
```diff
@@ -0,0 +1,89 @@
...
+            T.RandomGrayscale(p=random_gray_scale),
+            GaussianBlur(sigmas=sigmas, prob=gaussian_blur),
+        ]
+        if to_tensor:
+            transform.append(T.ToTensor())
+        if normalize:
+            transform += [T.Normalize(mean=normalize["mean"], std=normalize["std"])]
+        self.transform = T.Compose(transform)
+
+    def __call__(self, image):
```
guarin, line 88: Remember to add types before merging.

guarin (commented):
This looks awesome! I gave some feedback :)

lightly/transforms/rotation.py
```diff
@@ -46,3 +48,21 @@ def __call__(self, sample):
         if prob < self.prob:
             sample = TF.rotate(sample, self.angle)
         return sample
+
+class RandomRotationTransform(object):
```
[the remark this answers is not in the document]
Ra1Nik: I did that at the beginning, but then I created a class so I concatenate all classes in the T composition. Why should I prefer a function to a class in this case?

lightly/transforms/imagenet_normalize.py
```diff
@@ -0,0 +1 @@
+imagenet_normalize = {"mean": [0.485, 0.456, 0.406], "std": [0.229, 0.224, 0.225]}
```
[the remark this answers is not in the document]
philippmwirth: Big change: Requires a docs update.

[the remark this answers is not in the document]
guarin: I would still leave the old version in `collate.py` for backwards compatibility.

lightly/transforms/multi_view_transform.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __call__(self, image: Tensor) -> List[Tensor]:
+        """
+        Applies the transforms to the input image.
+
+        Args:
+            Image (Tensor): The input image to apply the transforms to.
+
+        Returns:
+            List[Tensor]: A list of transformed images.
```
guarin, line 25:
```suggestion
            List of views.
```

lightly/transforms/simclr_transform.py
```diff
@@ -0,0 +1,149 @@
...
+        rr_prob:
+            Probability that random rotation is applied.
+        rr_degrees:
+            Range of degrees to select from for random rotation. If rr_degrees is None,
+            images are rotated by 90 degrees. If rr_degrees is a (min, max) tuple,
+            images are rotated by a random angle in [min, max]. If rr_degrees is a
+            single number, images are rotated by a random angle in
+            [-rr_degrees, +rr_degrees]. All rotations are counter-clockwise.
+        normalize:
+            Dictionary with 'mean' and 'std' for torchvision.transforms.Normalize.
```
guarin, line 45: Docstring for `to_tensor` is missing.

lightly/transforms/simclr_transform.py
```diff
@@ -0,0 +1,149 @@
...
+            T.RandomGrayscale(p=random_gray_scale),
+            GaussianBlur(sigmas=sigmas, prob=gaussian_blur),
+        ]
+        if to_tensor:
+            transform.append(T.ToTensor())
+        if normalize:
+            transform += [T.Normalize(mean=normalize["mean"], std=normalize["std"])]
+        self.transform = T.Compose(transform)
+
+    def __call__(self, image: Tensor) -> Tensor:
```
guarin, line 138: Can be tensor or image.

lightly/transforms/utils.py
```diff
@@ -0,0 +1 @@
+IMAGENET_NORMALIZE = {"mean": [0.485, 0.456, 0.406], "std": [0.229, 0.224, 0.225]}
```
guarin, line 1: Lets set in `collate.py` `imagenet_normalize = IMAGENET_NORMALIZE` so we are backwards compatible but it is only defined in one place. Please also add a comment in `collate.py` that `imagenet_normalize` is kept for backwards compatibility.

lightly/transforms/rotation.py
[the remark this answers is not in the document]
guarin: Classes should have one clearly defined purpose. `RandomRotationTransform` currently covers multiple use-cases:
1. If rr_degrees is `None` -> Use RandomRotate with a fixed 90 degrees angle (or multiple of it)
2. if rr_degrees is not `None` -> Random rotation with angle in rr_degrees.

What we could do is add a class `RandomRotateDegrees`:
```
class RandomRotateDegrees:
    def __init__(self, prob: float, degrees: Union[float, Tuple[float, float]]):
        self.transform = T.RandomApply([T.RandomRotation(degrees=degrees)], p=prob)
    
    def __call__(self, image):
        return self.transform(image)
```

And then update the `get_random_rotation_transform` function to either return `RandomRotate` or `RandomRotateDegrees`.

lightly/transforms/simclr_transform.py
```diff
@@ -0,0 +1,149 @@
...
+    Examples:
+
+        >>> # SimCLR for ImageNet
+        >>> collate_fn = SimCLRCollateFunction()
+        >>>
+        >>> # SimCLR for CIFAR-10
+        >>> collate_fn = SimCLRCollateFunction(
+        >>>     input_size=32,
+        >>>     gaussian_blur=0.,
+        >>> )
```
guarin, line 56: Please update

guarin (commented):
This is starting to look great. Added some more comments. Please also add tests for the new classes.

lightly/data/multi_view_collate.py
```diff
@@ -0,0 +1,54 @@
...
+                    ...
+                ],
+                [label_0, label_1, ...],
+                [filename_0, filename_1, ...]
+            )
+
+
+        """
+        if len(batch) == 0:
+            warn("The batch is empty. Collate returned empty lists.")
```
guarin, line 40:
```suggestion
            warn("MultiViewCollate received empty batch.")
```

lightly/transforms/simclr_transform.py
```diff
@@ -135,7 +129,7 @@ def __init__(
             transform += [T.Normalize(mean=normalize["mean"], std=normalize["std"])]
         self.transform = T.Compose(transform)
 
-    def __call__(self, image: Tensor) -> Tensor:
+    def __call__(self, image: Union[Tensor, Image]) -> Tensor:
```
guarin, line 132: Can the output not be an image as well?

tests/data/test_MultiViewCollate.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def test_single_item_batch(self):
+        img = [torch.randn((3, 224, 224)) for _ in range(5)]
+        label = 1
+        fname = "image1.jpg"
+        batch = [(img, label, fname)]
+        views, labels, fnames = self.multi_view_collate(batch)
+        self.assertEqual(len(views), 5)
+        self.assertEqual(views[0].shape, (1, 3, 224, 224))
+        self.assertEqual(len(labels), 1)
```
guarin, line 28: We should also assert that the label is correct.

tests/data/test_MultiViewCollate.py
```diff
@@ -0,0 +1,43 @@
...
+        self.assertEqual(len(fnames), 1)
```
guarin, line 29: We should also assert that the filename is correct.

tests/data/test_MultiViewCollate.py
```diff
@@ -0,0 +1,43 @@
...
+        fname1 = "image1.jpg"
+        img2 = [torch.randn((3, 224, 224)) for _ in range(5)]
+        label2 = 2
+        fname2 = "image2.jpg"
+        batch = [(img1, label1, fname1), (img2, label2, fname2)]
+        views, labels, fnames = self.multi_view_collate(batch)
+        self.assertEqual(len(views), 5)
+        self.assertEqual(views[0].shape, (2, 3, 224, 224))
+        self.assertEqual(len(labels), 2)
+        self.assertEqual(len(fnames), 2)
```
guarin, line 43: Same as above, we should verify that the values are correct.

tests/transforms/test_MultiViewTransform.py
```diff
@@ -0,0 +1,17 @@
+from PIL import Image
+import unittest
+from lightly.transforms.multi_view_transform import MultiViewTransform
+import torchvision.transforms as T
+class TestMultiViewTransform(unittest.TestCase):
+
+    def test_multi_view_on_pil_image(self):
+            for w in range(1, 10):
+                for h in range(1, 10):
```
guarin, line 9: No need to test with different image sizes.

tests/transforms/test_Rotation.py
```diff
@@ -1,14 +1,30 @@
...
+from lightly.transforms.rotation import RandomRotate, RandomRotateDegrees, random_rotation_transform
 
 
-class TestGaussianBlur(unittest.TestCase):
+class TestRotation(unittest.TestCase):
 
-    def test_on_pil_image(self):
+    def test_RandomRotate_on_pil_image(self):
         for w in range(1, 100):
             for h in range(1, 100):
```
guarin, line 11: No need for different sizes.

tests/transforms/test_Rotation.py
```diff
@@ -1,14 +1,30 @@
...
                 random_rotate = RandomRotate()
                 sample = Image.new('RGB', (w, h))
                 random_rotate(sample)
+    
+    def test_RandomRotateDegrees_on_pil_image(self):
+        for w in range(1, 10):
+            for h in range(1, 10):
```
guarin, line 18: No need for different sizes.

tests/transforms/test_Rotation.py
```diff
@@ -1,14 +1,30 @@
...
+                for r in range(1, 90):
```
guarin, line 19: This will run 90 tests. We don't need that many as it will slow down the test suite a lot, I would propose something like this, then we also cover the tuple input types:
```
for degrees in [0, 1, 45, (0, 0), (-15, 30)]:
```

tests/transforms/test_SimCLRTransform.py
```diff
@@ -0,0 +1,22 @@
...
+import unittest
+
+from lightly.transforms.simclr_transform import SimCLRTransform, SimCLRViewTransform
+
+
+class TestSimCLRTransform(unittest.TestCase):
+
+    def test_view_on_pil_image(self):
+        for w in range(1, 10):
+            for h in range(1, 10):
```
guarin, line 11: No need for different input sizes.

tests/transforms/test_SimCLRTransform.py
```diff
@@ -0,0 +1,22 @@
...
+                single_view_transform = SimCLRViewTransform()
+                sample = Image.new('RGB', (w, h))
+                single_view_transform(sample)
+
+    def test_multi_view_on_pil_image(self):
+        for w in range(1, 10):
+            for h in range(1, 10):
```
guarin, line 18: No need for different input sizes

tests/transforms/test_SimCLRTransform.py
```diff
@@ -0,0 +1,22 @@
+from PIL import Image
+import unittest
+
+from lightly.transforms.simclr_transform import SimCLRTransform, SimCLRViewTransform
+
+
+class TestSimCLRTransform(unittest.TestCase):
+
```
guarin, line 8: Could you also add a test with `to_tensor = False` and passing a tensor instead of an image?

tests/transforms/test_SimCLRTransform.py
```diff
@@ -0,0 +1,22 @@
...
+    def test_view_on_pil_image(self):
+        for w in range(1, 10):
+            for h in range(1, 10):
+                single_view_transform = SimCLRViewTransform()
+                sample = Image.new('RGB', (w, h))
+                single_view_transform(sample)
```
guarin, line 14: Please assert that the output is a tensor with the correct shape.

tests/transforms/test_SimCLRTransform.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def test_multi_view_on_pil_image(self):
+        for w in range(1, 10):
+            for h in range(1, 10):
+                multi_view_transform = SimCLRTransform()
+                sample = Image.new('RGB', (w, h))
+                multi_view_transform(sample)
```
guarin, line 21: Please assert that the output is a tensor with the correct shape.

tests/transforms/test_SimCLRTransform.py
```diff
@@ -0,0 +1,22 @@
+from PIL import Image
```
guarin, line 1: Please rename the file to `test_simclr_transform.py`. It is standard to name the file with the tests the same as the original file but just with a `test_` prefix. We still have some old tests that follow a bad practice of naming files according to their class which we shouldn't do.

lightly/transforms/simclr_transform.py
Ra1Nik: No, it can't. Otherwise image normalisation gives error. The ToTensor is optional only if the input is already a tensor

guarin: Image normalization is also optional though.

Ra1Nik: That's also true, but I wouldn't let this possibility as it seems a bit tacky to me. Also in the collate we assume that the batches are of tensors.

lightly/transforms/rotation.py
```diff
@@ -50,10 +50,30 @@ def __call__(self, sample):
         return sample
 
 class RandomRotateDegrees(object):
+    """Random rotate image between two rotation angles with a random probability.
```
guarin, line 53: There is no need to inherit from `object` when defining a new class:
```
class RandomRotateDegrees(object):
```
You can just write it as:
```
class RandomRotateDegrees:
```

Inheritance from object was only required in the old python days. We sadly still have some of this old code around, as you can see in the `RandomRotate` class. Feel free to remove it from there as well.

lightly/transforms/rotation.py
```diff
@@ -50,10 +50,30 @@ def __call__(self, sample):
...
+
+    Attributes:
+        prob:
+            Probability with which image is rotated.
+        degrees (sequence or number): Range of degrees to select from.
+            If degrees is a number instead of sequence like (min, max), the range of degrees
+            will be (-degrees, +degrees).
```
guarin, line 60: We use a different docstring format that pytorch. We don't put the type in the docstring as it is already documented in the type hints and we always put the description on a new line. Like this:
```
        degrees: 
            Range of degrees to select from. If degrees is a number instead of sequence
            like (min, max), the range of degrees will be (-degrees, +degrees).
```

guarin, line 60: Also please add a description how the rotation works. -> We select a random angle in (min, max) or (-degrees, +degrees) and we rotate counter-clockwise.

lightly/transforms/rotation.py
```diff
@@ -50,10 +50,30 @@ def __call__(self, sample):
...
+
+    """
     def __init__(self, prob: float, degrees: Union[float, Tuple[float, float]]):
         self.transform = T.RandomApply([T.RandomRotation(degrees=degrees)], p=prob)
     
     def __call__(self, image):
+        """Rotates the images with a given probability.
+
+        Args:
+            sample:
```
guarin, line 70:
```suggestion
            image:
```

lightly/transforms/rotation.py
```diff
@@ -50,10 +50,30 @@ def __call__(self, sample):
...
+        degrees (sequence or number): Range of degrees to select from.
+            If degrees is a number instead of sequence like (min, max), the range of degrees
+            will be (-degrees, +degrees).
+
+    """
     def __init__(self, prob: float, degrees: Union[float, Tuple[float, float]]):
         self.transform = T.RandomApply([T.RandomRotation(degrees=degrees)], p=prob)
     
     def __call__(self, image):
+        """Rotates the images with a given probability.
```
guarin, line 67: Please add types to function definition.

tests/data/test_multi_view_collate.py
```diff
@@ -0,0 +1,46 @@
...
+    label = 1
+    fname = "image1.jpg"
+    batch = [(img, label, fname)]
+    views, labels, fnames = multi_view_collate(batch)
+    assert len(views) == 5
+    assert views[0].shape == (1, 3, 224, 224)
+    assert len(labels) == 1
+    assert len(fnames) == 1
+    assert labels[0] == label
+    assert fnames[0] == fname
```
guarin, line 27:
```suggestion
    assert labels == [1]
    assert fnames == ["image1.jpg]
```

tests/data/test_multi_view_collate.py
```diff
@@ -0,0 +1,46 @@
...
+    batch = [(img1, label1, fname1), (img2, label2, fname2)]
+    views, labels, fnames = multi_view_collate(batch)
+    assert len(views) == 5
+    assert views[0].shape == (2, 3, 224, 224)
+    assert len(labels) == 2
+    assert len(fnames) == 2
+    assert labels[0] == label1
+    assert fnames[0] == fname1
+    assert labels[1] == label2
+    assert fnames[1] == fname2
```
guarin, line 46:
```suggestion
    assert labels == [1, 2]
    assert fnames == ["image1.jpg", "image2.jpg"]
```

tests/transforms/test_simclr_transform.py
```diff
@@ -0,0 +1,18 @@
...
+    sample = Image.new('RGB', (100, 100))
+    output = single_view_transform(sample)
+    assert output.shape == (3, 32, 32)
+
+def test_multi_view_on_pil_image():
+    multi_view_transform = SimCLRTransform(input_size=32)
+    sample = Image.new('RGB', (100, 100))
+    output = multi_view_transform(sample)
+    assert len(output) == 2
+    assert output[0].shape == (3, 32, 32)
```
guarin, line 18: Let's add the check for the second view as well:
```
assert output[0].shape == (3, 32, 32)
```

guarin (commented):
Awesome! Some small changes are left and then we are good to merge.

Could you also run black on the new files?

lightly/transforms/rotation.py
```diff
@@ -24,58 +26,62 @@ class RandomRotate(object):
...
+
     """
 
     def __init__(self, prob: float = 0.5, angle: int = 90):
         self.prob = prob
         self.angle = angle
 
-    def __call__(self, sample):
-        """Rotates the images with a given probability.
+    def __call__(self, image: Union[Image, Tensor]):
```
guarin, line 38:
```suggestion
    def __call__(self, image: Union[Image, Tensor]) -> Union[Image, Tensor]:
```

lightly/transforms/rotation.py
```diff
@@ -24,58 +26,62 @@ class RandomRotate(object):
...
+            a random angle in the (min, max) range or in the (-degrees, +degrees) range.
 
     """
+
     def __init__(self, prob: float, degrees: Union[float, Tuple[float, float]]):
         self.transform = T.RandomApply([T.RandomRotation(degrees=degrees)], p=prob)
-    
-    def __call__(self, image):
+
+    def __call__(self, image: Union[Image, Tensor]):
```
guarin, line 71:
```suggestion
    def __call__(self, image: Union[Image, Tensor]) -> Union[Image, Tensor]:
```

guarin (commented):
Two more types and then it is perfect!

guarin (approved):

PR #1070, opened 2023-02-20
Add benchmarks results with new GaussianBlur implementation

Ra1Nik:
Add benchmarks results with the new GaussianBlur on imagenet100 and imagenette. Cifar10 is not affected by gaussian blur.


guarin:
Did some extra cleanup for the docs.

Also noticed that the CosineWarmupScheduler was not correctly tested and had a bug. LR went from 0 to 1 instead of 1 to 0.

examples/pytorch_lightning/vicregl.py
```diff
@@ -37,8 +37,8 @@ def forward(self, x):
 
     def training_step(self, batch, batch_index):
         (view_global, view_local, grid_global, grid_local), _, _ = batch
-        z_global, z_global_local_features = model(view_global)
-        z_local, z_local_local_features = model(view_local)
+        z_global, z_global_local_features = self(view_global)
```
philippmwirth, line 40:
```suggestion
        z_global, z_global_local_features = self.forward(view_global)
        z_local, z_local_local_features = self.forward(view_local)
```
to make it more explicit

examples/pytorch_lightning_distributed/vicregl.py
```diff
@@ -36,8 +36,8 @@ def forward(self, x):
     
     def training_step(self, batch, batch_index):
         (view_global, view_local, grid_global, grid_local), _, _ = batch
-        z_global, z_global_local_features = model(view_global)
-        z_local, z_local_local_features = model(view_local)
+        z_global, z_global_local_features = self(view_global)
```
philippmwirth, line 39: Same change as suggested above.

lightly/utils/scheduler.py
```diff
@@ -91,8 +100,8 @@ def scale_lr(self, epoch: int) -> float:
 
         """
         if epoch < self.warmup_epochs:
-            return epoch / self.warmup_epochs
+            return (epoch + 1) / self.warmup_epochs
         else:
             return cosine_schedule(
-                epoch - self.warmup_epochs, self.max_epochs - self.warmup_epochs, 0, 1
+                epoch - self.warmup_epochs, self.max_epochs - self.warmup_epochs, 1.0, 0.0
```
philippmwirth, line 106: Please use keyword arguments here to prevent similar errors from happening

philippmwirth (changes requested):
Looks great, just a few smaller suggestions.

philippmwirth (approved):

PR #1078, opened 2023-02-24
validate the config created by the api

japrescott:
when creating a selection configuration we now only allow to use the v2 config supported by worker2.5 which is type safe


guarin (approved):
LGTM!

PR #1079, opened 2023-02-24
Add deprecation warning to collate functions

Ra1Nik:
After https://github.com/lightly-ai/lightly/pull/1068, the old collate functions will be deprecated. Add deprecation warning.


lightly/data/collate.py
```diff
@@ -36,6 +37,12 @@ class BaseCollateFunction(nn.Module):
     """
 
     def __init__(self, transform: torchvision.transforms.Compose):
+        warn(
```
philippmwirth, line 40: Can you add this function to the bottom of the file and then use it wherever you need it:
```python
def _deprecation_warning_collate_functions() -> None:
    warn(
        "Collate functions are deprecated and will be removed in favor of transforms in v1.4.0.\n"
        "Please use MultiViewCollate in `lightly.data.multi_view_collate` together with the correct transform for your model instead.\n"
        "See https://docs.lightly.ai/self-supervised-learning/examples/models.html for examples."
        category=DeprecationWarning,
    )
```

I'm not super happy with the wording. Maybe you can improve.

Note the change to `DeprecationWarning` for better visibility.

philippmwirth (changes requested):

philippmwirth (commented):
Awesome! I'm not approving yet because of the dependency between this PR and the others but you can consider the issue done 👍

guarin:
@philippmwirth could you approve?

philippmwirth (approved):

PR #1083, opened 2023-02-28
Configv2 typings

japrescott:
better typing for the v2 config endpoint


lightly/api/api_workflow_compute_worker.py
```diff
@@ -114,9 +116,9 @@ def delete_compute_worker(self, worker_id: str):
 
     def create_compute_worker_config(
         self,
-        worker_config: Optional[Dict[str, Any]] = None,
-        lightly_config: Optional[Dict[str, Any]] = None,
-        selection_config: Optional[Union[Dict[str, Any], SelectionConfig]] = None,
+        worker_config: Optional[DockerWorkerConfigV2Docker] = None,
+        lightly_config: Optional[DockerWorkerConfigV2Lightly] = None,
+        selection_config: Optional[SelectionConfig] = None,
```
MalteEbner, line 121: This is the user interface. We had decided to require a `dict` here, not an instance of class `DockerWorkerConfigV2Docker`. Please keep that.

MalteEbner (commented):
One point to keep instead of change, otherwise it is fine.

PR #1089, opened 2023-03-06
add get_compute_worker_run_checkpoint_url

MalteEbner:
## Description

Add a new endpoint to get the read URL of a checkpoint. See https://linear.app/lightly/issue/LIG-2541#comment-e252b52f for code on how this will be used.

Also added a unittest for the new endpoint.

## Tests

- CI
- new unittest
- unmocked test: 

```python
from lightly.api import ApiWorkflowClient
from lightly.api.download import download_and_write_file

client = ApiWorkflowClient(token=token)
run = client.get_compute_worker_runs(dataset_id="63af0acd4719a554515701a0")[-1]
url = client.get_compute_worker_run_checkpoint_url(run=run)
download_and_write_file(url = url, output_path="./checkpoint_test")
```

### DIFF
--- a/lightly/api/api_workflow_artifacts.py
+++ b/lightly/api/api_workflow_artifacts.py
@@ -1,6 +1,9 @@
 import os
 from lightly.api import download
-from lightly.openapi_generated.swagger_client import DockerRunData
+from lightly.openapi_generated.swagger_client import (
+    DockerRunData,
+    DockerRunArtifactData,
+)
 from lightly.openapi_generated.swagger_client.models.docker_run_artifact_type import (
     DockerRunArtifactType,
 )
@@ -353,6 +356,17 @@ class _ArtifactsMixin:
         output_path: str,
         timeout: int,
     ) -> None:
+        artifact = self._get_artifact_by_type(artifact_type, run)
+        self._download_compute_worker_run_artifact(
+            run_id=run.id,
+            artifact_id=artifact.id,
+            output_path=output_path,
+            timeout=timeout,
+        )
+
+    def _get_artifact_by_type(
+        self, artifact_type: str, run: DockerRunData
+    ) -> DockerRunArtifactData:
         if run.artifacts is None:
             raise ArtifactNotExist(f"Run has no artifacts.")
         try:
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_artifacts.py
```diff
@@ -353,6 +356,17 @@ def _download_compute_worker_run_artifact_by_type(
         output_path: str,
         timeout: int,
     ) -> None:
+        artifact = self._get_artifact_by_type(artifact_type, run)
```
MalteEbner, line 359: I moved some code into the new `_get_artifact_by_type` function.

tests/api_workflow/test_api_workflow_artifacts.py
```diff
@@ -160,3 +161,28 @@ def test__download_compute_worker_run_artifact_by_type__no_artifact_with_type(
             output_path="output_dir/checkpoint.ckpt",
             timeout=0,
         )
+
+
+def test__get_compute_worker_run_checkpoint_url(
+    mocker: MockerFixture,
+) -> None:
+    mocked_client = mocker.MagicMock(spec=ApiWorkflowClient)
```
MalteEbner, line 169: Cannot be a `spec_set`, because we overwrite the `_compute_worker_api`

guarin: Did you consider not mocking the full client, like this:
```
def test__get_compute_worker_run_checkpoint_url(
    mocker: MockerFixture,
) -> None:
    client = ApiWorkflowClient(token="123")
    client._compute_worker_api.get_docker_run_artifact_read_url_by_id.return_value = (
        "some_read_url"
    )

    mocked_run = mocker.MagicMock(spec_set=DockerRunData)
    read_url = client.get_compute_worker_run_checkpoint_url(run=mocked_run)
    ...
```

This would be easier to understand and it is clearer what exactly has to be mocked.

guarin (commented):
The implementation looks good to me. I think we could make the test a bit more concise.

tests/api_workflow/test_api_workflow_artifacts.py
MalteEbner: Yes, I considered that option, because it is also used for the other tests, but decided against it:
1. It also tests the `ApiWorkflowClient.__init__`.
2. It calls the version check and thus the API inside `ApiWorkflowClient.__init__`
3.  I know from typing the last worker tests that that option does not work with mypy: It is NOT allowed to mock only some methods of an existing class:

```python
client = ApiWorkflowClient(token="123")
client.some_method = mocker.MagicMock() # raises a mypy assignment error, because you cannot assign a MagicMock() to a Callable.
```

guarin (approved):

PR #1091, opened 2023-03-06
Guarin lig 2684 make api workflow client picklable

guarin:
## What has changed and why?

* Make `ApiWorkflowClient` picklable
* Make `Configuration` picklable
* Add picklable `LightlySwaggerApiClient`
* Add picklable `LightlySwaggerRESTClientObject`
* Refactor `set_api_client_request_timeout` and `patch_rest_client ` and moved the logic to `PatchRESTClientObjectMixin`

The patching of the `ApiClient` and the `RESTClientObject` classes are now done using mixins. This allows us to reuse the mixins in the worker and makes it clearer which patches are applied. Previously, the patches were applied at different times and in different places (`set_api_client_request_timeout`, `patch_rest_client`, ...).

The drawback of this approach is that the original `ApiClient` and `RESTClientObject` classes are not patched, instead, new subclasses `LightlySwaggerApiClient`, and `LightlySwaggerRESTClientObject` were added.

The `Configuration` class is treated differently as many configuration objects are created and we cannot easily subclass it. The issue is that every swagger class takes a configuration as input or creates a default configuration. We would have to be super careful to never forget passing our custom config. Patching the existing class seems saver to me.

All swagger generated classes (except `ApiClient` and `RESTClientObject`) should now also be copyable with `copy.copy` and `copy.deepcopy`.

## How was it tested

* Added unit tests
* Made manual tests to verify that pickling/unpickling an api client works and that requests with the unpickled api client are successful.


lightly/api/patch.py
```diff
@@ -0,0 +1,130 @@
+import logging
+from typing import Any, Dict, Type
+
+
+def rest_client_flatten_array_query_parameters(rest_client_cls: Type):
```
guarin, line 5: Moved this method from `patch_rest_client.py` and renamed the `rest_client` argument to `rest_client_cls`

lightly/api/__init__.py
```diff
@@ -2,12 +2,19 @@
...
 # All Rights Reserved
-
+from lightly.api import patch as _patch
 from lightly.api.api_workflow_client import ApiWorkflowClient
 from lightly.api.api_workflow_artifacts import ArtifactNotExist
-from lightly.api.patch_rest_client import patch_rest_client
 
-from lightly.openapi_generated.swagger_client.rest import RESTClientObject
+from lightly.openapi_generated.swagger_client import ApiClient as _ApiClient, Configuration as _Configuration
+from lightly.openapi_generated.swagger_client.rest import RESTClientObject as _RESTClientObject
```
guarin, line 10: Made these imports private so they cannot be imported with `from lightly.api import Configuration` or similar.

lightly/api/api_workflow_client.py
```diff
@@ -291,11 +291,16 @@ def set_api_client_request_timeout(
 
     """
     request_fn = client.rest_client.request
+    client.rest_client.request = _request_with_timeout(timeout=timeout, request_fn=request_fn)
 
-    def new_request_fn(*args, **kwargs):
+
+class _request_with_timeout:
```
guarin, line 297: Had to refactor the `new_request_fn` function into a global object to make it picklable.

tests/api/test_patch.py
```diff
@@ -0,0 +1,74 @@
...
+    mocker: MockerFixture,
+) -> None:
+    mocker.patch.object(api_workflow_client, "is_compatible_version", return_value=True)
+    client = ApiWorkflowClient(token="123")
+    client._dataset_id = "my-dataset-id"
+    new_client = pickle.loads(pickle.dumps(client))
+
+    assert set(client.__dict__.keys()) == set(new_client.__dict__.keys())
+    assert all(
+        type(client.__dict__[key] == new_client.__dict__[key])
```
MalteEbner, line 19: Why `type()` ? Or should it be around each separately?

MalteEbner (approved):
I have one question regarding a unittest, otherwise well done!

lightly/api/patch.py
```diff
@@ -0,0 +1,131 @@
...
+        "lightly.openapi_generated.swagger_client"
+    )
+    self.logger["urllib3_logger"] = logging.getLogger("urllib3")
+    # Set logger_format and logger_file explicitly because they recreate
+    # logger_formatter, logger_file_handler, and logger_stream_handler which are removed
+    # before pickling.
+    self.logger_format = state["_Configuration__logger_format"]
+    self.logger_file = state["_Configuration__logger_file"]
+    # Set debug explicitly because it has side effects.
+    self.debug = state["_Configuration__debug"]
```
michal-lightly, line 100: How did you figure out what you need to unset and what to reset? Asking because the fields in `__getstate__` and `__setstate__`.

tests/api/test_patch.py
guarin: Great catch, thanks! That should indeed be individual types. Fixed it.

lightly/api/patch.py
guarin: What to unset: Try to pickle a class, if it fails remove the offending attribute.

What to reset: Everything that is removed must be set again (with some exceptions). Make sure that `new_object = unpickle(pickle(object))` has the same attributes as original object.

lightly/api/patch_rest_client.py
```diff
@@ -1,58 +0,0 @@
-from typing import Type
-
-
-def patch_rest_client(rest_client: Type):
-    """
```
guarin, line 5: Code moved to `PatchRESTClientObjectMixin`

tests/api/test_patch.py
```diff
@@ -0,0 +1,21 @@
...
+        type(config.__dict__[key]) == type(new_config.__dict__[key])
+        for key in config.__dict__.keys()
+    )
+
+    original_dict = config.__dict__.copy()
+    del original_dict["logger_stream_handler"]  # new object created on unpickle
+    del original_dict["logger_formatter"]  # new object created on unpickle
+    assert all(
+        original_dict[key] == new_config.__dict__[key] for key in original_dict.keys()
+    )
```
MalteEbner, line 21: Do you want to assert that the new_config is a deepcopied version of the old one ore the opposite? I don't get it.
e.g.
```python 
config.__dict__["logger_stream_handler"] = None
assert new_config.__dict__["logger_stream_handler"] is None # or is not None?
```

lightly/api/patch.py
```diff
@@ -0,0 +1,58 @@
+import logging
```
MalteEbner, line 1: The name of the file is a bit weird, as it only patches the swagger `Configuration`.

lightly/api/patch.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import Any, Dict, Type
+
+
+def make_swagger_configuration_picklable(
+    configuration_cls: Type,
+) -> None:
```
MalteEbner, line 7: Here you overwrite attributes of the original `Configuration` to create the patched version. Why is this done using a mixin for the `LightlySwaggerApiClient` and `LightlySwaggerRESTClientObject`? Why cannot you use the way to overwrite some attributes for all 3 cases?

tests/api_workflow/test_api_workflow_compute_worker.py
```diff
@@ -668,7 +706,7 @@ def test__snake_to_camel_case() -> None:
     assert _snake_to_camel_case("loremIpsum") == "loremIpsum"  # do nothing
 
 
-def test__validate_config__docker(mocker: MockerFixture) -> None:
+def test__validate_config__docker() -> None:
```
MalteEbner, line 709: Nice cleanup!

lightly/api/api_workflow_compute_worker.py
```diff
@@ -157,6 +156,8 @@ def create_compute_worker_config(
             )
             docker = deserialize_worker_config(worker_config_cc)
             _validate_config(cfg=worker_config, obj=docker)
+        else:
```
MalteEbner, line 159: Nice bugfix!

MalteEbner (commented):
Looks mostly good to me, but I have some questions.

tests/api/test_patch.py
guarin: You are right, I update the testing code, should be clearer now.

lightly/api/patch.py
guarin: I originally put different patches in there. Now only a single one is left. But I would keep the name if we add more patches in the future.

MalteEbner (approved):
LGTM

PR #1101, opened 2023-03-11
drop host trailing slash

shaundaley39:
Since k8s doesn't magically clean up double slashes, some of us hit problems by setting
```
export LIGHTLY_SERVER_LOCATION='https://api.dev.lightly.ai/'
```

With that, the ApiWorkflowClient would build urls like `https://api.dev.lightly.ai//v1/docker/runs/query/datasetId/6401d4534d2ed9112da782f5`.

Which is invalid. Resulting in a mix of 403 forbiddens and 404 not founds.

```
$ curl 'https://api.dev.lightly.ai/v1/versions/pip/minimum'
1.1.0
$ curl 'https://api.dev.lightly.ai//v1/versions/pip/minimum'
{
    "code": "FORBIDDEN",
    "error": "Not able to authenticate user. Please provide a valid token",
    "requestId": "f6f521264e78f1d8a575313f779fe1d9"
}
```

It does work with AppEngine magically stripping out double slashes. 

But we would rather the ApiWorkflowClient always build valid urls without the double slash after a hostname.

This PR just strips any user-defined trailing slash from a hostname.

### DIFF
--- a/lightly/api/utils.py
+++ b/lightly/api/utils.py
@@ -192,11 +192,17 @@ def get_signed_url_destination(signed_url: str = "") -> DatasourceType:
     return DatasourceType.LOCAL
 
 
+def get_lightly_server_location_from_env() -> string:
+    return (
+        getenv("LIGHTLY_SERVER_LOCATION", "https://api.lightly.ai").strip().rstrip("/")
+    )
+
+
 def get_api_client_configuration(
     token: Optional[str] = None,
     raise_if_no_token_specified: bool = True,
 ) -> Configuration:
-    host = getenv("LIGHTLY_SERVER_LOCATION", "https://api.lightly.ai").strip().rstrip("/")
+    host = get_lightly_server_location_from_env()
     ssl_ca_cert = getenv("LIGHTLY_CA_CERTS", None)
 
     if token is None:
--- a/lightly/api/utils.py
+++ b/lightly/api/utils.py
@@ -196,7 +196,7 @@ def get_api_client_configuration(
     token: Optional[str] = None,
     raise_if_no_token_specified: bool = True,
 ) -> Configuration:
-    host = getenv("LIGHTLY_SERVER_LOCATION", "https://api.lightly.ai").rstrip("/")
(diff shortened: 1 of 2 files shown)


lightly/api/utils.py
```diff
@@ -196,7 +196,7 @@ def get_api_client_configuration(
     token: Optional[str] = None,
     raise_if_no_token_specified: bool = True,
 ) -> Configuration:
-    host = getenv("LIGHTLY_SERVER_LOCATION", "https://api.lightly.ai")
+    host = getenv("LIGHTLY_SERVER_LOCATION", "https://api.lightly.ai").rstrip("/")
```
japrescott, line 199: maybe also add a `trim().rstrip("/")` to also safeguard for people adding space at the end by accident? just a thought

japrescott (approved):

lightly/api/utils.py
shaundaley39: Sure, that seems like a weirder mistake to make (it's often not easy and it's always obviously wrong to include whitespace in env variables!), but we might as well prevent that here too.

lightly/api/utils.py
```diff
@@ -196,7 +196,7 @@ def get_api_client_configuration(
     token: Optional[str] = None,
     raise_if_no_token_specified: bool = True,
 ) -> Configuration:
-    host = getenv("LIGHTLY_SERVER_LOCATION", "https://api.lightly.ai")
+    host = getenv("LIGHTLY_SERVER_LOCATION", "https://api.lightly.ai").strip().rstrip("/")
```
philippmwirth, line 199: Can you move this to a separate function, e.g. `get_lightly_server_location_from_env` and test it properly please?

philippmwirth (changes requested):

lightly/api/utils.py
shaundaley39: Thanks, broken out, could you take another look at this?

philippmwirth (approved):

PR #1102, opened 2023-03-13
support Label Box V4 export format

shaundaley39:
Add support for LabelBox V4 import format, using the endpoint added in https://github.com/lightly-ai/lightly-core/pull/2481


guarin:
@shaundaley39 can we close this? The feature was added in https://github.com/lightly-ai/lightly/pull/1108

PR #1112, opened 2023-03-20
Guarin lig 2895 Make Lightly compatible with Pytorch Lightning 2.0

guarin:
## What has changed?

* Convert `gpus` to `accelerator` and `devices` in train cli, examples, docs, and benchmarks.
* Use `strategy` instead of `distributed_backend` in train cli, examples, docs, and benchmarks.
* Use `use_distributed_sampler` instead of `replace_sampler_ddp` in distributed training examples.

PyTorch Lightning 2.0 removed the `gpus` argument from the `Trainer` class and replaced it with `devices` and `accelerator`. It also removed `distributed_backend` in favor of `strategy`. This PR makes the CLI commands compatible with PyTorch Lightning >= 2.0. The minimum version for the CLI remains 1.5.

Because PyTorch Lightning made breaking changes in 2.0, the examples for distributed training are no longer backwards compatible. I added comments for the required changes to make them work for versions <2.0.

PyTorch Lightning also deprecated the `training_epoch_end` hook which we heavily use in the benchmarks. Making the benchmarks compatible with the new lightning version will require more work (#1113). For now we can only support <2.0.

## How has it been tested?

* Tested the CLI manually on old and new PyTorch Lightning versions.
* Tested the examples on old and new PyTorch Lightning versions.

### DIFF
--- a/lightly/cli/train_cli.py
+++ b/lightly/cli/train_cli.py
@@ -7,6 +7,7 @@ command-line interface.
 
 # Copyright (c) 2020. Lightly AG and its affiliates.
 # All Rights Reserved
+import copy
 import os
 import warnings
 
@@ -119,7 +120,10 @@ def _train_cli(cfg, is_cli_call=True):
     encoder = SelfSupervisedEmbedding(model, criterion, optimizer, dataloader)
 
     # Create trainer config
-    trainer_kwargs = OmegaConf.to_container(cfg["trainer"])
+    if isinstance(cfg, dict):
+        trainer_kwargs = copy.deepcopy(cfg["trainer"])
+    else:
+        trainer_kwargs = OmegaConf.to_container(cfg["trainer"])
     if "gpus" in trainer_kwargs:
         # PyTorch Lightning >= 2.0 doesn't support the gpus trainer flag anymore.
         # We have to use accelerator and devices instead.


examples/pytorch_lightning/barlowtwins.py
```diff
@@ -55,7 +55,7 @@ def configure_optimizers(self):
     num_workers=8,
 )
 
-gpus = 1 if torch.cuda.is_available() else 0
+accelerator = "gpu" if torch.cuda.is_available() else "cpu"
 
-trainer = pl.Trainer(max_epochs=10, gpus=gpus)
+trainer = pl.Trainer(max_epochs=10, devices=1, accelerator=accelerator)
```
MalteEbner, line 60: Why not `devices="auto"`, `accelerator="auto"` ? This also applies at other places. Or just remove the arguments completely.

MalteEbner (commented):
I don't get why we need to specify the devices and accelerator at so many places and not just simply set it to default almost everywhere. Users needing that can just read up the Pytorch Lightning Docs.

Otherwise, it looks fine.

examples/pytorch_lightning/barlowtwins.py
guarin: > Why not devices="auto", accelerator="auto" ? This also applies at other places. Or just remove the arguments completely.

`"auto"` is not supported for some older PyTorch Lightning versions 😞

PR #1119, opened 2023-03-23
Malte lig 2899 remove the non tutorial docs using lightly-upload (also as part of lightly-magic)

MalteEbner:
## Description.

- Removing everything under
https://docs.lightly.ai/self-supervised-learning/getting_started/platform.html
as that is forwarded anyways.
- Change the explanations of `lightly-magic` to training + embedding (without upload)
- remove autodocs for lightly-upload


docs/source/getting_started/command_line_tool.rst
```diff
@@ -81,41 +81,30 @@ An example for the label names .yaml file:
...
-Training, Embedding, and Uploading in a go - Magic
+Training and Embedding in a go - Magic
 ---------------------------------------------------
-Lightly-magic is a singular command for training, embedding, and uploading to the Lightly Platform. 
-    
-* To start with, we need to input the directory of the dataset, pass it to input_dir 
+Lightly-magic is a singular command for training a self-supervised model and use it to compute embeddings
+
+* To start with, we need to input the directory of the dataset, pass it to input_dir
 * It requires information on the number of epochs to perform, set trainer.max_epochs,
```
guarin, line 89:
```suggestion
* To start with, we need to input the directory of the dataset, pass it to input_dir.
* It requires information on the number of epochs to perform, set trainer.max_epochs.
```

docs/source/getting_started/command_line_tool.rst
```diff
@@ -81,41 +81,30 @@ An example for the label names .yaml file:
 You can use the output of the lightly-crop command as the *input_dir* for your
 lightly-train command.
 
-Training, Embedding, and Uploading in a go - Magic
+Training and Embedding in a go - Magic
 ---------------------------------------------------
```
guarin, line 85:
```suggestion
Training and Embedding in a Go – Magic
---------------------------------------------------
```

guarin (changes requested):
I would keep some of the api docs but otherwise LGTM.

There are also some references to lightly-upload in `config.yaml`. Not sure if removing them should be part of this issue or not.

guarin:
> It is not accessible anyway, just try to access https://docs.lightly.ai/self-supervised-learning/getting_started/platform.html and you will be forwarded to the new docs by the proxy.

I meant the docs here: https://docs.lightly.ai/self-supervised-learning/lightly.api.html#module-lightly.api.api_workflow_client

MalteEbner:
> There are also some references to lightly-upload in `config.yaml`. Not sure if removing them should be part of this issue or not.

Good point, I removed them.

docs/source/lightly.api.rst
```diff
@@ -18,14 +18,5 @@ lightly.api
...
 
-.. automodule:: lightly.api.api_workflow_upload_dataset
-   :members:
-
-.. automodule:: lightly.api.api_workflow_upload_embeddings
-   :members:
-
-.. automodule:: lightly.api.api_workflow_upload_metadata
-   :members:
-
```
MalteEbner, line 29: @guarin 
I think that removing these makes sense, as we don't want customers to use the upload via pip directly anymore.

guarin: `api_workflow_upload_embeddings ` and `api_workflow_upload_metadata ` also contain methods to download things which we should still expose.

MalteEbner: I did not find any methods that users should use in those files. Which ones do you mean?
The download of embeddings is inside `api_workflow_download_dataset.py`.

guarin: Ah sorry, I though `verify_custom_metadata_format` was needed but it isn't, so `api_worklfow_upload_metadata` can be removed. From `api_workflow_upload_embeddings` I would have kept `get_embedding_by_name` or do we not support exporting emebeddings?

MalteEbner: `get_embedding_by_name` looks like a duplicate of [api_workflow_download_dataset.py:get_embedding_data_by_name](https://github.com/lightly-ai/lightly/blob/2a0f7e8edff98634bdd3bfd44df748565ace6883/lightly/api/api_workflow_download_dataset.py#L152)

guarin: Ah nice, then we can remove it.

guarin (approved):

PR #1120, opened 2023-03-23
Remove download large files tests

guarin:
## What has changed and why?

We accidentally ran one of the e2e tests in our unit tests suite because the test was executed when the file was imported. The test recently started failing, see https://github.com/lightly-ai/lightly/actions/runs/4500624600/jobs/7920079145?pr=1105

I removed the test because we shouldn't rely on external files.


michal-lightly (approved):
LGTM

PR #1121, opened 2023-03-24
Malte lig 2280 deprecate active learning in pip

MalteEbner:
## Description

- add function `raise_active_learning_deprecation_warning()`
- call it from the inits of the `ActiveLearningAgent`, `SelectionConfig` and `Scorer`
- format `lightly/active_learning` with black

## Other (see the [issue comment](https://linear.app/lightly/issue/LIG-2280#comment-3085ec56))

- replace `PendingDeprecationWarning` with `DeprecationWarning`


lightly/active_learning/__init__.py
```diff
@@ -0,0 +1,15 @@
+import warnings
+
+
+def raise_active_learning_deprecation_warning():
+    print("asdf")
```
huan-lightly-0, line 5: I suppose this should be removed

huan-lightly-0 (commented):
It looks okay. I'm thinking if there is a better way of throwing warnings instead of doing that in `super().__init__`.

huan-lightly-0 (approved):

PR #1122, opened 2023-03-27
Fix VICRegL

guarin:
## Changes

* Add support for multiple global and local views to `VICRegLLoss`
* Set default number of local views to 0 when using ResNet
* Generalize `VICRegLoss` to support multiple dimensions
* Update imagenette benchmark
* Cleanup

I thoroughly compared our loss implementation with the original loss. There is a slight difference in the nearest neighbor matching. The same neighbors are matched but the order in which they are returned is different. This results in small differences in the loss computation. Other than that the losses are equal. I trained the model once with our nearest neighbor implementation and once with the original one and there was no difference in performance.

Imagenette performance is a bit lower than before but it is very hard to say if this is expected or not. VICRegL should have similar performance as VICReg on 224px images but we train on 128px. Our old implementation was missing some terms in the loss and only compared one global with one local view. So the old results are also not really comparable.

### DIFF
--- a/docs/source/getting_started/benchmarks/imagenette_benchmark.py
+++ b/docs/source/getting_started/benchmarks/imagenette_benchmark.py
@@ -1192,7 +1192,7 @@ class VICRegLModel(BenchmarkModule):
         views_and_grids = batch[0]
         views = views_and_grids[: len(views_and_grids) // 2]
         grids = views_and_grids[len(views_and_grids) // 2 :]
-        features = [model(view) for view in views]
+        features = [self.forward(view) for view in views]
         z_a, z_a_local_features = features[0]
         grid_a = grids[0]
         loss = 0
--- a/docs/source/getting_started/benchmarks/imagenette_benchmark.py
+++ b/docs/source/getting_started/benchmarks/imagenette_benchmark.py
@@ -1205,8 +1205,8 @@ class VICRegLModel(BenchmarkModule):
         # Training diverges without LARS
         optim = LARS(
             self.parameters(),
-            lr=0.1 * lr_factor,
-            weight_decay=1e-6,
+            lr=0.3 * lr_factor,
+            weight_decay=1e-4,
             momentum=0.9,
         )
         cosine_scheduler = scheduler.CosineWarmupScheduler(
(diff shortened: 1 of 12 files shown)


docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -1177,29 +1176,28 @@ def __init__(self, dataloader_kNN, num_classes):
         self.projection_head = heads.BarlowTwinsProjectionHead(512, 2048, 2048)
         self.local_projection_head = heads.VicRegLLocalProjectionHead(512, 128, 128)
         self.average_pool = nn.AdaptiveAvgPool2d(output_size=(1, 1))
-        self.criterion = VICRegLLoss(alpha=0.75, num_matches=(16, 4))
+        self.criterion = VICRegLLoss(num_matches=(8, 4))
         self.backbone = nn.Sequential(self.train_backbone, self.average_pool)
-        self.warmup_epochs = 40 if max_epochs >= 800 else 20
+        self.warmup_epochs = 20 if max_epochs >= 800 else 10
```
guarin, line 1181: Reduced this because the paper uses only 10 epochs warmup for 300 epochs training.

examples/pytorch/vicregl.py
```diff
@@ -48,33 +48,29 @@ def forward(self, x):
...
+    batch_size=256,
     collate_fn=collate_fn,
     shuffle=True,
     drop_last=True,
     num_workers=8,
 )
 
 criterion = VICRegLLoss()
-optimizer = torch.optim.SGD(model.parameters(), momentum=0.9, lr=0.06)
+optimizer = torch.optim.SGD(model.parameters(), lr=0.01, momentum=0.9)
```
guarin, line 59: 0.01 is a bit more stable.

lightly/loss/vicregl_loss.py
```diff
@@ -1,232 +1,422 @@
...
+
     """
 
     def __init__(
         self,
         lambda_param: float = 25.0,
         mu_param: float = 25.0,
         nu_param: float = 1.0,
-        alpha: float = 0.25,
+        alpha: float = 0.75,
```
guarin, line 71: Default is 0.75

philippmwirth (approved):

PR #1123, opened 2023-03-28
Use config v3 endpoint

philippmwirth:
# Use config v3 endpoint

* Generate swagger client for config v3 endpoint
* Use config v3 endpoint for scheduling jobs
* Use config v3 endpoint in tests
* Add `user_id` to all `DockerRunData` and `DockerWorkerRegistryEntryData`.


tests/api_workflow/test_api_workflow_compute_worker.py
```diff
@@ -61,9 +61,7 @@ def test_delete_compute_worker(self):
     def test_create_compute_worker_config(self):
         config_id = self.api_workflow_client.create_compute_worker_config(
             worker_config={
-                "stopping_condition": {
-                    "n_samples": 10,
-                },
+                "pretagging": True,
```
MalteEbner, line 64: Could you change this to set something of the new config?
E.g. 
```suggestion
                "training": {"task_name": "lightly_pretagging"},
```

MalteEbner (approved):
Makes sense to me.

PR #1126, opened 2023-03-31
Automate release using Github actions

philippmwirth:
### What has changed and why?

Add automated release workflow for the Lightly PIP package:
- On dispatch for now (change to release later)
- Check out latest tag
- Install `wheel` and `twine`
- Build distribution and upload to PyPI
- Send Slack message indicating success or failure

### How has this been tested?

Action runs through successfully using `twine -h` instead of `twine upload ...`:
- [Action](https://github.com/lightly-ai/lightly/actions/runs/4575518821/jobs/8078446374?pr=1126)
- [Slack message](https://lightly-ai.slack.com/archives/C03T7GLHG8P/p1680268950577909)

### DIFF
--- a/.github/workflows/release_pypi.yml
+++ b/.github/workflows/release_pypi.yml
@@ -1,9 +1,10 @@
 name: Build and release lightly
 
 on:
-  push:
-  release:
-    types: [created]
+  workflow_dispatch:
+# TODO(Philipp, 03/23): Enable me after proper testing.
+#  release:
+#    types: [created]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/release_pypi.yml
```diff
@@ -0,0 +1,56 @@
+name: Build and release lightly
+
+on:
+  workflow_dispatch:
+# TODO(Philipp, 03/23): Enable me after proper testing.
+#  release:
+#    types: [created]
```
guarin, line 7: Are you sure we should use `created`? From the docs it seems to me that we should use `published`.

guarin (commented):
The changes look good to me, the only thing I am not sure about is the workflow dispatch.

.github/workflows/release_pypi.yml
philippmwirth: You might be right although I'm not 100% what the difference is. I'll change it.

PR #1130, opened 2023-04-04
Add FastSiam model

### LINKED ISSUE #972: Which framework to use for more resource-efficient training? Plus a recent paper in this regard, "FastSiam: Resource-Efficient Self-supervised Learning on a Single GPU"
For those of us with just a single GPU, what framework do you suggest using in general? I know that SimSiam is often a good choice in this regard since it generally requires smaller batch sizes than the other frameworks.

I recently came across the following paper, "FastSiam: Resource-Efficient Self-supervised Learning on a Single GPU": [https://link.springer.com/chapter/10.1007/978-3-031-16788-1_4](https://link.springer.com/chapter/10.1007/978-3-031-16788-1_4). I'm not well-versed in how to judge how good all these new self-supervised frameworks are, but it seems promising to have a framework like this that is targeted to those in the community who don't have access to big compute power. No code implementations of this anywhere 😞 Are there any plans for Lightly to include frameworks like these, which aim to be "lighter" as opposed to "more powerful"?

image
image
image

philippmwirth:
# Add FastSiam model

Closes #972.
* Add SimSiam and FastSiam transforms
* Add FastSiam model to benchmarks

I'll add the examples in a follow-up pull request.

TODO:
- [X] Imagenette @ 200 epochs
- [x] Imagenette @ 800 epochs
- [x] Cifar10 benchmarks
- [x] Imagenet100 benchmarks


philippmwirth:
Results for `12699915c0e8e2c669db2b7a3780ef8e706a7d7c`.
```
------------------------------------------------------------------------------------------
| Model         | Batch Size | Epochs |  KNN Test Accuracy |       Time | Peak GPU Usage |
------------------------------------------------------------------------------------------
| FastSiam      |        256 |    200 |              0.836 |   69.0 Min |      7.3 GByte |
------------------------------------------------------------------------------------------
```

docs/source/getting_started/benchmarks/cifar10_benchmark.py
```diff
@@ -413,6 +421,25 @@ def configure_optimizers(self):
         return [optim], [scheduler]
 
 
+class FastSiamModel(SimSiamModel):
+    def __init__(self, dataloader_kNN, num_classes):
+        super().__init__(dataloader_kNN, num_classes)
+
+    def training_step(self, batch, batch_idx):
+        xs, _, _ = batch
```
guarin, line 429: Maybe call `xs` `views` instead.

docs/source/getting_started/benchmarks/cifar10_benchmark.py
```diff
@@ -413,6 +421,25 @@ def configure_optimizers(self):
...
+        zs, ps = zip(*list(self.forward(x) for x in xs))
+        zs = torch.cat([z.unsqueeze(0) for z in zs], dim=0)
+        ps = torch.cat([p.unsqueeze(0) for p in ps], dim=0)
+
+        loss = 0.0
+        for i in range(len(xs)):
+            mask = torch.arange(len(xs)) != i
+            loss += self.criterion(ps[i], torch.mean(zs[mask], dim=0))
```
guarin, line 437: I am surprised that you don't have to specify the device for the mask. If you want to add it you can do:
```
mask = torch.arange(len(xs), device=self.device) != i
```

docs/source/getting_started/benchmarks/cifar10_benchmark.py
```diff
@@ -413,6 +421,25 @@ def configure_optimizers(self):
...
+
+        self.log("train_loss_ssl", loss)
+        return loss
```
guarin, line 440: Maybe divide the loss by the number of views.

docs/source/getting_started/benchmarks/cifar10_benchmark.py
```diff
@@ -167,6 +169,11 @@
     gaussian_blur=0.0,
 )
 
+# Multi crop augmentation for FastSiam
+fast_siam_transform = MultiViewTransform(
+    [SimCLRViewTransform(input_size=32, cj_strength=0.5, gaussian_blur=0.0)] * 4
+)
```
guarin, line 175: We could add a new fastsiam transform.

guarin (commented):
Oh, that is shorter than expected! Left a few comments :)

docs/source/getting_started/benchmarks/cifar10_benchmark.py
philippmwirth: I'm not sure this makes sense. It's just a lot of copy pasted code... But I can add it if you think it's a good idea.

guarin: SimSiam uses SimCLR transform right? In that case we can just make an alias. Does FastSiam change any transform parameters compared to SimSiam?

philippmwirth: No they don't change anything compared to SimSiam. But it's possible SimSiam uses a different min_scale than SimCLR

guarin (approved):
LGTM! Left two small comments but feel free to resolve and merge.

PR #1131, opened 2023-04-05
lig-2904: Remove hydra from lightly config

huan-lightly-0:
The `hydra` section in `config.yaml` is only for hydra and not used in any logic. Removing this redundant bit from the actual lightly config read from the config file.

### DIFF
--- a/lightly/cli/config/get_config.py
+++ b/lightly/cli/config/get_config.py
@@ -6,7 +6,7 @@ from omegaconf import DictConfig, OmegaConf
 def get_lightly_config() -> DictConfig:
     config_path = Path(__file__).with_name("config.yaml")
     conf = OmegaConf.load(config_path)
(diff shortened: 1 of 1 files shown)


lightly/cli/config/get_config.py
```diff
@@ -6,4 +6,8 @@
 def get_lightly_config() -> DictConfig:
     config_path = Path(__file__).with_name("config.yaml")
     conf = OmegaConf.load(config_path)
+    # FIXME: remove this when hydra is completely dropped
```
guarin, line 9: Nit: We usually put `TODO(name, date):`  for those kind of messages.

guarin (approved):

lightly/cli/config/get_config.py
huan-lightly-0: Updated

PR #1140, opened 2023-04-11
add datasourceUrl to export endpoint

MalteEbner:
## Description

The `ApiWorkflowClient.export_filenames_and_read_urls_by_tag_id` now also includes the datasource URL.

Other possible implementation options I tried but decided against:
1. As separate functions for the one with read_urls and datasource_urls -> lots of duplicate code.
2. By allowing to set the `FileNameFormat` -> makes it more complicated.

Why is the name of the method the same?
I know that it does not fit perfectly, but changing it would be breaking. Thus I would leave it as it is.

### DIFF
--- a/lightly/api/api_workflow_export.py
+++ b/lightly/api/api_workflow_export.py
@@ -250,7 +250,7 @@ class _ExportDatasetMixin:
             [
                 {
                     "fileName": "sample1.jpg",
-                    "readURL": "s3://my_datasource/sample1.jpg?read_url_key=EAIFUIENDLFN",
+                    "readUrl": "s3://my_datasource/sample1.jpg?read_url_key=EAIFUIENDLFN",
                     "datasourceUrl": "s3://my_datasource/sample1.jpg",
                 },
                 {
--- a/lightly/api/api_workflow_export.py
+++ b/lightly/api/api_workflow_export.py
@@ -308,7 +308,7 @@ class _ExportDatasetMixin:
                 Name of the tag which should exported.
 
         Returns:
-            A list of dictionaries with the keys "filename", "readURL" and "datasourceURL".
+            A list of dictionaries with the keys "filename", "readUrl" and "datasourceUrl".
 
         Examples:
             >>> # write json file which can be used to access the actual file contents.
--- a/lightly/api/api_workflow_export.py
+++ b/lightly/api/api_workflow_export.py
@@ -238,7 +238,7 @@ class _ExportDatasetMixin:
         self,
         tag_id: str,
     ) -> List[Dict[str, str]]:
-        """Export the samples filenames to map with their readURL and datasource URL.
+        """Export filenames, read URLs, and datasource URLs from the given tag.
 
         Args:
             tag_id:
--- a/lightly/api/api_workflow_export.py
+++ b/lightly/api/api_workflow_export.py
@@ -255,7 +255,7 @@ class _ExportDatasetMixin:
                 },
                 {
                     "fileName": "sample2.jpg",
-                    "readURL": "s3://my_datasource/sample2.jpg?read_url_key=JSBFIEUHVSJ",
+                    "readUrl": "s3://my_datasource/sample2.jpg?read_url_key=JSBFIEUHVSJ",
                     "datasourceUrl": "s3://my_datasource/sample2.jpg",
                 },
             ]
--- a/lightly/api/api_workflow_export.py
+++ b/lightly/api/api_workflow_export.py
@@ -245,7 +245,7 @@ class _ExportDatasetMixin:
                 Id of the tag which should exported.
 
         Returns:
-            A list of dictionaries with the keys "filename", "readURL" and "datasourceURL".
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_export.py
```diff
@@ -237,19 +237,30 @@ def export_filenames_by_tag_name(
...
+                {
+                    "fileName": "sample2.jpg",
+                    "readURL": "s3://my_datasource/sample2.jpg?read_url_key=JSBFIEUHVSJ",
+                    "datasourceUrl": "s3://my_datasource/sample2.jpg",
+                },
+            ]
 
         """
-        # TODO (Philipp, 10.01.2023): Switch to the exportTagToBasicFilenamesAndReadUrls
-        # when the read-urls are fixed.
```
MalteEbner, line 252: That cannot be used for datasourceURLs

tests/api_workflow/test_api_workflow_export.py
```diff
@@ -62,18 +64,34 @@ def test_export_label_studio_tasks_by_tag_name(self):
         self.assertTrue(all(isinstance(task, dict) for task in tasks))
 
     def test_export_tag_to_basic_filenames_and_read_urls(self):
```
MalteEbner, line 66: I switched to not using the MockedApiWorkflowClient anymore.

lightly/api/api_workflow_export.py
```diff
@@ -262,40 +273,50 @@ def export_filenames_and_read_urls_by_tag_id(
             tag_id=tag_id,
             file_name_format=FileNameFormat.REDIRECTED_READ_URL,
         )
+        datasource_urls_string = retry(
+            self._tags_api.export_tag_to_basic_filenames,
+            dataset_id=self.dataset_id,
+            tag_id=tag_id,
+            file_name_format=FileNameFormat.DATASOURCE_FULL,
+        )
```
guarin, line 281: What happens for this endpoint with old datasets that were created using the local workflow? My fear is that an empty string is returned which will result in `zip(filenames, read_urls, datasource_urls)` being and empty list.

lightly/api/api_workflow_export.py
```diff
@@ -237,19 +237,30 @@ def export_filenames_by_tag_name(
     def export_filenames_and_read_urls_by_tag_id(
         self,
         tag_id: str,
-    ) -> List[Dict]:
-        """Export the samples filenames to map with their readURL.
+    ) -> List[Dict[str, str]]:
+        """Export the samples filenames to map with their readURL and datasource URL.
```
guarin, line 241:
```suggestion
        """Export filenames, read URLs, and datasource URLs from the given tag.
```

lightly/api/api_workflow_export.py
```diff
@@ -237,19 +237,30 @@ def export_filenames_by_tag_name(
...
                 Id of the tag which should exported.
 
         Returns:
-            A list of mappings of the samples filenames and readURLs within a certain tag.
+            A list of dictionaries with the keys "filename", "readURL" and "datasourceURL".
+            An example:
+            [
+                {
+                    "fileName": "sample1.jpg",
+                    "readURL": "s3://my_datasource/sample1.jpg?read_url_key=EAIFUIENDLFN",
```
guarin, line 253:
```suggestion
                    "readUrl": "s3://my_datasource/sample1.jpg?read_url_key=EAIFUIENDLFN",
```

lightly/api/api_workflow_export.py
```diff
@@ -262,40 +273,50 @@ def export_filenames_and_read_urls_by_tag_id(
...
+            )
         ]
 
     def export_filenames_and_read_urls_by_tag_name(
         self,
         tag_name: str,
-    ) -> List[Dict]:
-        """Export the samples filenames to map with their readURL.
+    ) -> List[Dict[str, str]]:
+        """Export the samples filenames to map with their readURL and datasource URL.
```
guarin, line 304:
```suggestion
        """Export filenames, read URLs, and datasource URLs from the given tag name.
```

lightly/api/api_workflow_export.py
```diff
@@ -262,40 +273,50 @@ def export_filenames_and_read_urls_by_tag_id(
...
 
         Args:
             tag_name:
                 Name of the tag which should exported.
 
         Returns:
-            A list of mappings of the samples filenames and readURLs within a certain tag.
+            A list of dictionaries with the keys "filename", "readURL" and "datasourceURL".
```
guarin, line 312:
```suggestion
            A list of dictionaries with the keys "filename", "readUrl" and "datasourceUrl".

```

guarin (changes requested):
Looks good :) There is one concern about datasets without a datasource.

lightly/api/api_workflow_export.py
MalteEbner: It works, I tried it out for a dataset uploaded via CLI:
image

guarin: Awesome!

PR #1141, opened 2023-04-12
Enable installation of api dependencies via requirements.txt

michal-lightly:
### What has changed

Move torch dependencies to torch.txt. Lightly installation does not change, The dependencies for the install are now loaded from `base.txt` and `torch.txt`.

This enables to install `base.txt` manually without installing bulky torch.


guarin (approved):
LGTM!

requirements/torch.txt
```diff
@@ -0,0 +1,5 @@
+torch<2
```
michal-lightly, line 1: Note: Needed to limit torch version, otherwise the tests don't pass

PR #1143, opened 2023-04-12
rr_degrees: Update swagger codegen

philippmwirth:
# rr_degrees: Update swagger codegen

* Update the swagger codegen to include `rr_degrees`.
* Add a test to schedule a worker run with `rr_degrees = [-90, 90]`


huan-lightly-0:
I think it's better not to include generated code irrelevant to rr_degrees here.

Issue #1085, opened 2023-03-01
Move config files and setup.py to pyproject.toml


guarin:
We would like to move to `pyproject.toml` for package configuration. @SauravMaheshkar made a PR (#1081 ) for this but it turned out to be more involved than expected.

TODO:
- [x] Figure out how to move everything from `setup.py` to `pyproject.toml`
- [x] Move `setup.py` and test on old python versions (3.6, 3.7)
- [x] Check if we have to upgrade `setuptools` dependency (`setuptools` only supports `pyproject.toml` from version 61.0.0)
- [x] Move other configuration files to `pyproject.toml`
    - [x] (skip) `pylintrc`
    - [x] `.coveragerc` 
    - [x] `mypy.ini`
- [x] Check if we can also move requirement files to `pyproject.toml`

SauravMaheshkar:
@guarin Maybe we could break this down into smaller tasks and try to achieve this using subsequent smaller PRs. I think the easiest to start would be moving `coverage` and `mypy.ini` into `pyproject.toml`. Skipping `pylint` for now since it's rather verbose.

guarin:
Hi, yes `.coveragerc` or `mypy.ini` could be a good start and a PR would be very welcome. I agree that adding `pylint` doesn't make sense for now, especially because we don't enforce linting right now.

SauravMaheshkar:
@guarin with #1549 merged what parts of the above checklist are still to be fixed considering #1149 as well.

philippmwirth:
Thanks for your work, @SauravMaheshkar. I have updated the original comment to track progress.

philippmwirth:
I think a good next step would be to move the dependencies from `requirements/*.txt` to the `pyproject.toml`. Let's try to maintain the structure (base, dev, minimal, etc): https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements

SauravMaheshkar:
> I think a good next step would be to move the dependencies from `requirements/*.txt` to the `pyproject.toml`. Let's try to maintain the structure (base, dev, minimal, etc): https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements

Thanks will work on that next !

guarin:
Completed in #1559

PR #1149, opened 2023-04-16
Support newer setuptools

adamjstewart:
This PR makes two changes:

- [x] Setuptools is not needed at run-time, only at build-time (can use `setup_requires` if you prefer)
- [x] Newer versions of setuptools seem to have fixed the bug that required pinning setuptools

I was trying to install both lightly and the latest version of setuptools but this wasn't allowed due to either one of the above issues.

### DIFF
--- a/.github/workflows/test_setup.yml
+++ b/.github/workflows/test_setup.yml
@@ -32,7 +32,9 @@ jobs:
         path: ${{ env.pythonLocation }}
         key: cache_v2_${{ env.pythonLocation }}-${{ hashFiles('requirements/**') }}
     - name: Install Dependencies and lightly
-      run: pip install .
+      run: |
+        pip install .
+        pip list
     - name: basic tests of CLI
       run: |
         LIGHTLY_SERVER_LOCATION="localhost:-1"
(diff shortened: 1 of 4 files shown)


guarin:
Hi @adamjstewart, thanks a lot for the PR!

It looks like the changes don't pass our tests. I would assume this is due to removing setuptools from the requirements. Did you try just removing the pinned version instead?

guarin:
> Could also be that newer setuptools doesn't work for some reason. It seems like someone needs to approve CI after every single commit, so this may take a while to debug 😅

For debugging it is probably easiest if you try installing it in a Python 3.7 environment with `pip install .`

adamjstewart:
Was able to reproduce locally but not able to figure out why it wasn't working. Using `setup_requires` instead works, and gives behavior similar to what you previously had. It won't install setuptools for you if setuptools isn't already installed, which `pyproject.toml` supports, but I'll let someone handle that when working on #1085. I would suggest opening an issue with the setuptools devs if you want to get to the bottom of it.

Anyway, this should work now, and be ready to merge.

PR #1150, opened 2023-04-16
Add support for SimCLR v2

adamjstewart:
SimCLR v2 adds a deeper projection head (they experiment with 2–4 layers). By default, the input_dim, hidden_dim, and output_dim all match the number of outputs from the ResNet. All layers use a BatchNorm. 

See https://github.com/google-research/simclr/blob/master/model_util.py#L141 for the reference implementation.

### DIFF
--- a/lightly/models/modules/heads.py
+++ b/lightly/models/modules/heads.py
@@ -230,17 +230,14 @@ class SimCLRProjectionHead(ProjectionHead):
                 nn.ReLU(),
             )
         )
-        if num_layers > 2:
-            layers.extend(
-                [
-                    (
-                        hidden_dim,
-                        hidden_dim,
-                        nn.BatchNorm1d(hidden_dim) if batch_norm else None,
-                        nn.ReLU(),
-                    )
-                ]
-                * (num_layers - 2)
+        for layers in range(2, num_layers):
+            layers.append(
+                (
+                    hidden_dim,
+                    hidden_dim,
+                    nn.BatchNorm1d(hidden_dim) if batch_norm else None,
+                    nn.ReLU(),
+                )
             )
         layers.append(
             (
--- a/lightly/models/modules/heads.py
+++ b/lightly/models/modules/heads.py
@@ -219,7 +219,7 @@ class SimCLRProjectionHead(ProjectionHead):
             output_dim: Number of output dimensions.
             num_layers: Number of hidden layers.
         """
-        layers: list[tuple[int, int, Optional[nn.Module], Optional[nn.Module]]] = []
+        layers: List[Tuple[int, int, Optional[nn.Module], Optional[nn.Module]]] = []
         layers.append((input_dim, hidden_dim, nn.BatchNorm1d(hidden_dim), nn.ReLU()))
         if num_layers > 2:
             layers.extend(
--- a/lightly/models/modules/heads.py
+++ b/lightly/models/modules/heads.py
@@ -219,14 +219,14 @@ class SimCLRProjectionHead(ProjectionHead):
             output_dim: Number of output dimensions.
             num_layers: Number of hidden layers.
         """
-        super().__init__(
-            [
-                (input_dim, hidden_dim, nn.BatchNorm1d(hidden_dim), nn.ReLU()),
-                (hidden_dim, hidden_dim, nn.BatchNorm1d(hidden_dim), nn.ReLU())
-                * (num_layers - 2),
-                (hidden_dim, output_dim, nn.BatchNorm1d(output_dim), None),
-            ]
-        )
+        layers = [(input_dim, hidden_dim, nn.BatchNorm1d(hidden_dim), nn.ReLU())]
+        if num_layers > 2:
+            layers.extend(
+                [(hidden_dim, hidden_dim, nn.BatchNorm1d(hidden_dim), nn.ReLU())]
+                * (num_layers - 2)
+            )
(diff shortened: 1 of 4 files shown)


lightly/models/modules/heads.py
```diff
@@ -198,19 +198,36 @@ class SimCLRProjectionHead(ProjectionHead):
...
-        )
+        """Initialize a new SimCLRProjectionHead instance.
+
+        Args:
+            input_dim: Number of input dimensions.
+            hidden_dim: Number of hidden dimensions.
+            output_dim: Number of output dimensions.
+            num_layers: Number of hidden layers.
+        """
+        layers: list[tuple[int, int, Optional[nn.Module], Optional[nn.Module]]] = []
```
guarin, line 222: Please use `List` and `Tuple` instead of `list` and `tuple` for typing as the latter are only supported in newer Python versions.

lightly/models/modules/heads.py
```diff
@@ -198,19 +198,36 @@ class SimCLRProjectionHead(ProjectionHead):
...
+        layers.append((input_dim, hidden_dim, nn.BatchNorm1d(hidden_dim), nn.ReLU()))
```
guarin, line 223: This changes the logic, SimCLR v1 doesn't have batchnorm in the projection head. Maybe it would be easier to add a `SimCLRV2ProjectionHead` if there are significant differences between the two versions.

guarin (changes requested):
Hi, thanks a lot for the PR!

I believe the tests are failing because of the extra batchnorm layer this PR adds in the projection head. We try to load checkpoints in our tests that do not contain this layer and loading is now failing. Could you remove it again? If SimCLRV2 adds batchnorm I would propose to add a `SimCLRV2ProjectionHead` instead of modifying the original one.

lightly/models/modules/heads.py
adamjstewart: Are you sure SimCLR v1 doesn't use batchnorm? I see it right here: https://github.com/google-research/simclr/blob/1.0/model_util.py#L112

The v1 and v2 papers don't mention batchnorm, but the source code still uses it for all layers.

guarin: You are right, we must have missed this. Interesting that neither vissl nor solo-learn add batchnorm either:
* https://github.com/facebookresearch/vissl/blob/04788de934b39278326331f7a4396e03e85f6e55/configs/config/pretrain/simclr/simclr_4node_resnet.yaml#L53
* https://github.com/vturrisi/solo-learn/blob/6a891ee9af6e28ae071313b596f8e553b7227b64/solo/methods/simclr.py#L48-L52

The issue is that we cannot change it globally right now as it breaks some other code. I guess we have to add a `batch_norm: bool = True` argument for now and set it to `False` in those places:
* https://github.com/lightly-ai/lightly/blob/527572e2bef6d4df4b96dbe3c2a5e46e36ae1528/lightly/models/simclr.py#L35
* https://github.com/lightly-ai/lightly/blob/527572e2bef6d4df4b96dbe3c2a5e46e36ae1528/lightly/cli/_cli_simclr.py#L24

adamjstewart: > The issue is that we cannot change it globally right now as it breaks some other code.

Is that other code incorrect? Happy to fix this in other locations too if necessary. Can also set the number of layers to 2 by default if you want to maintain backwards compatibility.

adamjstewart: Found at least one test where I can set `batch_norm=False`, not sure if there are others. We could change the default to False if we need to, but we should change it to True someday once we've re-trained with the extra layers.

guarin: > Found at least one test where I can set batch_norm=False, not sure if there are others.

We also have to set `batch_norm=False` in these two places:
* https://github.com/lightly-ai/lightly/blob/527572e2bef6d4df4b96dbe3c2a5e46e36ae1528/lightly/models/simclr.py#L35
* https://github.com/lightly-ai/lightly/blob/527572e2bef6d4df4b96dbe3c2a5e46e36ae1528/lightly/cli/_cli_simclr.py#L24

I believe if you change this the tests should pass.

> We could change the default to False if we need to, but we should change it to True someday once we've re-trained with the extra layers.

I think setting it to True by default makes sense. We just have to set it to False in the places where we need backwards compatibility.

lightly/models/modules/heads.py
```diff
@@ -198,19 +198,59 @@ class SimCLRProjectionHead(ProjectionHead):
...
+            layers.extend(
+                [
+                    (
+                        hidden_dim,
+                        hidden_dim,
+                        nn.BatchNorm1d(hidden_dim) if batch_norm else None,
+                        nn.ReLU(),
+                    )
+                ]
+                * (num_layers - 2)
```
guarin, line 243: I believe this won't work because the same batchnorm instance will be used in multiple layers. Instead, a new batchnorm instance should be created in every layer.

guarin (changes requested):
> Let me know if and where you want me to add unit tests for the multi-layer case.

Could you add a test that passes a random tensor to the head and assert that the output dimensions are correct? And the number of layers.

You can add the test in a new method at the end of this file: https://github.com/lightly-ai/lightly/blob/master/tests/models/test_ProjectionHeads.py
The tests there are unfortunately a bit messy 😬

guarin (approved):
This looks great! Thanks a lot!

PR #1151, opened 2023-04-16
Make type hints public

adamjstewart:
This will allow downstream libraries to use lightly's type hints with mypy.

See [PEP 561](https://peps.python.org/pep-0561/)

### DIFF
--- a/lightly/py.typed
+++ b/lightly/py.typed
@@ -1,0 +1,1 @@
+partial
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,6 @@
 [metadata]
 # This includes the license file(s) in the wheel.
 license_files = LICENSE.txt
+
+[options.package_data]
+lightly = py.typed


guarin (commented):
Thanks for the PR!

I tried it in a downstream package that uses type checking with mypy and ran into the following error:
```
/opt/conda/envs/env/lib/python3.10/site-packages/lightly/__init__.py: error: Source file found twice under different module names: "lightly" and "lightly.__init__"
```

I am not sure if it is an issue with how mypy is configured or if it is due to the lightly package structure. Did you try it in a downstream package as well? The error seems related to https://github.com/python/mypy/issues/8944 but I couldn't get it working, will have to investigate a bit more.

adamjstewart:
It appears to be due to the lightly package structure. If I remove most of the stuff in `lightly/__init__.py`, mypy works. I'm not sure what that stuff is for, and it's really hard to search for. It seems to be based on PyTorch Lightning, but Lightning doesn't seem to do this anymore either.

guarin:
Thanks for investigating! I'll have a look once I have a bit more time. I believe we also have to wait a bit with marking the package as typed because many parts are not yet typed and some types might be outdated.

adamjstewart:
> I believe we also have to wait a bit with marking the package as typed because many parts are not yet typed

I think this can be addressed by https://peps.python.org/pep-0561/#partial-stub-packages

guarin:
> I think this can be addressed by https://peps.python.org/pep-0561/#partial-stub-packages

Thanks for pointing this out, we'll definitely have to add this as well. However, typing is currently not highest priority for us so this might take a while.

adamjstewart:
Usually what I do in this case is add type hints to the things I'm personally using one by one. If enough users do that, a significant portion of the library will have type hints without you doing anything.

adamjstewart:
Marked this as a draft until we figure out what's wrong with the "Source file found twice" issue.

adamjstewart:
Closing this until we solve the aforementioned packaging issue.

PR #1155, opened 2023-04-17
Fix dataset exists

philippmwirth:
# Fix dataset exists

* Make `dataset_exist` raise if the api exception is not `404` (not found).
* Add unit tests.

Note: `dataset_exist` works with `dataset_id` not with `dataset_name`. For the latter use `dataset_name_exists`.

### DIFF
--- a/tests/api_workflow/test_api_workflow_datasets.py
+++ b/tests/api_workflow/test_api_workflow_datasets.py
@@ -7,6 +7,7 @@ from lightly.openapi_generated.swagger_client import (
     DatasetsApi,
     DatasetType,
 )
+from lightly.openapi_generated.swagger_client.rest import ApiException
 from tests.api_workflow.mocked_api_workflow_client import MockedApiWorkflowSetup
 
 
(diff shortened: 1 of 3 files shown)


tests/api_workflow/test_api_workflow_datasets.py
```diff
@@ -29,6 +30,16 @@ def test_dataset_name_exists__own_not_existing(self):
...
 
+    def test_dataset_exists(self):
+        assert self.api_workflow_client.dataset_exists(dataset_id="dataset_1_id")
+        assert not self.api_workflow_client.dataset_exists(
+            dataset_id="non_existing_dataset_id"
+        )
+
+    def test_dataset_exists__raises_error(self):
+        with self.assertRaises(ApiException):
+            self.api_workflow_client.dataset_exists(dataset_id=None)
```
michal-lightly, line 41: Can you please also assert the status code of the raised exception? To make sure it is not 404.

michal-lightly (commented):
It is a bit opaque what is the api response in the test. Left a suggestion. Otherwise it looks good.

michal-lightly (approved):
LGTM, thanks!

PR #1162, opened 2023-04-24
Guarin lig 3049 add simclr imagenet resnet50 benchmark

### LINKED ISSUE #569: Write benchmark code for ImageNet
# Write benchmark code for ImageNet

Similar to the benchmarks we have for [Cifar10 and Imagenette](https://docs.lightly.ai/getting_started/benchmarks.html), we should write the benchmark code for `ImageNet` with the parameters from the respective papers. This will allow us to 
- provide a model zoo of different pre-trained ResNets on `ImageNet`
- validate our implementations on `ImageNet`

guarin:
## Changes

* Add `benchmarks` directory
* Add code for SimCLR ImageNet benchmark
* Add linear evaluation code
* Add knn evaluation code
* Add online linear evaluation code
* Add LARS optimizer

Closes #569

## How was it tested?

* Most parts were manually tested
* Will add unit tests later

## Runs

Initial run: https://tensorboard.dev/experiment/VD0GPdaoSLa5uonTp6k6dQ/#scalars
There seems to be an issue with the linear eval validation logic, top1 accuracy collapses after a couple epochs: 
Screen Shot 2023-04-27 at 09 38 45

Doesn't happen for train top1 but we get max 51%. I would expect 64% val top1 (https://github.com/facebookresearch/vissl/blob/main/MODEL_ZOO.md#simclr).

**Issues**:
* We accidentally increased SimCLR default output dimension from 128 to 2048 but this shouldn't have a big impact on the accuracy:  This was already fixed in #1173 
Screen Shot 2023-04-27 at 10 09 05
  

* Will check accuracy with pretrained supervised model to check if there is something off with our validation logic, would expect 76% val top1 (https://github.com/facebookresearch/vissl/blob/main/MODEL_ZOO.md#supervised)
--> Got 73.1% knn val top1 and 76.3% linear eval top1 [tensorboard](https://tensorboard.dev/experiment/w07oartYSECMy4o8E2SmCA/).

* Verified that ntx_ent_loss is correct for non-distributed training
* Will check if there if the results differ between distributed and non-distributed training to verify that loss/backprop/sync works as expected
* Other options are
  * Incorrect training parameters
  * Incorrect augmentations
  * Incorrect collate -> Verified that multi view collate works as expected
  * Incorrect projection head (our implementation slightly differs from solo-learn/vissl in that it has a batch norm layer at the end of the head, but this is the way it is implemented in the authors code).

Edit: Issues fixed in #1175


benchmarks/imagenet/resnet50/main.py
```diff
@@ -0,0 +1,200 @@
...
+        sync_batchnorm=True,
+    )
+
+    trainer.fit(
+        model=model,
+        train_dataloaders=train_dataloader,
+        val_dataloaders=val_dataloader,
+    )
+    for metric in ["val_online_cls_top1", "val_online_cls_top5"]:
+        print(f"max {metric}: {max(metric_callback.get(metric))}")
```
guarin, line 195: This is only temporary, will add proper results reporting later on.

benchmarks/imagenet/resnet50/simclr.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class SimCLR(LightningModule):
+    def __init__(self, batch_size: int, num_classes: int) -> None:
+        super().__init__()
+        self.save_hyperparameters()
+        self.batch_size = batch_size
+
+        resnet = resnet50()
+        resnet.fc = Identity()  # Ignore classification head
```
guarin, line 24: I used this instead of the `self.backbone = nn.Sequential(*list(resnet.children())[:-1])` trick as this should make it easier to load the saved checkpoints. And is IMHO also cleaner.

benchmarks/imagenet/resnet50/simclr.py
```diff
@@ -0,0 +1,100 @@
...
+        self.backbone = resnet
+        self.projection_head = SimCLRProjectionHead()
+        self.criterion = NTXentLoss(temperature=0.1, gather_distributed=True)
+
+        self.online_classifier = OnlineLinearClassifier(num_classes=num_classes)
```
guarin, line 29: Implemented online linear classifier as a standalone module that can be added to any model class. Decided against creating a superclass like the benchmark module as it makes setup a bit weird and is less composable.

benchmarks/imagenet/resnet50/simclr.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def forward(self, x: Tensor) -> Tensor:
+        return self.backbone(x)
```
guarin, line 32: Forward function should only return features, the evaluation code uses this forward function.

benchmarks/imagenet/resnet50/simclr.py
```diff
@@ -0,0 +1,100 @@
...
+                    "params": params_no_weight_decay,
+                    "weight_decay": 0.0,
+                },
+                {
+                    "name": "online_classifier",
+                    "params": self.online_classifier.parameters(),
+                    "weight_decay": 0.0,
+                },
+            ],
+            lr=0.3 * self.batch_size * self.trainer.world_size / 256,
```
guarin, line 82: Setup here is a bit complicated because weight decay shouldn't be used for batchnorm and bias parameters. Similarly, weight decay is not good for the classifier.

lightly/utils/benchmarking/benchmark_module.py
```diff
@@ -13,82 +11,7 @@
 from torch import Tensor
 from torch.utils.data import DataLoader
 
-# code for kNN prediction from here:
-# https://colab.research.google.com/github/facebookresearch/moco/blob/colab-notebook/colab/moco_cifar10_demo.ipynb
-
-
-def knn_predict(
```
guarin, line 20: Moved this to `knn.py`. Backwards compatibility is maintained by importing it in `benchmarking/__init__.py`

lightly/utils/benchmarking/knn_classifier.py
```diff
@@ -0,0 +1,87 @@
...
+            knn_k=self.knn_k,
+            knn_t=self.knn_t,
+        )
+        topk = mean_topk_accuracy(
+            predicted_classes=predicted_classes, targets=targets, k=self.topk
+        )
+        log_dict = {f"val_top{k}": acc for k, acc in topk.items()}
+        self.log_dict(log_dict, prog_bar=True, sync_dist=True, batch_size=len(targets))
+
+    def on_validation_epoch_start(self) -> None:
```
guarin, line 64: We have to do this on validation start because `on_train_epoch_end` is not called if there is a single training epoch.

lightly/utils/benchmarking/knn_classifier.py
```diff
@@ -0,0 +1,87 @@
...
+    def on_fit_start(self) -> None:
+        # Freeze model weights.
+        deactivate_requires_grad(model=self.model)
+
+    def on_fit_end(self) -> None:
+        # Unfreeze model weights.
+        activate_requires_grad(model=self.model)
+
+    def configure_optimizers(self) -> None:
+        pass
```
guarin, line 87: Required because PyTorch Lightning expects `configure_optimizers` to be defined. Note that this is not a hack, PyTorch Lightning actually allows training without an optimizer.

lightly/utils/benchmarking/metric_callback.py
```diff
@@ -0,0 +1,30 @@
+from collections import defaultdict
+from typing import List
+
+from pytorch_lightning import LightningModule, Trainer
+from pytorch_lightning.callbacks import Callback
+from torch import Tensor
+
+
+class MetricCallback(Callback):
```
guarin, line 9: This is needed to get max performance over all epochs. PyTorch Lightning only stores logs from last epoch. This callback stores logs at the end of every epoch.

lightly/utils/lars.py
```diff
@@ -0,0 +1,158 @@
+import torch
+from torch.optim.optimizer import Optimizer, required
+
+
+class LARS(Optimizer):
```
guarin, line 5: Had to add this because PyTorch Lightning Bolts is not compatible with newer PyTorch Lightning versions. It is also the only dependency we had from PyTorch Lightning Bolts, so it is easier to just maintain it ourselves instead of adding an additional dependency.

lightly/utils/scheduler.py
```diff
@@ -27,12 +29,9 @@ def cosine_schedule(
...
-        # Note: we allow step == max_steps even though step starts at 0 and should end
-        # at max_steps - 1. This is because Pytorch Lightning updates the LR scheduler
-        # always for the next epoch, even after the last training epoch. This results in
-        # Pytorch Lightning calling the scheduler with step == max_steps.
-        raise ValueError(
-            f"The current step cannot be larger than max_steps but found step {step} and max_steps {max_steps}."
+        warnings.warn(
+            f"Current step number {step} exceeds max_steps {max_steps}.",
+            category=RuntimeWarning,
         )
```
guarin, line 35: Changed this to warning to avoid accidentally crashing training in the very last epoch. This didn't happen in the benchmarks but could easily happen if a user passes slightly wrong arguments.

lightly/utils/scheduler.py
```diff
@@ -60,25 +59,33 @@ class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
...
         self,
         optimizer: torch.optim.Optimizer,
-        warmup_epochs: int,
-        max_epochs: int,
-        last_epoch: int = -1,
+        warmup_epochs: float,
+        max_epochs: float,
+        last_epoch: float = -1,
+        start_value: float = 1.0,
+        end_value: float = 0.001,
```
guarin, line 82: Set this to >0 to avoid training with LR=0 in last epoch/steps.

lightly/utils/scheduler.py
```diff
@@ -1,9 +1,11 @@
+import warnings
+
 import numpy as np
 import torch
 
 
 def cosine_schedule(
-    step: int, max_steps: int, start_value: float, end_value: float
+    step: float, max_steps: float, start_value: float, end_value: float
```
guarin, line 8: Relaxed the types as the code also works with floats.

benchmarks/imagenet/resnet50/simclr.py
```diff
@@ -0,0 +1,100 @@
...
+        features0 = self.forward(view0).flatten(start_dim=1)
+        features1 = self.forward(view1).flatten(start_dim=1)
+        z0 = self.projection_head(features0)
+        z1 = self.projection_head(features1)
+        loss = self.criterion(z0, z1)
+        self.log(
+            "train_loss", loss, prog_bar=True, sync_dist=True, batch_size=len(targets)
+        )
+
+        cls_loss, cls_log = self.online_classifier.training_step(
```
philippmwirth, line 47: The behavior here is not intuitive:
- Reading this code it's not clear what happens to the gradients from the online classification task
- Looking the the online classifier it's clear that the backpropagation is stopped
- I'd consider moving the `.detach` here

philippmwirth: I like that you went by composition over inheritance for the online classifier but it looks hacky to call a second training step from within the first one. Can we solve it by doing a forward pass through the online classifier and calculating the classification loss in the training step of SimCLR? I know we'd need to configure more optimizers etc but it's more transparent imo.

lightly/models/utils.py
```diff
@@ -529,3 +531,39 @@ def nearest_neighbors(
     )  # [bsz, num_matches, feature_dimension]
 
     return filtered_input_maps, filtered_candidate_maps
+
+
+def get_weight_decay_parameters(
+    modules: Iterable[Module],
+    no_batch_norm: bool = True,
```
philippmwirth, line 538: Always use verbs to describe boolean parameters: `decay_batch_norm`

lightly/utils/benchmarking/knn_classifier.py
```diff
@@ -0,0 +1,99 @@
+from typing import Tuple, Union
+
+import torch
+import torch.nn.functional as F
+from pytorch_lightning import LightningModule
+from torch import Tensor
+from torch.nn import Module
+
+from lightly.models.utils import activate_requires_grad, deactivate_requires_grad
+from lightly.utils.benchmarking.knn import knn_predict
```
philippmwirth, line 10:
```suggestion
from lightly.utils.benchmarking import knn_predict
...
knn_predict.knn_predict(...)
```

lightly/utils/benchmarking/knn_classifier.py
```diff
@@ -0,0 +1,99 @@
...
+from lightly.utils.benchmarking.topk import mean_topk_accuracy
+
+
+class KNNClassifier(LightningModule):
```
philippmwirth, line 14: Do we follow any reference implementation here or is this our own contribution?

benchmarks/imagenet/resnet50/knn_eval.py
```diff
@@ -0,0 +1,78 @@
+from pathlib import Path
```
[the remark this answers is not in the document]
guarin: Mmm, I don't think it depends on architecture. Could move it one level up indeed.

benchmarks/imagenet/resnet50/knn_eval.py
```diff
@@ -0,0 +1,78 @@
...
+    accelerator: str,
+    devices: int,
+    num_classes: int,
+) -> None:
+    print("Running KNN evaluation...")
+
+    # Setup training data.
+    train_transform = T.Compose(
+        [
+            T.RandomResizedCrop(224),
```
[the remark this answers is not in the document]
guarin: Good point, we should use same augmentation as for validation transform.

benchmarks/imagenet/resnet50/linear_eval.py
```diff
@@ -0,0 +1,92 @@
...
+    train_dir: Path,
+    val_dir: Path,
+    log_dir: Path,
+    batch_size: int,
+    num_workers: int,
+    accelerator: str,
+    devices: int,
+    precision: str,
+    num_classes: int,
+) -> None:
```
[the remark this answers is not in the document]
guarin: SimCLR :)

benchmarks/imagenet/resnet50/main.py
```diff
@@ -0,0 +1,200 @@
...
+    batch_size: int,
+    epochs: int,
+    num_workers: int,
+    accelerator: str,
+    devices: int,
+    precision: str,
+    compile_model: bool,
+    methods: Union[Sequence[str], None],
+    num_classes: int,
+    no_knn_eval: bool,
```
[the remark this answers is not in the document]
guarin: `skip_...` seems the most clear to me.

benchmarks/imagenet/resnet50/simclr.py
guarin: > I'd consider moving the .detach here

Yep, good point.

> Can we solve it by doing a forward pass through the online classifier and calculating the classification loss in the training step of SimCLR? 

Then we also have to add the accuracy calculations and extra logging logic to every model which I would like to avoid.

> I know we'd need to configure more optimizers etc but it's more transparent imo.

Not sure I understand what you mean. Just note that we cannot use multiple different optimizers in a single training step without writing the optimization logic ourselves.

benchmarks/imagenet/resnet50/simclr.py
```diff
@@ -0,0 +1,100 @@
...
+            (features0, targets), batch_idx
+        )
+        self.log_dict(cls_log, sync_dist=True, batch_size=len(targets))
```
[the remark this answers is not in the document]
guarin: https://tensorboard.dev/experiment/m6IGt6YDS62qCaqOgPCqmw/#scalars

lightly/utils/benchmarking/knn_classifier.py
guarin: Our own :)

lightly/utils/benchmarking/linear_classifier.py
```diff
@@ -0,0 +1,91 @@
...
+from torch import Tensor
+from torch.nn import CrossEntropyLoss, Linear, Module
+from torch.optim import SGD
+
+from lightly.models.utils import activate_requires_grad, deactivate_requires_grad
+from lightly.utils.benchmarking.topk import mean_topk_accuracy
+from lightly.utils.scheduler import CosineWarmupScheduler
+
+
+class LinearClassifier(LightningModule):
```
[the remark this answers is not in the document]
guarin: Default parameters are from SimCLR. Will add it with the docstrings.

lightly/utils/benchmarking/linear_classifier.py
```diff
@@ -0,0 +1,91 @@
...
+    def validation_step(self, batch, batch_idx) -> Tensor:
+        loss, topk = self.shared_step(batch=batch, batch_idx=batch_idx)
+        batch_size = len(batch[1])
+        log_dict = {f"val_top{k}": acc for k, acc in topk.items()}
+        self.log("val_loss", loss, prog_bar=True, sync_dist=True, batch_size=batch_size)
+        self.log_dict(log_dict, prog_bar=True, sync_dist=True, batch_size=batch_size)
+        return loss
+
+    def configure_optimizers(self):
+        optimizer = SGD(
```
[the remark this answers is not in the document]
guarin: My plan was to subclass if it is necessary to change parameters. For example for ViT finetuning we'll need AdamW optmizer and different scheduler/lr/weight decay, ... I believe in that case subclassing is easier and more explicit.

guarin:
Thanks a lot for the review!

> I think we will need to write some good docstrings to make the code usable by others, too.

My plan was to add docstrings to the modules in `lightly/utils/benchmarking` and a readme to `benchmarks/README.md`. Eventually also readmes for the differet datasets/architectures.

benchmarks/imagenet/resnet50/knn_eval.py
```diff
@@ -0,0 +1,78 @@
...
+    train_dir: Path,
+    val_dir: Path,
+    log_dir: Path,
+    batch_size: int,
+    num_workers: int,
+    accelerator: str,
+    devices: int,
+    num_classes: int,
+) -> None:
+    print("Running KNN evaluation...")
```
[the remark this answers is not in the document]
guarin: Actually, this will add even more setup code. I am not sure if it is worth it. I'll leave it as prints for now.

PR #1163, opened 2023-04-24
Add support for MoCo v3

adamjstewart:
Exact same design as #1150. Should we just subclass one in the other?

I'll let you decide whether you want to default to v1/2 or v3 hyperparams.

Reference implementations:

* v1 and v2: https://github.com/facebookresearch/moco/blob/main/moco/builder.py#L34
* v3: https://github.com/facebookresearch/moco-v3/blob/main/moco/builder.py#L36

### DIFF
--- a/lightly/models/modules/heads.py
+++ b/lightly/models/modules/heads.py
@@ -130,19 +130,60 @@ class MoCoProjectionHead(ProjectionHead):
     "(...) we replace the fc head in MoCo with a 2-layer MLP head (hidden layer
     2048-d, with ReLU)" [0]
 
-    [0]: MoCo, 2020, https://arxiv.org/abs/1911.05722
+    "The projection head is a 3-layer MLP. The prediction head is a 2-layer MLP. The
+    hidden layers of both MLPs are 4096-d and are with ReLU; the output layers of both
+    MLPs are 256-d, without ReLU. In MoCo v3, all layers in both MLPs have BN" [2]
 
+    [0]: MoCo v1, 2020, https://arxiv.org/abs/1911.05722
+    [1]: MoCo v2, 2020, https://arxiv.org/abs/2003.04297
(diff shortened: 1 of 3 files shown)


lightly/models/modules/heads.py
```diff
@@ -128,21 +128,62 @@ class MoCoProjectionHead(ProjectionHead):
...
+    [2]: MoCo v3, 2021, https://arxiv.org/abs/2104.02057
     """
 
     def __init__(
-        self, input_dim: int = 2048, hidden_dim: int = 2048, output_dim: int = 128
+        self,
+        input_dim: int = 2048,
+        hidden_dim: int = 4096,
+        output_dim: int = 256,
+        num_layers: int = 2,
```
philippmwirth, line 147: This should be 3 for v3 I believe.

philippmwirth (changes requested):
Hi @adamjstewart thanks a lot for contributing. The changes look great 🙂

The only thing we still need to do is update the projection heads used in the benchmarks, see:

https://github.com/lightly-ai/lightly/blob/0d26bd257e8699a2bf6dc825863515377b934189/docs/source/getting_started/benchmarks/imagenet100_benchmark.py#L222

https://github.com/lightly-ai/lightly/blob/0d26bd257e8699a2bf6dc825863515377b934189/docs/source/getting_started/benchmarks/imagenette_benchmark.py#L317

You can simply replace the hidden dimension and the output dimension with the default values of MoCoV3.

lightly/models/modules/heads.py
philippmwirth: I just discussed this with @guarin and changing the defaults will change user code so we'd prefer to use the v2 settings as default values for now.

adamjstewart:
Let me know if you also want me to change the defaults for SimCLR from v2 to v1.

philippmwirth (approved):
Thanks for the changes!

PR #1167, opened 2023-04-26
Guarin lig 3049 add topk tests

guarin:
## Changes

* Improve docstring for `mean_topk_accuracy`
* Add tests for `mean_topk_accuracy`

### DIFF
--- a/lightly/utils/benchmarking/topk.py
+++ b/lightly/utils/benchmarking/topk.py
@@ -7,7 +7,24 @@ from torch import Tensor
 def mean_topk_accuracy(
     predicted_classes: Tensor, targets: Tensor, k: Sequence[int]
 ) -> Dict[int, Tensor]:
-    """Computes the mean accuracy for the specified values of k."""
+    """Computes the mean accuracy for the specified values of k.
+
+    The mean is calculated over the batch dimension.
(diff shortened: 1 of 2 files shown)


lightly/utils/benchmarking/topk.py
```diff
@@ -7,7 +7,24 @@
...
+            in descending order of confidence.
+        targets:
+            Tensor of shape (batch_size) containing the target classes.
+        k:
+            Sequence of integers specifying the values of k for which the accuracy
+            should be computed.
+
+    Returns:
+        Dictionary containing the mean accuracy for each value of k. For example for
+        k=(1, 5) the dictionary could look like this: {1: 0.4, 5: 0.6}.
```
MalteEbner, line 26: I think this could be explained better:
The mean accuracy is the mean number of samples in which the sample's target class is within the first k predicted classes of that sample.

MalteEbner (approved):
LGTM, apart from the docstring, which could be better.

lightly/utils/benchmarking/topk.py
guarin: Updated the docstring, I hope it is clearer now.

PR #1169, opened 2023-04-26
Guarin lig 3049 add linear classifier tests

guarin:
## Changes

* Add tests for `LinearClassifier`
* Pass model parameters to optimizer if model is not frozen

### DIFF
--- a/lightly/utils/benchmarking/linear_classifier.py
+++ b/lightly/utils/benchmarking/linear_classifier.py
@@ -123,8 +123,11 @@ class LinearClassifier(LightningModule):
         return loss
 
     def configure_optimizers(self):
+        parameters = list(self.classification_head.parameters())
+        if not self.freeze_model:
+            parameters += self.model.parameters()
         optimizer = SGD(
-            self.classification_head.parameters(),
+            parameters,
             lr=0.1 * self.batch_size * self.trainer.world_size / 256,
             momentum=0.9,
             weight_decay=0.0,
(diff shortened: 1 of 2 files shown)


lightly/utils/benchmarking/linear_classifier.py
```diff
@@ -123,8 +123,11 @@ def validation_step(self, batch, batch_idx) -> Tensor:
         return loss
 
     def configure_optimizers(self):
+        parameters = list(self.classification_head.parameters())
+        if not self.freeze_model:
+            parameters += self.model.parameters()
         optimizer = SGD(
-            self.classification_head.parameters(),
+            parameters,
```
guarin, line 130: Fix, forgot to pass parameters for model as well if model is not frozen.

tests/utils/benchmarking/test_linear_classifier.py
```diff
@@ -0,0 +1,94 @@
...
+        assert trainer.callback_metrics["val_loss"].item() > 0
+        assert trainer.callback_metrics["val_top1"].item() >= 0
+        assert (
+            trainer.callback_metrics["val_top5"].item()
+            >= trainer.callback_metrics["val_top1"].item()
+        )
+        assert trainer.callback_metrics["val_top5"].item() <= 1
+
+        # Verify that weights were updated.
+        assert not torch.all(torch.eq(initial_weights, linear.weight))
```
MalteEbner, line 48: This line (and the` freeze_model=`) are the only ones differing between the tests, right? Don't you think that is is possible to merge the tests? Otherwise, it becomes hard to read understand the tests and their difference.

MalteEbner: Ok, you call.

MalteEbner (approved):
Approved

PR #1174, opened 2023-04-27
Update cifar10 benchmark results

guarin:
## Changes

* Update cifar10 results
* Make batch size, epochs order in result tables consistent
* Indicate that we report KNN accuracy


MalteEbner (approved):
I like the formatting change and great that we have new results!

PR #1180, opened 2023-05-02
lig 3180 Check if torch.distributed is available

guarin:
## Changes
* Check if `torch.distributed` is available before using it

This PR should make it possible to use lightly with torch versions without distributed support. This is, for example, the case on OSX.

### DIFF
--- a/lightly/loss/dcl_loss.py
+++ b/lightly/loss/dcl_loss.py
@@ -2,7 +2,9 @@ from functools import partial
 from typing import Callable, Optional
 
 import torch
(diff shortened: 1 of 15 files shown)


michal-lightly (commented):
I did a search for "distributed" in the codebase. These files are not included:
* loss/pmsn_loss.py
* loss/vicregl_loss.py
* utils/dist.py

Just checking, we don't need to change them?

michal-lightly (approved):
Lgtm, thanks!

PR #1186, opened 2023-05-04
Add finetune_eval.py

guarin:
## Changes

* Add `finetune_eval.py`

As for linear eval, the parameters follow the SimCLR settings.
I don't have results yet, will add them later.


philippmwirth (approved):
Looks great!

PR #1189, opened 2023-05-05
lig-3208: Update ApiWorkflowClient for relevant filenames

huan-lightly-0:
Updated ApiWorkflowClient for handling relevant filenames with relevant filenames artefact ID. Note that `relevant_filenames_artefact_id` and `run_id` must both be given at the same time. This is consistent with the logic on the API server side.

Also added some test cases to check this behaviour.

### DIFF
--- a/lightly/api/api_workflow_datasources.py
+++ b/lightly/api/api_workflow_datasources.py
@@ -102,7 +102,7 @@ class _DatasourcesMixin:
         to: Optional[int] = None,
         relevant_filenames_file_name: Optional[str] = None,
         run_id: Optional[str] = None,
-        relevant_filenames_artefact_id: Optional[str] = None,
+        relevant_filenames_artifact_id: Optional[str] = None,
         use_redirected_read_url: Optional[bool] = False,
         progress_bar: Optional[tqdm.tqdm] = None,
     ) -> List[Tuple[str, str]]:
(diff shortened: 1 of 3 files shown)


lightly/api/api_workflow_datasources.py
```diff
@@ -101,6 +101,8 @@ def download_raw_samples(
         from_: int = 0,
         to: Optional[int] = None,
         relevant_filenames_file_name: Optional[str] = None,
+        run_id: Optional[str] = None,
+        relevant_filenames_artefact_id: Optional[str] = None,
```
philippmwirth, line 105: We use "artifact" everywhere else so please change this for consistency.

lightly/api/api_workflow_datasources.py
```diff
@@ -136,6 +145,8 @@ def download_raw_samples(
             relevant_filenames_file_name=relevant_filenames_file_name,
             use_redirected_read_url=use_redirected_read_url,
             progress_bar=progress_bar,
+            relevant_filenames_run_id=run_id,
```
philippmwirth, line 148: Should we check before that either both are `None` or none are `None`?

philippmwirth (changes requested):

lightly/api/api_workflow_datasources.py
huan-lightly-0: Updated

lightly/api/api_workflow_datasources.py
huan-lightly-0: Yup updated.

philippmwirth (approved):

PR #1191, opened 2023-05-08
lig-3225: Add new "all in one" sample creation endpoint

huan-lightly-0:
Add the new "all in one" sample creation endpoint to the API client. Write URLs will be returned along with sample IDs


guarin (approved):
Thanks!

PR #1193, opened 2023-05-08
lig-3208: Revert changes to ApiWorkflowClient.download_raw_samples

huan-lightly-0:
Reverted some changes in #1189. Changes to `download_raw_samples` were incorrect as `relevant_filenames_artifact_id` does not apply to that endpoint.

### DIFF
--- a/lightly/api/api_workflow_datasources.py
+++ b/lightly/api/api_workflow_datasources.py
@@ -101,8 +101,6 @@ class _DatasourcesMixin:
         from_: int = 0,
         to: Optional[int] = None,
         relevant_filenames_file_name: Optional[str] = None,
-        run_id: Optional[str] = None,
-        relevant_filenames_artifact_id: Optional[str] = None,
         use_redirected_read_url: Optional[bool] = False,
         progress_bar: Optional[tqdm.tqdm] = None,
     ) -> List[Tuple[str, str]]:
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_datasources.py
```diff
@@ -216,6 +194,12 @@ def download_raw_predictions(
                 "'run_id' should not be `None` when 'relevant_filenames_artifact_id' "
                 "is specified."
             )
+        relevant_filenames_kwargs = {}
+        if run_id and relevant_filenames_artifact_id:
```
guarin, line 198:
```suggestion
        if run_id is not None and relevant_filenames_artifact_id is not None:
```

Would be better to also explicitly check for `None` in the checks above.

lightly/api/api_workflow_datasources.py
```diff
@@ -282,6 +265,12 @@ def download_raw_metadata(
...
```
guarin, line 269: Same as in other comment.

guarin (approved):
The changes look good but it would be better to check for `None` explicitly. Otherwise the behavior can be unexpected for empty strings.

lightly/api/api_workflow_datasources.py
huan-lightly-0: Updated

lightly/api/api_workflow_datasources.py
huan-lightly-0: Updated

PR #1196, opened 2023-05-09
Shaun lig 3228 fix pip build and release action

shaundaley39:
See discussion in issue:
https://linear.app/lightly/issue/LIG-3228/fix-pip-build-and-release-action

Test workflow run:
https://github.com/lightly-ai/lightly/actions/runs/4925523237/jobs/8799856074

### DIFF
--- a/.github/workflows/release_pypi.yml
+++ b/.github/workflows/release_pypi.yml
@@ -25,6 +25,7 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: "3.10"
+        check-latest: true
     - name: Build and release
       id: build_and_release
       run: |
--- a/.github/workflows/test_release_pypi.yml
+++ b/.github/workflows/test_release_pypi.yml
@@ -1,0 +1,36 @@
+name: Build and release lightly
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - shaun-lig-3228-fix-pip-build-and-release-action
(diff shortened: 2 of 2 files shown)


.github/workflows/release_pypi.yml
```diff
@@ -25,6 +25,7 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: "3.10"
+        check-latest: true
```
guarin, line 28: What is the expected behavior here? AFAIK `check-latest` checks if the latest python version is installed. But this will never be the case because the version is pinned to `3.10` just above.

guarin (changes requested):
Added a question :)

.github/workflows/release_pypi.yml
guarin: If it was just a one time error I am also fine with leaving as is.

shaundaley39: `3.10` fixes on the minor version, but still consumes patch version updates with `check-latest: true`. That's the effect.

Likewise, I'd also be ok with not having this change. This would eliminate one potential cause.

guarin: I would keep it as is. There a no longer 3.10 versions released as current version is 3.11.

shaundaley39: Fair, that settles it.

PR #1239, opened 2023-05-17
Fix torch.distributed import in vicregl_loss.py

ibro45:


guarin (approved):
Oh, good find! Thanks a lot for the PR.

PR #1246, opened 2023-05-24
Update self-supervised learning docs

philippmwirth:
# Update self-supervised learning docs

Update the self-supervised learning docs to use transforms instead of collate functions.

### DIFF
--- a/docs/source/getting_started/lightly_at_a_glance.rst
+++ b/docs/source/getting_started/lightly_at_a_glance.rst
@@ -26,7 +26,7 @@ For example, the transforms below will apply the SimCLR image transform to the i
 
     from lightly.transforms.simclr_transform import SimCLRTransform
 
-    # The following transform will return two augmented image per input image.
+    # The following transform will return two augmented images per input image.
     transform = SimCLRTransform()
 
 Let's now load an image dataset and create a PyTorch dataloader.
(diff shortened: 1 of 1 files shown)


docs/source/getting_started/lightly_at_a_glance.rst
```diff
@@ -7,49 +7,46 @@ Lightly is a computer vision framework for training deep learning models using s
...
+For example, the transforms below will apply the SimCLR image transform to the input images.
 
 .. code-block:: python
 
-    import lightly.data as data
+    from lightly.transforms.simclr_transform import SimCLRTransform
 
-    # the collate function applies random transforms to the input images
-    collate_fn = data.ImageCollateFunction(input_size=32, cj_prob=0.5)
+    # The following transform will return two augmented image per input image.
```
guarin, line 29:
```suggestion
    # The following transform will return two augmented images per input image.
```

guarin (approved):
One typo, otherwise LGTM! I'll already approve.

Note that this part is no longer true:
> .. note:: You can also use a custom PyTorch `Dataset` instead of the 
          `LightlyDataset`. Just make sure your `Dataset` implementation returns
          a tuple of **(sample, target, filename)** to support the basic functions
          for training models. See :py:class:`lightly.data.dataset`
          for more information.

With the switch to transforms one can also just use the default torchvision datasets or a custom dataset without the `(sample, target, filename)` restriction. To make it work we have to adapt the dataloader loops from this:
```
for (x0, x1), _, _ in dataloader:
    ...
```
To this:
```
for batch in dataloader:
    (x0, x1) = batch[0]
     ...
```
Then we automatically support lightly datasets that return `(views, target, filename)` or any other dataset that returns the views as first item.

We can change this either in this PR or a follow-up, as you prefer.

PR #1250, opened 2023-05-25
Deprecate ApiWorkflowClient.selection

huan-lightly-0:
Deprecating `ApiWorkflowClient.selection`. Triggering selections from the API client will not be supported soon.


lightly/api/api_workflow_selection.py
```diff
@@ -82,6 +82,13 @@ def selection(
 
         """
 
+        warnings.warn(
```
MalteEbner, line 85: If you want, you can already remove the selection() method in this PR, otherwise, it is telling to use another deprecated method.

MalteEbner (approved):
LGTM

lightly/api/api_workflow_selection.py
huan-lightly-0: Added this to #1251. Will get that merged first.

PR #1251, opened 2023-05-25
lig-3363: Remove active learning

huan-lightly-0:
Removing most code for active learning since it's deprecated.

`SelectionConfig` is kept for now and will be cleaned up when `ApiWorkflowClient.selection` is removed.


MalteEbner (commented):
I would delete the whole file and everywhere it is called. The dataset and embedding upload via pip package is deprecated anyways, users should use the docker for it instead.

huan-lightly-0:
I updated this PR and deleted most code for active learning. There's just one thing I'm not sure about: `BoundingBox` is referenced in other pieces of code, so I moved it to `lightly.utils.bounding_box`. I suppose this is okay?

MalteEbner (approved):
I love to see this <3

> BoundingBox is referenced in other pieces of code, so I moved it to lightly.utils.bounding_box

That is a good solution, well done.

PR #1253, opened 2023-05-26
Use per device batch size for benchmarks

guarin:
## Changes

* Use per device batch size for benchmarks

This was already the case before this PR but it was not obvious.

### DIFF
--- a/benchmarks/imagenet/resnet50/finetune_eval.py
+++ b/benchmarks/imagenet/resnet50/finetune_eval.py
@@ -52,6 +52,7 @@ def finetune_eval(
     Parameters follow SimCLR settings.
     """
     print("Running fine-tune evaluation...")
+
     # Setup trainer.
     metric_callback = MetricCallback()
     trainer = Trainer(
--- a/benchmarks/imagenet/resnet50/finetune_eval.py
+++ b/benchmarks/imagenet/resnet50/finetune_eval.py
@@ -20,7 +20,7 @@ class FinetuneLinearClassifier(LinearClassifier):
         parameters += self.model.parameters()
         optimizer = SGD(
             parameters,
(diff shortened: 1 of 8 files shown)


benchmarks/imagenet/resnet50/finetune_eval.py
```diff
@@ -52,6 +52,21 @@ def finetune_eval(
     Parameters follow SimCLR settings.
     """
     print("Running fine-tune evaluation...")
+    # Setup trainer.
+    metric_callback = MetricCallback()
+    trainer = Trainer(
```
guarin, line 58: Trainer must be setup before dataloaders to scale batch size properly.

benchmarks/imagenet/resnet50/simclr.py
```diff
@@ -85,9 +85,9 @@ def configure_optimizers(self):
             # Square root learning rate scaling improves performance for small
             # batch sizes (<=2048) and few training epochs (<=200). Alternatively,
             # linear scaling can be used for larger batches and longer training:
-            #   lr=0.3 * self.batch_size * self.trainer.world_size / 256
+            #   lr=0.3 * self.batch_size / 256
```
philippmwirth, line 88: There's a slight problem introduced by this: For certain models we need to know the batch size per GPU in the initializer (e.g. for the size of the queue in SwaV). Do you have a suggestion how to address this?

guarin: Hmm, we could also use the `devices` parameter.

philippmwirth: I think it would be the cleanest approach. And as soon as we have the world size we can calculate the global batch size.

philippmwirth (approved):

PR #1258, opened 2023-05-29
Shaun lig 3308 apiworkflowclients nightly tests iter

shaundaley39:
## Improve Pagination

In general, we paginate an API endpoint where we expect to (potentially) get a lot of stuff from it.

Lots of stuff can cause a few problems:
- we can end up with JSON objects larger than the maximum supported size (solution: pagination)
- we can end up waiting a long time for all the data to be fetched/ assembled, even while earlier records are available (solution: pagination plus using an iterator on the client side so we can begin processing the first pages before later pages arrive)
- we can end up waiting a long time for all the data to be fetched/ assembled, even when we are often only interested in the first few records (solution: pagination and a lazy iterator on the client side)
- we can end up fetching such a massive amount of stuff that we're using too much RAM locally, or we waste time with allocation of lots of memory/ even disk operations (solution: pagination plus a client-side lazy iterator).

With this PR, we get lazy iteration over a paginated endpoint. Each time we have exactly one page remaining in our local iterator queue, we pre-fetch the next page. Where the consuming process is faster than the API supplier, this degenerates to a sequence of immediately consecutive page fetches. Where the consuming process is slower than the API supplier, there is never more than 2 pages queued. When the API supplier is the slower process, it is more important to set the page_size appropriately (for the API and the entity being fetched).

Advantages include:
- scaling to receive massive amounts of stuff from a paginated API endpoint without having to allocate memory for all of that stuff simultaneously (low memory requirements)
- ability to begin processing stuff from the API as soon as the first page arrives, without having to first wait for all of it (low start latency)
- ability to stop fetching early once we have what we need, in cases where we didn't need to see everything the API can return

Disadvantages include:
- if a series of paginated fetches is slower than a single fetch, we obviously pay that cost (by doing any kind of pagination)
- for some APIs, many concurrent page fetches might complete faster than the same number of sequential page fetches. This is more likely to be true if there is lots of horizontal scaling on the API (and DB) side. Where that is the case, we don't use it: the fastest supported fetching from the API in the above PR is a series of directly sequential paginated fetches. For Lightly, that's probably reasonable.

### DIFF
--- a/lightly/api/api_workflow_compute_worker.py
+++ b/lightly/api/api_workflow_compute_worker.py
@@ -262,6 +262,31 @@ class _ComputeWorkerMixin:
         )
         return response.id
 
+    def get_compute_worker_runs_iter(
+        self,
+        dataset_id: Optional[str] = None,
+    ) -> Iterator[DockerRunData]:
+        """Fetches all Lightly Worker runs for the user.
+
+        Args:
+            dataset_id:
+                Target dataset ID. Optional. If set, only runs with the given dataset
+                will be returned.
+
+        Returns:
+            Runs sorted by creation time from the oldest to the latest.
+
+        """
+        if dataset_id is not None:
+            return utils.paginate_endpoint(
+                self._compute_worker_api.get_docker_runs_query_by_dataset_id,
+                dataset_id=dataset_id,
+            )
+        else:
+            return utils.paginate_endpoint(
+                self._compute_worker_api.get_docker_runs,
+            )
+
     def get_compute_worker_runs(
         self,
         dataset_id: Optional[str] = None,
@@ -277,15 +302,7 @@ class _ComputeWorkerMixin:
             Runs sorted by creation time from the oldest to the latest.
 
         """
-        if dataset_id is not None:
-            runs: List[DockerRunData] = utils.paginate_endpoint(
-                self._compute_worker_api.get_docker_runs_query_by_dataset_id,
-                dataset_id=dataset_id,
-            )
-        else:
-            runs: List[DockerRunData] = utils.paginate_endpoint(
-                self._compute_worker_api.get_docker_runs,
-            )
+        runs: List[DockerRunData] = list(self.get_compute_worker_runs_iter(dataset_id))
         sorted_runs = sorted(runs, key=lambda run: run.created_at or -1)
         return sorted_runs
 
--- a/lightly/api/api_workflow_compute_worker.py
+++ b/lightly/api/api_workflow_compute_worker.py
@@ -266,7 +266,7 @@ class _ComputeWorkerMixin:
         self,
         dataset_id: Optional[str] = None,
     ) -> Iterator[DockerRunData]:
-        """Fetches all Lightly Worker runs for the user.
+        """Fetches in iterator over all Lightly Worker runs for the user.
 
         Args:
             dataset_id:
@@ -274,7 +274,7 @@ class _ComputeWorkerMixin:
                 will be returned.
 
         Returns:
-            Runs sorted by creation time from the oldest to the latest.
+            Runs iterator.
 
         """
         if dataset_id is not None:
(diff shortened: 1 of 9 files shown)


shaundaley39:
@huan-lightly-0 yes, that is absolutely correct. Duplicates can happen when iterating over pages.

In a WebUI, you're already used to this and it's totally natural. You read through one page, you go to the next page, and you see some of the same entities again because new entities joined the first page in the meantime. Normal.

Where a process is consuming a paginated API, this should definitely be kept in mind. If you want deduplication, it's important to do that.

huan-lightly-0 (approved):

shaundaley39:
Thanks @guarin, this has been modified to avoid any breaking changes.

There are good arguments against multi-threading (N>1) fetch threads against a paginated endpoint, both in this case and in general. E.g.
- where page displacement occurs on insertion of entities, multithreaded fetch vastly increases the risk of dropping entities, where requests go to different (eventually consistent) replicas, and some have a new insertion (displacing pages) while others don't. Whereas sequential fetches (on one fetch thread) keep this risk extremely low (eventually consistent replicas do typically become consistent faster than two sequential requests can be performed, and sequential requests will far more often go to the same replica anyway).
- it doesn't help us in cases where we want to stop fetching prematurely because we have what we need (most use cases for?)
- it also doesn't help in cases where the consuming thread has lower throughput than a dedicated-thread API fetch (another common case if we aren't just searching over entities for something interesting and we're actually processing them somehow).
- multithreaded hitting of paginated endpoints would slaughter our API (and DB) with demand spikes. For third party APIs, it would usually get you rate limited (and you would have no control over which of your requests get answered first).

I suspect this one-background-thread handling is probably close to the best general solution for paginated endpoints. (Optional deduplication support might be nice, would require a lambda to extract a unique key for each entity... but not useful for us) And so, I also half-expect someone to point to an out-of-the-box component that does precisely this already.

tests/api/test_utils.py
```diff
@@ -90,3 +92,33 @@ def test_get_lightly_server_location_from_env(self):
...
+            else:
+                return page_size * ["a"]
+
+        page_size = 8
+        some_iterator = paginate_endpoint(some_function, page_size=page_size)
+        some_list = list(some_iterator)
+        self.assertEqual((4 * page_size - 1) * ["a"], some_list)
+        self.assertEqual(len(some_list), (4 * page_size - 1))
+
+    def test_paginate_endpoint_raises_error_from_fetch_thread(self):
```
guarin, line 111: Is this test still needed?

tests/api/test_utils.py
```diff
@@ -90,3 +92,33 @@ def test_get_lightly_server_location_from_env(self):
         os.environ["LIGHTLY_SERVER_LOCATION"] = "https://api.dev.lightly.ai/ "
         host = get_lightly_server_location_from_env()
         self.assertEqual(host, "https://api.dev.lightly.ai")
+
+    def test_paginate_endpoint(self):
```
guarin, line 96: Could you add one more test with an empty reply?

lightly/api/api_workflow_compute_worker.py
```diff
@@ -262,30 +262,47 @@ def schedule_compute_worker_run(
...
         return response.id
 
-    def get_compute_worker_runs(
+    def get_compute_worker_runs_iter(
         self,
         dataset_id: Optional[str] = None,
-    ) -> List[DockerRunData]:
-        """Fetches all Lightly Worker runs for the user.
+    ) -> Iterator[DockerRunData]:
+        """Fetches in iterator over all Lightly Worker runs for the user.
```
guarin, line 269:
```suggestion
        """Returns an iterator over all Lightly Worker runs for the user.
```

guarin (approved):
LGTM!

Left some comments but feel free to merge/ignore.

guarin (approved):
LGTM!

Left some comments but feel free to merge/ignore.

guarin (approved):
LGTM!

Left some comments but feel free to merge/ignore.

tests/api/test_utils.py
shaundaley39: removed

tests/api/test_utils.py
shaundaley39: added

PR #1263, opened 2023-05-31
Guarin lig 3056 add mae imagenet benchmark

### LINKED ISSUE #1255: MAE

guarin:
Closes #1255 


Reference values from paper:
<img width="582" alt="Screen Shot 2023-06-15 at 08 50 01" src="https://github.com/lightly-ai/lightly/assets/43336610/8c340ebd-ba04-4b22-9209-933a6ed878a1">

<img width="592" alt="Screen Shot 2023-06-15 at 09 03 04" src="https://github.com/lightly-ai/lightly/assets/43336610/3b115670-94a1-4c0d-9c4e-5e0ecc0d180d">



The paper uses ViT-L with 1600 epoch pretraining by default while we use only ViT-B with 100 epoch pretraining. So I would expect finetune performance to be <82.3% following Figure 7. solo-learn reports 81.6% finetune top1 for ViT-B after 100 epochs. The paper reports ViT-B finetune performance of 83.6% after 1600 epoch pretraining.  

From Figure 7 it looks like linear eval should be <57.3 as we use ViT-B instead of ViT-L. ViT-B linear eval after 1600 epoch pretraining is reported as 68% in Table 12.

## First Results
```
Imagenet
    max val_online_cls_top1: 0.17285999655723572                                                                                                                                                                                                                    
    max val_online_cls_top5: 0.36941999197006226 
    knn val_top1: 0.13728000223636627                                                                                                                                                                                                                               
    knn val_top5: 0.2959200143814087 
    max linear val_top1: 0.4907799959182739                                   
    max linear val_top5: 0.728380024433136
    max finetune val_top1: 0.7822399735450745                                 
    max finetune val_top5: 0.9412800073623657
```
Results are a bit worse than expected. We should get ~82% finetune top1 but got 78.2%. Similarly for linear top1 we should get something a bit below 57% but we only got 49%.

Interestingly, knn top1 is really bad. Could be a sign that pretraining doesn't work properly or that (linear) finetuning is very important.


guarin:
Closing in favor of: https://github.com/lightly-ai/lightly/pull/1468

PR #1264, opened 2023-06-01
Revert changes in config

huan-lightly-0:
Some of these config values are still referenced in other parts of the codebase. Revert the last changes so that nothing breaks and clean up the config later.


shaundaley39 (approved):
Ok, reverting the config change will do no harm - let's remove lightly worker dependencies on these CLI config values in a follow-up issue.

PR #1266, opened 2023-06-02
VicRegL Vertical Flip Probability

### LINKED ISSUE #1240: VICRegL drop default vertical flip
We accidentally pass the horizontal flip probability also as vertical flip probability in `RandomResizedCropAndFlip` transform from `VICRegL`: https://github.com/lightly-ai/lightly/blob/61127338eacd50ad977b9eebaa7a297c404fecb9/lightly/transforms/random_crop_and_flip_with_grid.py#L151-L152

We should fix this by adding vertical flip probability as argument to the class and passing it accordingly to `RandomVerticalFlipWithLocation`. The default vertical flip probability for `VICRegL` should be `0` as only horizontal flip is used.

iamharsha1999:
Fixes #1240


guarin (approved):
Thanks a lot for the PR! Everything looks good.

PR #1273, opened 2023-06-07
I-JEPA

Natyren:
closes #issue_number

## Description
- this pr adds ijepa implementation to lightly
## Documentation
- my implementation need docs

- added inept


Natyren (commented):
draft version of pr

IgorSusmelj:
Hey @Natyren, can we help you with the PR? Adding I-JEPA to the library would be awesome :)

Natyren:
Hey, @IgorSusmelj! Sorry for the long wait, I've been really busy. Just wanted to give you a heads up that I'll be adding a new model to the framework next week.

Natyren:
Added collator and model (encoder and predictor, according to original paper and code).

Natyren:
Furthermore, I have examined the current setup in the examples by training the model on an A100 for 10 epochs. During this training, the loss consistently decreases, indicating that the model is effectively learning.

guarin (commented):
This looks really good!

I would suggest that we merge it as is and mark it as experimental until we have a fully tested version. There are also some points we'll have to address in follow-up issues/PRs (let me know if you are interested in helping out @Natyren):

* Torchvision ViT has no stoachstic depth ([here](https://github.com/pytorch/vision/blob/bb3aae7b2543637191ad9c810f082eae622534b8/torchvision/models/vision_transformer.py#L114)) while IJEPA ViT has it ([here](https://github.com/facebookresearch/ijepa/blob/52c1ae95d05f743e000e8f10a1f3a79b10cff048/src/models/vision_transformer.py#L169-L170)). I had the same issue when working on MAE. The annoying thing is that drop path is not even used during pretraining but only during finetunig. Without it we'll sadly not be able to properly evaluate the model. For MAE I added drop path only during finetuning but it is pretty hacky, see [here](https://github.com/lightly-ai/lightly/blob/9c4a8cfbf4cecd1419d02de0a49c333398542347/benchmarks/imagenet/vitb16/finetune_eval.py#L40). I guess our best bet is to switch to timm vision transformer which already has support for it, see #1301 
* I believe torchvision ViT also doesn't do [this](https://github.com/facebookresearch/ijepa/blob/52c1ae95d05f743e000e8f10a1f3a79b10cff048/src/models/vision_transformer.py#L264) part of the initialization but I have to check in detail.
* The collator looks like it should be in the transform instead. The only thing I can find that depends on the whole batch is the number of masks [here](https://github.com/facebookresearch/ijepa/blob/52c1ae95d05f743e000e8f10a1f3a79b10cff048/src/masks/multiblock.py#L147). But maybe I am missing something.
* The positional embedding functions should be moved from `lightly/models/modules/i_jepa` to `lightly/models/utils`. Note that I already partially implemented them in the MAE PR: #1263 
* Some docstrings are missing or in the wrong format. For example the `apply_masks` method or `IJEPAMaskCollator`.
* We should also add references to the original paper/code in major classes/functions. For example in `IJEPAMaskCollator`.
* Use `ijepa` instead of `i_jepa` for filenames  to make it more consistent with the class names.
* Use camel case in all class names. For example, change `IJEPA_predictor` to `IJEPAPredictor` etc.
* Add tests
* Finish examples and docs
* Add imagenet benchmark to `lightly/benchmarks`

Natyren:
@guarin Hi! This is cool, let's advance and start working on it! What are our next steps?

guarin:
Let's do these and then merge:

* We should also add references to the original paper/code in major classes/functions. For example in `IJEPAMaskCollator`.
* Use `ijepa` instead of `i_jepa` for filenames  to make it more consistent with the class names.
* Use camel case in all class names. For example, change `IJEPA_predictor` to `IJEPAPredictor` etc.

Natyren:
@guarin I will do it today, thank you for quick reply

Natyren:
@guarin Done. Please take a look

guarin:
Awesome! Thanks for the super fast update. I added some small changes and the note about experimental support. Tested the code and it is running 🥳 

Will merge once the tests pass.

guarin (approved):

PR #1275, opened 2023-06-08
Fix exporting information with ApiWorkflowClient

huan-lightly-0:
Update the methods for exporting information so that the API client returns information in dictionaries properly.


michal-lightly (approved):
LGTM

PR #1277, opened 2023-06-08
Use transforms instead of collate function in tutorials

philippmwirth:
# Use transforms instead of collate function in tutorials

Fix all tutorials to use transforms instead of collate functions. I ran all tutorials to make sure they work as expected.

### DIFF
--- a/docs/source/tutorials_source/package/tutorial_moco_memory_bank.py
+++ b/docs/source/tutorials_source/package/tutorial_moco_memory_bank.py
@@ -313,7 +313,7 @@ class Classifier(pl.LightningModule):
         self.log("train_loss_fc", loss)
         return loss
 
-    def training_epoch_end(self, outputs):
+    def on_train_epoch_end(self):
         self.custom_histogram_weights()
 
     # We provide a helper method to log weights in tensorboard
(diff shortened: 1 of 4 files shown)


docs/source/tutorials_source/package/tutorial_moco_memory_bank.py
```diff
@@ -332,14 +332,15 @@ def validation_step(self, batch, batch_idx):
...
         num = predicted.shape[0]
         correct = (predicted == y).float().sum()
+        self.validation_step_outputs.append((num, correct))
         return num, correct
 
-    def validation_epoch_end(self, outputs):
+    def on_validation_epoch_end(self):
         # calculate and log top1 accuracy
-        if outputs:
+        if self.validation_step_outputs:
```
guarin, line 340: I think you have to call `self.validation_step_outputs.clear()` at the end of this function. Otherwise outputs are aggregated over all epochs.

docs/source/tutorials_source/package/tutorial_pretrain_detectron2.py
```diff
@@ -261,7 +261,7 @@ def forward(self, x):
 
 # %%
 #
-# The :py:class:`lightly.data.collate.SimCLRCollateFunction` applies an ImageNet
+# The :py:class:`lightly.transforms.simclr.SimCLRTransform` applies an ImageNet
```
guarin, line 264:
```suggestion
# The :py:class:`~lightly.transforms.simclr.SimCLRTransform` applies an ImageNet
```

This will show only `SimCLRTransform` instead of `lightly.transforms.simclr.SimCLRTransform` in the text.

guarin (changes requested):
Thanks! I think you have to update the validation accuracy calculation but otherwise LGTM.

guarin (approved):

PR #1285, opened 2023-06-14
lig-3492: Fetch datasets owned by team members

huan-lightly-0:
Updated `get_datasets_by_name` and `get_datasets_iter` so that team datasets are returned when the flag `shared` is not `False`.

Test:
* Create one dataset named "test-dataset" using my account.
* Create another dataset named "test-dataset" using a service account.
* Call `get_datasets_by_name` against the API server and get 2 datasets.

### DIFF
--- a/lightly/api/api_workflow_datasets.py
+++ b/lightly/api/api_workflow_datasets.py
@@ -1,6 +1,6 @@
 import warnings
 from itertools import chain
-from typing import Iterator, List, Optional
+from typing import Iterator, List, Optional, Set
 
 from lightly.api import utils
 from lightly.openapi_generated.swagger_client.models import (
@@ -111,8 +111,8 @@ class _DatasetsMixin:
                 one dataset will be returned.
                 If True, returns only datasets which have been shared with the user. Can
                 return multiple datasets.
-                If None, returns datasets the users has access to. Can return multiple
-                datasets. Defaults to False.
+                If None, returns datasets the users has access to, including team
+                datasets. Can return multiple datasets. Defaults to False.
 
         Returns:
             A list of datasets that match the name. If no datasets with the name exist,
@@ -134,7 +134,7 @@ class _DatasetsMixin:
             >>> client.get_datasets_by_name(dataset_name="random-name")
             []
         """
-        datasets = []
+        datasets: List[DatasetData] = []
         if not shared or shared is None:
             datasets.extend(
                 self._datasets_api.get_datasets_query_by_name(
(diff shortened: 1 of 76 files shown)


lightly/api/api_workflow_datasets.py
```diff
@@ -111,8 +111,8 @@ def get_datasets_by_name(
                 one dataset will be returned.
                 If True, returns only datasets which have been shared with the user. Can
                 return multiple datasets.
-                If None, returns datasets the users has access to. Can return multiple
-                datasets. Defaults to False.
+                If None, returns datasets the users has access to, including team
+                datasets. Can return multiple datasets. Defaults to False.
```
michal-lightly, line 115: What is the difference between `shared=True` and `shared=None`? Team datasets are shared with the user, no? This is very counterintuitive.

michal-lightly: Imho in the ideal case the api client would have almost no logic and just mimic the api spec parameters. But I understand that this was changed in another PR.

tests/api_workflow/test_api_workflow_datasets.py
```diff
@@ -234,9 +234,12 @@ def test_get_datasets__shared(mocker: MockerFixture) -> None:
...
     client.get_datasets(shared=True)
-    mocked_pagination.assert_called_once_with(
-        mock_datasets_api.get_datasets, shared=True
-    )
+    assert mocked_pagination.call_count == 2
+    call_args = mocked_pagination.call_args_list
+    assert call_args[0][0] == (mock_datasets_api.get_datasets,)
+    assert call_args[0][1] == {"shared": True}
+    assert call_args[1][0] == (mock_datasets_api.get_datasets,)
+    assert call_args[1][1] == {"get_assets_of_team": True}
```
michal-lightly, line 242: We should also test that the results are unique if the multiple calls return duplicate datasets.

michal-lightly (changes requested):

lightly/api/api_workflow_datasets.py
```diff
@@ -151,7 +153,24 @@ def get_datasets_by_name(
...
                 )
             )
-        return datasets
+            datasets.extend(
+                self._datasets_api.get_datasets_query_by_name(
+                    dataset_name=dataset_name,
+                    exact=True,
+                    get_assets_of_team=True,
+                )
+            )
```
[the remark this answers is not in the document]
huan-lightly-0: No. According to what Jeremy told me, `get_assets_of_team` takes precedence over `shared`, so two calls are needed.

lightly/api/api_workflow_datasets.py
```diff
@@ -151,7 +153,24 @@ def get_datasets_by_name(
...
+
+        # De-duplicate datasets because results from shared=True and
+        # those from get_assets_of_team=True might overlap
+        dataset_ids: Set[str] = set()
+        filtered_datasets: List[DatasetData] = []
+        for dataset in datasets:
+            if dataset.id not in dataset_ids:
+                dataset_ids.add(dataset.id)
+                filtered_datasets.append(dataset)
```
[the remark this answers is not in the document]
huan-lightly-0: I think it's safer to keep this logic here. Now we already struggle understanding what is actually returned. Remembering whether or not the returned sets are disjoint will be even worse. Keeping the logic here ensures that no duplicate is returned in any case.

tests/api_workflow/test_api_workflow_datasets.py
huan-lightly-0: Updated the test case.

lightly/api/api_workflow_datasets.py
huan-lightly-0: Updated relevant docstrings.

michal-lightly (approved):
LGTM, thanks! :)

PR #1291, opened 2023-06-19
Guarin lig 3062 add mocov2 imagenet benchmark

### LINKED ISSUE #1018: MemoryBankModule - no distributed all gather prior to queueing a batch
If I'm not mistaken, the batch is not `all_gather`ed before the queue is updated. I was just comparing the code against the one it was based on (MoCo's memory bank) and noticed this: https://github.com/facebookresearch/moco/blob/78b69cafae80bc74cd1a89ac3fb365dc20d157d3/moco/builder.py#L55

Was this done on purpose?

Is it because different processes would use the same queued batches, unlike the DistributedDataLoader batches which are indexed in DDP way?

### LINKED ISSUE #1288: MoCov2

guarin:
## Changes

* Add MoCoV2 to imagenet benchmarks
* Add `gather_distributed` option to `MemoryBankModule`
* Move `memory_bank.py` from `lightly/loss` to `lightly/modules/models`
  * This makes conceptually more sense as the memory bank is its own module, independent from loss functions
  * It was required to avoid circular imports (we import `MemoryBankModule` in `lightly/modules/models/nn_memory_bank.py`)

Closes #1288
Closes #1018 

### Results

Ran a benchmark using the code from #1009:
```
max val_online_cls_top1: 0.5466600060462952
max val_online_cls_top5: 0.7995399832725525
max linear val_top1: 0.6080999970436096
max linear val_top5: 0.835860013961792
max finetune val_top1: 0.739300012588501
max finetune val_top5: 0.9166200160980225
knn val_top1: 0.4095799922943115
knn val_top5: 0.711139976978302
```
The results are slightly worse than expected. I would expect similar performance as SimCLR which has 63.2% linear top1 accuracy while mocov2 has only 60.8 here. I figured out that the transform parameters are slightly different in the original code and made an additional PR to update them: #1441. Will run a new benchmark with those parameters.

New results using this branch:
```
max val_online_cls_top1: 0.5602999925613403
max val_online_cls_top5: 0.808139979839325
max linear val_top1: 0.6151000261306763
max linear val_top5: 0.8411200046539307
max finetune val_top1: 0.7437400221824646
max finetune val_top5: 0.9182599782943726
knn val_top1: 0.41255998611450195
knn val_top5: 0.7188000082969666
```

It is still slightly worse than SimCLR which has 63.2% linear top 1, while MoCov2 has 61.5%. I think we can merge it like that, I can't find any differences atm. A possible reason might be that we use the SimCLR version with optimized lr for smaller batch sizes.

### Review

When reviewing please only consider these files:
* `benchmarks/imagenet/resnet50/mocov2.py`
* `README.md`
* `benchmarks.rst`

The other files have already been reviewed when their respective PRs were merged into this branch.


docs/source/getting_started/benchmarks.rst
```diff
@@ -31,6 +31,7 @@ Evaluation settings are based on the following papers:
...
   "BYOL", "Res50", "256", "100", "62.4", "84.7", "74.0", "91.9", "45.6", "74.8", "`link <https://tensorboard.dev/experiment/Z0iG2JLaTJe5nuBD7DK1bg>`_", "`link <https://lightly-ssl-checkpoints.s3.amazonaws.com/imagenet_resnet50_byol_2023-07-10_10-37-32/pretrain/version_0/checkpoints/epoch%3D99-step%3D500400.ckpt>`_"
   "DINO", "Res50", "128", "100", "68.2", "87.9", "72.5", "90.8", "49.9", "78.7", "`link <https://tensorboard.dev/experiment/DvKHX9sNSWWqDrRksllPLA>`_", "`link <https://lightly-ssl-checkpoints.s3.amazonaws.com/imagenet_resnet50_dino_2023-06-06_13-59-48/pretrain/version_0/checkpoints/epoch%3D99-step%3D1000900.ckpt>`_"
+  "MoCoV2", "Res50", "256", "100", "61.5", "84.1", "74.4", "91.8", "41.3", "71.9", "\-", "`link <https://lightly-ssl-checkpoints.s3.amazonaws.com/imagenet_resnet50_mocov2_2023-12-06_15-06-19/pretrain/version_0/checkpoints/epoch%3D99-step%3D500400.ckpt>`_"
```
guarin, line 34: The `"\-"` is necessary to render a `-` in the docs. Sphinx renders dashes as lists if they are not escaped.

Note that we can no longer provide tensorboard links because tensorboard.dev is shutting down. We have to replace them eventually, see https://github.com/lightly-ai/lightly/issues/1435

benchmarks/imagenet/resnet50/mocov2.py
```diff
@@ -0,0 +1,145 @@
...
+        self.save_hyperparameters()
+        self.batch_size_per_device = batch_size_per_device
+
+        resnet = resnet50()
+        resnet.fc = Identity()  # Ignore classification head
+        self.backbone = resnet
+        self.projection_head = MoCoProjectionHead()
+        self.query_backbone = copy.deepcopy(self.backbone)
+        self.query_projection_head = MoCoProjectionHead()
+        self.criterion = NTXentLoss(
```
ersi-lightly, line 36: Looking at the NTXentLoss function, it seems that the memory bank size should be an int, not a tuple

benchmarks/imagenet/resnet50/mocov2.py
```diff
@@ -0,0 +1,145 @@
...
+        self, batch: Tuple[List[Tensor], Tensor, List[str]], batch_idx: int
+    ) -> Tensor:
+        views, targets = batch[0], batch[1]
+
+        # Encode queries.
+        query_projections = self.forward_query_encoder(views[1])
+
+        # Momentum update. This happens between query and key encoding, following the
+        # original implementation from the authors:
+        # https://github.com/facebookresearch/moco/blob/5a429c00bb6d4efdf511bf31b6f01e064bf929ab/moco/builder.py#L142
```
ersi-lightly, line 79: super minor: but maybe adjust docstring length

benchmarks/imagenet/resnet50/mocov2.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def configure_optimizers(self):
+        # Don't use weight decay for batch norm, bias parameters, and classification
+        # head to improve performance.
+        # NOTE: The original implementation from the authors uses weight decay for all
+        # parameters.
+        params, params_no_weight_decay = get_weight_decay_parameters(
+            [self.query_backbone, self.query_projection_head]
+        )
+        optimizer = SGD(
```
ersi-lightly, line 116: was SGD also used in the paper? I was a bit surprised by the choice of optimizer

benchmarks/imagenet/resnet50/mocov2.py
```diff
@@ -0,0 +1,145 @@
...
+    ) -> Tensor:
+        images, targets = batch[0], batch[1]
+        features = self.forward(images).flatten(start_dim=1)
+        cls_loss, cls_log = self.online_classifier.validation_step(
+            (features.detach(), targets), batch_idx
+        )
+        self.log_dict(cls_log, prog_bar=True, sync_dist=True, batch_size=len(targets))
+        return cls_loss
+
+    def configure_optimizers(self):
```
ersi-lightly, line 108: type of return parameters :)

benchmarks/imagenet/resnet50/mocov2.py
```diff
@@ -0,0 +1,145 @@
...
+                    "params": params_no_weight_decay,
+                    "weight_decay": 0.0,
+                },
+                {
+                    "name": "online_classifier",
+                    "params": self.online_classifier.parameters(),
+                    "weight_decay": 0.0,
+                },
+            ],
+            lr=0.03 * self.batch_size_per_device * self.trainer.world_size / 256,
```
ersi-lightly, line 130: I might be missing sth here, but why the choice of 0.3 scaling?

docs/source/getting_started/benchmarks.rst
ersi-lightly, line 34: ok! I think we can discuss in general how we report results (maybe we could have mean and std) and also not compare with fixed number of epochs, but rather a stopping criterion like patience using the validation set, but this is not sth to discuss for this PR :)

benchmarks/imagenet/resnet50/mocov2.py
guarin: This PR also included changes to the memory bank to make it accept a (num_features, feature_dim) tuple instead of a single `num_features` argument. The reason is that we have to know the `feature_dim` when creating the memory bank for distributed processing. Before, we would dynamically get the `feature_dim` from the first batch stored in the bank but in a distributed setting you cannot dynamically change the parameter size.

benchmarks/imagenet/resnet50/mocov2.py
guarin: Yes :) SGD with momentum is still a popular choice in self-supervised learning.

benchmarks/imagenet/resnet50/mocov2.py
guarin: 0.03 is the base learning rate from the paper: https://github.com/facebookresearch/moco/blob/1999df86de600027cd2702cbd7cd7b7d2a53559d/main_moco.py#L81

docs/source/getting_started/benchmarks.rst
guarin: Fixed epoch is standard in self-supervised learning, early stopping is usually not that helpful for such large datasets. And reporting mean/std is sadly super expensive as a single run takes multiple days :/

benchmarks/imagenet/resnet50/mocov2.py
guarin: Not sure if we mean the same thing, but it is already adjusted (first line wraps at 90 chars) and the link is on a new line (we don't wrap links over multiple lines).

ersi-lightly (approved):

PR #1293, opened 2023-06-22
Memory bank distributed training support

guarin:
## Changes

* Add feature dimension when setting memory bank size. This is required for distributed training because the buffer size cannot be changed dynamically.
* Cleanup memory bank implementation.
* Add `feature_dim_first` argument to memory bank. This is required for backwards compatibility. In the future we should be able to set `feature_dim_first=False` and remove the requirement for transposing features coming from the feature bank.
* Update classes/functions using memory bank  from `size=num_features` to `size=(num_features, dim)`.

## How was it tested?
* Added unit tests

### DIFF
--- a/benchmarks/imagenet/resnet50/mocov2.py
+++ b/benchmarks/imagenet/resnet50/mocov2.py
@@ -33,7 +33,7 @@ class MoCoV2(LightningModule):
         self.projection_head = MoCoProjectionHead()
         self.query_backbone = copy.deepcopy(self.backbone)
         self.query_projection_head = MoCoProjectionHead()
-        self.criterion = NTXentLoss(temperature=0.2, memory_bank_size=65536)
+        self.criterion = NTXentLoss(temperature=0.2, memory_bank_size=(65536, 128))
 
         self.online_classifier = OnlineLinearClassifier(num_classes=num_classes)
 
--- a/lightly/models/modules/memory_bank.py
+++ b/lightly/models/modules/memory_bank.py
@@ -22,14 +22,21 @@ class MemoryBankModule(Module):
 
     Attributes:
         size:
-            Number of keys the memory bank can store. If set to 0,
-            memory bank is not used.
+            Size of the memory bank as (num_features, dim) tuple. If num_features is 0
+            then the memory bank is disabled. Deprecated: If only a single integer is
+            passed, it is interpreted as the number of features and the feature
+            dimension is inferred from the first batch stored in the memory bank.
+            Leaving out the feature dimension might lead to errors in distributed
+            training.
         gather_distributed:
             If True then negatives from all gpus are gathered before the memory bank
             is updated. This results in more frequent updates of the memory bank and
             keeps the memory bank contents independent of the number of gpus. But it has
             the drawback that synchronization between processes is required and
             diversity of the memory bank content is reduced.
+        dim_first:
+            If True, the memory bank returns features with shape (dim, num_features).
+            If False, the memory bank returns features with shape (num_features, dim).
 
     Examples:
         >>> class MyLossFunction(MemoryBankModule):
--- a/lightly/models/modules/memory_bank.py
+++ b/lightly/models/modules/memory_bank.py
@@ -60,7 +60,7 @@ class MemoryBankModule(Module):
 
         if any(x < 0 for x in size_tuple):
             raise ValueError(
-                f"Illegal memory bank size {size}, entries must be non-negative."
+                f"Illegal memory bank size {size}, all entries must be non-negative."
             )
 
         self.size = size_tuple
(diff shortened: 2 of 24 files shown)


benchmarks/imagenet/resnet50/main.py
```diff
@@ -230,7 +230,7 @@ def pretrain(
         logger=TensorBoardLogger(save_dir=str(log_dir), name="pretrain"),
         precision=precision,
         strategy="ddp_find_unused_parameters_true",
-        sync_batchnorm=True,
+        sync_batchnorm=accelerator != "cpu",  # Sync batchnorm is not supported on CPU.
```
guarin, line 233: Added this to make benchmarks runnable on CPU, makes testing distributed training easier.

lightly/models/modules/memory_bank.py
```diff
@@ -20,26 +22,30 @@ class MemoryBankModule(torch.nn.Module):
...
+            dimension is inferred from the first batch stored in the memory bank.
+            Leaving out the feature dimension might lead to errors in distributed
+            training.
         gather_distributed:
             If True then negatives from all gpus are gathered before the memory bank
             is updated. This results in more frequent updates of the memory bank and
             keeps the memory bank contents independent of the number of gpus. But it has
             the drawback that synchronization between processes is required and
             diversity of the memory bank content is reduced.
+        feature_dim_first:
```
guarin, line 37: `feature_dim_first=False` makes working with memory bank much easier.

lightly/utils/benchmarking/metric_callback.py
```diff
@@ -60,7 +60,7 @@ def _append_metrics(
         self, metrics_dict: Dict[str, List[float]], trainer: Trainer
     ) -> None:
         for name, value in trainer.callback_metrics.items():
-            if isinstance(value, float) or (
-                isinstance(value, Tensor) and value.numel() == 1
-            ):
-                metrics_dict.setdefault(name, []).append(float(value))
+            if isinstance(value, Tensor) and value.numel() != 1:
+                # Skip non-scalar tensors.
+                continue
```
guarin, line 65: `trainer.callback_metrics.items()`  returns `(str, Tensor)` tuples and mypy pointed out that `isinstance(value, float)` is impossible if the value is always a tensor. As this check previously passed CI, I believe that either pytorch lightning changed the typing of this function, or I have a different mypy version locally. In any case, I changed the test slightly to make it pass in all cases. The functionality should remain exactly the same :)

lightly/models/modules/memory_bank.py
```diff
@@ -48,38 +54,67 @@ class MemoryBankModule(torch.nn.Module):
...
+                ),
+                UserWarning,
+            )
+        else:
+            self._init_memory_bank(dim=None)
+
     @torch.no_grad()
-    def _init_memory_bank(self, dim: int) -> None:
-        """Initialize the memory bank if it's empty
+    def _init_memory_bank(self, dim: Union[Sequence[int], None]) -> None:
```
philippmwirth, line 101: This is a bit convoluted I think it'd be a better solution to pass the `size` explicitly:
```python
    @torch.no_grad()
    def _init_memory_bank(self, size: Sequence[int]) -> None:
        """Initialize the memory bank if it's empty.
        Args:
            size:
                The dimensions of the memory bank.
        """
        self.bank = torch.randn(size).type_as(self.bank)
        self.bank = torch.nn.functional.normalize(self.bank, dim=-1)
        self.bank_ptr = torch.zeros(1).type_as(self.bank_ptr)
```

Also, we should initialize the memory bank when the size is passed in the constructor is a tuple.

philippmwirth (changes requested):

lightly/models/modules/memory_bank.py
guarin: Updated :)

> Also, we should initialize the memory bank when the size is passed in the constructor is a tuple.

This is already implemented.

benchmarks/imagenet/resnet50/mocov2.py
```diff
@@ -33,7 +33,11 @@ def __init__(self, batch_size_per_device: int, num_classes: int) -> None:
         self.projection_head = MoCoProjectionHead()
         self.query_backbone = copy.deepcopy(self.backbone)
         self.query_projection_head = MoCoProjectionHead()
-        self.criterion = NTXentLoss(temperature=0.2, memory_bank_size=(65536, 128))
+        self.criterion = NTXentLoss(
+            temperature=0.2,
+            memory_bank_size=(65536, 128),
+            gather_distributed=self.trainer.num_devices > 1,
```
guarin, line 39: This makes sure that the memory bank is identical over all processes (keys are gathered before the memory bank is updated). This follows the original implementation: https://github.com/facebookresearch/moco/blob/1999df86de600027cd2702cbd7cd7b7d2a53559d/moco/builder.py#L68

PR #1299, opened 2023-06-28
Correctly create SelectionConfig with repeated object references

michal-lightly:
Fixes a bug when `selection_config_from_dict` would fail if the input dictionary contains multiple references to the same strategy object.

### DIFF
--- a/lightly/api/api_workflow_compute_worker.py
+++ b/lightly/api/api_workflow_compute_worker.py
@@ -618,13 +618,12 @@ class _ComputeWorkerMixin:
 
 def selection_config_from_dict(cfg: Dict[str, Any]) -> SelectionConfig:
     """Recursively converts selection config from dict to a SelectionConfig instance."""
-    strategies = [
-        SelectionConfigEntry(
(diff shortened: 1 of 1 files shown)


lightly/api/api_workflow_compute_worker.py
```diff
@@ -618,12 +618,14 @@ def get_compute_worker_run_tags(self, run_id: str) -> List[TagData]:
 
 def selection_config_from_dict(cfg: Dict[str, Any]) -> SelectionConfig:
     """Recursively converts selection config from dict to a SelectionConfig instance."""
+    strategies = [
+        SelectionConfigEntry(
+            input=SelectionConfigEntryInput(**entry["input"]),
+            strategy=SelectionConfigEntryStrategy(**entry["strategy"]),
```
guarin, line 624: This silently discards any additional keys in `entry` that are not `input` or `strategy`. Users would then not notice that there is something wrong with the config.

guarin (changes requested):

guarin (approved):
LGTM!

PR #1300, opened 2023-06-29
lig-3581: Update `PredictionTaskSchema` in API client

huan-lightly-0:
Class `PredictionTaskSchema` in the API client is updated to leverage OpenAPI feature `oneOf`.


guarin (approved):
LGTM!

PR #1302, opened 2023-07-01
Igor lig 3059 add dcl dclw imagenet benchmark

### LINKED ISSUE #1303: DCL & DCLW

IgorSusmelj:
- Add ImageNet benchmark for DCL and DCLW models
Closes #1303 

Results for DCL:
```
knn val_top1: 0.49592000246047974
knn val_top5: 0.7745599746704102

max linear val_top1: 0.6510999798774719                                   
max linear val_top5: 0.862339973449707

max finetune val_top1: 0.7349600195884705
max finetune val_top5: 0.9171800017356873
```

Results for DCLW:
```
knn val_top1: 0.4851999878883362
knn val_top5: 0.7682399749755859

max linear val_top1: 0.6449599862098694                                   
max linear val_top5: 0.8599600195884705

max finetune val_top1: 0.7316399812698364                                 
max finetune val_top5: 0.9147400259971619
```

For comparison:
These are the SimCLR results:
```
knn val_top1: 0.44764000177383423                                         
knn val_top5: 0.7394000291824341

max linear val_top1: 0.6315199732780457                                   
max linear val_top5: 0.8517000079154968

max finetune val_top1: 0.7389600276947021                                 
max finetune val_top5: 0.9190400242805481
```

### DIFF
--- a/benchmarks/imagenet/resnet50/main.py
+++ b/benchmarks/imagenet/resnet50/main.py
@@ -10,6 +10,7 @@ import linear_eval
 import simclr
 import swav
 import dcl
+import dclw
 import torch
 from pytorch_lightning import LightningModule, Trainer
 from pytorch_lightning.callbacks import (
@@ -47,6 +48,7 @@ METHODS = {
     "simclr": {"model": simclr.SimCLR, "transform": simclr.transform},
     "swav": {"model": swav.SwAV, "transform": swav.transform},
     "dcl": {"model": dcl.DCL, "transform": simclr.transform},
+    "dclw": {"model": dclw.DCLW, "transform": simclr.transform},
 }
 
 
--- a/benchmarks/imagenet/resnet50/main.py
+++ b/benchmarks/imagenet/resnet50/main.py
@@ -44,11 +44,11 @@ parser.add_argument("--skip-linear-eval", action="store_true")
 parser.add_argument("--skip-finetune-eval", action="store_true")
 
 METHODS = {
+    "dcl": {"model": dcl.DCL, "transform": simclr.transform},
+    "dclw": {"model": dclw.DCLW, "transform": simclr.transform},
     "dino": {"model": dino.DINO, "transform": dino.transform},
     "simclr": {"model": simclr.SimCLR, "transform": simclr.transform},
     "swav": {"model": swav.SwAV, "transform": swav.transform},
-    "dcl": {"model": dcl.DCL, "transform": simclr.transform},
-    "dclw": {"model": dclw.DCLW, "transform": simclr.transform},
 }
 
 
--- a/benchmarks/imagenet/resnet50/main.py
+++ b/benchmarks/imagenet/resnet50/main.py
@@ -44,8 +44,8 @@ parser.add_argument("--skip-linear-eval", action="store_true")
 parser.add_argument("--skip-finetune-eval", action="store_true")
 
 METHODS = {
-    "dcl": {"model": dcl.DCL, "transform": simclr.transform},
-    "dclw": {"model": dclw.DCLW, "transform": simclr.transform},
+    "dcl": {"model": dcl.DCL, "transform": dcl.transform},
+    "dclw": {"model": dclw.DCLW, "transform": dclw.transform},
     "dino": {"model": dino.DINO, "transform": dino.transform},
     "simclr": {"model": simclr.SimCLR, "transform": simclr.transform},
     "swav": {"model": swav.SwAV, "transform": swav.transform},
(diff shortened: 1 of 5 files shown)


IgorSusmelj:
From the paper (https://arxiv.org/pdf/2110.06848.pdf):

> result on ImageNet-1K using linear evaluation. With DCL, SimCLR can achieve 64.6% top-1 accuracy with only 100 epochs compared to SimCLR baseline: 57.5%;

- @guarin, which SimCLR version did we benchmark so far? It can't be the original one as our accuracy is much higher `63.2` vs `57.5`.

Both, the original SimCLR paper and the DCL paper refer to the experiment with batch size 256 and training for 100 epochs which results in `57.5` accuracy for linear evaluation.

guarin:
It is the original one but with square root lr scaling, see appendix B.1 in the paper:
Screen Shot 2023-07-10 at 17 20 36

IgorSusmelj:
> It is the original one but with square root lr scaling, see appendix B.1 in the paper: Screen Shot 2023-07-10 at 17 20 36

Ok, in the paper they state that they use the default settings 
image

Let's do the following. 
- We update the SimCLR ImageNet code to use linear lr and call it SimCLRv1. We add a comment that one could also use square root learning rate for the model and get slightly better results for smaller batch sizes/fewer epochs
- We make a SimCLRv2 version that uses the modifications from the follow up paper (3 layer projection head etc.)
- I'll then rerun the experiments for DCL and DCLW to use SimCLRv1. the SimCLR models are trained quickly.

In the future we should be more careful with the reference implementations. If we don't follow the settings in the paper we will have issues with all the other papers that build up on the previous one.

guarin:
> In the future we should be more careful with the reference implementations. Maybe we review the PRs and check for the parameters before we start the training run. If we don't follow the settings in the paper we will have issues with all the other papers that build up on the previous one.

We have chosen those parameters on purpose because the paper reported that those work best for small batch sizes. There are also comments in the code for the relevant parameters, why they were chosen, and what to do for larger batch sizes. We also did the same for the other models. Just using default parameters without adjusting for batch size/epochs results in bad performance for most models and I don't think this is helpful as the reported numbers wouldn't allow a fair comparison between the different methods. If we truly want to reproduce the models we also have to use the same batch size and training epochs.

> We make a SimCLRv2 version that uses the modifications from the follow up paper (3 layer projection head etc.)

Could you add an issue for it?

> I'll then rerun the experiments for DCL and DCLW to use SimCLRv1. the SimCLR models are trained quickly.

Sure

> I'd not discard the current results but put them to a section of "additional experiments". I'd also drop all the others we did. E.g. the BYOL one with the wrong learning rate. It would be a pity to discard the results completely.
 
I think it is ok to leave them in the PR, otherwise this will be super hard to maintain.

IgorSusmelj:
I understand your points. How about at least marking them in the readme?
Something like this
image

guarin:
> I understand your points. How about at least marking them in the readme? Something like this @guarin

I like it! Maybe we could even add a `See Appendix B.1 in [SimCLR paper](https://arxiv.org/abs/2002.05709)`.

benchmarks/imagenet/resnet50/main.py
```diff
@@ -45,6 +47,8 @@
     "dino": {"model": dino.DINO, "transform": dino.transform},
     "simclr": {"model": simclr.SimCLR, "transform": simclr.transform},
     "swav": {"model": swav.SwAV, "transform": swav.transform},
+    "dcl": {"model": dcl.DCL, "transform": simclr.transform},
+    "dclw": {"model": dclw.DCLW, "transform": simclr.transform},
```
guarin, line 51:
```suggestion
    "dcl": {"model": dcl.DCL, "transform": dcl.transform},
    "dclw": {"model": dclw.DCLW, "transform": dclw.transform},
```

Could you also move them to the top of the dict? Then everything stays alphabetically sorted.

guarin (changes requested):
The code and results LGTM!

Could you add the results also to the benchmarks docs?

benchmarks/imagenet/resnet50/main.py
```diff
@@ -42,6 +44,8 @@
 parser.add_argument("--skip-finetune-eval", action="store_true")
 
 METHODS = {
+    "dcl": {"model": dcl.DCL, "transform": simclr.transform},
+    "dclw": {"model": dclw.DCLW, "transform": simclr.transform},
```
guarin, line 48:
```suggestion
    "dcl": {"model": dcl.DCL, "transform": dcl.transform},
    "dclw": {"model": dclw.DCLW, "transform": dclw.transform},
```

guarin (approved):
One last thing but I'll already approve

benchmarks/imagenet/resnet50/main.py
IgorSusmelj: I changed that!

PR #1306, opened 2023-07-03
lig-3522: Add `use_datapool` to config

huan-lightly-0:
Add a new flag `use_datapool` to Lightly config. This flag tells the worker if it should use the datapool during a run.


guarin (approved):
LGTM

PR #1310, opened 2023-07-06
lig-3659: align import style

huan-lightly-0:
Some functions are still imported directly, which is against the convention where we prefer to import modules and then reference the functions through the modules. This PR updates those lines.


guarin (approved):
Thanks! LGTM

PR #1317, opened 2023-07-14
Add timm dependency

### LINKED ISSUE #1301: Add timm dependency?
Lightly currently only relies on torch, torchvision, and pytorch lightning for the model implementations. This has worked really well for all convnet/resnet based SSL methods but is much harder for the newer, vision transformer based methods. For example, we built MAE on top of the torchvision vision transformer which is lacking some key features (such as stochastic depth dropout) from the [original implementation](https://github.com/facebookresearch/mae) based on [timm](https://github.com/huggingface/pytorch-image-models).

Although our implementation works reasonably well, it doesn't (yet) match the original paper's implementation (see #1263). To make things worse, other popular transformer based SSL methods do not rely on torchvision either. This is understandable as torchvision currently only provides vanilla [vit](https://pytorch.org/vision/main/models/vision_transformer.html) and [swin](https://pytorch.org/vision/main/models/swin_transformer.html). Most newer method either implement their own vision transformer, build on top of timm, or mix the two, see:
* [dinov2](https://github.com/facebookresearch/dinov2/blob/main/dinov2/layers/attention.py)
* [i-jepa](https://github.com/facebookresearch/ijepa/blob/main/src/models/vision_transformer.py)
* [mae](https://github.com/facebookresearch/mae/blob/main/models_mae.py)

I want to avoid having to implement all these vision transformers in lightly and would prefer to use another framework for the backbones. I think timm is the best option we have there. But even with timm we might not exactly match the original implementations… Trying to find differences between two vision transformer implementations is not straight forward.

guarin:
Closes #1301 

## Changes
* Add timm dependency to `torch.txt`

Added it to `torch.txt` to avoid installing `timm` if only the api part of the package is used.

### DIFF
--- a/requirements/torch.txt
+++ b/requirements/torch.txt
@@ -3,3 +3,4 @@ torchvision
 # Note: pytorch_lightning>=1.5 is required for CLI
 # https://github.com/lightly-ai/lightly/issues/912
 pytorch_lightning>=1.0.4
+timm


requirements/torch.txt
```diff
@@ -3,3 +3,4 @@ torchvision
 # Note: pytorch_lightning>=1.5 is required for CLI
 # https://github.com/lightly-ai/lightly/issues/912
 pytorch_lightning>=1.0.4
+timm
```
IgorSusmelj, line 6: Why can't we make it an optional dependency? Timm now installs huggingface hub and has restrictions with older PyTorch versions. See: https://github.com/huggingface/pytorch-image-models/blob/main/requirements.txt

IgorSusmelj (changes requested):
I'd make it optional and write in the code where users would need to install it.

guarin:
Will keep it optional for now :)

PR #1318, opened 2023-07-14
remove _UploadDatasetMixin

MalteEbner:
## Description

- remove the `_UploadDatasetMixin`, which is deprecated
- remove the corresponding tests
- I did not remove anything in the `MockedApiWorkflowClient`, because the functions there might be used by other methods. In the first draft I tried, but that led to tests succeeding on my machine and failing in CI.
Removing the `MockedApiWorkflowClient` is a separate issue in my eyes.


guarin:
Unit tests failed because of an unrelated codecov 502 error. Will merge.

PR #1322, opened 2023-07-15
Moving positional embedding functions from ijepa to utils

Natyren:
Move positional embedding functions from ijepa to utils


lightly/models/utils.py
```diff
@@ -596,3 +597,72 @@ def repeat_interleave_batch(x, B, repeat):
         dim=0,
     )
     return x
+
+
+def _get_2d_sincos_pos_embed(embed_dim, grid_size, cls_token=False):
```
guarin, line 602:
```suggestion
def get_2d_sincos_pos_embed(embed_dim: int, grid_size: int, cls_token: bool = False) -> NDArray[np.float_]:
```

We use `_function_name` for private functions (only used in the same file) and `function_name` for public functions that can be used in different files. Could you make all the functions public?

I also added type hints, could you add them to the other functions as well?

lightly/models/utils.py
```diff
@@ -596,3 +597,72 @@ def repeat_interleave_batch(x, B, repeat):
...
+    """
+    Made 2d sincos positial embeddings. Code inspired by [0].
+
+    - [0]: https://github.com/facebookresearch/ijepa
+    grid_size: int of the grid height and width
+    return:
+    pos_embed: [grid_size*grid_size, embed_dim] or [1+grid_size*grid_size, embed_dim] (w/ or w/o cls_token)
```
guarin, line 609:
```suggestion
    Returns 2D sin-cos embeddings. Code from [0].

    - [0]: https://github.com/facebookresearch/ijepa
    
    Args:
        embed_dim:
            Embedding dimension.
        grid_size:
            Grid height and width. Should usually be set to sqrt(sequence length).
        cls_token:
            If True, a positional embedding for the class token is prepended to the returned
            embeddings.
    
    Returns:
        Positional embeddings array with size (grid_size * grid_size, embed_dim) if cls_token is False. 
        If cls_token is True, a (1 + grid_size * grid_size, embed_dim) array is returned.
```

guarin: Could you adapt the other comments as well?

Natyren: @guarin Hi! I will do it, thanks for review

Natyren: @guarin please take a look.

guarin (approved):
LGTM! Thanks for the PR!

I formatted the files and cleaned up git history. You can run formatting with `make format`, see our contribution guide: https://github.com/lightly-ai/lightly/blob/master/CONTRIBUTING.md

Regarding git history, could you next time make a PR from a clean branch based on lightly/master? The we don't get all the additional commits in the history. Thanks!

Natyren:
@guarin Thank you.  Next time I'll do it like according to contribution guide.

PR #1325, opened 2023-07-17
Add transforms for BYOL

### LINKED ISSUE #1324: Clarify SimCLR vs BYOL augmentations
The BYOL paper is a bit misleading as it states that it uses SimCLR augmentations:
image

However, BYOL additionally uses Solarization:
image

At the same time they also made an ablation with using symmetric augmentations or for not using e.g. solarization:
image

The difference without solarize seems to be only `0.2`. With fewer epochs as we use it might be less. We should probably introduce a BYOLTransform that uses the correct augmentations.

IgorSusmelj:
## Closes #1324 

- BYOL augmentations are now separate
- Weaker color jitter as described in the paper
- Asymmetric transformations between the views
- Additional solarization

I'm not very excited about the design as we use just dicts. We don't get any typing. It's the same setup as the original BYOL code uses. But having to make another class just for the parameters sounds also not ideal.

### DIFF
--- a/lightly/transforms/byol_transform.py
+++ b/lightly/transforms/byol_transform.py
@@ -1,0 +1,192 @@
+from typing import Optional, Tuple, Union
+
+import torchvision.transforms as T
+from PIL.Image import Image
+from torch import Tensor
+
+from lightly.transforms.gaussian_blur import GaussianBlur
+from lightly.transforms.multi_view_transform import MultiViewTransform
+from lightly.transforms.rotation import random_rotation_transform
+from lightly.transforms.solarize import RandomSolarization
+from lightly.transforms.utils import IMAGENET_NORMALIZE
+
+
+class BYOLTransform(MultiViewTransform):
+    """Implements the transformations for BYOL[0].
+
+    Input to this transform:
+        PIL Image or Tensor.
+
+    Output of this transform:
+        List of Tensor of length 2.
+
+    Applies the following augmentations by default:
+        - Random resized crop
+        - Random horizontal flip
+        - Color jitter
+        - Random gray scale
+        - Gaussian blur
+        - Solarization
+        - ImageNet normalization
+
+    Note that SimCLR v1 and v2 use similar augmentations. In detail, BYOL has
+    asymmetric gaussian blur and solarization. Furthermore, BYOL has weaker
+    color jitter compared to SimCLR.
+
+    - [0]:Bootstrap Your Own Latent, 2020, https://arxiv.org/pdf/2006.07733.pdf
+
+    Input to this transform:
+        PIL Image or Tensor.
+
+    Output of this transform:
+        List of [tensor, tensor].
+
+    Attributes:
+        input_size:
+            Size of the input image in pixels.
+        cj_prob:
+            Probability that color jitter is applied.
+        cj_strength:
+            Strength of the color jitter. `cj_bright`, `cj_contrast`, `cj_sat`, and
+            `cj_hue` are multiplied by this value. For datasets with small images,
+            such as CIFAR, it is recommended to set `cj_strenght` to 0.5.
+        cj_bright:
+            How much to jitter brightness.
+        cj_contrast:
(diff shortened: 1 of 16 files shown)


guarin (commented):
I agree that dicts as arguments are not ideal. Did you consider the following setup:

```
class BYOLTransform(MultiViewTransform):
    def __init__(self, view_1: BYOLView1Transform = BYOLView1Transform(), view_2: BYOLView2Transform = BYOLView2Transform()):
        ...

class BYOLView1Transform:
    def __init__(self, ...) # defaults for view 1

class BYOLView2Transform:
    def __init__(self, ...) # defaults for view 2
```

In other parts of the code we have sometimes prefixed arguments like this:
```
global_input_size: int = 224
local_input_size: int = 96
...
```
For example in DINOTransform. But this mostly works if not all arguments can be different. For SMOG we have lists for every argument:
```
input_sizes: [224, 224],
cj_probs: [0.8, 0.8],
...
```
Might also be an option.

IgorSusmelj:
> I agree that dicts as arguments are not ideal. Did you consider the following setup:
> 
> ```
> class BYOLTransform(MultiViewTransform):
>     def __init__(self, view_1: BYOLView1Transform = BYOLView1Transform(), view_2: BYOLView2Transform = BYOLView2Transform()):
>         ...
> 
> class BYOLView1Transform:
>     def __init__(self, ...) # defaults for view 1
> 
> class BYOLView2Transform:
>     def __init__(self, ...) # defaults for view 2
> ```
> 
> In other parts of the code we have sometimes prefixed arguments like this:
> 
> ```
> global_input_size: int = 224
> local_input_size: int = 96
> ...
> ```
> 
> For example in DINOTransform. But this mostly works if not all arguments can be different. For SMOG we have lists for every argument:
> 
> ```
> input_sizes: [224, 224],
> cj_probs: [0.8, 0.8],
> ...
> ```
> 
> Might also be an option.

I did consider it but was not sure as we would have to have two defaults as the Augmentations are asymmetric. But I'll probably go with this then. The other issue I discovered is that with the current dict setup the type hints mess up some of the floats as being integers (e.g. `1.0` turns into `1` and `0.0` turns into `0`)

lightly/transforms/byol_transform.py
```diff
@@ -0,0 +1,172 @@
...
+                The input image to apply the transforms to.
+
+        Returns:
+            The transformed image.
+
+        """
+        return self.transform(image)
+
+
+class BYOLTransform(MultiViewTransform):
```
guarin, line 132: We put the main transform at the top of the file for the other transforms.

lightly/transforms/byol_transform.py
```diff
@@ -0,0 +1,172 @@
...
+        - Random gray scale
+        - Gaussian blur
+        - Solarization
+        - ImageNet normalization
+
+    Note that SimCLR v1 and v2 use similar augmentations. In detail, BYOL has
+    asymmetric gaussian blur and solarization. Furthermore, BYOL has weaker
+    color jitter compared to SimCLR.
+
+    - [0]:Bootstrap Your Own Latent, 2020, https://arxiv.org/pdf/2006.07733.pdf
```
guarin, line 154:
```suggestion
    - [0]: Bootstrap Your Own Latent, 2020, https://arxiv.org/pdf/2006.07733.pdf
```

guarin (approved):
LGTM!

Left some minor comments but feel free to resolve and merge.

lightly/transforms/byol_transform.py
IgorSusmelj: I'm not sure whether I can switch the order as we use the `BYOLView1Transform` in the type hint and it has to be initialized before using it.

IgorSusmelj:
Looks like the corrected augmentations could indeed have a slightly impact on the final model accuracy
Red is the model with SImCLR augmentation. Orange is the one with the correct BYOL augmentations.
image

guarin (approved):
LGTM!

PR #1330, opened 2023-07-18
little fix with deletion of pos_embeddings from ijepa

Natyren:
Delete private functions like _get_2d_sincos_pos_embed because where is no need for them


guarin (approved):
LGTM!

PR #1338, opened 2023-07-21
Guarin lig 3696 figure out add timeout that does not use sigalarm

guarin:
## Changes

* Add version check request timeout
* Run __init__ version check in background
* Remove SIGALRM based timeout

## How was it tested?
* Added unit tests
* Manual tests


lightly/api/_version_checking.py
guarin, line 1: Renamed file from `version_checking.py` to  `_version_checking.py` as it really shouldn't be used from outside the package.

lightly/api/api_workflow_client.py
```diff
@@ -104,10 +104,14 @@ def __init__(
                     )
                 )
         except (
+            # TODO(guarin, 07/23): Why do we have to catch value errors?
             ValueError,
```
guarin, line 108: Do you remember @philippmwirth?

tests/api/test_version_checking.py
```diff
@@ -1,75 +1,87 @@
...
-    get_minimum_compatible_version,
-    is_compatible_version,
-    is_latest_version,
-    pretty_print_latest_version,
-)
-from tests.api_workflow.mocked_api_workflow_client import MockedVersioningApi
-
-
-class TestVersionChecking(unittest.TestCase):
-    def setUp(self) -> None:
```
guarin, line 18: Rewrote the tests as pytest tests

tests/api_workflow/mocked_api_workflow_client.py
```diff
@@ -982,14 +982,6 @@ def update_scheduled_docker_run_state_by_id(
         raise NotImplementedError()
 
 
-class MockedVersioningApi(VersioningApi):
-    def get_latest_pip_version(self, **kwargs):
```
guarin, line 986: No longer needed

tests/conftest.py
```diff
@@ -41,8 +41,8 @@ def pytest_collection_modifyitems(config, items):
             item.add_marker(skip_slow)
 
 
-@pytest.fixture(scope="session", autouse=True)
-def mock_versioning_api():
+@pytest.fixture(scope="module", autouse=True)
```
guarin, line 44: Applies fixture per test module. Cannot use per session as we have to overwrite fixture for the version check tests.

lightly/api/api_workflow_client.py
philippmwirth: `version_compare` raises a `ValueError` if the version is in an unexpected format.

lightly/api/api_workflow_client.py
```diff
@@ -104,10 +104,14 @@ def __init__(
...
+            # Any error by API client if status not in [200, 299].
             ApiException,
-            LightlyAPITimeoutException,
-            AttributeError,
```
philippmwirth, line 110: Do you remember why we caught the `AttributeError`? If not, I'd keep it in here.

lightly/api/api_workflow_client.py
guarin: Aah, thanks!

lightly/api/api_workflow_client.py
guarin: Yes, it was because SIGALRM was not supported on windows.

philippmwirth (approved):

PR #1339, opened 2023-07-21
Enable setting local datasource purpose

philippmwirth:
# Enable setting local datasource purpose

* Expose the datasource purpose for local datasources.
* Remove inaccurate comments about when an output datasource is needed.


guarin (approved):
LGTM!

PR #1346, opened 2023-08-04
Adding py.typed so that type informations from lightly works

furkanbiten:
Hey, 

We have been using lightly for our SSL models. However, it does seem that the type information is not being accessed and read by mypy properly. 

It is because of the missing empty py.typed file. 

In this PR, I have added an empty file `py.typed` and added a keyword argument `package_data` in setup.

More info can be found here: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/


michal-lightly:
Hi @furkanbiten, thanks for your contribution!

Unfortunately, we can't mark the package as typed yet. The type annotations currently serve as a form of documentation. However we don't run a type checker, and there are certainly some types annotated incorrectly. If we mark the package as typed this could lead to problems for users that have lightly integrated and already use a type checker.

However, even though we don't have a timeline yet, typing the package properly is definitely something that we do consider.

furkanbiten:
I see, thanks for the swift response.

furkanbiten:
I will close the PR then.

PR #1351, opened 2023-08-14
Add VICReg benchmark

### LINKED ISSUE #1350: VICReg

philippmwirth:
Part of #1350.

TODOs:
- [X] Add benchmark code.
- [x] Schedule benchmark on cluster.
- [x] Fix problems if there are any.
- [x] Report results


benchmarks/imagenet/resnet50/vicreg.py
```diff
@@ -0,0 +1,114 @@
...
+        return self.backbone(x)
+
+    def training_step(
+        self, batch: Tuple[List[Tensor], Tensor, List[str]], batch_idx: int
+    ) -> Tensor:
+        views, targets = batch[0], batch[1]
+        features = self.forward(torch.cat(views)).flatten(start_dim=1)
+        z = self.projection_head(features)
+        z_a, z_b = z.chunk(len(views))
+        loss = self.criterion(z_a=z_a, z_b=z_b)
```
MalteEbner, line 43: In the simclr.py, it uses instead z0 and z1 instead. Is there a reason for this inconsistency?

```python
views, targets = batch[0], batch[1]
features = self.forward(torch.cat(views)).flatten(start_dim=1)
z = self.projection_head(features)
z0, z1 = z.chunk(len(views))
oss = self.criterion(z0, z1)
```

benchmarks/imagenet/resnet50/vicreg.py
```diff
@@ -0,0 +1,114 @@
...
+        return cls_loss
+
+    def configure_optimizers(self):
+        # Don't use weight decay for batch norm, bias parameters, and classification
+        # head to improve performance.
+        params, params_no_weight_decay = get_weight_decay_parameters(
+            [
+                self.student_backbone,
+                self.student_projection_head,
+                self.student_prediction_head,
```
MalteEbner, line 74: Where are these attributes set? I only found them set for the DINO model.

I would have expected

```python
params, params_no_weight_decay = get_weight_decay_parameters(
  [self.backbone, self.projection_head]
)
```

benchmarks/imagenet/resnet50/vicreg.py
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    "name": "online_classifier",
+                    "params": self.online_classifier.parameters(),
+                    "weight_decay": 0.0,
+                },
+            ],
+            # Linear learning rate scaling with a base learning rate of 0.2.
+            # See https://arxiv.org/pdf/2105.04906.pdf for details.
+            lr=0.2 * self.batch_size_per_device * self.trainer.world_size / 256,
+            momentum=0.9,
```
MalteEbner, line 94: Where does the momentum come from? I did not find it in the paper.

MalteEbner (commented):
Looks good, but I have a few questions.

benchmarks/imagenet/resnet50/vicreg.py
philippmwirth: Because the criterion expects `z_a` and `z_b` so I wanted to be consistent there.

benchmarks/imagenet/resnet50/vicreg.py
philippmwirth: Thanks, great catch! I'll fix it.

benchmarks/imagenet/resnet50/vicreg.py
philippmwirth: It's always `0.9`. Check for example the reference implementation: https://github.com/facebookresearch/vicreg/blob/4e12602fd495af83efd1631fbe82523e6db092e0/main_vicreg.py#L245

lightly/models/modules/heads.py
```diff
@@ -694,6 +694,38 @@ def __init__(
         )
 
 
+
+class VICRegProjectionHead(ProjectionHead):
```
philippmwirth, line 698: VICReg actually uses a two-layer projection head for experiments with 100 epochs of training.

benchmarks/imagenet/resnet50/vicreg.py
```diff
@@ -0,0 +1,128 @@
...
+        }
+        return [optimizer], [scheduler]
+
+
+# VICReg transform
+transform = VICRegTransform()
+
+
+def _get_base_learning_rate(global_batch_size: int) -> float:
+    """Returns the base learning rate for training 100 epochs with a given batch size.
```
philippmwirth, line 116: They did a gridsearch for best base learning rates at 100 epochs of training.

README.md
```diff
@@ -283,6 +283,7 @@ Training a model takes around 4 days for 100 epochs (35 min per epoch), includin
...
 | SimCLR* + DCLW | Res50    |        256 |    100 |        64.5 |          73.2 |     48.5 |      [link](https://tensorboard.dev/experiment/TrALnpwFQ4OkZV3uvaX7wQ/) |       [link](https://lightly-ssl-checkpoints.s3.amazonaws.com/imagenet_resnet50_dclw_2023-07-07_14-57-13/pretrain/version_0/checkpoints/epoch%3D99-step%3D500400.ckpt) |
 | SwAV           | Res50    |        256 |    100 |        67.2 |          75.4 |     49.5 |      [link](https://tensorboard.dev/experiment/Ipx4Oxl5Qkqm5Sl5kWyKKg) |       [link](https://lightly-ssl-checkpoints.s3.amazonaws.com/imagenet_resnet50_swav_2023-05-25_08-29-14/pretrain/version_0/checkpoints/epoch%3D99-step%3D500400.ckpt) |
+| VICReg         | Res50    |        256 |    100 |        63.0 |          73.7 |     46.3 |      [link](https://tensorboard.dev/experiment/qH5uywJbTJSzgCEfxc7yUw) |       [link](https://lightly-ssl-checkpoints.s3.amazonaws.com/imagenet_resnet50_vicreg_2023-09-11_10-53-08/pretrain/version_0/checkpoints/epoch%3D99-step%3D500400.ckpt) |
```
philippmwirth, line 286: The paper reports `67.9` for a batch size of `256` and `100` epochs of training. They use a different linear evaluation protocol but other than that I'm not certain where the difference is coming from.

guarin: The original code also uses a different learning rate of `0.02` (not scaled by batch size) whereas we use `0.1` (scaled by batch size).

lightly/models/modules/heads.py
guarin: Thanks! Made an issue to also change it in the docs #1397

guarin (approved):
This looks great! Let's get it merged.

PR #1353, opened 2023-08-15
Update scheduler.py

anishacharya:
Here is a quick test to verify the **fixed scheduler**:

```
### Test
# Parameters for the scheduler and training
import matplotlib.pyplot as plt
import torch
from lightly.utils.scheduler import CosineWarmupScheduler
warmup_epochs = 10
max_epochs = 1000
start_value = 2.0
end_value = 0.0001

# Create a dummy optimizer (no need to attach it to a model)
dummy_optimizer = torch.optim.SGD([torch.tensor(0.0)], lr=start_value)

# Create the scheduler instance
scheduler = CosineWarmupScheduler(
    optimizer=dummy_optimizer,
    warmup_epochs=warmup_epochs,
    max_epochs=max_epochs,
    start_value=start_value,
    end_value=end_value,
    verbose=False
)

# List to store learning rates
learning_rates = []

# Simulate epoch progression and track learning rates
for epoch in range(max_epochs):
    current_lr = scheduler.get_last_lr()[0]
    learning_rates.append(current_lr)
    scheduler.step()

# Plotting the learning rate schedule
# Find the lowest learning rate and its index
lowest_lr = min(learning_rates)
lowest_lr_index = learning_rates.index(lowest_lr)
highest_lr = max(learning_rates)
highest_lr_index = learning_rates.index(highest_lr)

plt.figure(figsize=(10, 6))
plt.plot(range(max_epochs), learning_rates, linestyle='-', color='b')
plt.scatter(lowest_lr_index, lowest_lr, color='r', label=f'Lowest LR: {lowest_lr:.5f}')
plt.scatter(highest_lr_index, highest_lr, color='r', label=f'Highest LR: {highest_lr:.5f}')
plt.xlabel('Epoch', fontsize=14)
plt.ylabel('Learning Rate', fontsize=14)
plt.title('Learning Rate Schedule', fontsize=16)
plt.xticks(fontsize=12)
plt.yticks(fontsize=12)
plt.grid(True)
plt.legend()
plt.tight_layout()
```
lightly_lrs

To see the old scheduler issue:
Run the same **test with the existing version** of the scheduler. It should produce the following plot:
lightly_lrs_old

Hope this helps.

### DIFF
--- a/lightly/utils/scheduler.py
+++ b/lightly/utils/scheduler.py
@@ -48,7 +48,7 @@ def cosine_schedule(
             * (np.cos(np.pi * step / (max_steps - 1)) + 1)
             / 2
         )
-    return decay / start_value 
+    return decay / start_value
 
 
 class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
@@ -106,7 +106,7 @@ class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
 
         """
         if epoch < self.warmup_epochs:
-            return (epoch + 1) / self.warmup_epochs
+            return self.start_value * (epoch + 1) / self.warmup_epochs
         else:
             return cosine_schedule(
                 step=epoch - self.warmup_epochs,
--- a/lightly/utils/scheduler.py
+++ b/lightly/utils/scheduler.py
@@ -48,7 +48,7 @@ def cosine_schedule(
             * (np.cos(np.pi * step / (max_steps - 1)) + 1)
             / 2
         )
-    return decay
+    return decay / start_value 
 
 
 class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
--- a/lightly/utils/scheduler.py
+++ b/lightly/utils/scheduler.py
@@ -48,7 +48,7 @@ def cosine_schedule(
             * (np.cos(np.pi * step / (max_steps - 1)) + 1)
             / 2
         )
-    return decay / start_value
+    return decay
 
 
 class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
--- a/tests/utils/test_scheduler.py
+++ b/tests/utils/test_scheduler.py
@@ -62,3 +62,22 @@ class TestScheduler(unittest.TestCase):
             RuntimeWarning, msg="Current step number 7 exceeds max_steps 6."
         ):
             scheduler.step()
+
+    def test_CosineWarmupScheduler__warmup(self):
+        model = nn.Linear(10, 1)
+        optimizer = torch.optim.SGD(
+            model.parameters(), lr=1.0, momentum=0.0, weight_decay=0.0
+        )
+        scheduler = CosineWarmupScheduler(
+            optimizer,
+            warmup_epochs=3,
+            max_epochs=6,
+            start_value=2.0,
+            end_value=0.0,
+        )
+        # Linear warmup
+        self.assertAlmostEqual(scheduler.scale_lr(epoch=0), 2.0 * 1 / 3)
+        self.assertAlmostEqual(scheduler.scale_lr(epoch=1), 2.0 * 2 / 3)
+        self.assertAlmostEqual(scheduler.scale_lr(epoch=2), 2.0 * 3 / 3)
+        # Cosine decay
+        self.assertLess(scheduler.scale_lr(epoch=3), 2.0)
--- a/tests/utils/test_scheduler.py
+++ b/tests/utils/test_scheduler.py
@@ -80,4 +80,5 @@ class TestScheduler(unittest.TestCase):
         self.assertAlmostEqual(scheduler.scale_lr(epoch=1), 2.0 * 2 / 3)
         self.assertAlmostEqual(scheduler.scale_lr(epoch=2), 2.0 * 3 / 3)
         # Cosine decay
-        self.assertLess(scheduler.scale_lr(epoch=3), 2.0)
+        self.assertAlmostEqual(scheduler.scale_lr(epoch=3), 2.0 * 3 / 3)
+        self.assertLess(scheduler.scale_lr(epoch=4), 2.0)


philippmwirth:
Hi @anishacharya, I'll reply to your question here. The plot in your comment actually shows that our implementation works as expected. You are using `start_value=2.0` for the scheduler and an initial learning rate of `lr=2.0`. The scheduler decays the multiplicative factor from `2.0` to `0.0001` over `1000` steps so the actual learning rate ranges from `4.0` just after warmup to `0.00020` at the end.

guarin:
Added the change to scale the learning rate by `start_value` during warmup.

lightly/utils/scheduler.py
```diff
@@ -48,7 +48,7 @@ def cosine_schedule(
             * (np.cos(np.pi * step / (max_steps - 1)) + 1)
             / 2
         )
-    return decay
+    return decay / start_value
```
philippmwirth, line 51: I think we need to drop this.

philippmwirth (changes requested):

philippmwirth (approved):

anishacharya:
The issue is not fixed i think with the proposed change --- please see the following test code

import matplotlib.pyplot as plt
import torch
from lightly.utils.scheduler import CosineWarmupScheduler
warmup_epochs = 10
max_epochs = 1000
start_value = 2.0
end_value = 0.0001

# Create a dummy optimizer (no need to attach it to a model)
dummy_optimizer = torch.optim.SGD([torch.tensor(0.0)], lr=start_value)

# Create the scheduler instance
scheduler = CosineWarmupScheduler(
    optimizer=dummy_optimizer,
    warmup_epochs=warmup_epochs,
    max_epochs=max_epochs,
    start_value=start_value,
    end_value=end_value,
    verbose=False
)

# List to store learning rates
learning_rates = []

# Simulate epoch progression and track learning rates
for epoch in range(max_epochs):
    current_lr = scheduler.get_last_lr()[0]
    learning_rates.append(current_lr)
    scheduler.step()

# Plotting the learning rate schedule
# Find the lowest learning rate and its index
lowest_lr = min(learning_rates)
lowest_lr_index = learning_rates.index(lowest_lr)
highest_lr = max(learning_rates)
highest_lr_index = learning_rates.index(highest_lr)

plt.figure(figsize=(10, 6))
plt.plot(range(max_epochs), learning_rates, linestyle='-', color='b')
plt.scatter(lowest_lr_index, lowest_lr, color='r', label=f'Lowest LR: {lowest_lr:.5f}')
plt.scatter(highest_lr_index, highest_lr, color='r', label=f'Highest LR: {highest_lr:.5f}')
plt.xlabel('Epoch', fontsize=14)
plt.ylabel('Learning Rate', fontsize=14)
plt.title('Learning Rate Schedule', fontsize=16)
plt.xticks(fontsize=12)
plt.yticks(fontsize=12)
plt.grid(True)
plt.legend()


plt.tight_layout()

and the output:
download


Do you understand the issue ? 
At the end of warmup -- the learning rate should be 2.0 and not 4.0

anishacharya:
**The issue is in the cosine decay function** as I pointed out before ~~~ 

since you are returning decay to superclass lambdaLR the resulting 
`lr = decay * start_value`

example when  the cosine schedule kicks off i.e. step=0 
```
decay = ( 
end_value - (end_value - start_value)* (np.cos(np.pi * step / (max_steps - 1)) + 1) / 2
) 
= start_value
```

which should be the learning rate not decay

anishacharya:
The issue persists....
@guarin @philippmwirth @IgorSusmelj 
The issue was not in your warmup -- the issue is in your decay.... Please note that you should be returning "scaling factor" as  you are using torch.optim.lr_scheduler.LambdaLR 

so an easy fix is:

```
class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
	"""
    Cosine warmup scheduler for learning rate.
    Args:
        optimizer:
            Optimizer object to schedule the learning rate.
        warmup_epochs:
            Number of warmup epochs or steps.
        max_epochs:
            Total number of training epochs or steps.
        last_epoch:
            The index of last epoch or step. Default: -1
        start_value:
            Starting learning rate scale. Default: 1.0
        end_value:
            Target learning rate scale. Default: 0.001
        verbose:
            If True, prints a message to stdout for each update. Default: False.
    """

	def __init__(
			self,
			optimizer: torch.optim.Optimizer,
			warmup_epochs: int,
			max_epochs: int,
			last_epoch: int = -1,
			start_value: float = 1.0,
			end_value: float = 0.001,
			verbose: bool = False,
	) -> None:
		self.warmup_epochs = warmup_epochs
		self.max_epochs = max_epochs
		self.start_value = start_value
		self.end_value = end_value
		super().__init__(
			optimizer=optimizer,
			lr_lambda=self.scale_lr,
			last_epoch=last_epoch,
			verbose=verbose,
		)

	def cosine_schedule(self, step: int, max_steps: int) -> float:
		"""
		Use cosine decay to gradually modify start_value to reach target end_value during iterations.
		@param step
		@param max_steps
		Returns: Cosine decay value.
		"""
		if step < 0:
			raise ValueError("Current step number can't be negative")
		if max_steps < 1:
			raise ValueError("Total step number must be >= 1")
		if step > max_steps:
			warnings.warn(
				f"Current step number {step} exceeds max_steps {max_steps}.",
				category=RuntimeWarning,
			)
		if max_steps == 1:
			# Avoid division by zero
			decay = self.end_value
		elif step == max_steps:
			# Special case for Pytorch Lightning which updates LR scheduler also for epoch
			# after last training epoch.
			decay = self.end_value
		else:
			decay = (
					self.end_value
					- (self.end_value - self.start_value)
					* (np.cos(np.pi * step / (max_steps - 1)) + 1)
					/ 2
			)
		**return decay / self.start_value**. <----- This line changed 

	def scale_lr(self, epoch: int) -> float:
		"""
        Scale learning rate according to the current epoch number.
        Args:
            epoch:
                Current epoch number.
        Returns:
            Scaled learning rate scaling:
        """
		if epoch < self.warmup_epochs:
			**decay = (epoch + 1) / self.warmup_epochs**. < ---- Revert back to this
		else:
			decay = self.cosine_schedule(
				step=epoch - self.warmup_epochs,
				max_steps=self.max_epochs - self.warmup_epochs
			)
		return decay 
```


Now the test I attached would produce the correct thing ....
download (1)

philippmwirth:
Hi @anishacharya 

Here's an extract from PyTorch's documentation of `LambdaLR`:
> lr_lambda (function or [list](https://docs.python.org/3/library/stdtypes.html#list)) – A function which computes a multiplicative factor given an integer parameter epoch, or a list of such functions, one for each group in optimizer.param_groups.

Note that this is exactly what the current implementation does. In the example you showed above you set `start_value = 2.0` and `start_lr = 2.0` so after warmup you get the scaled learning rate `scaled_lr = start_value * start_lr = 2 * 2 = 4` which is intended becasue the multiplicative factor of `LambdaLR` is greater than one. If you want the learning rate to scale from two to one please set `start_lr = 2.0` and the `start_value` of the scheduler to `1.0`:
```python
# This should produce the desired result
import matplotlib.pyplot as plt
import torch
from lightly.utils.scheduler import CosineWarmupScheduler

warmup_epochs = 10
max_epochs = 1000
start_value = 1.0     # <-- Changed this from 2.0 to 1.0
end_value = 0.0001
start_lr = 2.0        # <-- Added this


dummy_optimizer = torch.optim.SGD([torch.tensor(0.0)], lr=start_lr) # <-- Changed this to start_lr
scheduler = CosineWarmupScheduler(
    optimizer=dummy_optimizer,
    warmup_epochs=warmup_epochs,
    max_epochs=max_epochs,
    start_value=start_value,
    end_value=end_value,
    verbose=False
)



```

PR #1355, opened 2023-08-16
Update OpenAPI client. Configure local datasource.

michal-lightly:
### What has changed and why?

Update OpenAPI client with newest specs.

Allow configuring new fields for local datasource.

How to review: Skip `openapi_generated` folder.

### How is it tested?

Modified the unit test.


lightly/api/api_workflow_datasources.py
```diff
@@ -517,7 +516,8 @@ def set_gcs_config(
 
     def set_local_config(
         self,
-        resource_path: str,
+        relative_path: str,
+        web_server_location: Optional[str] = "http://localhost:3456",
```
japrescott, line 520: 🏆

japrescott (approved):

PR #1357, opened 2023-08-16
Fix barlow twins loss

IgorSusmelj:
## Changes

The existing implementation of the loss function was very slow (training could take 5x longer) because we used slow operations. This change uses a more efficient way of computing the loss.

### DIFF
--- a/lightly/loss/barlow_twins_loss.py
+++ b/lightly/loss/barlow_twins_loss.py
@@ -67,10 +67,9 @@ class BarlowTwinsLoss(torch.nn.Module):
                 c = c / world_size
                 dist.all_reduce(c)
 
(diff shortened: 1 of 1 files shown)


IgorSusmelj:
I closed this PR and added the changes directly to the PR that adds the imagenet benchmark

MalteEbner (commented):
While the code looks correct, I am not too sure about it. Can you either

- write a unittest for it?
- ask GPT4 if the two code samples are the same and post the answer?

Furthermore, I would rename the variables:
- rename `on_diag` to `invariance_loss`
- rename `off_diag` to `redundancy_reduction_loss`

These are the terms in the paper:

image

PR #1362, opened 2023-08-24
Guarin lig 3083 cleanup pip setup.py

### LINKED ISSUE #1215: Cleanup setup.py
## Tasks

- [ ] Move as many options as possible to setup.cfg
- [ ] Update outdated options (for example Python 10 and 11 are missing in classifiers )
- [ ] Remove unused parts

guarin:
Closes #1215 

Contrary to the issue description I didn't move the options to `setup.cfg` but instead moved the single option that was in `setup.cfg` to `setup.py`. This results in less changes and `setup.py` allows a bit more flexibility than `setup.cfg`, especially regarding listing requirements in `requirements.txt` files.

In the future we should directly migrate to `pyproject.toml` once it is the standard. This is not yet possible as `pyproject.toml` only works with recent Python versions.

### Changes
* Remove unused code from `setup.py`
* Cleanup `setup.py`
* Update authors list
* Update pypi classifiers
* Load version in `setup.py` without importing `lightly`
* Automatically find packages


.gitignore
```diff
@@ -5,6 +5,7 @@ lightning_logs/
 **lightning_logs/
 **/__MACOSX
 datasets/
+dist/
```
guarin, line 8: Added this because I built lightly a couple of times and the build outputs are stored in `dist/`. We never want to commit these files to git.

lightly/__init__.py
```diff
@@ -79,40 +79,25 @@
...
-    __LIGHTLY_SETUP__
-except NameError:
-    __LIGHTLY_SETUP__ = False
-
-
-if __LIGHTLY_SETUP__:
-    # setting up lightly
-    msg = f"Partial import of {__name__}=={__version__} during build process."
-    print(msg)
-else:
```
guarin, line 96: This code was necessary to figure out if lightly is imported from `setup.py` or from somewhere else. If it was imported from `setup.py`, then we didn't want to run the version checks because they failed due to missing dependencies. Dependencies were missing because they are not yet available at when `setup.py` runs.

This code has been replaced by the `load_version` function added to `setup.py`. See its docstring for more info.

setup.py
```diff
@@ -1,36 +1,19 @@
...
-import sys
+from pathlib import Path
+from typing import List
 
 import setuptools
 
-try:
-    import builtins
-except ImportError:
-    import __builtin__ as builtins
```
guarin, line 9: No longer needed because `__LIGHTLY_SETUP__` hack was removed.

setup.py
```diff
@@ -1,36 +1,19 @@
...
+_PATH_ROOT = Path(os.path.dirname(__file__))
 
-PATH_ROOT = PATH_ROOT = os.path.dirname(__file__)
-builtins.__LIGHTLY_SETUP__ = True
 
-import lightly
```
guarin, line 14: The `lightly` import was only used to read `lightly.__version__`. This is no longer necessary as we now read the value directly from the `lightly/__init__.py` file without imports. See `load_version` function for more info.

setup.py
```diff
@@ -1,36 +1,19 @@
...
-
-
-def load_description(path_dir=PATH_ROOT, filename="DOCS.md"):
-    """Load long description from readme in the path_dir/ directory"""
-    with open(os.path.join(path_dir, filename)) as f:
-        long_description = f.read()
-    return long_description
```
guarin, line 21: Inlined this code further down.

setup.py
```diff
@@ -41,29 +24,43 @@ def load_requirements(path_dir=PATH_ROOT, filename="base.txt", comment_char="#")
...
+            if line.startswith("__version__"):
+                version = line.split("=")[-1].strip().strip('"')
+                return version
+    raise RuntimeError("Unable to find version string in '{version_filepath}'.")
+
+
 if __name__ == "__main__":
     name = "lightly"
-    version = lightly.__version__
-    description = lightly.__doc__
```
guarin, line 47: This variable was overwritten a couple lines further down and is therefore not used.

setup.py
```diff
@@ -41,29 +24,43 @@ def load_requirements(path_dir=PATH_ROOT, filename="base.txt", comment_char="#")
...
             "lightly-crop = lightly.cli.crop_cli:entry",
-            "lightly-train = lightly.cli.train_cli:entry",
+            "lightly-download = lightly.cli.download_cli:entry",
             "lightly-embed = lightly.cli.embed_cli:entry",
             "lightly-magic = lightly.cli.lightly_cli:entry",
-            "lightly-download = lightly.cli.download_cli:entry",
-            "lightly-version = lightly.cli.version_cli:entry",
             "lightly-serve = lightly.cli.serve_cli:entry",
+            "lightly-train = lightly.cli.train_cli:entry",
+            "lightly-version = lightly.cli.version_cli:entry",
```
guarin, line 60: Just sorted the entries alphabetically.

setup.py
```diff
@@ -41,29 +24,43 @@ def load_requirements(path_dir=PATH_ROOT, filename="base.txt", comment_char="#")
...
         ]
     }
 
-    long_description = load_description()
-
```
philippmwirth, line 66: We should list the packages below automatically (line 82, I can't comment there)

philippmwirth (commented):
Looks much better. I have one more request.

philippmwirth (approved):

PR #1382, opened 2023-09-07
Add mypy and partial typing

philippmwirth:
# Add mypy and partial typing

Changes:
* Add mypy to dev dependencies.
* Add CI to type check on push or pull-request.
* Type check all of `lightly.transforms` and tests (and declare with `py.typed`).
* Type check parts of `lightly.utils` and tests.
* Change all `__init__.py` files to only import submodules at runtime.
* Spent some time trying to figure out how to type check `ApiWorkflowClient` (hint: Use Protocols).

Fun facts:
* Only `160` files left to type check.
* `58` files type checked in this pull-request.
* `99` tears shed fighting hacks and legacy code.
* `15` instances of `Any`.
* Openapi generated code is not typed. In fact, it throws `3434` errors in `251` files.


.github/workflows/test_code_format.yml
```diff
@@ -21,7 +20,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: "3.10"
+        python-version: "3.7"
```
philippmwirth, line 23: We want to make sure types are correct in Python 3.7.

lightly/transforms/random_crop_and_flip_with_grid.py
```diff
@@ -28,7 +28,7 @@ class Location:
     vertical_flip: bool = False
 
 
-class RandomResizedCropWithLocation(T.RandomResizedCrop):
+class RandomResizedCropWithLocation(T.RandomResizedCrop):  # type: ignore
```
michal-lightly, line 31: Can you add a comment why we need the ignore?

lightly/transforms/random_crop_and_flip_with_grid.py
```diff
@@ -128,7 +128,7 @@ class RandomResizedCropAndFlip(nn.Module):
         hf_prob:
             The probability of applying horizontal flipping to the image.
         normalize:
-            A dictionary containing the mean and std values for normalizing the image.
+            A Dictionary containing the mean and std values for normalizing the image.
```
michal-lightly, line 131: Minor: This seems like and artefact of replacing dict with Dict globally :)

mypy.ini
```diff
@@ -0,0 +1,215 @@
...
+
+[mypy]
+ignore_missing_imports = True
+python_version = 3.10
+warn_unused_configs = True
+strict_equality = True
+
+# Disallow dynamic typing
+disallow_any_decorated = True
+# TODO(Philipp, 09/23): Remove me!
```
michal-lightly, line 11: Minor: Instead of `disallow_any_explicit = False` we can have a commented out version:
```
# disallow_any_explicit = True
```

guarin (commented):
This looks great! Left a bunch of comments but nothing major.

I also took the liberty to replace the `TYPE_CHECKING` checks in the `__init__` files by `follow_imports = skip` settings in `mypy.ini`. I believe this is a bit a safer approach.

mypy.ini
```diff
@@ -0,0 +1,215 @@
+# Global options:
+
+[mypy]
+ignore_missing_imports = True
+python_version = 3.10
```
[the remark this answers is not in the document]
philippmwirth: No, it needs to be Python 3.10 because if it's 3.7 and we test on 3.10, type checking fails somewhere in numpy probably due to some logic they have implemented based on the Python version.

guarin (approved):
LGTM!

PR #1384, opened 2023-09-11
Optional relative path for set local config

michal-lightly:
Make relative path optional for `set_local_config`.


guarin (approved):
LGTM!

PR #1403, opened 2023-09-20
Remove platform tutorials

guarin:
## Changes

* List the "Python Package" directly as "Tutorials" (no longer a subpage)
* Remove platform tutorials
* Old links to platform tutorials are forwarded to new docs at https://docs.lightly.ai/

I removed all tutorial files/code that are not referenced from somewhere else in the docs. Some tutorials have sphinx references to them from other outdated parts of the docs. For example the labelstudio export tutorial has references to it from the docker archive. I left these tutorials to avoid breaking the links (they are now forwarded to the new docs). We can remove the tutorials once we remove the docker archive.

## How was it tested?

* Built docs and checked that links work
* Verified that removed tutorials are not linked from anywhere else in the docs

[Documentation — lightly 1.4.19 documentation.pdf](https://github.com/lightly-ai/lightly/files/12671554/Documentation.lightly.1.4.19.documentation.pdf)


docs/source/tutorials/package.rst
```diff
@@ -1,22 +0,0 @@
-.. _lightly-tutorials: 
-
-Python Package
-===================================
-
-With the Lightly SSL framework you can use the power of self-supervised learning
```
japrescott, line 6: Q: would it be possible to keep this file (and info) but have the tutorials unfolded in the navigation? Unless you find the info about pytorch and lighting irrelevant, its good for me

guarin: I actually tried this and the only way I could get it working was by adding it as a "Tutorial" page. But I found this a bit confusing. I would also consider the information on this page not crucial, from the package and the tutorials it is pretty clear that lightly is built on pytorch (lightning), and the `toctree` is now part of the main tree so this information is not lost.

japrescott (approved):

PR #1412, opened 2023-09-29
Jeremy lig 4022 add reportv2json artifact to api webapp pip package

japrescott:
closes lig-4022
- update schema for reportv2


MalteEbner (approved):
Approved

Issue #1319, opened 2023-07-14
Add SimCLR with temperature schedule


guarin:
https://arxiv.org/abs/2303.13664 shows that varying the temperature in the contrastive loss during pretraining can help models learn better representations for datasets with long-tail distributions and doesn't hurt performance for datasets with uniform distributions. This could be an interesting method to add to the package.

I would start with running a benchmark on a normal dataset (ImageNet) as we don't have standard long-tail datasets in our benchmarks. It would also be interesting to see if it also gives good performance on full ImageNet as the paper only reports results for ImageNet100-LT. If it also works on default ImageNet it could be a good default method to add to most contrastive models.

TODO:
- [ ] Add optional period to [`cosine_schedule`](https://github.com/lightly-ai/lightly/blob/df004d288fa07816e55975fa0af92868f9867436/lightly/utils/scheduler.py#L7)
- [ ] Add benchmark with SimCLR and temperature schedule
- [ ] Run benchmark and report results

guarin:
Added it to #1172

PR #1413, opened 2023-10-01
Add simCLR with temperature schedule

JannikWirtz:
Hi, 

this is my first PR to an open source project so I am starting it out as a draft.

I am looking to contribute to #1319. This PR should cover the first task. I will probably be able to do most of the tasks except for actually running the benchmark, since I don't think I have the neccessary hardware.

Thanks for reviewing :)


lightly/utils/scheduler.py
```diff
@@ -19,6 +24,9 @@ def cosine_schedule(
             Starting value.
         end_value:
             Target value.
+        period (optional):
+            The number of steps over which the cosine function completes its half-cycle.
```
guarin, line 28: Are you sure this is only for a half cycle? Shouldn't `np.cos(2 * pi * step / period)` give a full cycle once `step == period`

lightly/utils/scheduler.py
```diff
@@ -28,7 +36,7 @@ def cosine_schedule(
         raise ValueError("Current step number can't be negative")
     if max_steps < 1:
         raise ValueError("Total step number must be >= 1")
-    if step > max_steps:
+    if not period and step > max_steps:
```
guarin, line 39:
```suggestion
    if period is None and step > max_steps:
```

We usually check directly for `None` to avoid issues when the variable is `0` or another value that returns `False` when converted to boolean.

guarin, line 39: Could you add a check below that raises a `ValueError` if `period is not None and period <= 0`. We should catch this to avoid division by zero errors and unexpected behavior.

lightly/utils/scheduler.py
```diff
@@ -37,6 +45,11 @@ def cosine_schedule(
     if max_steps == 1:
         # Avoid division by zero
         decay = end_value
+    elif period:  # "cycle" based on period, if provided
```
guarin, line 48:
```suggestion
    if period is not None:  # "cycle" based on period, if provided
```

You could even move this above the `if max_steps == 1:` check

guarin (changes requested):
Thanks a lot of the PR!

Could you add a `test_cosine_schedule__period` method to the tests in `tests/utils/test_scheduler.py` to check that the new functionality works as expected?

lightly/utils/scheduler.py
JannikWirtz: Sorry, yes you are absolutely correct. It is a full cycle as in the paper. Will Fix.

JannikWirtz:
Thanks a lot for the feedback! I will take a look at the benchmark next.

lightly/utils/scheduler.py
```diff
@@ -48,7 +63,7 @@ def cosine_schedule(
             * (np.cos(np.pi * step / (max_steps - 1)) + 1)
             / 2
         )
-    return decay
+    return float(decay)
```
guarin, line 66: Are you sure the conversion to float is required?

guarin (commented):
Thanks a lot for the update! This looks pretty much good to merge. Left one small question.

lightly/utils/scheduler.py
JannikWirtz: I had gotten a warning from isort when running the make checks, regarding a potential return path which is not of type float. Honestly I have no idea how there could be one, but I added this type to fix it for now.

guarin: Ah I see, this is an issue from mypy because `decay` is defined in multiple places using `if/else`. I added a line to define it before the `if/else` which solves the issue.

guarin (approved):

petroslk:
Hello there! I stumbled upon this and was wondering, has there been any docs added to document the use of the temperature scheduler with NTXentLoss for simCLR? 
Thanks a lot for your work!

Issue #105, opened 2021-01-21
Add: Histogram equalization augmentation


philippmwirth:
As implemented in [PIL.ImageOps](https://pillow.readthedocs.io/en/3.0.x/reference/ImageOps.html#PIL.ImageOps.equalize). May be used to replace ImageNet normalization when working on X-ray images. This normalization increases contrast in the image.

Example:
- [COVID-19 DETERIORATION PREDICTION VIA SELF-SUPERVISED REPRESENTATION LEARNING  AND MULTI-IMAGE PREDICTION](https://arxiv.org/pdf/2101.04909.pdf)

guarin:
I'll close this as the feature has been added to torchvision: https://pytorch.org/vision/main/generated/torchvision.transforms.RandomEqualize.html#randomequalize

PR #1415, opened 2023-10-04
Add histogram equalization

JannikWirtz:
Hi, this should take care of issue #105.

Note: This implementation assumes that all images are of the same size as the mask, in case one is provided.

### DIFF
--- a/lightly/transforms/equalize.py
+++ b/lightly/transforms/equalize.py
@@ -1,0 +1,36 @@
+# Copyright (c) 2021. Lightly AG and its affiliates.
+# All Rights Reserved
+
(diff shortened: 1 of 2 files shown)


guarin:
Thanks for the PR! I just checked torchvision transforms and it seems that they have in the meantime added histogram normalization: https://pytorch.org/vision/main/generated/torchvision.transforms.RandomEqualize.html#randomequalize It also uses `ImageOps.equalize`: https://pytorch.org/vision/main/_modules/torchvision/transforms/functional.html#equalize

I believe in this case it is best if we let users use the torchvision transform and not provide our own version.

PR #1426, opened 2023-11-13
new specs with correct v3

japrescott:
regenerate the specs with correct v3 config


MalteEbner (changes requested):
Please remove all `__init__.py-e` files

MalteEbner (approved):
Generation looks good.

tests/api_workflow/test_api_workflow_compute_worker.py
guarin, line 1: I think there is another test that has to be updated around line 218 (`test_selection_config`). It should also use the new `V3` classes. When merging this PR we should not have to import any of the old `SelectionCon`

tests/api_workflow/test_api_workflow_compute_worker.py
```diff
@@ -43,6 +43,18 @@
     SelectionStrategyType,
     TagData,
 )
+from lightly.openapi_generated.swagger_client.models.selection_config_v3 import (
+    SelectionConfigV3,
+)
```
guarin, line 48: Sorry, I forgot to mention this, but we usually import the swagger generated classes directly from `swagger_client.models` like this:
```
from lightly.openapi_generated.swagger_client.models import (
    SelectionConfigV3,
    SelectionConfigV3Entry,
    ...
)

It makes the code a bit easier to read, the drawback is that the VSCode auto-import function does not use this format.

**Review by guarin — CHANGES_REQUESTED** [review]
Nice! I think one more tests needs to be changed but otherwise this looks good to me :)

File: tests/api_workflow/test_api_workflow_compute_worker.py [hunk 2]
```diff
@@ -101,16 +101,16 @@ def test_create_compute_worker_config__selection_config_is_class(self) -> None:
...
-                        input=SelectionConfigEntryInput(
+                    SelectionConfigV3Entry(
+                        input=SelectionConfigV3EntryInput(
                             type=SelectionInputType.EMBEDDINGS,
                             dataset_id=utils.generate_id(),
                             tag_name="some-tag-name",
                         ),
-                        strategy=SelectionConfigEntryStrategy(
+                        strategy=SelectionConfigV3EntryStrategy(
                             type=SelectionStrategyType.SIMILARITY,
```
guarin, line 114: Could you update `SelectionStrategyType` also to `SelectionStrategyTypeV3`?

guarin (approved):
One last small thing otherwise everything LGTM! I already approved so you can merge whenever you are ready :)

PR #1427, opened 2023-11-15
Add discord release notification

guarin:
### Changes

* Automatically send message to discord for new releases

The message will look like this and be sent to the general channel:
Screen Shot 2023-11-15 at 10 12 23

I tried to keep the action simple and did not add any fancy checks. In some weird scenarios we might get additional or no messages:
* If a release is first released as a pre-release and only later changed to a latest release no notification will be sent
* If a release is released but not marked as the latest release a notification will still be sent

Note that no message will be sent for draft releases, edits, deleted releases, etc.

### How was it tested?

Tested it in a test repo and test discord channel.

### DIFF
--- a/.github/workflows/discord_release_notification.yml
+++ b/.github/workflows/discord_release_notification.yml
@@ -1,0 +1,37 @@
+name: Discord Release Notification
+
+on:
+  release:
+    types: [published]
+
+jobs:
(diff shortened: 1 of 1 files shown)


philippmwirth (approved):
Nice 👍

PR #1429, opened 2023-11-19
Typecheck part of utils is

IgorSusmelj:
## Changes

- Adds typing and enables type check for some files in utils

### DIFF
--- a/lightly/utils/benchmarking/linear_classifier.py
+++ b/lightly/utils/benchmarking/linear_classifier.py
@@ -1,15 +1,9 @@
-from typing import Dict, List, Tuple, Union
+from typing import Any, Dict, List, Tuple, Union
 
 from pytorch_lightning import LightningModule
 from torch import Tensor
 from torch.nn import CrossEntropyLoss, Linear, Module
 from torch.optim import SGD, Optimizer
-
-# Try to import the new LRScheduler from torch 2.0.
-try:
-    from torch.optim.lr_scheduler import LRScheduler
-except ImportError:
-    from torch.optim.lr_scheduler import _LRScheduler as LRScheduler
 
 from lightly.models.utils import activate_requires_grad, deactivate_requires_grad
 from lightly.utils.benchmarking.topk import mean_topk_accuracy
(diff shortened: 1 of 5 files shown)


lightly/utils/benchmarking/linear_classifier.py
```diff
@@ -94,17 +94,20 @@ def __init__(
 
     def forward(self, images: Tensor) -> Tensor:
         features = self.model.forward(images).flatten(start_dim=1)
-        return self.classification_head(features)
+        output: Tensor = self.classification_head(features)
+        return output
 
-    def shared_step(self, batch, batch_idx) -> Tuple[Tensor, Dict[int, Tensor]]:
+    def shared_step(
+        self, batch: Tuple[Tensor, Tensor], batch_idx: int
```
guarin, line 101: This should be `Tuple[Tensor, ...]`

lightly/utils/benchmarking/metric_callback.py
```diff
@@ -58,8 +60,4 @@ def _append_metrics(
         self, metrics_dict: Dict[str, List[float]], trainer: Trainer
     ) -> None:
         for name, value in trainer.callback_metrics.items():
-            # Only store scalar values.
-            if isinstance(value, float) or (
-                isinstance(value, Tensor) and value.numel() == 1
-            ):
-                metrics_dict.setdefault(name, []).append(float(value))
+            metrics_dict.setdefault(name, []).append(float(value))
```
guarin, line 63: I am not 100% sure if we can make this change because this code is called from code that is not yet type-checked and I am not sure if all metrics are really floats. There might also be tensors in there with more than 1 element which cannot be converted to floats. That's why we did the `value.numel() == 1` check.

lightly/utils/embeddings_2d.py
```diff
@@ -53,13 +59,22 @@ def transform(self, X: np.ndarray):
...
+        Raises:
+            ValueError: If PCA was not fitted before.
         """
+        if self.mean is None or self.w is None:
+            raise ValueError("PCA not fitted yet. Call fit() before transform().")
         X = X.astype(np.float32)
         X = X - self.mean + self.eps
-        return X.dot(self.w)[:, : self.n_components]
+        transformed = X.dot(self.w)[:, : self.n_components]
+        return np.asarray(transformed, dtype=np.float32)
```
guarin, line 70: Is this needed because numpy could widen it to `np.float64`? Or why can we not do `transformed: NDArray[np.float32] = X.dot(...)`

guarin (commented):
Nice! Looking forward to this!

Left two questions :)

guarin (approved):
LGTM!

PR #1430, opened 2023-11-20
Typecheck part of models

IgorSusmelj:
## Changes

- add types and enabled type checking for part of models

### DIFF
--- a/lightly/loss/memory_bank.py
+++ b/lightly/loss/memory_bank.py
@@ -4,6 +4,7 @@
 # All Rights Reserved
 
 import functools
+from typing import Optional, Tuple, Union
 
 import torch
 
@@ -55,7 +56,7 @@ class MemoryBankModule(torch.nn.Module):
         )
 
     @torch.no_grad()
-    def _init_memory_bank(self, dim: int):
+    def _init_memory_bank(self, dim: int) -> None:
         """Initialize the memory bank if it's empty
 
         Args:
@@ -67,12 +68,12 @@ class MemoryBankModule(torch.nn.Module):
         # we could use register buffers like in the moco repo
         # https://github.com/facebookresearch/moco but we don't
         # want to pollute our checkpoints
-        self.bank = torch.randn(dim, self.size).type_as(self.bank)
+        self.bank = torch.randn(dim, self.size).type_as(self.bank) # type: ignore
         self.bank = torch.nn.functional.normalize(self.bank, dim=0)
-        self.bank_ptr = torch.zeros(1).type_as(self.bank_ptr)
+        self.bank_ptr = torch.zeros(1).type_as(self.bank_ptr) # type: ignore
 
     @torch.no_grad()
-    def _dequeue_and_enqueue(self, batch: torch.Tensor):
+    def _dequeue_and_enqueue(self, batch: torch.Tensor) -> None:
         """Dequeue the oldest batch and add the latest one
 
         Args:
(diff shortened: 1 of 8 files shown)


lightly/models/batchnorm.py
```diff
@@ -63,10 +65,11 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
                 self.eps,
             ).view(N, C, H, W)
         else:
+            assert self.running_mean is not None and self.running_var is not None
```
IgorSusmelj, line 68: I would avoid having assert statements in time critical places like forward passes. That's why I added the ignore in the first place.

lightly/loss/memory_bank.py
```diff
@@ -68,9 +68,9 @@ def _init_memory_bank(self, dim: int) -> None:
         # we could use register buffers like in the moco repo
         # https://github.com/facebookresearch/moco but we don't
         # want to pollute our checkpoints
-        self.bank = torch.randn(dim, self.size).type_as(self.bank) # type: ignore
-        self.bank = torch.nn.functional.normalize(self.bank, dim=0)
-        self.bank_ptr = torch.zeros(1).type_as(self.bank_ptr) # type: ignore
+        bank: Tensor = torch.randn(dim, self.size).type_as(self.bank)
```
IgorSusmelj, line 71: Let's be consistent everywhere? I'd use `Tensor` or `torch.Tensor` everywhere.

lightly/models/modules/nn_memory_bank.py
```diff
@@ -38,9 +39,16 @@ class NNMemoryBankModule(MemoryBankModule):
...
     def __init__(self, size: int = 2**16):
+        if size <= 0:
+            raise ValueError(f"Memory bank size must be positive, got {size}.")
         super(NNMemoryBankModule, self).__init__(size)
 
-    def forward(self, output: torch.Tensor, update: bool = False):
+    def forward(
+        self,
+        output: torch.Tensor,
+        labels: Optional[torch.Tensor] = None,
```
guarin, line 49: Why was the labels argument added?

lightly/models/batchnorm.py
```diff
@@ -21,15 +25,18 @@ class SplitBatchNorm(nn.BatchNorm2d):
 
     """
 
-    def __init__(self, num_features, num_splits, **kw):
+    running_mean: torch.Tensor
+    running_var: torch.Tensor
```
guarin, line 29: I think this should be moved to the `__init__` method as those are not class variables.

guarin (approved):
Left one comment, otherwise LGTM! I'll let you resolve and merge.

One minor thing you could do is replace all `torch.Tensor` types with `Tensor` only. Then it would be more consistent with the rest of the codebase.

lightly/models/modules/nn_memory_bank.py
philippmwirth: I have no idea. Will remove it again.

philippmwirth: It's because of mismatching signatures. I added a TODO instead.

lightly/models/batchnorm.py
philippmwirth: Unfortunately it doesn't work if I move it to `__init__`. I have reverted to using type ignores for now.

guarin: This is weird, I tried it and it worked for me.

PR #1441, opened 2023-12-04
Fix MoCov2 Transform Parameters

guarin:
## Changes

* Increase min_scale from 0.08 to 0.2
* Decrease color jitter strengths

Modifies the default parameters for the MoCov2 transform to match the parameters from the original implementation: https://github.com/facebookresearch/moco/blob/5a429c00bb6d4efdf511bf31b6f01e064bf929ab/main_moco.py#L328-L337


IgorSusmelj (approved):
LGTM

PR #1442, opened 2023-12-06
Allow empty file upload

guarin:
## Changes

* Allow empty files to be uploaded with the ApiWorkflowClient

Requests expects the file in `requests.put(data=file)` to not be empty. To upload an empty file we have to set `data=None`. As the file is an `IOBase` object we cannot infer the size from it. We therefore have to rely on the calling code to set the file to `None` if it is empty.

## How was it tested?

Tested it manually. Writing a test for it is difficult as it would require mocking something deep in the requests library.


michal-lightly (approved):
LGTM!

PR #1450, opened 2023-12-13
Fix VICRegLoss (remove extra de-mean)

### LINKED ISSUE #1445: VICReg Loss De-Means Twice?
https://github.com/lightly-ai/lightly/blob/66ad1b40ebf3b53512703c774988211ce283211f/lightly/loss/vicreg_loss.py#L128-L129

I think the VICReg loss removes the mean, then calls `.var()` which also de-means (see: https://pytorch.org/docs/stable/generated/torch.var.html). 

If I understand correctly, that seems unnecessary?

philippmwirth:
# Fix VICRegLoss (remove extra de-mean)

Closes #1445.
* Remove extra de-mean.
* The loss doesn't change according to our regression test.

### DIFF
--- a/lightly/loss/vicreg_loss.py
+++ b/lightly/loss/vicreg_loss.py
@@ -125,7 +125,6 @@ def variance_loss(x: Tensor, eps: float = 0.0001) -> Tensor:
         eps:
             Epsilon for numerical stability.
     """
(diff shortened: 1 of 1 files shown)


guarin (approved):
LGTM

Issue #1425, opened 2023-11-12
Clarification: Why does KNN eval normalize the representations?


RylanSchaeffer:
Why does KNN eval normalize the representations?

I'm specifically referring to [here](https://github.com/lightly-ai/lightly/blob/master/lightly/utils/benchmarking/knn_classifier.py#L103) and [here](https://github.com/lightly-ai/lightly/blob/master/lightly/utils/benchmarking/knn_classifier.py#L113).

Is this commonly done in the literature? It seems odd to normalize the representations before applying KNN.

guarin:
Our implementation follows [InstDisc](https://arxiv.org/pdf/1805.01978v1.pdf) and [MoCo](https://github.com/facebookresearch/moco) which both apply normalization. For MoCo please see the colab notebook: https://colab.research.google.com/github/facebookresearch/moco/blob/colab-notebook/colab/moco_cifar10_demo.ipynb

crazyboy9103:
When euclidean distance is used for the distance metric, it is in fact necessary to normalize the representations before applying knn. Take a look at: https://stats.stackexchange.com/questions/287425/why-do-you-need-to-scale-data-in-knn. However, for cosine similarity, it is not. But lightly's implementation of [knn_predict](https://github.com/lightly-ai/lightly/blob/master/lightly/utils/benchmarking/knn.py#L8) looks like it expects normalized representations, so i think their implementation is correct.

RylanSchaeffer:
What that post/link does is different from what the code does, if I
understand correctly. The code normalizes to the hypersphere. The post does
not.

RylanSchaeffer:
To clarify, I believe the post standardizes each dimension independently,
whereas the code performs L2 normalization. If I'm correctly understanding,
these are not the same.

crazyboy9103:
You're correct, the post illustrates the importance of scaling when using euclidean distance as the similarity metric. For continuous values like representations, it is common to use cosine similarity, which includes L2 normalization (onto a unit hypersphere) to compare a pair of representations.

guarin:
[...]

PR #1457, opened 2023-12-15
Make KNN feature normalization optional

guarin:
## Changes

* Make KNN feature normalization optional
* Always reset KNN features at beginning of training epoch

## How was it tested?
* Added unit tests

Closes: #1425

### DIFF
--- a/lightly/utils/benchmarking/knn_classifier.py
+++ b/lightly/utils/benchmarking/knn_classifier.py
@@ -151,6 +151,12 @@ class KNNClassifier(LightningModule):
         # Set model to eval mode to disable norm layer updates.
         self.model.eval()
 
+        # Reset features and targets.
+        self._train_features = []
+        self._train_targets = []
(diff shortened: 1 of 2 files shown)


lightly/utils/benchmarking/knn_classifier.py
```diff
@@ -145,6 +153,12 @@ def on_train_epoch_start(self) -> None:
         # Set model to eval mode to disable norm layer updates.
         self.model.eval()
 
+        # Reset features and targets.
+        self._train_features = []
+        self._train_targets = []
+        self._train_features_tensor = None
+        self._train_targets_tensor = None
```
guarin, line 160: This is not strictly necessary because one should not "train" the KNNClassifier for more than one epoch. But I still added it to avoid any issues if someone actually uses it over multiple epochs.

philippmwirth (approved):

Issue #1343, opened 2023-08-02
Request for New Method - EMP-SSL


surajpaib:
Hi Team,

I recently came across a paper introducing a novel method of self-supervised learning known as EMP-SSL (Link: [EMP-SSL Paper](https://arxiv.org/abs/2304.03977)). The authors have provided their implementation on Github [(EMP-SSL Repository)](https://github.com/tsb0601/EMP-SSL).

TLDR of the method: The key towards efficient self-supervised learning is to increase the number of crops from each image instance.

Key contributions: 
- EMP-SSL can reduce the training epochs by two orders of magnitude.
- The method seems to transfer well to out-of-domain tasks.
- Due to the reduction in training epochs, the overall training time is considerably decreased. This, however, may come at the cost of increased memory usage.

Considering the promising results this method has shown, I think it could be a valuable addition to the collection of self-supervised learning methods so that it is accessible to the larger community.

philippmwirth:
Hi @surajpaib thanks for bringing this up! I'll add it to the tracker #1172.

johnsutor:
I can implement this.

philippmwirth:
@johnsutor let me know if you need any assistance 👍

deepsthewarrior:
> I can implement this.

Are you already working on this?

johnsutor:
@deepsthewarrior yep!

PR #1467, opened 2024-01-04
Add EMP-SSL Loss

johnsutor:
This commit includes the code implementing the EMP-SSL Loss from [this repository](https://github.com/tsb0601/EMP-SSL/tree/main) and [this paper](https://arxiv.org/abs/2304.03977), and proposed in #1343 . The transform from the paper is not implemented, as it is simply a multi-crop transform based on the VICReg Transform (and a high-level implementation is given in the example code within the loss doc string)


guarin:
Thanks for the PR! Will try to take a look ASAP.

lightly/loss/emp_ssl_loss.py
```diff
@@ -0,0 +1,111 @@
...
+        >>> # feed through encoder head
+        >>> z = torch.cat([model(s) for s in samples])
+        >>>
+        >>> # calculate loss
+        >>> loss = loss_fn(z)
+    """
+
+    def __init__(
+        self,
+        eps: float = 0.2,
```
guarin, line 81: Let's call this `tcr_eps` to make it clear that it is a parameter for the `tcr_loss`.

lightly/loss/emp_ssl_loss.py
```diff
@@ -0,0 +1,111 @@
...
+        num_patches: int = 100,
+        tcr_coef: float = 1.0,
+        patch_coef: float = 200.0,
```
guarin, line 84: Let's call this `inv_coef` to make it clear that is a parameter for the `invariance_loss`.

lightly/loss/emp_ssl_loss.py
```diff
@@ -0,0 +1,111 @@
...
+        >>>
+        >>> # calculate loss
+        >>> loss = loss_fn(z)
+    """
+
+    def __init__(
+        self,
+        eps: float = 0.2,
+        num_patches: int = 100,
+        tcr_coef: float = 1.0,
```
guarin, line 83: I think we can drop both of these parameters. `num_patches` can be dropped because we can infer the number of patches based on the tensor passed to the loss function.

`tcr_coef` can be dropped because the number of coefficients of the total loss should be 1 - number of loss terms. The total loss `loss = patch_coef * patch_loss + tcr_coef * tcr_loss` could also be written as `loss = patch_coef / tcr_coef * patch_loss + tcr_loss`, thus "capturing" the value of the `tcr_coef` in the coefficient for the patch loss. Note that the paper uses only a single coefficient as well (lambda for the patch loss):
Screen Shot 2024-01-05 at 12 15 42

lightly/loss/emp_ssl_loss.py
```diff
@@ -0,0 +1,111 @@
...
+        self.patch_coef = patch_coef
+
+    def forward(self, z: Tensor) -> Tensor:
+        z_list = z.chunk(self.num_patches, dim=0)
+        contractive_loss = self.similarity_loss(z_list)
+        tcr_loss = torch.tensor(0.0, device=z.device, requires_grad=True)
+        for i in range(self.num_patches):
+            tcr_loss = tcr_loss + self.tcr(z_list[i])
+        tcr_loss = tcr_loss / self.num_patches
+        return Tensor(self.patch_coef * contractive_loss + self.tcr_coef * tcr_loss)
```
guarin, line 111: If there are many views, we usually pass a list of views to the loss (see [DINOLoss](https://github.com/lightly-ai/lightly/blob/534cdf42a0feebe10a917bbdaacc6a533c1104fa/lightly/loss/dino_loss.py#L80-L85) or [VICRegLLoss](https://github.com/lightly-ai/lightly/blob/534cdf42a0feebe10a917bbdaacc6a533c1104fa/lightly/loss/vicregl_loss.py#L98-L104)). I suggest to change the function definition to `def forward(self, z_views: Sequence[Tensor]) -> Tensor`.

And if we go for the approach where the sub-losses are defined as functions the forward method can be rewritten as:
```
    def forward(self, z_views: Sequence[Tensor]) -> Tensor:
        # z has shape (num_views, batch_size, dim)
        z = torch.stack(z_views)
        return tcr_loss(z, eps=self.tcr_eps) + self.inv_coef * invariance_loss(z)
```

guarin (changes requested):
Thanks a lot for the PR! I left some comments on how we could structure the code a bit to be more composable. Let me know what you think!

lightly/loss/emp_ssl_loss.py
```diff
@@ -0,0 +1,111 @@
...
+        z_stacked = torch.stack(z_list, dim=0)
+        z_avg = z_stacked.mean(dim=0)
+
+        z_sim = torch.tensor(0.0, device=z_stacked.device, requires_grad=True)
+        for i in range(num_patch):
+            z_sim = z_sim + F.cosine_similarity(z_stacked[i], z_avg, dim=1).mean()
+
+        z_sim = z_sim / num_patch
+
+        return -z_sim
```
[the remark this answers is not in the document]
johnsutor: That seems fine by me, for simplicity, and is consistent with other loss functions like WMSE.

lightly/loss/emp_ssl_loss.py
```diff
@@ -0,0 +1,111 @@
...
+class TotalCodingRate(nn.Module):
+    def __init__(self, eps: float = 0.2) -> None:
+        super(TotalCodingRate, self).__init__()
+        self.eps = eps
+
+    def compute_discrimn_loss(self, W: Tensor) -> Tensor:
+        """Discriminative Loss."""
+        d, b = W.shape
+        I = torch.eye(d, device=W.device)
+        scalar = d / (b * self.eps)
```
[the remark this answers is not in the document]
johnsutor: I'll mention that, and I think it's fine to keep it as just epsilon instead of performing a redundant multiplication each time the loss is calculated.

lightly/loss/emp_ssl_loss.py
johnsutor: Sounds good. I was trying to stay faithful to the code implementation, but since the lambda term can be refactored as a combination of the coefficients, We can just have a lambda term.

guarin:
@johnsutor would you like to make the changes or do you prefer if we make the changes on our side?

johnsutor:
@guarin  I’m currently on vacation until Friday, I will push the updates then!

guarin (approved):
Awesome, thanks a lot for the changes! This is ready to be merged.

PR #1472, opened 2024-01-11
Add codecov config

IgorSusmelj:
## Changes

* Added a codecov.yml config file to make codecov pass even if the coverage slightly goes down by 0.5% as recommended. This will allow us to enforce again that all tests have to pass in order to merge a PR (we disabled this as codecov got annoying)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    project:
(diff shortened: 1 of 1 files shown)


guarin (approved):
LGTM! Are you sure this are the only required fields?

PR #1474, opened 2024-01-11
Igor lig 4447 w mse benchmark

IgorSusmelj:
## Changes

* Adds WMSE ImageNet benchmark
* Adds missing projection head


IgorSusmelj:
image
Seems to be training.

benchmarks/imagenet/resnet50/wmse.py
```diff
@@ -0,0 +1,109 @@
...
+        self.save_hyperparameters()
+        self.batch_size_per_device = batch_size_per_device
+
+        resnet = resnet50()
+        resnet.fc = Identity()  # Ignore classification head
+        self.backbone = resnet
+
+        # we use a projection head with output dimension 64
+        # and w_size of 128 to support a batch size of 256
+        self.projection_head = WMSEProjectionHead(output_dim=64)
```
philippmwirth, line 31: I think the output dimension is wrong here. From the paper:
> Finally, we use an embedding size
of 64 for CIFAR-10 and CIFAR-100, and an embedding of
size of 128 for STL-10 and Tiny ImageNet. For ImageNet-
100 we use a configuration similar to the Tiny ImageNet
experiments, and 240 epochs of training. Finally, in the
ImageNet experiments (Tab. 3), we use the implementation
and the hyperparameter configuration of (Chen et al., 2020b)
(same number of layers in the projection head, etc.) based
on their open-source implementation2, the only difference
being the learning rate and the loss function (respectively,
0.075 and the contrastive loss in (Chen et al., 2020b) vs. 0.1
and Eq. 6 in W-MSE 4

So they're using a SimCLR2 projection head.

benchmarks/imagenet/resnet50/wmse.py
```diff
@@ -0,0 +1,109 @@
...
+
+        self.criterion_WMSE4loss = WMSELoss(
+            w_size=128, embedding_dim=64, num_samples=4, gather_distributed=True
```
philippmwirth, line 34: For ImageNet they probably use `w_size=256`:
>  For CIFAR-10
and CIFAR-100, the slicing sub-batch size is 128, for Tiny
ImageNet and STL-10, it is 256

lightly/loss/wmse_loss.py
```diff
@@ -59,10 +61,18 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
 
         f_cov_shrinked = (1 - self.eps) * f_cov + self.eps * eye
 
+        # get type of f_cov_shrinked and temporary convert to full precision
+        # to support chelosky decomposition
+        f_cov_shrinked_type = f_cov_shrinked.dtype
```
philippmwirth, line 66: This looks super duper hacky. Why is it necessary?

IgorSusmelj: Yes, as written in the comment. The original code is not using half precision.

lightly/loss/wmse_loss.py
```diff
@@ -173,13 +194,23 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
             ValueError:
                 If the batch size is smaller than w_size.
         """
+        # gather all batches
+        if self.gather_distributed and dist.is_initialized():
+            world_size = dist.get_world_size()
+            if world_size > 1:
+                input = torch.cat(gather(input), dim=0)
```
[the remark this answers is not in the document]
IgorSusmelj: I removed it but I will add it again. That seems the most easy and proper way to support multi-GPU training. I'll make sure we divide the loss by the number of devices to make runs more comparable between different multi-gpu setups.

lightly/models/modules/heads.py
```diff
@@ -699,6 +699,27 @@ def __init__(
         )
 
 
+class WMSEProjectionHead(SimCLRProjectionHead):
```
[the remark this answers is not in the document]
IgorSusmelj: @guarin, we should make sure things are consistent. I'm not sure what we agreed on. AFAIK, The same goes for the transforms.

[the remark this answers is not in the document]
guarin: Aren't the default values different?

In any case, I prefer if all components of the `WMSE` model are called `WMSESomething`. Mixing components from different models is always confusing and it makes the components harder to discover in the code. If two models have the same head then we can just subclass from the first model and update the docstring.

lightly/models/modules/heads.py
```diff
@@ -699,6 +699,27 @@ def __init__(
...
+
+    def __init__(
+        self,
+        input_dim: int = 2048,
+        hidden_dim: int = 2048,
+        output_dim: int = 128,
+        num_layers: int = 2,
+        batch_norm: bool = True,
+    ):
+        super(WMSEProjectionHead, self).__init__(
```
guarin, line 718:
```suggestion
        super().__init__(
```

In general the class should not be passed to the super method.

PR #1476, opened 2024-01-12
Update install dependencies

IgorSusmelj:
## Changes

* removes openapi.txt requirements file
* move openapi dependencies to base.txt
* use restore keys (apparently this should speed up the tests as we can use the cache from the master branch when creating PRs!?)
* added a test to check minimal dependencies so we can track whenever we add new things that require version updates
* update the readme and remove support for Python 3.6

This allows to install the lightly package with minimal dependencies (no PyTorch and torchvision)

New minimal dependencies (to run all the unittests) are the following:
```
pytorch_lightning==1.7.1
torch==1.11.0
torchvision==0.12.0
```


guarin (approved):
LGTM!

.github/workflows/test_minimal_deps.yml
```diff
@@ -0,0 +1,39 @@
+name: Minimal Dependency Tests
+
+on:
+  push:
+    paths-ignore:
+      - 'docs/**'
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
```
guarin, line 9: We can also ignore `examples/**` and `benchmarks/**`

.github/workflows/test_minimal_deps.yml
```diff
@@ -0,0 +1,39 @@
...
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Cache Python Dependencies
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.pythonLocation }}
+        key: minimal-env-${{ runner.os }}-python-${{ matrix.python }}-${{ hashFiles('requirements/minimal_requirements.txt') }}
+        restore-keys: |
+          minimal-env-${{ runner.os }}-python-${{ matrix.python }}-
```
guarin, line 33: Could this not accidentally use a cache that has different minimal requirements than in `minimal_requirements.txt`?

requirements/minimal_requirements.txt
```diff
@@ -0,0 +1,40 @@
...
+docutils<=0.16
+sphinx-copybutton
+sphinx-design
+sphinx-gallery
+sphinx-tabs
+sphinx-reredirects
+sphinx_rtd_theme
+
+pre-commit
+opencv-python==4.4.0.46
```
guarin, line 35: We should be able to drop this in the future. Opencv is only used in some old tests.

requirements/minimal_requirements.txt
```diff
@@ -0,0 +1,40 @@
...
+python_dateutil==2.5.3
+requests==2.23.0
+six==1.10
+tqdm==4.57.0
+urllib3==1.25.3
+pydantic==1.10.5
+aenum==3.1.11
+pytorch_lightning==1.7.1
+torch==1.11.0
+torchvision==0.12.0
```
guarin, line 15: Could you put these as minimal version in `requirements/torch.txt`?

requirements/minimal_requirements.txt
```diff
@@ -0,0 +1,40 @@
...
+sphinx-reredirects
+sphinx_rtd_theme
+
+pre-commit
+opencv-python==4.4.0.46
+scikit-learn==0.24.2
+pandas
+matplotlib
+
+av==8.0.3
```
guarin, line 40: Could you put this as minimal requirement in `video.txt`? I think we can also drop `torchvision` from `video.txt`

.github/workflows/test_minimal_deps.yml
IgorSusmelj: If we update the requirements.txt file we would always get a new hash and therefore don't use the cache.

The only complication that can happen here is that if we have two branches and one of them introduces a requirements change it might update the cache and then the second branch might use the newly cached dependencies. The question is if we think that this is a likely scenario that might break things.

requirements/minimal_requirements.txt
IgorSusmelj: I thought about this but decided against it. There are still other modules that can be used without the latest PyTorch and torchvision packages. E.g. if someone just cares about a loss function from lightly but uses still older PyTorch or Torchvision packages, we would force them to upgrade.

On the torchvision side, many dependency constraints come from the video processing (a feature not many users use).

requirements/minimal_requirements.txt
IgorSusmelj: I updated the video.txt file

.github/workflows/test_minimal_deps.yml
guarin: I thought the `restore-keys` are alternative keys that can be used if there is a cache miss. Maybe I didn't understand correctly. But I would be very conservative with caching. Accidentally using wrong dependencies might be difficult to spot.

requirements/minimal_requirements.txt
guarin: Made an issue for it #1477

guarin (approved):
LGTM! Thanks for the changes!

.github/workflows/test_minimal_deps.yml
```diff
@@ -29,11 +33,11 @@ jobs:
...
-        restore-keys: |
-          minimal-env-${{ runner.os }}-python-${{ matrix.python }}-
     - name: Install Minimal Dependencies
       run: pip install -r requirements/minimal_requirements.txt
     - name: Install Package Without Dependencies
       run: pip install --no-deps .
     - name: Run Tests
-      run: python -m pytest -s -v --runslow
+      run:  |
+        export LIGHTLY_SERVER_LOCATION="localhost:-1"
```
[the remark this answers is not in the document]
IgorSusmelj: We use it in the other test as well. To make sure we can run all tests offline and don't require the lightly api online.

PR #1479, opened 2024-01-19
Add AIM Model from Scalable Pre-training of Large Autoregressive Image Models

guarin:
This PR implements the `AIM` model proposed in [Scalable Pre-training of Large Autoregressive Image Models](https://arxiv.org/abs/2401.08541). The implementation is based on the [original code](https://github.com/apple/ml-aim) but uses a modified version of the vision transformer from [timm](https://github.com/huggingface/pytorch-image-models) as backbone. The backbone is fully compatible with the timm vision transformer and pretrained weights from our backbone should be loadable with the timm vision transformer (state dicts are identical).

~~The implementation is a best effort. The paper and reference code miss some crucial information. Specifically, the prefix length and detailed description of the MLP architecture for the prediction head are missing. Nevertheless, the current implementation is running and is hopefully a good start.~~ I checked with the authors, the head and prefix masking should be correct now :)

## Changes
* Add `MaskedCausalVisionTransformer`
* Add `AIMPredictionHead`
* Add `AIMTransform`
* Add `AIM` benchmark module

TODO:

- [x] causal vision transformer
- [x] prediction head
- [x] benchmark model
- [ ] run benchmark
- [ ] add evaluation code 

We also have to figure out whether we want to add this to `benchmarks/imagenet/vitb16` because the backbone is clearly not vitb16 😅 

## How was it tested?

* Manually, I tested a smaller version of the model and it runs well. Couldn't benchmark the full model due to compute limitations as the smallest model version requires 600M params for the backbone and 400M params for the head.
* Will add unit tests in a follow-up PR.

## For Review

Review is only required for the following files/functions:
* `benchmarks/imagenet/vitb16/aim.py`
* `lightly/models/modules/__init__.py`
* `lightly/models/modules/heads_timm.py`
* `lightly/models/modules/masked_causal_vision_transformer.py`
* `lightly/models/utils.py` -> `random_prefix_mask` function
* `lightly/transforms/aim_transform.py`

The other files/functions have already been reviewed in other PRs but are not yet on master.


lightly/models/modules/masked_causal_vision_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+        # allows us to use bidirectional instead of causal attention during evaluation
+        # and fine-tuning.
+        attn_mask = None
+        if mask is not None:
+            attn_mask = x.new_ones(size=(B, N, N), dtype=torch.bool).tril(diagonal=0)
+            # mask has shape (B, N)
+            mask = (~mask).unsqueeze(1).expand(B, N, N).bool()
+            attn_mask = torch.logical_or(attn_mask, mask)
+            attn_mask = attn_mask.unsqueeze(1)  # (B, 1, N, N)
+
```
ersi-lightly, line 47: I think here it would make sense to have a separate method that handles the masking (with a comment also pointing to the code of the paper) and then use the `forward` method of TIMM's [`Attention`](https://github.com/huggingface/pytorch-image-models/blob/53a48883280743f5c878897bdfd3d471f1c191e4/timm/models/vision_transformer.py#L84).

lightly/models/modules/masked_causal_vision_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+        qkv = (
+            self.qkv(x)
+            .reshape(B, N, 3, self.num_heads, self.head_dim)
+            .permute(2, 0, 3, 1, 4)
```
ersi-lightly, line 51: Are we sure that here `self.head_dim=C//num_heads`?

lightly/models/modules/masked_causal_vision_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+
+        x = x.transpose(1, 2).reshape(B, N, C)
+        x = self.proj(x)
+        x = self.proj_drop(x)
+        return x
+
+
+# Type ignore because superclass has Any types.
+class MaskedCausalBlock(Block):  # type: ignore[misc]
+    """Idential to timm.models.vision_transformer.Block, but uses PrefixCausalAttention
```
ersi-lightly, line 82: Identical

lightly/models/modules/masked_causal_vision_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+        x = self._pos_embed(x)
+        x = self.patch_drop(x)
+        x = self.norm_pre(x)
+        if self.grad_checkpointing and not jit.is_scripting():
+            # TODO: This probably doesn't work correctly as it doesn't consider the
+            # mask.
+            x = _manipulate.checkpoint_seq(self.blocks, x)
+        else:
+            for block in self.blocks:
+                x = block(x, mask=mask)
```
ersi-lightly, line 245: I assume that you have to loop through each `block` here because of the `mask`?

lightly/models/utils.py
```diff
@@ -474,6 +478,40 @@ def random_token_mask(
...
+    Returns:
+        A mask tensor with shape (batch_size, sequence_length) where each entry
+        is True if the token should be masked and False otherwise.
+
+    """
+    batch_size, sequence_length = size
+    arange = torch.arange(sequence_length, device=device).expand(
+        batch_size, sequence_length
+    )
+    indices = torch.randint(0, max_prefix_length, (batch_size, 1), device=device)
```
ersi-lightly, line 510: I think this should be `indices = torch.randint(0, max_prefix_length+1, (batch_size, 1), device=device)` because the high value in `torch.randint` is not included.

benchmarks/imagenet/vitb16/aim.py
```diff
@@ -0,0 +1,161 @@
...
+        pos_embed = get_2d_sincos_pos_embed(
+            embed_dim=vit.embed_dim,
+            grid_size=int(self.num_patches**0.5),
+            cls_token=False,
+        )
+        vit.pos_embed.requires_grad = False
+        vit.pos_embed.data.copy_(torch.from_numpy(pos_embed).float().unsqueeze(0))
+
+        self.backbone = vit
+        self.projection_head = AIMPredictionHead(
```
ersi-lightly, line 49: It could make sense to have `num_channels` as an attribute.

benchmarks/imagenet/vitb16/aim.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+
+        self.criterion = MSELoss()
+
+        self.online_classifier = OnlineLinearClassifier(
+            feature_dim=vit.embed_dim, num_classes=num_classes
+        )
+
+    def forward(self, x: Tensor) -> Tensor:
+        features = self.backbone.forward_features(x)
```
ersi-lightly, line 60: This is equivalent to `mask=None`,right? Shouldn't we have `mask` as an argument here?

benchmarks/imagenet/vitb16/aim.py
```diff
@@ -0,0 +1,161 @@
...
+        batch_size = images.shape[0]
+
+        mask = random_prefix_mask(
+            size=(batch_size, self.num_patches),
+            max_prefix_length=self.num_patches - 1,
+            device=images.device,
+        )
+        features = self.backbone.forward_features(images, mask=mask)
+        # Add positional embedding before head.
+        features = self.backbone._pos_embed(features)
```
ersi-lightly, line 81: It seems like `_pos_embed` is already called in `forward_features` . Why is the second call needed here?

lightly/models/modules/masked_causal_vision_transformer.py
guarin: Good point, a separate method definitely makes sense. But I cannot use timm's attention as it doesn't pass the attention mask to the attention calculation: https://github.com/huggingface/pytorch-image-models/blob/53a48883280743f5c878897bdfd3d471f1c191e4/timm/models/vision_transformer.py#L92

lightly/models/modules/masked_causal_vision_transformer.py
guarin: Yes :) https://github.com/huggingface/pytorch-image-models/blob/53a48883280743f5c878897bdfd3d471f1c191e4/timm/models/vision_transformer.py#L86

What happens is:
```
x # shape (B, N, C)
x = self.qkv(x) # (B, N, C * 3)
x = x.reshape(B, N, 3, self.num_heads, self.head_dim) # --> C * 3 = 3 * self.num_heads * self.head_dim
```
Because `self.qkv(x)` is a linear layer from `C -> 3*C` we also know that `C == self.dim` because the layer is created based on `self.dim` here: https://github.com/huggingface/pytorch-image-models/blob/53a48883280743f5c878897bdfd3d471f1c191e4/timm/models/vision_transformer.py#L77

lightly/models/modules/masked_causal_vision_transformer.py
guarin: Yes, `self.blocks` is a `Sequential` object which only take a single input `x` in the forward method: https://pytorch.org/docs/stable/_modules/torch/nn/modules/container.html#Sequential We have to pass the mask manually :)

lightly/models/utils.py
guarin: We have `max_prefix_length == num_patches - 1 == 255`. This will generate indices in `[0, 254]`. Then we set the mask with `mask = arange >= indices`, which means that all indices up-to and including `245` can be unmasked while only index 255 is always masked. I think this is what we want.

Note that we pass `num_patches - 1` here: https://github.com/lightly-ai/lightly/blob/aa3dde81c46950f43cbd3c7af8163e9dcd453c78/benchmarks/imagenet/vitb16/aim.py#L76

benchmarks/imagenet/vitb16/aim.py
guarin: We always assume that we have RGB images and the number of channels is 3. I think I'll leave it hardcoded for now.

benchmarks/imagenet/vitb16/aim.py
guarin: Yes it is equivalent. The `forward` method in all the benchmarking modules is used to get the features for evaluation, in that case we want bidirectional attention as described in the paper.

benchmarks/imagenet/vitb16/aim.py
guarin: This is on purpose, adding positional embeddings again can help the model make better predictions because sometimes the positional embedding information gets lost in the transformer layers. It follows the description in the paper:

Screen Shot 2024-01-23 at 15 34 14

lightly/models/modules/masked_causal_vision_transformer.py
ersi-lightly: ok, thanks!

benchmarks/imagenet/vitb16/aim.py
guarin: Updated :)

ersi-lightly (approved):

adamjstewart:
FWIW, this PR caused a bit of a headache for us in TorchGeo: https://github.com/microsoft/torchgeo/issues/1824

At the moment, the changes here make lightly v1.4.26 incompatible with any version of segmentation-models-pytorch. This isn't necessarily your fault, but it would help if you could check the version of timm available before importing everything else.

PR #1482, opened 2024-01-22
Masked vision transformer

ersi-lightly:
Creates a common backbone for any architecture that uses some sort of masked variant of a Vision Transformer.


lightly/models/modules/masked_vision_transformer_timm.py
```diff
@@ -0,0 +1,140 @@
...
+import torch.nn as nn
+from torch.nn import LayerNorm, Linear, Module, Parameter, Sequential
+
+
+class MaskedVisionTransformerTIMM(MaskedVisionTransformer):
+    def __init__(
+        self,
+        vit: VisionTransformer,  # here assume that we always initialize with a vit model from timm
+        mask_token: Union[bool, Parameter],
+        device : str
```
guarin, line 22: It is usually not necessary to store the device on the class. In general it can be retrieved from some other tensor/parameter once it is needed.

lightly/models/modules/masked_vision_transformer_timm.py
```diff
@@ -0,0 +1,140 @@
...
+        return class_token
+
+    def encode(
+        self,
+        images: torch.Tensor,
+        idx_keep: Optional[torch.Tensor] = None,
+        idx_mask: Optional[torch.Tensor] = None,
+    ) -> torch.Tensor:
+        # convert images to tokens
+        input: torch.Tensor = self.patch_embed(images)
```
guarin, line 53:
```suggestion
        input = self.patch_embed(images)
```

You can leave out the type hint because `self.patch_embed` is typed.

lightly/models/modules/masked_vision_transformer_timm.py
```diff
@@ -0,0 +1,140 @@
...
+    ) -> Tensor:
+        out = self.encode(images, idx_keep, idx_mask)
+        class_token = out[:, 0]
+        return class_token
+
+    def encode(
+        self,
+        images: torch.Tensor,
+        idx_keep: Optional[torch.Tensor] = None,
+        idx_mask: Optional[torch.Tensor] = None,
```
guarin, line 50:
```suggestion
        idx_mask: Optional[torch.Tensor] = None,
        idx_keep: Optional[torch.Tensor] = None,
```

Better to keep the same order as in `forward` method. Otherwise it is easy to mess up :)

lightly/models/modules/masked_vision_transformer_timm.py
```diff
@@ -0,0 +1,140 @@
...
+        self.sequence_length = vit.patch_embed.num_patches + vit.num_prefix_tokens
+        self.device = device
+
+    def forward(
+        self,
+        images: Tensor,
+        idx_mask: Optional[Tensor] = None,
+        idx_keep: Optional[Tensor] = None,
+    ) -> Tensor:
+        out = self.encode(images, idx_keep, idx_mask)
```
guarin, line 42:
```suggestion
        out = self.encode(images, idx_keep=idx_keep, idx_mask=idx_mask)
```

lightly/models/modules/masked_vision_transformer_timm.py
```diff
@@ -0,0 +1,140 @@
...
+        class_token = out[:, 0]
+        return class_token
```
guarin, line 44:
```suggestion
         if self.vit.attn_pool is not None:
            x = self.vit.attn_pool(x)
        elif self.vit.global_pool == 'avg':
            x = x[:, self.vit.num_prefix_tokens:].mean(dim=1)
        elif self.vit.global_pool:
            x = x[:, 0]  # class token
```
We should handle the different cases as in VisionTransformer:
https://github.com/huggingface/pytorch-image-models/blob/53a48883280743f5c878897bdfd3d471f1c191e4/timm/models/vision_transformer.py#L681-L686

lightly/models/modules/masked_vision_transformer_timm.py
```diff
@@ -0,0 +1,140 @@
...
+                (H, W),
+                num_prefix_tokens=0 if self.vit.no_embed_class else self.vit.num_prefix_tokens,
+            )
+            x = x.view(B, -1, C)
+        else:
+            pos_embed = self.vit.pos_embed
+        # TODO: give option to either have positional encoding for the prefix tokens or not using the self.vit.no_embed_class. 
+        # Here it is assumed that the prefix tokens always have positional encodings.
+        x = x + pos_embed
+        return self.vit.pos_drop(x)
```
guarin, line 100:
```suggestion
        if self.no_embed_class:
            x[:, self.vit.num_prefix_tokens:] += pos_embed
        else:
            x = x + pos_embed
        return self.vit.pos_drop(x)
```

lightly/models/modules/masked_vision_transformer_timm.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def _init_weights(module: Module) -> None:
+    if isinstance(module, Linear):
+        nn.init.xavier_uniform_(module.weight)
+        if isinstance(module, Linear) and module.bias is not None:
+            nn.init.constant_(module.bias, 0)
+    elif isinstance(module, LayerNorm):
+        nn.init.constant_(module.bias, 0)
+        nn.init.constant_(module.weight, 1.0)
```
guarin, line 139: Weight initialization should already be handled by the `VisionTransformer` superclass. I think we should move the MAE specific initialization outside the class. For example we can make a function `init_weights_mae` that does all this. And I think we can move `_initialize_2d_sine_cosine_positional_embedding` to `models/utils.py`  as we already have the other positional embedding functions there.

lightly/models/modules/masked_vision_transformer.py
```diff
@@ -0,0 +1,24 @@
...
+
+import torch
+
+
+class MaskedVisionTransformer(ABC):
+    def __init__(self):
+        pass
+
+    @abstractmethod
+    def forward(self):
```
guarin, line 11: The functions in the ABC class should have the same arguments and return types as the functions in `MaskedVisionTransformerTIMM`.

guarin (changes requested):
Left some comments :)

ersi-lightly:
> Left some comments :)

Many thanks Guarin! Although the PR was not ready for review yet :)

lightly/models/modules/masked_vision_transformer_timm.py
ersi-lightly: I don't think that the weight initialization can be handled by the `VisionTransformer` class because depending on whether we use torchvision or TIMM for example the methods would be different. I think I will keep the initialization as it is for now, because this is what was being done essentially so far (when the `Backbone` of the `masked_autoencoder.py` was used for other architectures). It will also make more straightforward the comparison of the examples with this new backbone and with the one that we had before. WDYT?

benchmarks/imagenet/vitb16/mae.py
```diff
@@ -22,14 +24,16 @@ def __init__(self, batch_size_per_device: int, num_classes: int) -> None:
         self.batch_size_per_device = batch_size_per_device
 
         decoder_dim = 512
-        vit = vit_base_patch16_224()
+        vit = vit_base_patch16_224(drop_path_rate=0.1)
```
guarin, line 27: This should only be set during finetuning but not during pretraining.

lightly/models/modules/masked_vision_transformer_timm.py
```diff
@@ -0,0 +1,169 @@
...
+            x = torch.cat(prefix_tokens + [x], dim=1)
+        return x
+
+    def add_pos_embed(self, x: Tensor) -> Tensor:
+        x_prefix = (
+            x[:, : self.vit.num_prefix_tokens, :]
+            if self.vit.num_prefix_tokens
+            else torch.empty()
+        )
+        x = x[:, self.vit.num_prefix_tokens :, :] if self.vit.num_prefix_tokens else x
```
guarin, line 108:
```suggestion
        x_prefix = x[:, : self.vit.num_prefix_tokens, :]
        x = x[:, self.vit.num_prefix_tokens :, :]
```

This is simpler and does not create any overhead. Note that using `torch.empty()` here is dangerous because it will potentially create a tensor on the wrong device and with the wrong datatype. If you really need a new tensor then `x.new_empty(size)` should be used instead. This makes sure that the new tensor is on the correct device with the correct datatype.

lightly/models/modules/masked_vision_transformer_timm.py
guarin: We can leave it for now but eventually we should move it outside the class (not outside the file though!). The reason is that different methods use different initialization schemes. Having a single `_init_weights` method will not cover the different cases.

guarin (changes requested):
Thanks for the update!

lightly/models/modules/masked_autoencoder_timm.py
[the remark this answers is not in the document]
ersi-lightly: Yes, I was going to do that after merging to the `ersi-lig-3910-update-mae-benchmark-code` branch

benchmarks/imagenet/vitb16/mae.py
ersi-lightly: This is already taken care of. See:
https://github.com/huggingface/pytorch-image-models/blob/88889de923149533593482879a564e7f7e0dabbd/timm/layers/drop.py#L139

benchmarks/imagenet/vitb16/mae.py
```diff
@@ -123,7 +123,7 @@ def configure_optimizers(self):
         params, params_no_weight_decay = utils.get_weight_decay_parameters(
             [self.backbone, self.decoder]
         )
-        params.append(self.mask_token)
+        params.append(self.backbone.mask_token)
```
guarin: params should already contain `self.backbone.mask_token` as we pass `self.backbone` to `utils.get_weight_decay_parameters`.

benchmarks/imagenet/vitb16/finetune_eval.py
```diff
@@ -35,6 +35,9 @@ def __init__(
         super().__init__(
             model, batch_size_per_device, feature_dim, num_classes, topk, freeze_model
         )
+        # Add path dropout.
+        model.__dict__["backbone"].__dict__["vit"].drop_path_rate = 0.1
```
guarin: I just checked and it looks like this is more complicated than expected. While this change sets `model.backbone.vit.drop_path_rate` to `0.1` it doesn't actually add path dropout. The reason is that timm creates the path dropout layers when the vision transformer class is created ([here](https://github.com/huggingface/pytorch-image-models/blob/88889de923149533593482879a564e7f7e0dabbd/timm/models/vision_transformer.py#L151) and [here](https://github.com/huggingface/pytorch-image-models/blob/88889de923149533593482879a564e7f7e0dabbd/timm/models/vision_transformer.py#L161)). Changing the parameter afterwards does not affect the layers anymore. An additional problem is that the dropout parameter is scaled with every layer [here](https://github.com/huggingface/pytorch-image-models/blob/88889de923149533593482879a564e7f7e0dabbd/timm/models/vision_transformer.py#L504). 

I am not sure what the best way is to enable path dropout after creating the model. Our setup does not support his usecase well. I guess we would have to add a `set_path_dropout` method to the model and call it from here. To set the path dropout we would have to go through the model and create the layers and make sure the drop rate is correctly scaled. Maybe we can add a todo comment for now?

guarin (approved):
🎉

PR #1484, opened 2024-01-23
Fix TiCoLoss

guarin:
## Changes

* Detach auxiliary matrix B in loss computation

The figure in the paper suggests that no gradients are back-propagated through the batch covariance matrix B:

Screen Shot 2024-01-23 at 15 54 28


IgorSusmelj (approved):
LGTM

PR #1487, opened 2024-01-25
Add timm version check

guarin:
## Changes

* Add version check for timm
* Only import AIM model parts if timm >= 0.9.9

This fixes import issues if users have an older version of timm installed. For details see:
* https://github.com/microsoft/torchgeo/issues/1824
* https://github.com/lightly-ai/lightly/pull/1479#issuecomment-1909750194

### DIFF
--- a/lightly/__init__.py
+++ b/lightly/__init__.py
@@ -79,10 +79,11 @@ __version__ = "1.4.26"
 
 
 import os
-import packaging.version
-from packaging.version import Version as _Version
-from packaging.version import InvalidVersion
 from typing import Optional
+
+import packaging.version
+from packaging.version import InvalidVersion
+from packaging.version import Version as _Version
 
 # see if torchvision vision transformer is available
 try:
--- a/lightly/__init__.py
+++ b/lightly/__init__.py
@@ -79,6 +79,10 @@ __version__ = "1.4.26"
 
 
 import os
+import packaging.version
+from packaging.version import Version as _Version
+from packaging.version import InvalidVersion
+from typing import Optional
 
 # see if torchvision vision transformer is available
 try:
@@ -97,10 +101,13 @@ try:
     import timm
 except ImportError:
     _timm_available = False
-    _timm_version = ""
+    _timm_version: Optional[_Version] = None
 else:
     _timm_available = True
-    _timm_version = timm.__version__
+    try:
+        _timm_version = packaging.version.parse(timm.__version__)
+    except InvalidVersion:
+        _timm_version = None
 
 
 if os.getenv("LIGHTLY_DID_VERSION_CHECK", "False") == "False":
--- a/lightly/__init__.py
+++ b/lightly/__init__.py
@@ -79,37 +79,6 @@ __version__ = "1.4.26"
 
 
 import os
-from typing import Optional
-
-import packaging.version
-from packaging.version import InvalidVersion
-from packaging.version import Version as _Version
-
-# see if torchvision vision transformer is available
-try:
-    import torchvision.models.vision_transformer
-
-    _torchvision_vit_available = True
-except (
-    RuntimeError,  # Different CUDA versions for torch and torchvision
-    OSError,  # Different CUDA versions for torch and torchvision (old)
-    ImportError,  # No installation or old version of torchvision
-):
-    _torchvision_vit_available = False
-
-# Check if timm is available.
-try:
-    import timm
-except ImportError:
-    _timm_available = False
-    _timm_version: Optional[_Version] = None
-else:
-    _timm_available = True
-    try:
-        _timm_version = packaging.version.parse(timm.__version__)
-    except InvalidVersion:
-        _timm_version = None
-
 
 if os.getenv("LIGHTLY_DID_VERSION_CHECK", "False") == "False":
     os.environ["LIGHTLY_DID_VERSION_CHECK"] = "True"
--- a/lightly/__init__.py
+++ b/lightly/__init__.py
@@ -97,8 +97,10 @@ try:
     import timm
 except ImportError:
     _timm_available = False
+    _timm_version = ""
 else:
     _timm_available = True
+    _timm_version = timm.__version__
 
 
 if os.getenv("LIGHTLY_DID_VERSION_CHECK", "False") == "False":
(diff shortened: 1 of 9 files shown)


guarin:
FYI @adamjstewart

lightly/models/modules/__init__.py
```diff
@@ -35,8 +35,8 @@
         MAEDecoder,
         MAEEncoder,
     )
-if _timm_available:
-    # Requires timm.
+if _timm_available and _timm_version >= "0.9.9":
```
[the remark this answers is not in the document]
guarin: Ah yes good point, thanks!

lightly/__init__.py
```diff
@@ -79,6 +79,11 @@
 
 
 import os
+from typing import Optional
+
+import packaging.version
```
guarin, line 84: Didn't want to import `from packaging import version` because it could be confusing to have a `ligthly.version` attribute.

lightly/__init__.py
```diff
@@ -79,6 +79,11 @@
...
+from packaging.version import InvalidVersion
+from packaging.version import Version as _Version
```
guarin, line 86: Didn't want to import `from packing.version import Version` because it could be confusing to have a `lightly.Version` attribute.

requirements/base.txt
```diff
@@ -9,4 +9,5 @@ six>=1.10
 tqdm>=4.44
 urllib3 >= 1.25.3
 pydantic >= 1.10.5, < 2
-aenum >= 3.1.11
\ No newline at end of file
+aenum >= 3.1.11
+packaging
```
guarin, line 13: Didn't specify minimal version as the package should already be installed in 99% of the cases as it is a dependency of many other packages. The functionality we require from the package is also available in very old versions. I wouldn't expect any issues here.

adamjstewart (approved):
LGTM. The try-except is probably overkill if you want to simplify things. Thanks for addressing this so quickly!

requirements/minimal_requirements.txt
```diff
@@ -14,6 +14,7 @@ pytorch_lightning==1.7.1
 torch==1.11.0
 torchvision==0.12.0
 pillow==7.1.2
+packaging
```
IgorSusmelj, line 17: Let's put a version here and not leave it open as we otherwise don't test "minimal" requirements. Most other dependencies are from 2021. So maybe we can just use version "21.0"?

IgorSusmelj (changes requested):
Thanks for the changes. Everything looks good. I have a few suggestions:

- We should add an explicit version for packaging in the minimal requirements (suggestion in comment).
- Since we use now a lib for version comparison we should get rid of the old code we wrote ourselves in the past: https://github.com/lightly-ai/lightly/blob/master/lightly/utils/version_compare.py. I'd do it in the same PR to avoid any tech debt for later and to make sure the new code really works as it should.

requirements/minimal_requirements.txt
adamjstewart: In TorchGeo, when we had a dependency on packaging, we used packaging>=17, as 17 was the minimum version of packaging supported by our minimum version of lightning. Like you said, the version comparison/parsing stuff has always been there, it's more about matching your other minimum requirement versions.

adamjstewart:
Packaging is no longer needed now right?

guarin:
> Packaging is no longer needed now right?

Yes exactly :) Plus we actually verify whether the required classes are there instead of relying on the version.

The change should also speed up importing `lightly` in environments where torchvision and/or timm are installed as we now only import them when needed.

lightly/utils/dependency.py
```diff
@@ -0,0 +1,27 @@
...
+        return False
+    else:
+        return True
+
+
+@functools.lru_cache(maxsize=1)
+def timm_vit_available() -> bool:
+    try:
+        # Requires timm >= 0.9.9
+        import timm.models.vision_transformer
```
adamjstewart, line 23: This import has existed in timm since v0.1, so this won't actually test whether or not newer timm versions are installed. The specific import that failed for us in TorchGeo was `timm.layers.LayerType`, which was added in 0.9.9. I _think_ that's the newest feature you're using.

guarin: Thanks! I added a check for `LayerType` specifically. With timm 0.9.8 we now indeed skip importing the MaskedCausalVisionTransformer and with timm 0.9.9 we import it.

tests/models/modules/test_masked_autoencoder.py
```diff
@@ -84,7 +88,9 @@ def test_forward_cuda(self):
         self._test_forward(torch.device("cuda"))
 
 
-@unittest.skipUnless(_torchvision_vit_available, "Torchvision ViT not available")
+@unittest.skipUnless(
+    dependency.torchvision_vit_available(), "Torchvision ViT not available"
+)
 class TestMAEDecoder(unittest.TestCase):
     def test_init(self):
         return MAEDecoder(
```
adamjstewart, line 96: FWIW, we should prob remove this return statement:
```
.../lib/python3.11/unittest/case.py:678: DeprecationWarning: It is deprecated to return a value that is not None from a test case (<bound method TestMAEDecoder.test_init of <test_masked_autoencoder.TestMAEDecoder testMethod=test_init>>)
    return self.run(*args, **kwds)
```

adamjstewart (approved):
This works for me now! The edited test file passes and correctly skips certain tests because I have timm 0.9.2 installed.

tests/models/modules/test_masked_autoencoder.py
guarin: Thanks, fixed! We anyways want to move test to pytest since a while :)

IgorSusmelj (approved):
This updated version is much more elegant and solves our need without new dependencies! Well done!

adamjstewart:
Thanks for the quick fix @guarin, glad we'll be able to use future timm versions again!

lightly/models/modules/__init__.py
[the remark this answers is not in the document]
Sanderway: good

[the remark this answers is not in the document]
Sanderway: > Ah yes good point, thanks!

PR #1490, opened 2024-01-30
upgrade requirements/base.txt minimum numpy version to 1.20 for typing

MalteEbner:
We are using the `numpy.typing` subpackage, which is only available in numpy >= 1.20: https://numpy.org/devdocs/reference/typing.html#module-numpy.typing

Thus we need to update the minimum requirement.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,7 +2,7 @@
 certifi>=14.05.14
 hydra-core>=1.0.0
 lightly_utils~=0.0.0
-numpy>=1.18.1
+numpy>=1.20.0
 python_dateutil>=2.5.3
 requests>=2.23.0
 six>=1.10


IgorSusmelj (changes requested):
Let's not merge this as of now. We need a proper strategy to deal with the dependencies.

For context:
- We just added https://github.com/lightly-ai/lightly/blob/master/requirements/minimal_requirements.txt which contains the actual minimum requirements needed to use all the features. We have a new test for this in the CI pipeline
- We did not want to change all the base requirements yet, as we have many users using only very specific features due to the modular way lightly works. We will have to figure out how we want to deal with that. The issue is that we accidentally broke many requirements because our previous tests were always installing latest packages. So we never tested if the requirements we once established are still valid or not.

japrescott (changes requested):
lets quickly review the minimum requirements

SauravMaheshkar:
How does https://github.com/lightly-ai/lightly/pull/1559 affect this ?

guarin:
@MalteEbner what is the state of this PR?

IMO there are two options:
* Handle `from numpy.typing import NDArray` depending on whether it is available or not. If it is not available we could define a generic alias for `NDArray` instead. Not sure how that would work for older Python versions though.
* Increase minimal numpy version in `pyproject.toml`. Numpy 1.18 has reached end of life >2 years ago.

The current state with `numpy>=1.18.1, <2` as [default requirement](https://github.com/lightly-ai/lightly/blob/6ebde88201ff8e647a809221617f2cf76ae81554/pyproject.toml#L41) and `numpy==1.26.6` as [minimal requirement](https://github.com/lightly-ai/lightly/blob/6ebde88201ff8e647a809221617f2cf76ae81554/pyproject.toml#L95C4-L95C17) is not ideal.

guarin:
This has been fixed in https://github.com/lightly-ai/lightly/pull/1625

We still support numpy<1.20 and only import from numpy.typing when type-checking.

PR #1491, opened 2024-02-02
Jeremy lig 4438 new pip package with v4 endpoints

japrescott:
closes lig-4438
- switch from scheduling v3 to v4


MalteEbner (approved):
LGTM

However, please make sure you already generated from the new specs in case you updated them in-between.

Issue #1481, opened 2024-01-22
Add MMCRProjectionHead


guarin:
We should add the projection head for the MMCR paper: https://arxiv.org/abs/2303.03307
It should have the same architecture as SImCLR but with dimensions (8192, 8192, 512).

guarin:
Completed in #1493

PR #1493, opened 2024-02-04
Add mmcr projection head

LukeSutor:
Add MMCR Projection Head as proposed in #1481

### DIFF
--- a/lightly/models/modules/heads.py
+++ b/lightly/models/modules/heads.py
@@ -641,13 +641,13 @@ class DINOProjectionHead(ProjectionHead):
         x = nn.functional.normalize(x, dim=-1, p=2)
         x = self.last_layer(x)
         return x
-    
+
 
 class MMCRProjectionHead(ProjectionHead):
     """Projection head used for MMCR.
 
-    "Following Chen et al. (14), we append a small perceptron to the output 
-    of the average pooling layer of the ResNet so that zi = g(h(xi)), where 
+    "Following Chen et al. (14), we append a small perceptron to the output
+    of the average pooling layer of the ResNet so that zi = g(h(xi)), where
     h is the ResNet and g is the MLP." [0]
 
     - [0]: MMCR, 2023, https://arxiv.org/abs/2303.03307
--- a/lightly/models/modules/heads.py
+++ b/lightly/models/modules/heads.py
@@ -641,6 +641,62 @@ class DINOProjectionHead(ProjectionHead):
         x = nn.functional.normalize(x, dim=-1, p=2)
         x = self.last_layer(x)
         return x
+    
+
+class MMCRProjectionHead(ProjectionHead):
+    """Projection head used for MMCR.
+
+    "Following Chen et al. (14), we append a small perceptron to the output 
+    of the average pooling layer of the ResNet so that zi = g(h(xi)), where 
+    h is the ResNet and g is the MLP." [0]
+
+    - [0]: MMCR, 2023, https://arxiv.org/abs/2303.03307
+    """
+
+    def __init__(
+        self,
+        input_dim: int = 8192,
+        hidden_dim: int = 8192,
+        output_dim: int = 512,
+        num_layers: int = 2,
+        batch_norm: bool = True,
+    ):
+        """Initialize a new MMCRProjectionHead instance.
+
+        Args:
+            input_dim: Number of input dimensions.
+            hidden_dim: Number of hidden dimensions.
+            output_dim: Number of output dimensions.
+            num_layers: Number of hidden layers (2 for v1, 3+ for v2).
+            batch_norm: Whether or not to use batch norms.
+        """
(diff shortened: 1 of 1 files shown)


lightly/models/modules/heads.py
```diff
@@ -643,6 +643,62 @@ def forward(self, x: Tensor) -> Tensor:
...
+        layers: List[Tuple[int, int, Optional[nn.Module], Optional[nn.Module]]] = []
+        layers.append(
+            (
+                input_dim,
+                hidden_dim,
+                nn.BatchNorm1d(hidden_dim) if batch_norm else None,
+                nn.ReLU(),
+            )
+        )
+        for _ in range(2, num_layers):
```
guarin, line 682:
```suggestion
        for _ in range(num_layers - 1):
```

According to the original code there should be 3 linear layers in total. As we add one before and after the loop we should only add a single one in the loop: https://github.com/ThomasYerxa/mmcr/blob/fafd3a98e7f03b48d6ea62f8e476921770c5a235/mmcr/imagenet/models.py#L124-L132

lightly/models/modules/heads.py
```diff
@@ -643,6 +643,62 @@ def forward(self, x: Tensor) -> Tensor:
...
+        num_layers: int = 2,
+        batch_norm: bool = True,
+    ):
+        """Initialize a new MMCRProjectionHead instance.
+
+        Args:
+            input_dim: Number of input dimensions.
+            hidden_dim: Number of hidden dimensions.
+            output_dim: Number of output dimensions.
+            num_layers: Number of hidden layers (2 for v1, 3+ for v2).
```
guarin, line 670:
```suggestion
            num_layers: Number of hidden layers.
```

lightly/models/modules/heads.py
```diff
@@ -643,6 +643,62 @@ def forward(self, x: Tensor) -> Tensor:
...
+                    hidden_dim,
+                    nn.BatchNorm1d(hidden_dim) if batch_norm else None,
+                    nn.ReLU(),
+                )
+            )
+        layers.append(
+            (
+                hidden_dim,
+                output_dim,
+                nn.BatchNorm1d(output_dim) if batch_norm else None,
```
guarin, line 695: The last layer is a simple linear layer without batchnorm: https://github.com/ThomasYerxa/mmcr/blob/fafd3a98e7f03b48d6ea62f8e476921770c5a235/mmcr/imagenet/models.py#L132

guarin (changes requested):
Thanks a lot for the PR!

I had a look at the original code and the architecture is actually not the same as in  SimCLR. They made quite a lot of changes, I should have checked this when writing the issue, sorry for that. Left some comments on what we have to change though. 

I also noticed that they disabled the bias term in all `Linear` layers in the head. We currently don't have this functionality in the `ProjectionHead` class. Would you be interested in adding that as well? I think we could do it by adding an extra, optional argument to the `blocks` tuples when creating the `ProjectionHead`. Something like this:
```
    def __init__(
        self, 
        blocks: List[
			Union[
				Tuple[int, int, Optional[nn.Module], Optional[nn.Module]], # w/o use_bias term
				Tuple[int, int, Optional[nn.Module], Optional[nn.Module], bool], # with use_bias term
			]
		],

    ) -> None:
        super(ProjectionHead, self).__init__()
        layers: List[nn.Module] = []
        for block in blocks:
			input_dim, output_dim, batch_norm, non_linearity in block[:-4]:
            use_bias = block[4] if len(block) > 4 else not bool(batch_norm)
            layers.append(nn.Linear(input_dim, output_dim, bias=use_bias))
            if batch_norm:
                layers.append(batch_norm)
            if non_linearity:
                layers.append(non_linearity)
        self.layers = nn.Sequential(*layers)
```

LukeSutor:
When implementing the use_bias functionality for the ProjectionHead class, won't it now be necessary to change all of the types for the lists inside of the other ProjectionHead classes to include the union of the two types? (e.g. https://github.com/lightly-ai/lightly/blob/e84f85896bcaedeb1ff63b9c4bcf7c9d2885bcdf/lightly/models/modules/heads.py#L162).

guarin:
> @guarin When implementing the use_bias functionality for the ProjectionHead class, won't it now be necessary to change all of the types for the lists inside of the other ProjectionHead classes to include the union of the two types? (e.g.

I don't think it is necessary if the `ProjectionHead.__init__` method has the following type:
```
    def __init__(
        self, 
        blocks: List[
			Union[
				Tuple[int, int, Optional[nn.Module], Optional[nn.Module]], # w/o use_bias term
				Tuple[int, int, Optional[nn.Module], Optional[nn.Module], bool], # with use_bias term
			]
		],
```
This will allow it to accept either the old lists with 4 entries or the new ones with 5 entries.

guarin:
Thanks for the update! The code looks good but it seems that there is now a duplicate `MMCRProjectionHead` in the file.

guarin (approved):
LGTM! Thanks a lot!

PR #1501, opened 2024-02-14
Jeremy lig 4589 lightlypathregex pip

japrescott:
partially closes lig-4589
- generate new specs with v4 config and lightlyPathRegex for future proofing


MalteEbner (approved):
LGTM

PR #1506, opened 2024-02-19
paginate mapping endpoints

japrescott:
closes lig-4613
- paginate mapping endpoint to allow >400k samples without overloading the API


MalteEbner (commented):
Code itself looks good. Did you test it once manually?

PR #1509, opened 2024-02-26
Add AIM examples and docs

### LINKED ISSUE #1497: Add AIM examples and docs
An implementation for the AIM model (Scalable Pre-training of Large Autoregressive Image Models) was added in #1479. This issue is about adding examples and docs for the model.

**Todo**
- [ ] Add examples
- [ ] Add docs

guarin:
### Changes

* Add AIM examples
* Add AIM docs
* Add AIM to README

[Uploading AIM — lightly 1.5.0 documentation.pdf…]()

Closes #1497

### DIFF
--- a/examples/pytorch/aim.py
+++ b/examples/pytorch/aim.py
@@ -15,7 +15,7 @@ class AIM(nn.Module):
     def __init__(self, vit):
         super().__init__()
         utils.initialize_2d_sine_cosine_positional_embedding(
-            pos_embedding=vit.pos_embed
+            pos_embedding=vit.pos_embed, has_class_token=vit.has_class_token
         )
         self.patch_size = vit.patch_embed.patch_size[0]
         self.num_patches = vit.patch_embed.num_patches
--- a/examples/pytorch/aim.py
+++ b/examples/pytorch/aim.py
@@ -39,7 +39,7 @@ class AIM(nn.Module):
 
         # Convert images to patches and normalize them.
         patches = utils.patchify(images, self.patch_size)
-        patches = utils.normalize_mean_var(patches)
+        patches = utils.normalize_mean_var(patches, dim=-1)
 
         return predictions, patches
 
--- a/examples/pytorch/aim.py
+++ b/examples/pytorch/aim.py
@@ -14,11 +14,12 @@ from lightly.transforms import AIMTransform
 class AIM(nn.Module):
     def __init__(self, vit):
         super().__init__()
-        self.patch_size = vit.patch_embed.patch_size[0]
-        self.num_patches = vit.patch_embed.num_patches
         utils.initialize_2d_sine_cosine_positional_embedding(
             pos_embedding=vit.pos_embed
         )
+        self.patch_size = vit.patch_embed.patch_size[0]
+        self.num_patches = vit.patch_embed.num_patches
+
         self.backbone = vit
         self.projection_head = AIMPredictionHead(
             input_dim=vit.embed_dim, output_dim=3 * self.patch_size**2, num_blocks=1
--- a/examples/pytorch/aim.py
+++ b/examples/pytorch/aim.py
@@ -1,0 +1,107 @@
+# Note: The model and training settings do not follow the reference settings
+# from the paper. The settings are chosen such that the example can easily be
+# run on a small dataset with a single GPU.
+
+import torch
+import torchvision
+from torch import nn
+
+from lightly.models import utils
+from lightly.models.modules import AIMPredictionHead, MaskedCausalVisionTransformer
+from lightly.transforms import AIMTransform
+
+
+class AIM(nn.Module):
+    def __init__(self, vit):
+        super().__init__()
+
+        self.patch_size = vit.patch_embed.patch_size[0]
+        self.num_patches = vit.patch_embed.num_patches
+
+        # Use absolute positional embedding.
+        pos_embed = utils.get_2d_sincos_pos_embed(
+            embed_dim=vit.embed_dim,
+            grid_size=int(self.num_patches**0.5),
+            cls_token=False,
+        )
+        vit.pos_embed.requires_grad = False
+        vit.pos_embed.data.copy_(torch.from_numpy(pos_embed).float().unsqueeze(0))
+
+        self.backbone = vit
+        self.projection_head = AIMPredictionHead(
+            input_dim=vit.embed_dim, output_dim=3 * self.patch_size**2, num_blocks=1
+        )
+
+    def forward(self, images):
+        batch_size = images.shape[0]
+
+        mask = utils.random_prefix_mask(
+            size=(batch_size, self.num_patches),
+            max_prefix_length=self.num_patches - 1,
+            device=images.device,
+        )
+        features = self.backbone.forward_features(images, mask=mask)
+        # Add positional embedding before head.
+        features = self.backbone._pos_embed(features)
+        predictions = self.projection_head(features)
+
+        # Convert images to patches and normalize them.
+        patches = utils.patchify(images, self.patch_size)
+        mean = patches.mean(dim=-1, keepdim=True)
+        var = patches.var(dim=-1, keepdim=True)
+        patches = (patches - mean) / (var + 1.0e-6) ** 0.5
+
+        return predictions, patches
+
+
+vit = MaskedCausalVisionTransformer(
+    img_size=224,
+    patch_size=32,
+    embed_dim=768,
+    depth=12,
+    num_heads=12,
+    qk_norm=False,
+    class_token=False,
+    no_embed_class=True,
+)
+model = AIM(vit)
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+model.to(device)
+
+transform = AIMTransform()
+# we ignore object detection annotations by setting target_transform to return 0
+dataset = torchvision.datasets.VOCDetection(
+    "datasets/pascal_voc",
+    download=True,
+    transform=transform,
+    target_transform=lambda t: 0,
+)
+# or create a dataset from a folder containing images or videos:
+# dataset = LightlyDataset("path/to/folder")
+
+dataloader = torch.utils.data.DataLoader(
+    dataset,
+    batch_size=256,
+    shuffle=True,
+    drop_last=True,
+    num_workers=8,
+)
+
+criterion = nn.MSELoss()
+optimizer = torch.optim.AdamW(model.parameters(), lr=1.5e-4)
+
+print("Starting Training")
+for epoch in range(10):
+    total_loss = 0
+    for batch in dataloader:
+        views = batch[0]
+        images = views[0].to(device)  # views contains only a single view
+        predictions, targets = model(images)
+        loss = criterion(predictions, targets)
+        total_loss += loss.detach()
+        loss.backward()
+        optimizer.step()
+        optimizer.zero_grad()
+    avg_loss = total_loss / len(dataloader)
+    print(f"epoch: {epoch:>02}, loss: {avg_loss:.5f}")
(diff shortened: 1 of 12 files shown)


examples/pytorch/aim.py
```diff
@@ -0,0 +1,107 @@
...
+        batch_size = images.shape[0]
+
+        mask = utils.random_prefix_mask(
+            size=(batch_size, self.num_patches),
+            max_prefix_length=self.num_patches - 1,
+            device=images.device,
+        )
+        features = self.backbone.forward_features(images, mask=mask)
+        # Add positional embedding before head.
+        features = self.backbone._pos_embed(features)
```
ersi-lightly, line 45: It would be great if we could avoid calling here the private method.. I suppose it is the best we can do though right now, as we need to add again the positional embedding after `forward_features`

examples/pytorch/aim.py
```diff
@@ -0,0 +1,107 @@
...
+        predictions = self.projection_head(features)
+
+        # Convert images to patches and normalize them.
+        patches = utils.patchify(images, self.patch_size)
+        mean = patches.mean(dim=-1, keepdim=True)
+        var = patches.var(dim=-1, keepdim=True)
+        patches = (patches - mean) / (var + 1.0e-6) ** 0.5
```
ersi-lightly, line 52: Did you find that there were numerical instabilities here? Is the addition of 1.0e-6 necessary here?
Also, you could probably use `torch.nn.functional.batch_norm`.

examples/pytorch/aim.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+vit = MaskedCausalVisionTransformer(
+    img_size=224,
+    patch_size=32,
+    embed_dim=768,
+    depth=12,
+    num_heads=12,
+    qk_norm=False,
+    class_token=False,
```
ersi-lightly, line 64: This will set the `vit.cls_token` that I mentioned before

examples/pytorch/aim.py
```diff
@@ -0,0 +1,107 @@
...
+dataloader = torch.utils.data.DataLoader(
+    dataset,
+    batch_size=256,
+    shuffle=True,
+    drop_last=True,
+    num_workers=8,
+)
+
+criterion = nn.MSELoss()
+optimizer = torch.optim.AdamW(model.parameters(), lr=1.5e-4)
```
ersi-lightly, line 92: I would add the learning rate as a variable in the example.

examples/pytorch_lightning/aim.py
```diff
@@ -0,0 +1,103 @@
...
+from lightly.transforms import AIMTransform
+
+
+class AIM(pl.LightningModule):
+    def __init__(self) -> None:
+        super().__init__()
+
+        img_size = 224
+        self.patch_size = 32
+        self.num_patches = (img_size // self.patch_size) ** 2
```
ersi-lightly, line 21: You can define the `self.num_patches`, after creating an object `vit` of `MaskedCausalVisionTransformer` below using `self.num_patches = vit.patch_embed.num_patches`

ersi-lightly, line 21: Similar for `patch_size` and you can use `vit.patch_embed.patch_size[0]`. This way you wouldn't need to have the attribute `patch_size` for the class `AIM`

examples/pytorch_lightning/aim.py
```diff
@@ -0,0 +1,103 @@
...
+
+        vit = MaskedCausalVisionTransformer(
+            img_size=img_size,
+            patch_size=self.patch_size,
+            embed_dim=768,
+            depth=12,
+            num_heads=12,
+            qk_norm=False,
+            class_token=False,
+            no_embed_class=True,
```
ersi-lightly, line 31: It would be better to pass the values of these parameters at the `init` of `AIM`

examples/pytorch_lightning/aim.py
```diff
@@ -0,0 +1,103 @@
...
+        patches = utils.patchify(images, self.patch_size)
+        mean = patches.mean(dim=-1, keepdim=True)
+        var = patches.var(dim=-1, keepdim=True)
+        patches = (patches - mean) / (var + 1.0e-6) ** 0.5
+
+        loss = self.criterion(predictions, patches)
+        return loss
+
+    def configure_optimizers(self):
+        optim = torch.optim.AdamW(self.parameters(), lr=1.5e-4)
```
ersi-lightly, line 75: Similar to before, I would pass the learning rate as a parameter to `configure_optimizers`. We could also pass the optimizer and only by default use Adam (but this is not necessary for the example TBH)

examples/pytorch_lightning/aim.py
```diff
@@ -0,0 +1,103 @@
...
+        return optim
+
+
+model = AIM()
```
ersi-lightly, line 79: Here I would pass the values of the parameters of the model as I mentioned before

examples/pytorch_lightning_distributed/aim.py
```diff
@@ -0,0 +1,111 @@
...
+from lightly.transforms import AIMTransform
+
+
+class AIM(pl.LightningModule):
+    def __init__(self) -> None:
+        super().__init__()
+
+        img_size = 224
+        self.patch_size = 32
+        self.num_patches = (img_size // self.patch_size) ** 2
```
ersi-lightly, line 21: Similar to the pytorch lightning example

examples/pytorch_lightning_distributed/aim.py
```diff
@@ -0,0 +1,111 @@
...
+            qk_norm=False,
+            class_token=False,
+            no_embed_class=True,
+        )
+
+        # Use absolute positional embedding.
+        pos_embed = utils.get_2d_sincos_pos_embed(
+            embed_dim=vit.embed_dim,
+            grid_size=int(self.num_patches**0.5),
+            cls_token=False,
```
ersi-lightly, line 38: Similar to before

examples/pytorch_lightning/aim.py
```diff
@@ -0,0 +1,103 @@
...
+        self.backbone = vit
+        self.projection_head = AIMPredictionHead(
+            input_dim=vit.embed_dim, output_dim=3 * self.patch_size**2, num_blocks=1
+        )
+
+        self.criterion = nn.MSELoss()
+
+    def training_step(self, batch, batch_idx):
+        images, targets = batch[0], batch[1]
+        images = images[0]  # images is a list containing only one view
```
ersi-lightly, line 52: Could you rename to reflect that?

examples/pytorch_lightning_distributed/aim.py
```diff
@@ -0,0 +1,111 @@
...
+        features = self.backbone.forward_features(images, mask=mask)
+        # Add positional embedding before head.
+        features = self.backbone._pos_embed(features)
+        predictions = self.projection_head(features)
+
+        # Convert images to patches and normalize them.
+        patches = utils.patchify(images, self.patch_size)
+        mean = patches.mean(dim=-1, keepdim=True)
+        var = patches.var(dim=-1, keepdim=True)
+        patches = (patches - mean) / (var + 1.0e-6) ** 0.5
```
ersi-lightly, line 69: similar to before

examples/pytorch_lightning_distributed/aim.py
```diff
@@ -0,0 +1,111 @@
...
+        self.backbone = vit
+        self.projection_head = AIMPredictionHead(
+            input_dim=vit.embed_dim, output_dim=3 * self.patch_size**2, num_blocks=1
+        )
+
+        self.criterion = nn.MSELoss()
+
+    def training_step(self, batch, batch_idx):
+        images, targets = batch[0], batch[1]
+        images = images[0]  # images is a list containing only one view
```
ersi-lightly, line 52: similar to before

examples/pytorch_lightning_distributed/aim.py
```diff
@@ -0,0 +1,111 @@
...
+        patches = utils.patchify(images, self.patch_size)
+        mean = patches.mean(dim=-1, keepdim=True)
+        var = patches.var(dim=-1, keepdim=True)
+        patches = (patches - mean) / (var + 1.0e-6) ** 0.5
+
+        loss = self.criterion(predictions, patches)
+        return loss
+
+    def configure_optimizers(self):
+        optim = torch.optim.AdamW(self.parameters(), lr=1.5e-4)
```
ersi-lightly, line 75: similar to before

examples/pytorch_lightning_distributed/aim.py
```diff
@@ -0,0 +1,111 @@
...
+        return optim
+
+
+model = AIM()
```
ersi-lightly, line 79: similar to before

ersi-lightly (changes requested):

examples/pytorch/aim.py
```diff
@@ -0,0 +1,107 @@
...
+    def __init__(self, vit):
+        super().__init__()
+
+        self.patch_size = vit.patch_embed.patch_size[0]
+        self.num_patches = vit.patch_embed.num_patches
+
+        # Use absolute positional embedding.
+        pos_embed = utils.get_2d_sincos_pos_embed(
+            embed_dim=vit.embed_dim,
+            grid_size=int(self.num_patches**0.5),
```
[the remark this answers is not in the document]
guarin: I changed it to use your `utils.initialize_2d_sine_cosine_positional_embedding` function instead. Much better!

examples/pytorch/aim.py
guarin: No this is from the MAE codebase: https://github.com/facebookresearch/mae/blob/efb2a8062c206524e35e47d04501ed4f544c0ae8/models_mae.py#L206-L209

I added `utils.normalize_mean_var` and used this instead. Note that we can't use `torch.nn.functional.batch_norm` because it expects more inputs, see: https://pytorch.org/docs/stable/generated/torch.nn.functional.batch_norm.html

examples/pytorch/aim.py
guarin: Yes, maybe we can change it in the future to use something like `add_pos_embed` from the masked vision transformer.

examples/pytorch/aim.py
guarin: I would leave it as is, we have it always hardcoded in the examples.

examples/pytorch_lightning/aim.py
guarin: Good point!

examples/pytorch_lightning/aim.py
guarin: I will keep them hardcoded. The examples are optimized to be short and concise, we do the same for all the other examples.

examples/pytorch/aim.py
ersi-lightly: hahaha I feel like we keep having the same conversation everytime!

lightly/models/utils.py
```diff
@@ -638,13 +639,15 @@ def add_stochastic_depth_to_blocks(vit: Module, prob: float = 0.0, mode="row") -
             mod.mlp = Sequential(mod.mlp, StochasticDepth(p=prob, mode=mode))
 
 
-def initialize_2d_sine_cosine_positional_embedding(pos_embedding: Parameter) -> None:
+def initialize_2d_sine_cosine_positional_embedding(
+    pos_embedding: Parameter, has_class_token: bool
```
ersi-lightly, line 643: Could you make `has_class_token` optional? In the other parts of the code where we use this we do not pass a value for `has_class_token`

guarin: I updated the other parts of the code. I think it is better to not have it optional because it is easy to miss.

examples/pytorch/aim.py
ersi-lightly: great, thanks!

lightly/models/utils.py
ersi-lightly: ok! If you are sure that you have updated all affected parts, feel free to resolve.

ersi-lightly (approved):

PR #1520, opened 2024-04-04
draft of PrettyPrintApiException

MalteEbner:
## Description (version 2):

Change `ApiException` to`PrettyPrintApiException(ApiException)`, which prints it much more nicely (formatting, print relevant part in bold and red). I kept it inheriting from `ApiException` to ensure that the calling code can still catch it if needed.

## Effects
1) Shortens the stacktrace a bit 
2) Prints the ApiException more nicely.

## Stacktrace still there

We will not overwrite the `sys.excepthook`, see discussion with Guarin [here](https://lightly-ai.slack.com/archives/C01KSD8EE65/p1712741150247569?thread_ts=1712242200.360349&cid=C01KSD8EE65), as it is too hacky and might cause problems among users of the library.

## Screenshots
image
image
image

### DIFF
--- a/lightly/api/api_workflow_compute_worker.py
+++ b/lightly/api/api_workflow_compute_worker.py
@@ -250,24 +250,8 @@ class _ComputeWorkerMixin:
                 "creator": self._creator,
             }
         )
-        try:
-            response = self._compute_worker_api.create_docker_worker_config_vx(request)
-            return response.id
-        except ApiException as e:
-            if e.body is None:
-                raise e
-            eb = json.loads(e.body)
-            eb_code = eb.get("code")
-            eb_error = eb.get("error")
-            if str(e.status)[0] == "4" and eb_code is not None and eb_error is not None:
-                raise ValueError(
-                    f"Trying to schedule your job resulted in\n"
-                    f">> {eb_code}\n>> {json.dumps(eb_error, indent=4)}\n"
-                    f">> Please fix the issue mentioned above and see our docs "
-                    f"https://docs.lightly.ai/docs/all-configuration-options for more help."
-                ) from e
-            else:
-                raise e
+        response = self._compute_worker_api.create_docker_worker_config_vx(request)
+        return response.id
 
     def schedule_compute_worker_run(
         self,
--- a/lightly/api/swagger_rest_client.py
+++ b/lightly/api/swagger_rest_client.py
@@ -18,7 +18,7 @@ class PrettyPrintApiException(ApiException):
         error_message += f"Error Reason: {self.reason}\n"
         error_body_dict = json.loads(self.body)
         if "error" in error_body_dict:
-            error_message += f"Error Message: {error_body_dict['error']}\n"
+            error_message += f"Error Message: \033[1m\033[31m{error_body_dict['error']}\033[0m\n"
 
         return error_message
 
--- a/lightly/api/swagger_rest_client.py
+++ b/lightly/api/swagger_rest_client.py
@@ -1,12 +1,9 @@
 import json
+from json import JSONDecodeError
 from typing import Any, Dict, Optional, Tuple, Union
 
 from lightly.openapi_generated.swagger_client.api_client import Configuration
-from lightly.openapi_generated.swagger_client.exceptions import (
-    ApiException,
-    UnauthorizedException,
-)
-from lightly.openapi_generated.swagger_client.models.api_error_code import ApiErrorCode
+from lightly.openapi_generated.swagger_client.exceptions import ApiException
 from lightly.openapi_generated.swagger_client.rest import RESTClientObject
 
 
@@ -19,18 +16,24 @@ class PrettyPrintApiException(ApiException):
 
     def __str__(self):
         error_message = "\n"
-        error_message += "#" * 30
-        error_message += f"\nError Code: {self.status}\n"
-        error_message += f"Error Reason: {self.reason}\n"
-        if self.body is not None:
-            try:
-                error_body_dict = json.loads(self.body)
-            except JSONDecodeError:
-                pass
-            else:
-                if "error" in error_body_dict:
-                    error_message += f"Error Message: {error_body_dict['error']}\n"
-        error_message += "#" * 30 + "\n"
+        error_message += "#" * 100
+        error_message += "\n"
+        error_message += f"Error Code: {self.status}"
+        error_message += "\n"
+        error_message += f"Error Reason: {self.reason}"
+        error_message += "\n"
+        error_message += "\n"
+        try:
+            error_body_dict = json.loads(self.body)
+        except JSONDecodeError:
+            pass
+        else:
+            if "error" in error_body_dict:
+                error_message += f"Error Message: {error_body_dict['error']}"
+
+        error_message += "\n"
+
+        error_message += "#" * 100
 
         # make the error message red
         error_message = f"\033[91m{error_message}\033[0m"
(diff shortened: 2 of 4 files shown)


lightly/api/swagger_rest_client.py
```diff
@@ -1,8 +1,27 @@
...
+
+    def __init__(self, current_exception: ApiException):
+        super().__init__(current_exception.status, current_exception.reason)
+        self.body = current_exception.body
+        self.headers = current_exception.headers
+
+    def __str__(self):
+        error_message = f"\nError Code: {self.status}\n"
+        error_message += f"Error Reason: {self.reason}\n"
+        error_body_dict = json.loads(self.body)
```
japrescott, line 19: this can easily fail if its not a well formatted error code from the server as an object.

japrescott (commented):
Looks like a good first attempt which for sure enhances the readability of the errors happening. However it seems to mainly focus on removing the "double error" with improved error printing and not on actually getting rid of the stacktrace for well-known errors.

I am wondering if you would get the desired result faster (including having a debug flag etc) by using https://docs.python.org/2/library/sys.html#sys.excepthook as a general "error handler" where you handle ApiException differently if the `reason`/`error` aligns as you did in the PrettyPrintApiException?

MalteEbner:
Update after discussion with Jeremy:

- Changing an error message to not include traceback is hardly possible, as the traceback is set by the calling functions (on a higher level), not the error message itself. It is only possible by changing global functions, e.g. the https://docs.python.org/2/library/sys.html#sys.excepthook 
- Because lightly is a library that might be used by other python code (e.g. voxels integrates it into their own python pipeline), we should not change global functions or rely on them being set to something special. Examples:
  - Voxel might also want to change the excepthook or rely on certain behaviour
  - Certain python modes (ipython, jupyter notebook) also change the exception behaviour
  
  Thus the idea is to keep the direction of this draft: Make the error message formatting nicer, but don't mess with global functions and thus keep the traceback. However, we should
  - use the `PrettyPrintApiException` for all `ApiException`
  - fix the formatting for the cases that the error code and error reason are not the same
  - remove any changes in the retry.
  
  Wdyt? @guarin

lightly/api/api_workflow_compute_worker.py
```diff
@@ -250,24 +250,8 @@ def create_compute_worker_config(
...
-            eb_error = eb.get("error")
-            if str(e.status)[0] == "4" and eb_code is not None and eb_error is not None:
-                raise ValueError(
-                    f"Trying to schedule your job resulted in\n"
-                    f">> {eb_code}\n>> {json.dumps(eb_error, indent=4)}\n"
-                    f">> Please fix the issue mentioned above and see our docs "
-                    f"https://docs.lightly.ai/docs/all-configuration-options for more help."
-                ) from e
-            else:
-                raise e
```
MalteEbner, line 270: No need anymore to add a custom error message twice.

lightly/api/utils.py
```diff
@@ -55,6 +57,8 @@ def retry(func, *args, **kwargs):  # type: ignore
             # return on success
             return func(*args, **kwargs)
         except Exception as e:
+            if isinstance(e, (PrettyPrintApiException, KeyboardInterrupt)):
```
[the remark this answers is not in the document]
MalteEbner: As we discussed to not change the retry here, I switched to only check for `KeyboardInterrupt`

lightly/api/swagger_rest_client.py
MalteEbner: I put additional checks and a try-catch around it.

lightly/api/swagger_rest_client.py
```diff
@@ -61,16 +94,26 @@ def request(
...
+                body=body,
+                post_params=post_params,
+                _preload_content=_preload_content,
+                _request_timeout=_request_timeout,
+            )
+        except ApiException as e:
+            if e.reason in [
+                ApiErrorCode.BAD_REQUEST,
+                ApiErrorCode.CONFLICT,
+                ApiErrorCode.MALFORMED_RESPONSE,
```
japrescott, line 112: The reason to originally exclude these from the "prettyprint" was because we were going to eliminate the stack trace. Now that we are keeping the stracktrace but formatting it nicely, we can format it nicely for all errors codes.

lightly/api/swagger_rest_client.py
```diff
@@ -1,9 +1,42 @@
...
+class PrettyPrintApiException(ApiException):
+
+    def __init__(self, current_exception: ApiException):
+        super().__init__(current_exception.status, current_exception.reason)
+        self.body = current_exception.body
+        self.headers = current_exception.headers
+
+    def __str__(self):
+        error_message = "\n"
+        error_message += "#" * 30
```
japrescott, line 22: did you see this style somewhere or is this Malteshe schema? :) a bit extreme but at least its clear where an error starts and where it ends :)

lightly/api/swagger_rest_client.py
```diff
@@ -61,16 +94,26 @@ def request(
...
+            )
+        except ApiException as e:
+            if e.reason in [
+                ApiErrorCode.BAD_REQUEST,
+                ApiErrorCode.CONFLICT,
+                ApiErrorCode.MALFORMED_RESPONSE,
+                ApiErrorCode.MALFORMED_REQUEST,
+            ]:
+                raise e
+            raise PrettyPrintApiException(e).with_traceback(None) from None
```
guarin, line 116: Has `with_traceback` an additional benefit over `from None`?

lightly/api/api_workflow_compute_worker.py
MalteEbner: We still do and get an even better link:
image

lightly/api/swagger_rest_client.py
MalteEbner: I removed them.

lightly/api/swagger_rest_client.py
MalteEbner: Guarin had proposed it: https://lightly-ai.slack.com/archives/C01KSD8EE65/p1712324976264859?thread_ts=1712242200.360349&cid=C01KSD8EE65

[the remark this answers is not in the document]
MalteEbner: Done

lightly/api/swagger_rest_client.py
```diff
@@ -1,9 +1,42 @@
...
+        error_message += f"\nError Code: {self.status}\n"
+        error_message += f"Error Reason: {self.reason}\n"
+        if self.body is not None:
+            try:
+                error_body_dict = json.loads(self.body)
+                if "error" in error_body_dict:
+                    error_message += f"Error Message: {error_body_dict['error']}\n"
+            except json.JSONDecodeError:
+                pass
+        error_message += "#" * 30 + "\n"
```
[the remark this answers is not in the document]
MalteEbner: Done

lightly/api/swagger_rest_client.py
```diff
@@ -1,9 +1,42 @@
...
+    def __init__(self, current_exception: ApiException):
+        super().__init__(current_exception.status, current_exception.reason)
+        self.body = current_exception.body
+        self.headers = current_exception.headers
+
+    def __str__(self):
+        error_message = "\n"
+        error_message += "#" * 30
+        error_message += f"\nError Code: {self.status}\n"
+        error_message += f"Error Reason: {self.reason}\n"
```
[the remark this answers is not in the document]
MalteEbner: However, then the error message is not last. Do you really think this makes it clearer?
image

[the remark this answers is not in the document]
MalteEbner: I went with this design:
image

lightly/api/swagger_rest_client.py
MalteEbner: No, it seems to not change anything, I removed it.

lightly/api/swagger_rest_client.py
```diff
@@ -1,9 +1,46 @@
...
 
 
+class PrettyPrintApiException(ApiException):
+
+    def __init__(self, current_exception: ApiException):
+        super().__init__(current_exception.status, current_exception.reason)
+        self.body = current_exception.body
+        self.headers = current_exception.headers
+
+    def __str__(self):
```
guarin, line 17:
```suggestion
    def __str__(self) -> str:
```

guarin (approved):
LGTM

lightly/api/swagger_rest_client.py
japrescott: gracias

lightly/api/swagger_rest_client.py
japrescott: GuarAlte code :)

japrescott (approved):

PR #1537, opened 2024-05-10
Validate input and lightly mount

guarin:
### Changes

* Validate input and lightly mount

This gives the user better feedback whether the input and lightly mounts were correctly passed to lightly-serve 

### How was it tested?

* Added unit tests
* Ran manual tests


lightly/api/serve.py
```diff
@@ -50,6 +50,35 @@ def send_response_only(self, code, message=None):
     return HTTPServer((host, port), _LocalDatasourceRequestHandler)
 
 
+def validate_input_mount(input_mount: Path) -> None:
+    """Validates that the input mount is a directory and contains files."""
+    input_mount = input_mount.resolve()
+    if not input_mount.exists():
+        raise ValueError(f"Input mount '{input_mount}' does not exist.")
```
japrescott, line 57: I think it would be helpful if the error messages would contain the actual parameter used? Something like `Path for 'input_mount' argument '{input_mount}' does not exist.`? Same for lightly_mount. Wdyt? Or too much?

japrescott (commented):
looks amazing! Thanks! Just one UX question with the error msgs but already approved for you to make the decision

japrescott (approved):

PR #1540, opened 2024-05-13
Malte lig 4952 add test for all components using gatherlayer

MalteEbner:
## Tests for losses

- Moved the `test_dist__gather.py` to `test_dist__gather__losses.py`.
- Extended it to not only test `NTXentLoss`, but also `TiCoLoss`, `VICRegLoss` and `DCLLoss`
- Not included `VICRegLLoss`, as it requires the global and local features and grids as input, which does not fit into the current framework. I think the effort to also test it is not worth it for the moment. If we want to do it, we should do it in a follow-up issue.
- This covers all losses using `gather(...)`
image

Aligning the import style is another topic. Here it is a bit more difficult, as there is the name clash between `lightly.utils.dist` and `torch.dist`. Currently some files import the one and some the other as `dist`.

## Test for Benchmark module

Added `test_dist__gather__benchmark_module.py` for benchmarking the `BenchmarkModule`. Also fixed bug in the `BenchmarkModule.on_validation_epoch_start `: As the `self.dataloader_kNN` is not split by the DDP, all devices have the full dataloader and no gathering is necessary.

## Other changes

Some typing updates

### No missing test for `lightly.utils.dist.gather`

All usages of lightly.utils.dist.gather are covered now (except `VICRegLLoss`):
image
image

### Missing test for other DDP functions

We also have more gather-like functions, e.g. `concat_all_gather` in `lightly/models/utils.py`. 
image

And the kNN also use some gather functions:
image

However, these are out of scope for this issue.

### DIFF
--- a/tests/utils/test_dist__gather__losses.py
+++ b/tests/utils/test_dist__gather__losses.py
@@ -1,4 +1,4 @@
-from typing import Any, Generator, Tuple
+from typing import Any, Generator, Tuple, Type
 
 import pytest
 import pytorch_lightning as pl
@@ -147,9 +147,10 @@ class TestGatherLayer_Losses:
         ys = torch.arange(n_samples * 2 * 2).reshape(n_samples, 2, 2).float()
         dataset = TensorDataset(xs, ys)
 
+        # type ignore because 'Unexpected keyword argument "gather_distributed" for "Module"'
         model = Model(
             gather=gather,
-            criterion=loss(gather_distributed=gather),
+            criterion=loss(gather_distributed=gather),  # type: ignore[call-arg]
             learning_rate=learning_rate,
         )
 
--- a/tests/utils/test_dist__gather__losses.py
+++ b/tests/utils/test_dist__gather__losses.py
@@ -85,7 +85,7 @@ class TestGatherLayer_Losses:
     def test_loss_ntxent(self, close_torch_distributed: None) -> None:
         self._test_ddp(
             NTXentLoss,
-            torch.Tensor(
+            torch.tensor(
                 [
                     [0.1675, 0.2676, 0.3678, 0.4679, 0.5680],
                     [0.5763, 0.6742, 0.7721, 0.8700, 0.9679],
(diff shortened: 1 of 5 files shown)


tests/utils/test_dist__gather__losses.py
```diff
@@ -0,0 +1,173 @@
...
+    - Any randomness in the transform, as the order the samples are processed and
+    thus the random seed differ.
+    - Any batch normalization in the model, as the batch size is split between
+    the devices and there is no gather layer for the batch normalization.
+    """
+
+    def test_loss_ntxent(self, close_torch_distributed: None) -> None:
+        self._test_ddp(
+            NTXentLoss,
+            torch.Tensor(
```
guarin, line 88:
```suggestion
            torch.tensor(
```

`torch.Tensor` to create a new tensor is deprecated

tests/utils/test_dist__gather__losses.py
```diff
@@ -0,0 +1,173 @@
...
+                    [0.1484, 0.2335, 0.3185, 0.4036, 0.4886],
+                    [0.6198, 0.7093, 0.7988, 0.8884, 0.9779],
+                ]
+            ),
+            learning_rate=0.1,
+        )
+
+    def _test_ddp(
+        self,
+        loss: Module,
```
guarin, line 135:
```suggestion
        loss: Type[Module],
```

Not sure how that passed mypy

tests/utils/test_dist__gather__losses.py
```diff
@@ -0,0 +1,173 @@
...
+        ys = torch.arange(n_samples * 2 * 2).reshape(n_samples, 2, 2).float()
+        dataset = TensorDataset(xs, ys)
+
+        model = Model(
+            gather=gather,
+            criterion=loss(gather_distributed=gather),
+            learning_rate=learning_rate,
+        )
+
+        dataloader = torch.utils.data.DataLoader(
```
guarin, line 156:
```suggestion
        dataloader = DataLoader(
```

guarin (approved):
Left some comments regarding formatting and imports but the code looks great!

PR #1545, opened 2024-05-17
Make codecov only run on Python3.10

IgorSusmelj:
## Changes

- Updates CI to only run codecov part for Python 3.10

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,7 @@ jobs:
         if: matrix.python == '3.10'
         run: |
           export LIGHTLY_SERVER_LOCATION="localhost:-1"
-          pip install pytest-cov==4.1.0
+          pip install pytest-cov==5.0.0
           python -m pytest -s -v --runslow --cov=./lightly --cov-report=xml --ignore=./lightly/openapi_generated/
       - name: Upload coverage to Codecov
         if: matrix.python == '3.10'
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -53,15 +53,23 @@ jobs:
       - name: Install Dependencies and lightly
         run: pip install -e '.[all]'
       - name: Clean Previous Coverage Data
+        if: matrix.python == '3.10'
```
MalteEbner, line 56: I am not sure about this. Why not alway remove it?

MalteEbner (approved):
One question, otherwise LGTM^2

tests/utils/test_dist__gather__benchmark_module.py
```diff
@@ -35,6 +35,7 @@ def close_torch_distributed() -> Generator[None, None, None]:
     torch.distributed.destroy_process_group()
 
 
+@pytest.mark.skip(reason="This test is running in parallel")
```
guarin, line 38:
```suggestion
@pytest.mark.skipif(os.getenv("GITHUB_ACTIONS"), reason="This test is running in parallel and breaks codecov")
```

tests/utils/test_dist__gather__losses.py
```diff
@@ -61,6 +61,7 @@ def close_torch_distributed() -> Generator[None, None, None]:
     torch.distributed.destroy_process_group()
 
 
+@pytest.mark.skip(reason="This test is running in parallel")
```
guarin, line 64: Same as above

guarin (approved):

.github/workflows/test.yml
IgorSusmelj: Removed it

PR #1552, opened 2024-06-03
Lightly server https

japrescott:
closes lig-5082
- allow to pass ssl certs to lightly-serve to start it in HTTPS mode

### DIFF
--- a/lightly/cli/serve_cli.py
+++ b/lightly/cli/serve_cli.py
@@ -1,5 +1,5 @@
-import sys
 import ssl
+import sys
 from pathlib import Path
 
 import hydra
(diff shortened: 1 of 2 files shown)


lightly/cli/serve_cli.py
```diff
@@ -19,9 +20,13 @@ def lightly_serve(cfg):
         lightly_mount:
             Path to the Lightly directory.
         host:
-            Hostname for serving the data (defaults to localhost).
+            Hostname for serving the data (defaults to localhost). If you want to expose this to the world, use '0.0.0.0'.
```
MalteEbner, line 23: [Similar to Ersi](https://lightly-ai.slack.com/archives/C01KSD8EE65/p1717435699756949?thread_ts=1717432537.232419&cid=C01KSD8EE65), I don't get what "exposing to the world" means. When would I use this more specifically? Exposing to the world sounds very bad from a security perspective, when is it ok?
```suggestion
            Hostname for serving the data (defaults to localhost). Set it to '0.0.0.0' to ...
```

MalteEbner (approved):
Apart from the docstring, I approve.

lightly/cli/serve_cli.py
japrescott: It **is** "bad" from a security perspective if you don't have any other security measures in place, yes. E.g if you allow other people to access the same network as you or you are in a public WLAN, other people **will** be able to see your server.

japrescott: I changed it to `If you want to expose it to the internet or your local network, use '0.0.0.0'.`. Does that explain it better to you Malte?

MalteEbner: Alternatively, don't explain it, but link to the new docs.

PR #1555, opened 2024-06-08
Torchvision transforms v2

### LINKED ISSUE #1547: Can we switch to torchvision v2 transforms?
My understanding is that lightly uses the "old" v1 torchvision transforms, see e.g. https://github.com/lightly-ai/lightly/blob/8878b5b5e816bbd1f11b5c0a866eaee45a9a6004/lightly/transforms/dino_transform.py#L4

The v1 transforms have some drawbacks:
- only work on PIL images, not on torch tensors (which are the output of `litdata.StreamingDataset`)
- don't support segmentation datasets, see https://pytorch.org/blog/extending-torchvisions-transforms-to-object-detection-segmentation-and-video-tasks/#the-new-transforms-api

Would you accept a PR where we add the v2 versions? My understanding is that it's a drop-in replacement and that everything *should* stay the same, but I'd double check.

siemdejong:
Resolves #1547

It deviates a little from the proposed method by @guarin in that lightly makes the importer think `torchvision.transforms` is `torchvision.transforms.v2`. I chose for this option as this would only require a change in `lightly.transforms.__init__.py`.

### DIFF
--- a/lightly/transforms/__init__.py
+++ b/lightly/transforms/__init__.py
@@ -8,6 +8,10 @@ transforms.
 # Copyright (c) 2020. Lightly AG and its affiliates.
 # All Rights Reserved
 
+import importlib
+import logging
+import sys
+
 from lightly.transforms.aim_transform import AIMTransform
 from lightly.transforms.byol_transform import (
     BYOLTransform,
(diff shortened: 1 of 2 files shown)


guarin:
Thanks a lot for your PR!

I think we shouldn't modify `sys.modules` as this also impacts code outside of lightly. Users might get really surprised if `torchvision.transforms` is suddenly `v2` instead of `v1` because they imported lightly. I would prefer the approach where we define a new `torchvision_transforms` variable. I would even put it in a new file in `lightly/transforms/torchvision_transforms.py`.

I also noticed that the v2 `ToTensor` transform is deprecated and shows a warning when used. We should replace it with a `ToTensor` transform that doesn't show the warning. Something like this should work:

```
from typing import Callable
import torch
from torch import Tensor
from PIL.Image import Image as PILImage

try:
    from torchvision.transforms import v2 as torchvision_transforms
    _TRANSFORMS_V2 = True
except ImportError:
    from torchvision import transforms as torchvision_transforms
    _TRANSFORMS_V2 = False


def ToTensor() -> Callable[[PILImage], Tensor]:
    T = torchvision_transforms
    if _TRANSFORMS_V2:
        # v2.transforms.ToTensor is deprecated and will be removed in the future.
        # This is the new recommended way to convert a PIL Image to a tensor:
        return T.Compose([T.ToImage(), T.ToDtype(dtype=torch.float32, scale=True)])
    else:
        return T.ToTensor()
```

PR #1568, opened 2024-07-01
feat: add `from_hf_dataset` method

SauravMaheshkar:
Adds a new helper method to create a LightlyDataset from a HuggingFace Dataset.

### DIFF
--- a/lightly/data/dataset.py
+++ b/lightly/data/dataset.py
@@ -216,6 +216,34 @@ class LightlyDataset:
         dataset_obj.dataset = dataset
         return dataset_obj
 
+    @classmethod
+    def from_hf_dataset(cls, dataset, transform=None, key: str):
+        """Builds a LightlyDataset from a HuggingFace dataset.
+
+        Args:
+            dataset:
+                HuggingFace dataset
+            transform:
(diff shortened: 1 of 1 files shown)


lightly/data/dataset.py
```diff
@@ -216,6 +216,34 @@ def from_torch_dataset(cls, dataset, transform=None, index_to_filename=None):
...
+        Returns:
+            A LightlyDataset object.
+
+        """ 
+        def apply_transform(batch, transform=tranform, key=key):
+            assert key in batch.keys(), f"the provided key, {key} does not exist in the dataset"
+
+            batch[key] = [transform(image) for image in batch[key]]
+
+            return batch
```
guarin, line 240: Not sure if this works with distributed training as local functions cannot be pickled and sent to other processes. We might have to define a transform class that takes the transform and key as input.

guarin:
@SauravMaheshkar is it ok if we close this as we now have the new tutorial from #1569 ?

SauravMaheshkar:
> @SauravMaheshkar is it ok if we close this as we now have the new tutorial from #1569 ?

Yup closed 😄

PR #1572, opened 2024-07-05
Fix Install API Client Dependencies Only

guarin:
## Changes
* Add `requirements/base.txt` and `requirements/openapi.txt` again
* Add GitHub Action to verify that package can be installed with API dependencies only
* Add tests to verify that `requirements/base.txt` and `requirements/openapi.txt` are in sync with `pyproject.toml`
* Add `toml` and `types-toml` dev dependencies

See `requirements/README.md` why the changes are needed.
`toml` and `types-toml` are necessary because the official `tomllib` standard library is only available from Python 3.11 on.

## How was it tested?
* Added tests and action

### DIFF
--- a/.github/workflows/test_api_deps_only.yml
+++ b/.github/workflows/test_api_deps_only.yml
@@ -1,0 +1,49 @@
+
+name: Install API Dependencies Only
+
+on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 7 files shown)


MalteEbner (approved):
LGTM

PR #1580, opened 2024-07-11
use latest gh actions

japrescott:
upgrade the github workflow actions to use latest versions and thus, supported environments

### DIFF
--- a/.github/workflows/release_pypi.yml
+++ b/.github/workflows/release_pypi.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout code
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         fetch-depth: 0
     - name: Checkout latest release tag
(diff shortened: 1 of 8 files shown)


SauravMaheshkar (commented):
Perhaps we should also look into adding a [dependabot](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide) configuration to the repository?

This would check for the latest available versions as specified and make PRs accordingly. 

An example file could be as follows `.github/dependabot.yml`

```yaml
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "daily"
```

japrescott:
Thanks @SauravMaheshkar for the suggestion. I agree that this could be helpful but am afraid to just activate it without careful consideration. Will add this to our backlog for further evaluation 👍

guarin (approved):
LGTM!

SauravMaheshkar (approved):
LGTM 🙂

Issue #1511, opened 2024-03-04
Allow skipping weight initialisation for Masked Vision Transformers.


ersi-lightly:
Allow skipping weight initializsation for `MaskedVisionTransformerTIMM`, `MaskedVisionTransformerTorchvision` and `MAEDecoderTIMM`. This will enable to  load a pretrained vit model.

PR #1581, opened 2024-07-11
LIG-5222 Allow skipping weight initialisation for Masked Vision Transformers

ersi-lightly:
Addresses issue #1511 and allows to skip weight initialization for  `MaskedVisionTransformerTIMM`, `MaskedVisionTransformerTorchvision` and `MAEDecoderTIMM`. This will enable to load a pretrained vit model.

### DIFF
--- a/lightly/models/modules/masked_autoencoder_timm.py
+++ b/lightly/models/modules/masked_autoencoder_timm.py
@@ -62,7 +62,7 @@ class MAEDecoderTIMM(Module):
         proj_drop_rate: float = 0.0,
         attn_drop_rate: float = 0.0,
         norm_layer: Callable[..., nn.Module] = partial(LayerNorm, eps=1e-6),
-        initialize_weights: bool = False,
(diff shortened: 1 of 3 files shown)


lightly/models/modules/masked_autoencoder_timm.py
```diff
@@ -60,6 +62,7 @@ def __init__(
         proj_drop_rate: float = 0.0,
         attn_drop_rate: float = 0.0,
         norm_layer: Callable[..., nn.Module] = partial(LayerNorm, eps=1e-6),
+        initialize_weights: bool = False,
```
guarin, line 65:
```suggestion
        initialize_weights: bool = True,
```
I think this should be True by default, otherwise the existing tutorials/examples will break as weights will no longer be correctly initialized.

guarin (changes requested):

guarin (approved):
LGTM!

PR #1592, opened 2024-07-22
Add uniform drop path rate for timm ViT

guarin:
## Changes

* Add function to set uniform drop path rates for TIMM ViT

This corresponds to: https://github.com/facebookresearch/dinov2/blob/e1277af2ba9496fbadf7aec6eba56e8d882d1e35/dinov2/models/vision_transformer.py#L116-L119

We'll use it like this:
```
vit = timm.VisionTransformer()
update_drop_path_rate(vit, drop_path_rate=0.1, mode="uniform")
masked_vit = MaskedVisionTransformerTIMM(vit)
```

Maybe I'll also add it as an extra argument to `MaskedVisionTransformerTIMM` instead, so it can be used like:
```
vit = timm.VisionTransformer()
masked_vit = MaskedVisionTransformerTIMM(vit, drop_path_rate=0.1, drop_path_mode="uniform")
```

## How was it tested?
* Added unit tests

### DIFF
--- a/lightly/models/utils.py
+++ b/lightly/models/utils.py
@@ -6,17 +6,7 @@ from __future__ import annotations
 
 import math
 import warnings
-from typing import (
-    TYPE_CHECKING,
-    Dict,
-    Iterable,
-    List,
-    Literal,
-    Optional,
-    Tuple,
-    Type,
-    Union,
-)
+from typing import TYPE_CHECKING, Dict, Iterable, List, Optional, Tuple, Type, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


lightly/models/utils.py
```diff
@@ -848,3 +852,41 @@ def normalize_mean_var(x: Tensor, dim: int = -1, eps: float = 1.0e-6) -> Tensor:
     mean = x.mean(dim=dim, keepdim=True)
     var = x.var(dim=dim, keepdim=True)
     return (x - mean) / (var + eps).sqrt()
+
+
+def update_drop_path_rate(
+    model: VisionTransformer,
+    drop_path_rate: float,
+    mode: str = "linear",
+) -> None:
```
MalteEbner, line 861: Should this be in the general `utils.py` if it only works for timm models?

MalteEbner (approved):
LGTM

PR #1594, opened 2024-07-22
Add timm optional dependency

guarin:
## Changes

* Add timm as optional dependency to pyproject.toml
* Increase timm version from 0.9.9 to 1.0.3
  * 0.9.9 is still supported but now that timm 1.0 is out we should encourage installing a >1 version

## How was it tested?

* Ran tests and type-check locally with and without timm

### DIFF
--- a/tests/models/modules/test_masked_autoencoder_timm.py
+++ b/tests/models/modules/test_masked_autoencoder_timm.py
@@ -1,108 +1,115 @@
 import unittest
 
+import pytest
 import torch
 
 from lightly.models import utils
 from lightly.utils import dependency
 
-if dependency.timm_vit_available():
-    from timm.models.vision_transformer import VisionTransformer, vit_base_patch32_224
+if not dependency.timm_vit_available():
+    # We do not use pytest.importorskip on module level because it makes mypy unhappy.
+    pytest.skip("TIMM vision transformer is not available", allow_module_level=True)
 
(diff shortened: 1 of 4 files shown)


tests/models/modules/test_masked_autoencoder_timm.py
```diff
@@ -1,108 +1,115 @@
...
+import pytest
 import torch
 
 from lightly.models import utils
 from lightly.utils import dependency
 
-if dependency.timm_vit_available():
-    from timm.models.vision_transformer import VisionTransformer, vit_base_patch32_224
-
-    from lightly.models.modules import MAEDecoderTIMM, MaskedVisionTransformerTIMM
```
guarin, line 11: The only change I did was replace these lines:
```
if dependency.timm_vit_available():
    from timm.models.vision_transformer import VisionTransformer, vit_base_patch32_224

    from lightly.models.modules import MAEDecoderTIMM, MaskedVisionTransformerTIMM
```

with
```
if not dependency.timm_vit_available():
    # We do not use pytest.importorskip on module level because it makes mypy unhappy.
    pytest.skip("TIMM vision transformer is not available", allow_module_level=True)


from timm.models.vision_transformer import VisionTransformer, vit_base_patch32_224

from lightly.models.modules import MAEDecoderTIMM, MaskedVisionTransformerTIMM
```

MalteEbner (approved):
LGTM

PR #1595, opened 2024-07-22
Add antialias option to masked vit

guarin:
## Changes

* Add antialias option for `MaskedVisionTransformer` classes
* Rename `initialize_weights` to `weight_initialization`
  * This makes it a bit more obvious what the parameter does and allows for different schemes depending on the model as is supported in TIMM: https://github.com/huggingface/pytorch-image-models/blob/a1996ec0f43dc7fcc053de7048be6a51cff2ee08/timm/models/vision_transformer.py#L450
  * I changed this as cleanup as I was anyways working on this class. Note that this is not a breaking change as the feature has not yet been released.

This allows disabling antialiasing as done in Dinov2: https://github.com/facebookresearch/dinov2/blob/e1277af2ba9496fbadf7aec6eba56e8d882d1e35/dinov2/models/vision_transformer.py#L206

### DIFF
--- a/lightly/models/modules/masked_vision_transformer_timm.py
+++ b/lightly/models/modules/masked_vision_transformer_timm.py
@@ -21,8 +21,8 @@ class MaskedVisionTransformerTIMM(MaskedVisionTransformer, Module):
         mask_token:
             The mask token.
         weight_initialization:
-            The weight initialization method. Valid options are ['', 'skip']. 'skip'
-            skips the weight initialization and '' uses the default weight
+            The weight initialization method. Valid options are ['', 'skip']. '' uses
+            the default MAE weight initialization and 'skip' skips the weight
             initialization.
         antialias:
             Whether to use antialiasing when resampling the positional embeddings.
--- a/lightly/models/modules/masked_vision_transformer_timm.py
+++ b/lightly/models/modules/masked_vision_transformer_timm.py
@@ -33,8 +33,8 @@ class MaskedVisionTransformerTIMM(MaskedVisionTransformer):
         self,
         vit: VisionTransformer,
         mask_token: Optional[Parameter] = None,
-        antialias: bool = True,
         weight_initialization: str = "",
+        antialias: bool = True,
     ) -> None:
         super().__init__()
         self.vit = vit
--- a/lightly/models/modules/masked_vision_transformer_timm.py
+++ b/lightly/models/modules/masked_vision_transformer_timm.py
@@ -1,5 +1,5 @@
 import math
-from typing import Optional
+from typing import List, Optional, Tuple
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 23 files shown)


lightly/models/modules/masked_vision_transformer_timm.py
```diff
@@ -20,14 +20,21 @@ class MaskedVisionTransformerTIMM(MaskedVisionTransformer, Module):
             The VisionTransformer object of TIMM.
         mask_token:
             The mask token.
+        weight_initialization:
+            The weight initialization method. Valid options are ['', 'skip']. 'skip'
+            skips the weight initialization and '' uses the default weight
```
MalteEbner, line 25: I would use the same order
```suggestion
            The weight initialization method. Valid options are ['skip', '']. 'skip'
            skips the weight initialization and '' uses the default weight.
```

Furthermore, why not use an enum?

MalteEbner (commented):
Can you also add tests? Codecov complains about uncovered lines and I think it complains correctly.

lightly/models/modules/masked_vision_transformer_timm.py
guarin: Updated :) I would like to keep the str instead of enum to be consistent with TIMM and the rest of the library. Enum makes it just a little bit harder to use because you have to remember to import it from the correct place etc. for very little gain.

MalteEbner (approved):
Looks great, thanks a lot for adding many tests!

tests/models/modules/masked_vision_transformer_test.py
```diff
@@ -0,0 +1,238 @@
...
+import torch
+from pytest_mock import MockerFixture
+from torch.nn import Parameter
+
+from lightly.models import utils
+from lightly.models.modules.masked_vision_transformer import MaskedVisionTransformer
+
+
+class MaskedVisionTransformerTest(ABC):
+    """Common tests for all classes implementing MaskedVisionTransformer."""
```
guarin, line 14: Tests mostly moved from `TestMaskedVisionTransformerTIMM`

lightly/models/modules/masked_vision_transformer_torchvision.py
```diff
@@ -35,9 +41,16 @@ def __init__(
             if mask_token is not None
             else Parameter(torch.zeros(1, 1, self.vit.hidden_dim))
         )
-        if initialize_weights:
+        if weight_initialization not in ("", "skip"):
+            raise ValueError(
```
MalteEbner, line 45: Optional: add a test asserting this `ValueError`.
Feel free to resolve.

lightly/models/modules/masked_vision_transformer_timm.py
```diff
@@ -36,9 +43,17 @@ def __init__(
             if mask_token is not None
             else Parameter(torch.zeros(1, 1, self.vit.embed_dim))
         )
-        if initialize_weights:
+
+        if weight_initialization not in ("", "skip"):
+            raise ValueError(
```
MalteEbner, line 48: Optional: add a test asserting this `ValueError`.
Feel free to resolve.

MalteEbner (approved):
LGTM, Great that this is now tested much better :)

PR #1598, opened 2024-07-23
chore(pre-commit): add linters in pre-commit

SauravMaheshkar:
Adds `black`, `isort` and `mypy` to pre-commit (with the versions specified in `pyproject.toml`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,6 +82,7 @@ dev = [
   "toml",
   "torchmetrics",
   "lightning-bolts", # for LARS optimizer
+  # should be the same version as defined in .pre-commit-config.yaml
   "black==23.1.0", # frozen version to avoid differences between CI and local dev machines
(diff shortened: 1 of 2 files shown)


MalteEbner (commented):
I would add a comment both here and at https://github.com/lightly-ai/lightly/blob/309ec17a76a1adf15c2a8e759904b25dc1dda1e0/pyproject.toml#L85-L87
that the versions for the 3 packages should align.

pyproject.toml
```diff
@@ -82,6 +82,7 @@ dev = [
   "toml",
   "torchmetrics",
   "lightning-bolts", # for LARS optimizer
+  # should be the same version as defined in .pre-commit-config.yaml
```
MalteEbner, line 85: Optional:
```suggestion
  # black, isort and mypy should be the same version as defined in .pre-commit-config.yaml
```

MalteEbner (approved):
LGTM

PR #1601, opened 2024-07-24
Add block masking

guarin:
## Changes

* Add block masking based on DINOv2 and iBOT

The implementation follows the referenced code. I did some minor modifications:
* Rename variables
* Add value checks
* Simplify some code

## How was it tested?
* Added unit tests


lightly/models/utils.py
```diff
@@ -569,6 +570,176 @@ def random_prefix_mask(
...
+        )
+
+    B, H, W = size
+    num_images_masked = int(B * mask_ratio)
+    probs = torch.linspace(
+        min_image_mask_ratio, max_image_mask_ratio, num_images_masked + 1
+    ).tolist()
+    masks = []
+    for prob_min, prob_max in zip(probs[:-1], probs[1:]):
+        num_mask = int(H * W * random.uniform(prob_min, prob_max))
```
MalteEbner, line 643: Just to get this correctly: There is an increasing number of masked blocks per image?

MalteEbner, line 643: I would make it clearer over what the masks are:
```suggestion
    masks_over_images = []
    # Iterate over images to be masked.
    for prob_min, prob_max in zip(probs[:-1], probs[1:]):
        num_masked_blocks = int(H * W * random.uniform(prob_min, prob_max))
```

lightly/models/utils.py
```diff
@@ -569,6 +570,176 @@ def random_prefix_mask(
...
+        for _ in range(max_attempts_per_block):
+            target_area = random.uniform(min_num_mask_per_block, max_new_masked)
+            aspect_ratio = math.exp(random.uniform(log_min_aspect, log_max_aspect))
+            h = int(round(math.sqrt(target_area * aspect_ratio)))
+            w = int(round(math.sqrt(target_area / aspect_ratio)))
+            if w < W and h < H:
+                top = random.randint(0, H - h)
+                left = random.randint(0, W - w)
+                num_masked = mask[top : top + h, left : left + w].sum().item()
+                new_masked = h * w - num_masked
```
MalteEbner, line 730: With type hints and other variable names, this could be clearer.
```suggestion
                num_already_masked: int = mask[top : top + h, left : left + w].sum().item()
               num_newly_masked: int = h * w - num_already_masked
```

lightly/models/utils.py
```diff
@@ -569,6 +570,176 @@ def random_prefix_mask(
...
+                max_num_mask_per_block=max_num_mask_per_block,
+                min_aspect=min_aspect,
+                max_aspect=max_aspect,
+                max_attempts_per_block=max_attempts_per_block,
+                device=device,
+            )
+        )
+    for _ in range(num_images_masked, B):
+        masks.append(torch.zeros((H, W), dtype=torch.bool, device=device))
+    # random.shuffle(masks)
```
MalteEbner, line 658: Why is this commented out?

MalteEbner (commented):
Could you call `random_block_mask` on e.g. a batch of 16 images, get the masked and visualise them, e.g. with matplotlib? I would do a quick visual test to see if the algorithm actually works, especially as the computational part is not checked by unittests.

lightly/models/utils.py
guarin: Yep, but the final masks are shuffled again so within a batch there are images with different masking ratios but in a random order.

lightly/models/utils.py
guarin: Fixed

lightly/models/utils.py
guarin: Didn't add the type hints as it makes it more verbose and IMO should be sufficiently clear that `sum().item()` returns an int.

guarin:
> Could you call random_block_mask on e.g. a batch of 16 images, get the masked and visualise them, e.g. with matplotlib? I would do a quick visual test to see if the algorithm actually works, especially as the computational part is not checked by unittests.

Example masks generated with:
```
utils.random_block_mask(size=(16, 20, 20))
```

# Batch 1
mask2

# Batch 2
mask1

# Batch 3
mask

MalteEbner (approved):

Issue #1470, opened 2024-01-11
Add TiCoTransform


guarin:
TiCo uses the BYOL transform for image augmentations. To make the method easier to use and more consistent with other methods, we should add a `TiCoTransform` that is a subclass of `BYOLTransform` and just add a new docstring for TiCo.

TODO

- [x] Add `TiCoTransform` as subclass of `BYOLTransform`
- [x] Update TiCo examples and benchmark code to use `TiCoTransform` instead of `BYOLTransform`

PR #1603, opened 2024-07-24
Add TiCoTransform

radiradev:
# Changes 

Adresses #1470. Created a separate file with `TiCoTransform` that simply inherits the `BYOLTransform` class and slightly updated the docstring. Also created corresponding`TiCoView...` classes.

I have updated the TiCo example to use this transform.


guarin (approved):
Thanks a lot for the PR! 

Closes #1470

PR #1608, opened 2024-07-25
Add KoLeoLoss

guarin:
## Changes

* Add KoLeoLoss

## How was it tested?

* Added unit test

### DIFF
--- a/tests/loss/test_koleo_loss.py
+++ b/tests/loss/test_koleo_loss.py
@@ -1,3 +1,5 @@
+import math
+
 import pytest
 import torch
 import torch.nn as nn
@@ -14,7 +16,8 @@ class TestKoLeoLoss:
         "x, expected_loss",
         [
             (torch.tensor([[1.0]]), 17.7275),
-            (torch.tensor([[1.0, 1.0]]), 17.5393),
+            (torch.tensor([[0.0, 1.0], [0.0, -1.0]]), -math.log(2)),
+            (torch.tensor([[0.0, 1.0], [1.0, 0.0]]), -math.log(2**0.5)),
             (
                 torch.tensor(
                     [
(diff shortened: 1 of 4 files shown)


tests/loss/test_koleo_loss.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class TestKoLeoLoss:
+    # Test values generated using the original implementation from:
+    # https://github.com/facebookresearch/dinov2/blob/main/dinov2/loss/koleo_loss.py
+    @pytest.mark.parametrize(
+        "x, expected_loss",
+        [
+            (torch.tensor([[1.0]]), 17.7275),
+            (torch.tensor([[1.0, 1.0]]), 17.5393),
```
MalteEbner, line 17: These two don't make sense, as they only calculate `-ln(eps + float_inaccuracies)`. I would only keep one of them, to ensure that a batch size of 1 does not break.

However, can you add these?
- [[0.0], [1.0]] -> should give mean([-ln(1), -ln(1)])
- [[0.0], [1.0], [2.0] -> should give mean([-ln(1), -ln(2)])

MalteEbner (commented):
Can you add a bit more tests? Ideally, they should not only test equality with the original implementation, but also against known easy values.

Otherwise looks good.

tests/loss/test_koleo_loss.py
guarin: Added more tests :)

Note that tests like this don't really make sense: `[[0.0], [1.0], [2.0]]`. The reason is that the loss first normalizes the vectors (`[0.0]`, `[1.0]`, `[2.0]`) to unit norm. So you'll always end up with a `[1.0]` vector (except for 0.0 which remains 0.0).

MalteEbner (approved):

PR #1609, opened 2024-07-25
Allow learnable positional embeddings in MaskedVisionTransformer

guarin:
## Changes

* Allow learnable positional embeddings in MaskedVisionTransformer

## How was it tested?
* Added unit tests

### DIFF
--- a/lightly/models/utils.py
+++ b/lightly/models/utils.py
@@ -371,7 +371,8 @@ def set_at_index(
 def mask_at_index(
     tokens: torch.Tensor, index: torch.Tensor, mask_token: torch.Tensor
 ) -> torch.Tensor:
-    """Copies mask token into the input tensor at the given indices.
+    """Returns a tensor where the tokens at the given indices are replaced by the
+    mask token.
 
     Args:
         tokens:
@@ -388,6 +389,27 @@ def mask_at_index(
     """
     mask = tokens.new_zeros(tokens.shape)
     mask = set_at_index(mask, index, 1)
+    return (1 - mask) * tokens + mask * mask_token
+
+
+def mask_bool(tokens: Tensor, mask: Tensor, mask_token: Tensor) -> Tensor:
+    """Returns a tensor with tokens replaced by the mask tokens in all positions where
+    the mask is True.
+
+    Args:
+        tokens:
+            Tokens tensor with shape (batch_size, sequence_length, dim).
+        mask:
+            Boolean mask tensor with shape (batch_size, sequence_length).
+        mask_token:
+            Mask token with shape (1, 1, dim).
+
+    Returns:
+        Tokens tensor with shape (batch_size, sequence_length, dim) where tokens[i, j]
+        is replaced by the mask token if mask[i, j] is True.
+    """
+    # Convert to int for multiplication.
+    mask = mask.unsqueeze(-1).to(torch.bool).to(torch.int)
     return (1 - mask) * tokens + mask * mask_token
 
 
--- a/lightly/models/utils.py
+++ b/lightly/models/utils.py
@@ -783,6 +783,10 @@ def initialize_positional_embedding(
 def initialize_learnable_positional_embedding(pos_embedding: Parameter) -> None:
     """Initializes a learnable positional embedding.
 
+    Uses standard initialization for ViT models, see [0].
+
+    - [0]: https://github.com/huggingface/pytorch-image-models/blob/cec70b6779ea81cec0ca08ee4a257b52affd235a/timm/models/vision_transformer.py#L590
+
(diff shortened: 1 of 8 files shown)


tests/models/test_ModelUtils.py
```diff
@@ -573,6 +574,35 @@ def test_get_named_leaf_modules() -> None:
...
+    if expected_fn is not None:
+        mock_fn.assert_called_once()
+
+
+def test_initialize_learnable_positional_embedding() -> None:
+    pos_embedding = Parameter(torch.ones(1, 1, 64) * 10)
+    utils.initialize_learnable_positional_embedding(pos_embedding)
+    assert pos_embedding.requires_grad
+    # Trunc normal should clip values to small size.
+    assert torch.all(pos_embedding.abs() < 5)
```
MalteEbner, line 603: Why not assert that [init.trunc_normal](https://pytorch.org/docs/stable/nn.init.html#torch.nn.init.trunc_normal_) was called?

If you want to keep it like this, I would rather assert that all are <=2, as this is the default value.

MalteEbner (commented):
Looks good, but I have a few questions.

lightly/models/utils.py
```diff
@@ -742,6 +742,55 @@ def add_stochastic_depth_to_blocks(vit: Module, prob: float = 0.0, mode="row") -
...
+
+
+def initialize_learnable_positional_embedding(pos_embedding: Parameter) -> None:
+    """Initializes a learnable positional embedding.
+
+    Args:
+        pos_embedding:
+            Positional embedding parameter.
+    """
+    init.trunc_normal_(pos_embedding, std=0.02)
```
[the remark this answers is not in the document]
guarin: It is standard for ViT: https://github.com/huggingface/pytorch-image-models/blob/cec70b6779ea81cec0ca08ee4a257b52affd235a/timm/models/vision_transformer.py#L590

lightly/models/utils.py
```diff
@@ -742,6 +742,55 @@ def add_stochastic_depth_to_blocks(vit: Module, prob: float = 0.0, mode="row") -
...
+    """Initializes the positional embedding with the given strategy.
+
+    Args:
+        pos_embedding:
+            Positional embedding parameter.
+        strategy:
+            Positional embedding initialization strategy. Valid options are:
+            ['learn', 'sincos', 'skip']. 'learn' makes the embedding learnable,
+            'sincos' creates a fixed 2D sine-cosine positional embedding, and 'skip'
+            does not initialize the positional embedding.
```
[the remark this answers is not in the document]
guarin: I would keep strategy required as it doesn't make sense to call the function without a strategy. I would also like to keep `'skip'` to be consistent with weight initialization where there is also a skip option.

tests/models/test_ModelUtils.py
guarin: I want to avoid mocking if not necessary. I changed now the test to just check that the embedding was updated, this avoids relying on implementation details of how the embedding is initialized.

lightly/models/utils.py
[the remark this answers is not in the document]
MalteEbner: I would add a comment that this the standard init for ViT. Optional: also add the link.

tests/models/test_ModelUtils.py
```diff
@@ -573,6 +574,37 @@ def test_get_named_leaf_modules() -> None:
...
+
+
+def test_initialize_learnable_positional_embedding() -> None:
+    pos_embedding = Parameter(torch.ones(1, 1, 64))
+    orig_pos_embedding = pos_embedding.clone()
+    utils.initialize_learnable_positional_embedding(pos_embedding)
+    # Embedding must be learnable.
+    assert pos_embedding.requires_grad
+    # Embedding must have changed.
+    assert torch.any(pos_embedding != orig_pos_embedding)
```
MalteEbner, line 605: This could fail randomly if the rng only gives out 0. Thus I would seed it.
Furthermore, I would rather check that e.g. at least 90% have changed (given a fixed seed).

MalteEbner (commented):
Approved

tests/models/test_ModelUtils.py
guarin: It would only fail if all values were randomly 0 which should be impossible. Let's keep the test as is.

MalteEbner (approved):

tests/models/test_ModelUtils.py
MalteEbner: Ok, but if it fails because all values are randomly 0, you're gonna hear a big "I told you so!" ;)

Feel free to resolve.

PR #1611, opened 2024-07-26
Remove lightning-bolts dependency

guarin:
## Changes

* Remove lightning-bolts dependency

lightning-bolts has a dependency on pytorch lightning <2 (see [here](https://github.com/Lightning-Universe/lightning-bolts/blob/2c4602aa684e7b90e7ffdcea1d3f93a20f9c2ead/requirements/base.txt#L2)) which blocks #1607


MalteEbner (approved):
Approved

Issue #1367, opened 2023-08-25
Refactor I-JEPA to use timm ViT


guarin:
As we want to use timm ViT backbones for transformer models we should refactor the I-JEPA code to use a timm ViT backbone. Ideally, the backbone can be changed easily and replaced with another ViT implementation.

### Todo

- [x] Refactor I-JEPA to use timm ViT backbone
- [x] Make sure all timm imports are optional and do not break the package. We want to keep timm dependency optional, see #1301

guarin:
Completed in https://github.com/lightly-ai/lightly/pull/1612

PR #1612, opened 2024-07-26
Refactor IJEPA to use timm. 

radiradev:
#Changes 
This PR adresses #1367. I have refactored IJEPA to use timm. I have tried to stay closer to the original implementation, and also added typing. There might be some structural changes needed - for instance the `apply_masks` function should probably be moved to `utils`? Any suggestions on how to improve this are welcome. 

Also like the MAE timm implementation, I have created a separate file, instead of directly replacing the `torchvision` implementation. I assume once this is benchmarked the plan would be to completely replace it. 

#Tests
Unit tests for the predictor, encoder and backbone classes.

### DIFF
--- a/lightly/models/modules/__init__.py
+++ b/lightly/models/modules/__init__.py
@@ -45,6 +45,11 @@ if _dependency.torchvision_vit_available():
 if _dependency.timm_vit_available():
     # Requires timm >= 0.9.9
     from lightly.models.modules.heads_timm import AIMPredictionHead
+    from lightly.models.modules.ijepa_timm import (
+        IJEPABackboneTIMM,
+        IJEPAEncoderTIMM,
+        IJEPAPredictorTIMM,
+    )
     from lightly.models.modules.masked_autoencoder_timm import MAEDecoderTIMM
     from lightly.models.modules.masked_causal_vision_transformer import (
         MaskedCausalVisionTransformer,
--- a/lightly/models/modules/ijepa_timm.py
+++ b/lightly/models/modules/ijepa_timm.py
@@ -1,0 +1,421 @@
+import math
+from functools import partial
+from typing import Callable, List, Optional, Union
+
+import torch
+import torch.nn as nn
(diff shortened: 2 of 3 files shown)


guarin:
Hi! Thanks a lot for this extensive PR, it looks really well made!

Looking at the code I see many parallels to our `MaskedVisionTransformer` implementation.  I didn't have time yet to go through the full PR but I have a suspicion that `IJEPABackboneTIMM` and `IJEPAEncoderTIMM` are compatible with our implementation of the `MaskedVisionTransformer` (see [here](https://github.com/lightly-ai/lightly/blob/master/lightly/models/modules/masked_vision_transformer_timm.py) and [here](https://github.com/lightly-ai/lightly/blob/master/lightly/models/modules/masked_vision_transformer.py)). Do you think it would be possible to use `MaskedVisionTransformer` directly instead of `IJEPABackboneTIMM` or am I missing something? If possible this simplify the code a lot.

What I imagine is something like this:
```
target_encoder = MaskedVisionTransformer()
context_encoder = MaskedVisionTransformer()
predictor = IJEPAPredictorTIMM()

# This encodes all patches.
target = target_encoder.encode(images)
# This encodes only the unmasked context patches.
context = context_encoder.encode(images, idx_keep=idx_keep)

prediction = predictor(context, masks_x, masks)
target = get_targets_at_masks(target, masks_x, masks)
loss(predictions, target)
```

radiradev:
Hi @guarin. I think you are correct and we can reuse the `MaskedVisionTransformer` making this PR a lot smaller. I have updated it with your suggested changes. Also added `drop_path_rate` for `IJepaPredictor`.

guarin (approved):
Awesome, thanks so much!

guarin:
Created two follow-up issues:
* https://github.com/lightly-ai/lightly/issues/1614
* https://github.com/lightly-ai/lightly/issues/1615

Closes: #1367

PR #1618, opened 2024-07-29
feat: drop duplicate code in modules

SauravMaheshkar:
* Remove duplicate code from `lightly/models/modules/masked_autoencoder_timm.py`.
* Drop redundant import + add type hint in `forward`.
* The `mypy` pre-commit hook was not recognising the plugin, so I added the additional dependency [source](https://github.com/numpy/numpy/issues/19215#issuecomment-1367434689)

### DIFF
--- a/lightly/models/modules/masked_autoencoder_timm.py
+++ b/lightly/models/modules/masked_autoencoder_timm.py
@@ -8,7 +8,7 @@ from torch import Tensor
 from torch.nn import LayerNorm, Module, Parameter, Sequential
 
 from lightly.models import utils
-from lightly.models.modules.masked_vision_transformer_timm import _init_weights
+from lightly.models.modules.masked_vision_transformer_timm import init_weights
 
 
 class MAEDecoderTIMM(Module):
@@ -114,7 +114,7 @@ class MAEDecoderTIMM(Module):
             Tensor with shape (batch_size, seq_length, out_dim).
 
         """
-        out: Tensor = self.embed(input)
+        out = self.embed(input)
         out = self.decode(out)
         return self.predict(out)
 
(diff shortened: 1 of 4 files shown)


lightly/models/modules/masked_autoencoder_timm.py
```diff
@@ -5,9 +5,10 @@
 import torch.nn as nn
 from timm.models.vision_transformer import Block
 from torch import Tensor
-from torch.nn import LayerNorm, Linear, Module, Parameter, Sequential
+from torch.nn import LayerNorm, Module, Parameter, Sequential
 
 from lightly.models import utils
+from lightly.models.modules.masked_vision_transformer_timm import _init_weights
```
guarin, line 11: Functions starting with an underscore are considered "private" and should not be accessed from outside the module/file. Could you rename the function in `masked_vision_transformer_timm.py` to `init_weights` instead to make it public?

lightly/models/modules/masked_autoencoder_timm.py
```diff
@@ -113,7 +114,7 @@ def forward(self, input: Tensor) -> Tensor:
             Tensor with shape (batch_size, seq_length, out_dim).
 
         """
-        out = self.embed(input)
+        out: Tensor = self.embed(input)
```
guarin, line 117: Are you sure this is necessary? `embed` has a type annotation that says it returns a `Tensor`.

guarin (changes requested):
Thanks for the PR :) Left some comments

lightly/models/modules/masked_autoencoder_timm.py
SauravMaheshkar: Apparently not. `mypy` was failing locally for me without this, but I cleared my cache and it works. Dropped in latest commit.

guarin (approved):
LGTM!

PR #1620, opened 2024-07-30
Download finetune tutorial checkpoint directly from url

guarin:
## Changes

* Load checkpoint directly from url

This fixes issues in docs CI which now no longer requires the checkpoint to be available locally.

FYI @SauravMaheshkar


SauravMaheshkar (approved):

MalteEbner (approved):
LGTM

PR #1625, opened 2024-08-07
Use uv for dev and CI environments

guarin:
### Changes

* Use uv for dev and CI environments
* Make library compatible with torch 1.10, torchvision 0.11, and pytorch lightning 1.6
* Add optional matplotlib dependency
* Remove unused minimal dependencies
* Only import from numpy.typing when type checking to keep backwards compatibility with old numpy versions

This PR does not change the minimal required dependencies to use the package. In the future we should make a breaking change that moves the torch, torchvision, and pytorch lightning dependencies from `minimal` to the main dependencies.

The PR contains quite a lot of smaller changes that were necessary to get everything working using uv and CI. I left comments to explain the changes.

This PR also fixes the issue regarding numpy versions from https://github.com/lightly-ai/lightly/pull/1490


.github/workflows/test.yml
```diff
@@ -42,29 +42,31 @@ jobs:
...
             echo "DISTRIB_RELEASE=18.04" > /etc/lsb-release
           fi
-      - name: Set up Python
+      - name: Set Up Python
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
-          cache: pip
-          cache-dependency-path: |
-            pyproject.toml
```
guarin, line 51: I dropped caching for now as installing with uv is super fast.

.github/workflows/test_minimal_deps.yml
```diff
@@ -33,23 +33,23 @@ jobs:
     strategy:
       matrix:
         python: ["3.7"]
-
+        dependencies: ["minimal", "minimal-extras"]
```
guarin, line 36: Test minimal dependencies once with and once without extras

Makefile
```diff
@@ -78,15 +78,7 @@ dist: clean
 	python -m build
 	ls -l dist
 
-## install the package to active site
-install: clean 
-	pip install .
```
guarin, line 83: Remove this

Makefile
```diff
@@ -78,15 +78,7 @@ dist: clean
...
 
-# Should be same command as in the docs:
-# https://docs.lightly.ai/docs/install-lightly#install-the-lightly-python-client
-install-api-only: clean
-	pip install -r requirements/base.txt
-	pip install . --no-deps
```
guarin, line 89: Moved further down

lightly/api/api_workflow_selection.py
```diff
@@ -1,9 +1,10 @@
+from __future__ import annotations
```
guarin, line 1: This is required to not get an error for unknown `NDArray` type hint if we are not type checking.

lightly/api/api_workflow_selection.py
```diff
@@ -16,6 +17,9 @@
     TagData,
 )
 
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
+
```
guarin, line 22: `NDArray` is only available since numpy `1.20`. To remain backwards compatible we only import it while type-checking.

lightly/embedding/_base.py
```diff
@@ -9,7 +9,7 @@
 import omegaconf
 from omegaconf import DictConfig
 from pytorch_lightning import LightningModule, Trainer
-from pytorch_lightning.callbacks.callback import Callback
+from pytorch_lightning.callbacks import Callback
```
guarin, line 12: Importing directly from `pytorch_lightning.callbacks` is backwards compatible with older lightning versions

pyproject.toml
```diff
@@ -88,54 +86,24 @@ dev = [
   "types-python-dateutil",
   "types-toml"
 ]
+# Minimal dependencies against which we test. Older versions might work depending on the
+# functionality used.
 minimal = [
-  "certifi==2017.4.17",
-  "hydra-core==1.0.0",
-  "lightly_utils~=0.0.0",
```
guarin, line 94: I removed all these deps as they are handled by uv.

pyproject.toml
```diff
@@ -88,54 +86,24 @@ dev = [
...
-  "aenum >= 3.1.11"
+  "python_dateutil>=2.5.3",
+  "setuptools>=21.0.0",
+  "urllib3>=1.25.3",
+  "pydantic>=1.10.5",
+  "aenum>=3.1.11"
 ]
 timm = ["timm>=0.9.9"]
 video = ["av>=8.0.3"]
+matplotlib = ["matplotlib>=3"]
```
guarin, line 105: Added because matplotlib is optionally used in `lightly/utils/debug.py`

pyproject.toml
```diff
@@ -88,54 +86,24 @@ dev = [
...
-  "pre-commit",
-  "opencv-python==4.4.0.46",
-  "scikit-learn==0.24.2",
-  "pandas",
-  "toml",
-  "matplotlib",
-  "av==8.0.3"
+  "torch>=1.10.0",
+  "torchvision>=0.11.0",
+  "pytorch_lightning>=1.6",
```
guarin, line 94: I also tested lower versions but there are issues that would be very hard to make backwards compatibly.

Makefile
```diff
@@ -100,3 +92,99 @@ tox:
...
+install-dev:
+	uv pip install ${EDITABLE} . --all-extras --requirement pyproject.toml
+
+
+# Install package with API dependencies only.
+# Should be same command as in the docs with some extra flags for CI:
+# https://docs.lightly.ai/docs/install-lightly#install-the-lightly-python-client
+.PHONY: install-api-only
+install-api-only:
+	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} --requirement requirements/base.txt
```
philippmwirth, line 140: Why do we have to exclude newer here?

Makefile
```diff
@@ -100,3 +92,99 @@ tox:
...
+# 
+# This command is split into multiple steps:
+# 1. Install the dev dependencies to be able to run tests. We don't want to use
+#    the minimal versions for these dependencies.
+# 2. Then we reinstall the package with minimal dependencies.
+# 3. Finally we install setuptools<50. This is necessary for compatibility with old
+#    PyTorch Lightning versions that do not include the correct setuptools dependencies.
+#
+# Explanation of flags:
+# --exclude-newer: We don't want to install dependencies released after that date to
```
philippmwirth, line 153: But then we'll miss breaking dependency updates.

Makefile
```diff
@@ -100,3 +92,99 @@ tox:
...
+# want to install the minimal versions.
+.PHONY: install-minimal-extras
+install-minimal-extras:
+	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} ${EDITABLE} ".[dev]"
+	uv pip install --resolution=lowest-direct --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall ${EDITABLE} ".[matplotlib,minimal,timm,video]" --requirement pyproject.toml
+	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall "setuptools<50"
+
+# Install package with fixed dependencies.
+.PHONY: install-fixed
+install-fixed:
```
philippmwirth, line 179: I think I said this earlier to Malte. The name. `fixed` is not descriptive enough. What are the dependencies fixed to? And why?

philippmwirth (approved):

Makefile
guarin: Because we don't want CI to break if new dependencies are released.

Makefile
guarin: Breaking dependencies are tested in `weekly_dependency_test.yml`. I updated it to use `make install-latest`

Makefile
guarin: I changed it to `install-pinned` and described it in more detail in the comment.

PR #1626, opened 2024-08-08
chore: drop `tox`

SauravMaheshkar:
Remove `tox` in favor of CI matrix tests


guarin (approved):
Makes sense as tox tests are outdated and not used anymore 👍🏼

PR #1628, opened 2024-08-08
Add CenterCrop to Val Transform in Finetuning Tutorial

guarin:
### Changes

* Add CenterCrop to val transform in finetuning tutorial
* This fixes size issues because the Resize transform resizes the shorter edge to the target length

It is standard to use Resize + CenterCrop for validation transforms.


SauravMaheshkar (approved):
LGTM 👍🏽

PR #1629, opened 2024-08-12
Refactor DINOLoss to not use Center module

guarin:
### Changes

* Fix issue that DINOLoss.to(device) didn't copy the center tensor in the Center module to the correct device

The issue was that module.to() copies tensors instead of modfiying them inplace. The `DINOLoss.center` buffer would get copied to the correct device, but the `Center.center` tensor would remain on the original device and not receive updates anymore. 

There is a global pytorch flag to always modify buffers inplace but I don't think we should force everyone to use this flag. I did some more checks and there is no simple way to keep the buffers in DINOLoss.center and Center synchronized. So instead I refactored the code a bit and removed Center from DINOLoss for now.

### How was it tested?
* Added unit test
* Note that the new public `center_momentum` and `center_mean` functions are already tested through the `Center` module

Closes: https://github.com/lightly-ai/lightly/issues/1630


lightly/models/modules/center.py
```diff
@@ -74,20 +60,33 @@ def update(self, x: Tensor) -> None:
...
+    return batch_center
+
+
+@torch.no_grad()
+def center_momentum(center: Tensor, batch_center: Tensor, momentum: float) -> Tensor:
+    """Returns the new center with momentum update."""
+    return center * momentum + batch_center * (1 - momentum)
+
+
+CENTER_MODE_TO_FUNCTION = {
```
philippmwirth, line 90: Why is this necessary? Do we plan to extend this in the future?

lightly/loss/dino_loss.py
```diff
@@ -66,17 +67,21 @@ def __init__(
         center_mode: str = "mean",
     ):
         super().__init__()
+        if center_mode not in CENTER_MODE_TO_FUNCTION:
```
philippmwirth, line 70: Suggestion: I find this slightly more readable but I'll leave it up to you:
```python
self._center_fn = CENTER_MODE_TO_FUNCTION.get(center_mode)
if self._center_fn is None:
    raise ValueError(...)

philippmwirth (approved):

lightly/models/modules/center.py
guarin: Yes, DINOv2 to uses sinkhorn centering for some models

PR #1644, opened 2024-09-12
Add cosine_warmup_schedule

guarin:
### Changes

* Add `cosine_warmup_schedule` function

### How was it tested?

The change is covered by the tests for `CosineWarmupScheduler`. I'll make a follow-up PR to refactor those tests.

### DIFF
--- a/lightly/utils/scheduler.py
+++ b/lightly/utils/scheduler.py
@@ -32,16 +32,16 @@ def cosine_schedule(
 
     """
     if step < 0:
-        raise ValueError("Current step number can't be negative")
+        raise ValueError(f"Current step number {step} can't be negative")
     if max_steps < 1:
-        raise ValueError("Total step number must be >= 1")
+        raise ValueError(f"Total step number {max_steps} must be >= 1")
     if period is None and step > max_steps:
         warnings.warn(
             f"Current step number {step} exceeds max_steps {max_steps}.",
             category=RuntimeWarning,
         )
     if period is not None and period <= 0:
-        raise ValueError("Period must be >= 1")
+        raise ValueError(f"Period {period} must be >= 1")
 
     decay: float
     if period is not None:  # "cycle" based on period, if provided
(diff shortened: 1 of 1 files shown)


lightly/utils/scheduler.py
```diff
@@ -67,6 +66,76 @@ def cosine_schedule(
...
+        warmup_end_value:
+            Target value for warmup. Defaults to start_value.
+        period:
+            The number of steps over which the cosine function completes a full cycle.
+            Defaults to max_steps - warmup_steps.
+    Returns:
+        Cosine decay value.
+    """
+    if warmup_steps < 0:
+        raise ValueError("Warmup steps can't be negative")
```
philippmwirth, line 105: Minor: 
```suggestion
        raise ValueError(f"Warmup steps ({warmup_steps}) can't be negative")
```

And you can add it to the other error messages below too

lightly/utils/scheduler.py
```diff
@@ -67,6 +66,76 @@ def cosine_schedule(
...
+    if warmup_steps > max_steps:
+        raise ValueError("Warmup steps must be <= max_steps")
+    if step > max_steps:
+        warnings.warn(
```
philippmwirth, line 109: We should add a logger...

lightly/utils/scheduler.py
```diff
@@ -67,6 +66,76 @@ def cosine_schedule(
...
+            category=RuntimeWarning,
+        )
+
+    if warmup_end_value is None:
+        warmup_end_value = start_value
+
+    if step < warmup_steps:
+        return (
+            warmup_start_value
+            + (warmup_end_value - warmup_start_value) * (step + 1) / warmup_steps
```
philippmwirth, line 120: Can you add a comment why `step + 1` is used?

lightly/utils/scheduler.py
```diff
@@ -67,6 +66,76 @@ def cosine_schedule(
...
+        )
+    elif period is not None:
+        return cosine_schedule(
```
philippmwirth, line 123: We can make this a bit more concise:
```python
else:
    # TODO: Explain why it's set to 1 if the period is not None.
    max_steps =  max_steps - warmup_steps if period is None else 1
    return cosine_schedule(
        step=step - warmup_steps,
        max_steps=max_steps,
        start_value=start_value,
        end_value=end_value,
        period=period,
    )
```

lightly/utils/scheduler.py
philippmwirth: Not in this PR of course 😄

guarin: I think warnings is fine for library code. They can be captured by any user defined logger.

philippmwirth (approved):

PR #1645, opened 2024-09-12
Add update_param_groups

guarin:
### Changes

* Add `update_param_groups` to `lightly/utils/optim`
 
This is useful for updating values in parameter groups when following training schedules.

### How was it tested?

* Added unit tests

### DIFF
--- a/lightly/utils/optim.py
+++ b/lightly/utils/optim.py
@@ -1,0 +1,90 @@
+from typing import Any, Dict, Iterable, Optional
+
+from torch.optim import Optimizer
+
+
+def update_param_groups(
+    optimizer: Optimizer,
+    default_update: Optional[Dict[str, Any]] = None,
+    updates: Optional[Iterable[Dict[str, Any]]] = None,
+) -> None:
+    """Update the values in the param groups of the optimizer.
+
+    This is useful to update values following a schedule during training.
+
+    Args:
+        optimizer:
+            A PyTorch optimizer.
+        default_update:
+            Key value pairs that will be updated for all param groups. The value is only
+            updated if the respective key already exists in the param group.
+        updates:
+            A list of dicts with key value pairs. Each dict must contain a key "name"
+            that specifies the name of the param group(s) that should be updated. All
+            values of the dict will be updated in the param group(s) with the same name.
+
+    Examples:
+        >>> optimizer = torch.optim.SGD(
+        >>>     [
+        >>>         {
+        >>>             "name": "model",
+        >>>             "params": params,
+        >>>         },
+        >>>         {
+        >>>             "name": "model_no_weight_decay",
+        >>>             "params": params_no_weight_decay,
+        >>>             "weight_decay": 0.0,
+        >>>         },
+        >>>         {
+        >>>             "name": "head",
+        >>>             "params": head_params,
+        >>>             "weight_decay": 0.5,
+        >>>         },
+        >>>     ],
+        >>>     lr=0.1,
+        >>>     weight_decay=0.2,
+        >>> )
+        >>>
+        >>> update_param_groups(
+        >>>     optimizer=optimizer,
+        >>>     default_update={"weight_decay": 0.3},
+        >>>     updates=[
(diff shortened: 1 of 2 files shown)


lightly/utils/optim.py
```diff
@@ -0,0 +1,90 @@
...
+        >>> update_param_groups(
+        >>>     optimizer=optimizer,
+        >>>     default_update={"weight_decay": 0.3},
+        >>>     updates=[
+        >>>         {"name": "model_no_weight_decay", "weight_decay": 0.0},
+        >>>         {"name": "head", "weight_decay": 0.6},
+        >>>     ]
+        >>> )
+        >>>
+        >>> # Param group "model" has now weight decay 0.3
```
philippmwirth, line 57: According to the docstring this is incorrect because `model` doesn't have a `weight_decay` in it's param group before the update.

Edit: My comment is wrong because we set a `weight_decay` default value.

lightly/utils/optim.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def update_param_groups(
+    optimizer: Optimizer,
+    default_update: Optional[Dict[str, Any]] = None,
+    updates: Optional[Iterable[Dict[str, Any]]] = None,
+) -> None:
+    """Update the values in the param groups of the optimizer.
+
+    This is useful to update values following a schedule during training.
```
philippmwirth, line 13: We should mention here which update is applied first so it's clear which one overrides the other.

lightly/utils/optim.py
```diff
@@ -0,0 +1,90 @@
+from typing import Any, Dict, Iterable, Optional
+
+from torch.optim import Optimizer
+
+
+def update_param_groups(
```
philippmwirth, line 6: I think this interface works. It's also similar to how the optimizer is initialized so that's good. 

Did you also consider something like this?
```python
def update_param_groups(
    optimizer: Optimizer,
    update: Optional[Dict[str, Any]] = None,
    filter_fn: Callable[[Dict[str, Any]], bool] = lamba x: x, # lazy default value, we can do better
) -> None:
    for param_group in param_groups:
        if not filter_fn(param_group):
            # Do update here.
            pass
```


Usage could look like this:
```python
update_param_groups(
    optimizer=optimizer,
    update={"weight_decay": 0.5},
    filter_fn=lambda x: "weight_decay" not in x or x["weight_decay"] == 0.0
)

File: lightly/utils/optim.py [hunk 4]
```diff
@@ -0,0 +1,90 @@
...
+        >>> # Param group "head" has now weight decay 0.6
+
+    """
+    if default_update is None:
+        default_update = {}
+    if updates is None:
+        updates = []
+
+    # Update the optimizer's param_groups with the provided default values.
+    for param_group in optimizer.param_groups:
```
philippmwirth, line 68: We go through this whole loop even if the default update is `None` or `{}`. Let's only do it if `default_update` contains an actual update.

lightly/utils/optim.py
guarin: Added a note that `updates` has precedence over `default_updates`.

lightly/utils/optim.py
guarin: I didn't consider it, but I am also not sure if the lambda approach is preferable. I think it will be difficult to specify different updates. For example, how would we update `{"name": "head1", "weight_decay": 0.1}` and `{"name": "head2", "weight_decay": 0.2},`?

lightly/utils/optim.py
guarin: Good catch, I switched the loop order when cleaning up. I switched it back now 👍🏼

lightly/utils/optim.py
philippmwirth: You would use two calls.

guarin: No you can do it in a single call:
```
update_param_groups(
    optimizer,
    updates=[
        {"name": "head1", "weight_decay": 0.1},
        {"name": "head2", "weight_decay": 0.2},
    ]
)
```

PR #1647, opened 2024-09-16
Update cosine warmup tests

guarin:
### Changes

* Update cosine warmup tests
* Add `warmup_start_value` and `warmup_end_value` arguments to CosineWarmupScheduler

Follow-up from #1644 

### How was it tested?

* Added new tests while keeping the old tests

Will remove the old tests in a follow-up.

### DIFF
--- a/lightly/utils/scheduler.py
+++ b/lightly/utils/scheduler.py
@@ -144,13 +144,17 @@ class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
         max_epochs:
             Total number of training epochs or steps.
         last_epoch:
-            The index of last epoch or step. Default: -1
+            The index of last epoch or step.
         start_value:
-            Starting learning rate scale. Default: 1.0
+            Starting learning rate.
         end_value:
-            Target learning rate scale. Default: 0.001
+            Target learning rate.
         verbose:
-            If True, prints a message to stdout for each update. Default: False.
+            If True, prints a message to stdout for each update.
+        warmup_start_value:
+            Starting learning rate for warmup.
+        warmup_end_value:
+            Target learning rate for warmup. Defaults to start_value.
 
     Note: The `epoch` arguments do not necessarily have to be epochs. Any step or index
     can be used. The naming follows the Pytorch convention to use `epoch` for the steps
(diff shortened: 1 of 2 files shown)


lightly/utils/scheduler.py
```diff
@@ -144,13 +144,17 @@ class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
...
             Total number of training epochs or steps.
         last_epoch:
-            The index of last epoch or step. Default: -1
+            The index of last epoch or step.
         start_value:
-            Starting learning rate scale. Default: 1.0
+            Starting learning rate.
         end_value:
-            Target learning rate scale. Default: 0.001
+            Target learning rate.
```
guarin, line 151: We don't put default values in docstrings anymore as they are already in the function definition.

lightly/utils/scheduler.py
```diff
@@ -144,13 +144,17 @@ class CosineWarmupScheduler(torch.optim.lr_scheduler.LambdaLR):
...
         verbose:
-            If True, prints a message to stdout for each update. Default: False.
+            If True, prints a message to stdout for each update.
+        warmup_start_value:
+            Starting learning rate for warmup.
+        warmup_end_value:
+            Target learning rate for warmup. Defaults to start_value.
```
philippmwirth, line 157: Remove defaults message.

lightly/utils/scheduler.py
```diff
@@ -167,12 +171,16 @@ def __init__(
...
         verbose: bool = False,
+        warmup_start_value: float = 0.0,
+        warmup_end_value: Optional[float] = None,
     ) -> None:
         self.warmup_epochs = warmup_epochs
         self.max_epochs = max_epochs
         self.start_value = start_value
         self.end_value = end_value
         self.period = period
+        self.warmup_start_value = warmup_start_value
```
philippmwirth, line 182: I'm surprised this works but ok 😄

philippmwirth: Does it also work if I change `warmup_start_value`?

philippmwirth (approved):

lightly/utils/scheduler.py
guarin: This is on purpose to let the user know that `warmup_end_value` is set to `start_value` if not specified.

lightly/utils/scheduler.py
guarin: Yes, `self.warmup_start_value` is passed later on to `cosine_warmup_schedule(warmup_start_value=self.warmup_start_value)`

PR #1648, opened 2024-09-17
Support python 3.12

### LINKED ISSUE #1473: Support for Python 3.12
This is just for people curious about why they can't use lightly with Python 3.12

The issue is that lightly uses PyTorch and there is not yet a stable release for Python 3.12. There is experimental support when you manually install the nightly build. You can read up more on that in the corresponding [PyTorch issue](https://github.com/pytorch/pytorch/issues/110436).

We expect support to be there sometime in Q1 2024.

MalteEbner:
closes #1473

### CI Updates

Updates the  tests to use python 3.12 instead of 3.10. 

### Testing updates

These were necessary because local tests with python 3.12 were failing (on Mac).

Replaces `self.assertRaisesRegexp` with `self.assertRaisesRegex`, as the former was deprecated in python 3.2, see [here](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex) and removed in python 3.12.

Replaces `self.assertTrue(torch.allclose(a, b))` with `torch.testing.assert_close(a,b)`, as the latter has better debugging support and more stability under other python / pytorch versions. `torch.testing.assert_close` was added in April 2021 and is thus available for our oldest supported torch version (1.10 was released in late 2021).
Also uses it for one more test.

## Testing

CI changes test themselves ;)

### DIFF
--- a/.github/workflows/release_pypi.yml
+++ b/.github/workflows/release_pypi.yml
@@ -21,10 +21,10 @@ jobs:
         LATEST_TAG=$(git describe --tags `git rev-list --tags --max-count=1`)
         git checkout $LATEST_TAG
         echo "tag_name=$LATEST_TAG" >> $GITHUB_OUTPUT;
-    - name: Set up Python 3.10
+    - name: Set up Python 3.12
       uses: actions/setup-python@v5
       with:
-        python-version: "3.10"
+        python-version: "3.12"
     - name: Build and release
       id: build_and_release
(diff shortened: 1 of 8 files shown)


MalteEbner:
TODO: Change our branch protection to require the 3.12 instead of 3.10
image

guarin (approved):
LGTM 🎉

Issue #1635, opened 2024-08-16
Type Check Package


guarin:
We would like to type check all files in the package. Adding types helps to document the code and makes it easier to use.

## How to work on this issue

1. Go to [pyproject.toml](https://github.com/lightly-ai/lightly/blob/master/pyproject.toml#L184) to the `exclude` section of the mypy settings. There should be a list of files like this:
```
exclude = '''(?x)(
    lightly/cli/version_cli.py |
    lightly/cli/crop_cli.py |
    lightly/cli/serve_cli.py |
    lightly/cli/embed_cli.py |
    lightly/cli/lightly_cli.py |
...
```
2. Pick a file you would like to type. Best are files from `lightly/loss`, `lightly/data`, `lightly/utils`, `lightly/models`, and the corresponding test files.
3. Run `mypy <filename>` (make sure that you have installed lightly following the [contribution guide](https://github.com/lightly-ai/lightly/blob/master/CONTRIBUTING.md)). This should show a list with all the missing types/errors.
4. Add types until no more errors are shown. This is a good example on how the types should look like: https://github.com/lightly-ai/lightly/blob/master/lightly/models/modules/memory_bank.py
5. Go to `pyproject.toml` and remove the filename of the file you just typed from the mypy `exclude` list.
6. Create a new PR named `Add types for <filename>` and push your changes. Make sure all Github actions pass.

ishaanagw:
Hey, can I work on this issue ?

ishaanagw:
Hey @guarin, some of the mypy tests are failing in local, 48 errors in 23 files in master, do you think I can still push my changes for the fix of some of the files ?

guarin:
Hi @ishaanagw! Thanks for contributing, feel free to work on any file :)

> Hey @guarin, some of the mypy tests are failing in local, 48 errors in 23 files in master, do you think I can still push my changes for the fix of some of the files ?

Interesting, will have a look at this on Monday. Which mypy version are you using? CI uses mypy 1.4.1 with Python 3.7. The Python version should not be too important. We use 3.7 in CI to make sure we don't introduce any types that are incompatible with old Python versions.

ishaanagw:
Hey, @guarin Can you please review this PR - https://github.com/lightly-ai/lightly/pull/1651

guarin:
> Hey @guarin, some of the mypy tests are failing in local, 48 errors in 23 files in master, do you think I can still push my changes for the fix of some of the files ?

[...]

PR #1654, opened 2024-09-24
Guarin lig 5512 fix local type check

guarin:
### Changes

* Fix typing inconsistencies between Python version
* Remove `python_version` mypy flag

This fixes all [typing issues](https://github.com/lightly-ai/lightly/issues/1635#issuecomment-2364651925) with mypy 1.4.1 on Python 3.7, 3.8, and 3.10. I removed the `python_version` mypy setting in `pyproject.toml` because it is not needed anymore. Mypy automatically uses the installed python version. This should make type checking more robust.


MalteEbner (approved):
LGTM

PR #1668, opened 2024-10-06
Add Documentation for lightly/models/utils.py

### LINKED ISSUE #1663: Cleanup docstrings in lightly/models/utils.py
See #1216 for more information

ayush22iitbhu:
fixes #1663 
I have done the following:
- Added documentation wherever necessary. For example, many functions had the error description part missing in their docstrings.
- Corrected any grammatical error that I found
- Added more comments for clarity


lightly/models/utils.py
```diff
@@ -501,7 +524,7 @@ def random_token_mask(
             Size of the token batch for which to generate masks.
             Should be (batch_size, sequence_length).
         mask_ratio:
-            Percentage of tokens to mask.
+            Proportion of tokens to mask. Default is 0.6 .
```
guarin, line 527:
```suggestion
            Proportion of tokens to mask.
```

The default value should not be documented as it is already part of the function definition:
```
def random_token_mask(
...
    mask_ratio: float = 0.6,
...
):
```

Also documenting it in the docstring creates a lot of overhead because it is difficult to verify that the actual default value and the documented default are always the same. This is especially a problem when code gets updated and changes to the docstrings get forgotten.

Could you also remove the default values in the docstrings in the other functions?

lightly/models/utils.py
```diff
@@ -678,20 +715,24 @@ def random_block_mask_image(
...
+            Maximum number of attempts to find a valid block mask. Default is 10.
         device:
             Device on which to create the mask.
     Returns:
         A boolean tensor with shape (height, width) where each entry is True if the
         patch should be masked and False otherwise.
+
+    Raises:
+        ValueError: If 'max_num_masks_per_block' is less than 'min_num_masks_per_block' or
+                    if 'max_block_aspect_ratio' is less than 'min_block_aspect_ratio'
```
guarin, line 735:
```suggestion
    Raises:
        ValueError: If 'max_num_masks_per_block' is less than 'min_num_masks_per_block' or
            if 'max_block_aspect_ratio' is less than 'min_block_aspect_ratio'    
```

We usually don't align indentation with text from previous line. Instead, we just indent one tab more.

lightly/models/utils.py
```diff
@@ -1115,11 +1178,15 @@ def update_drop_path_rate(
         mode:
             Drop path rate update mode. Can be "linear" or "uniform". Linear increases
             the drop path rate from 0 to drop_path_rate over the depth of the model.
-            Uniform sets the drop path rate to drop_path_rate for all blocks.
+            Uniform sets the drop path rate to drop_path_rate for all blocks. Default is "linear".
+    Raises:
+        ValueError : If an unknown mode is provided.
```
guarin, line 1183:
```suggestion
        ValueError: If an unknown mode is provided.        
```

guarin (changes requested):
Thanks a lot for the contribution! I left some small comments.

Could you also run `make format` to format the files? The github actions failed because `utils.py` is not well formatted: https://github.com/lightly-ai/lightly/actions/runs/11210971225/job/31161605179?pr=1668 The steps on how to create a PR and format it are described here: https://github.com/lightly-ai/lightly/blob/master/CONTRIBUTING.md#pull-requests

Also, are you planning to add/update docstrings for the remaining functions as well? In that case I would keep the issue assigned to you.

guarin:
@ayush22iitbhu let me know when this is ready for review again

lightly/models/utils.py
ayush22iitbhu: @guarin done

lightly/models/utils.py
ayush22iitbhu: @guarin done

lightly/models/utils.py
ayush22iitbhu: @guarin done

PR #1677, opened 2024-10-07
Update lightly_at_a_glance.rst

jizhang02:
add `import lightly.data as data` before using it in code example:
```
import torch
import lightly.data as data

# Create a dataset from your image folder.
dataset = data.LightlyDataset(
    input_dir='./my/cute/cats/dataset/',
    transform=transform,
)

# Build a PyTorch dataloader.
dataloader = torch.utils.data.DataLoader(
    dataset,                # Pass the dataset to the dataloader.
    batch_size=128,         # A large batch size helps with learning.
    shuffle=True,           # Shuffling is important!
)
```


guarin (approved):
Thanks for the fix!

Issue #1477, opened 2024-01-15
Remove opencv dependency from tests


guarin:
We still have an old opencv dependency in the tests, see here: https://github.com/lightly-ai/lightly/blob/2b215aa5d19b50515f280c1b0181f2f135cbc93f/tests/data/test_VideoDataset.py#L11

We should refactor the test(s) to not rely on opencv and remove opencv from all requirements files.

ishaanagw:
Hey @guarin, Can I work on this ?

SauravMaheshkar:
Sure @ishaanagw give it a go 😃

ishaanagw:
Can you please explain the task a little more please ? Do I have to replace the videoWriters from OpenCV to some other library ?

guarin:
> Can you please explain the task a little more please ? Do I have to replace the videoWriters from OpenCV to some other library ?

Yes the idea would be to replace `cv2.VideoWriter` with [`write_video`](https://pytorch.org/vision/main/generated/torchvision.io.write_video.html) from torchvision. Looking at it this could actually be a more tricky issue than anticipated. If you would like to work on it, could you try the following:
* Replace all `cv2.VideoWriter` with torchvision `write_video` calls instead
* Replace the numpy arrays containing the frames (see [here](https://github.com/lightly-ai/lightly/blob/2b215aa5d19b50515f280c1b0181f2f135cbc93f/tests/data/test_VideoDataset.py#L53-L56)) with tensors because `write_video` expects tensors
* Try to run the tests with `pytest tests/data/test_VideoDataset.py`

If you run into errors regarding video timestamps we most likely can't remove cv2.

ishaanagw:
Getting this error -  ValueError: Expected numpy array with ndim `3` but got `2`
in self._test_video_dataset_non_increasing_timestamps()

Is this what you were talking about ?

guarin:
Interesting, could you post the full stack trace? 

> Is this what you were talking about ?

No I believe the ValueError already happens before. With the timestamps I meant that depending on which library is used when creating a new video file, the timestamps stored in the file can be different. And I am not sure if `write_video` creates the exact same timestamps as `cv2.VideoWriter`.

ishaanagw:
tests/data/test_VideoDataset.py:238: in _test_video_dataset_non_increasing_timestamps
    self.create_dataset(n_videos=2, n_frames_per_video=5)
tests/data/test_VideoDataset.py:83: in create_dataset
[...]

PR #1678, opened 2024-10-07
Removing open CV from test_VideoDataset.py

ishaanagw:


ishaanagw:
Party fixes - https://github.com/lightly-ai/lightly/issues/1477

Using torchvision.io.write_video instead of cv2.

Issue #1660, opened 2024-10-02
Add RandomFrequenceMask transform


guarin:
See https://github.com/lightly-ai/lightly/issues/1646 for information

payo101:
Hi @guarin , I'm pretty new to open source and would like to contribute to this. I have been looking at the Contributions.md for basic guidelines, where should I go next on this?
Also are you on the hacktober's discord? (If it's comfortable with you to disclose that? If I had some doubts, I thought I could contact you their itself)

philippmwirth:
Hi @payo101, the next steps are: 1) We'll assign the issue to you; 2) You open a pull request as outlined in [CONTRIBUTIONS.md](https://github.com/lightly-ai/lightly/blob/master/CONTRIBUTING.md#pull-requests). For more information on the implementation details please check out the issue referenced in the description. For reference you can also take a look at the augmentations we have already implemented (see, e.g. [random rotation](https://github.com/lightly-ai/lightly/blob/master/lightly/transforms/rotation.py)). Could you please quickly confirm if you're interested in working on this?

> Also are you on the hacktober's discord? (If it's comfortable with you to disclose that? If I had some doubts, I thought I could contact you their itself)

Forwarding this to @guarin :)

payo101:
Yeah I am

guarin:
> Also are you on the hacktober's discord? (If it's comfortable with you to disclose that? If I had some doubts, I thought I could contact you their itself)

Yes you can find me under the same name :)

We also have a discord for lightly related questions: https://discord.gg/xvNJW94

payo101:
Hi @guarin , So I have 2 doubts to ask
- Will the input be the RFFT of the Image, or the Image itself? I have been currently working with assuming the former
- Should the output also return the binary mask that's being applied on the FFT?

payo101:
Also, I am facing a mypy error while proceeding to commit
<img width="835" alt="Screenshot 2024-10-06 235119" src="https://github.com/user-attachments/assets/5dd0a599-01ce-4c40-998e-8dd68469b1b3">

After looking into it, all necessary steps that can be done to prevent this has been done, but it's still showing this error, any chance anyone could help me with it?

philippmwirth:
[...]

PR #1680, opened 2024-10-07
Addition of Random frequency mask transform

payo101:
**Modified file(s)**
- `lightly/transforms/__init__.py`: Added new transform imports
- `lightly/transforms/random_crop_and_flip_with_grid.py`: Resolved `mypy` issue for `Any` subclassing

**Created file(s)**
- `lightly/transforms/random_frequency_mask_transform.py`
- `tests/transforms/test_random_frequency_mask_transform.py`

**Closes**: https://github.com/lightly-ai/lightly/issues/1660


lightly/transforms/random_frequency_mask_transform.py
```diff
@@ -0,0 +1,39 @@
+from typing import Tuple
+
+import numpy as np
+import torch
+from torch import Tensor
+
+
+class RFMTransform:
```
guarin, line 8: Let's use the full name as RFM is not a well known abbreviation:

```suggestion
class RandomFrequencyMaskTransform:
```

lightly/transforms/random_frequency_mask_transform.py
```diff
@@ -0,0 +1,39 @@
...
+    """2D Random Frequency Mask Transformation.
+
+    This transformation applies a binary mask on the fourier transform,
+    across all channels. k% of frequencies are set to 0 with this.
+    k ranges [0.01, 0.1)
```
guarin, line 13: Let's make `k` configurable:

```
def __init__(self, k: Tuple[float, float] = (0.01, 0.2)) -> None:
    self.k = k
```

lightly/transforms/random_frequency_mask_transform.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Input
+        - Tensor: RFFT of a 2D Image (C, H, W) C-> No. of Channels
+    Output
+        - Tensor: The masked RFFT of the image
+
+    """
+
+    def __call__(self, fft_image: Tensor) -> Tensor:
+        k = np.random.uniform(0.01, 0.1)
```
guarin, line 23:
```suggestion
        k = np.random.uniform(low=self.k[0], high=self.k[1])
```

lightly/transforms/random_frequency_mask_transform.py
```diff
@@ -0,0 +1,39 @@
...
+        # Mask: (C, H, W)
+        mask = torch.ones_like(fft_image)
```
guarin, line 25: I believe the random mask creation can be simplified a bit:

```suggestion
        # Create mask with k entries set to zero.
        mask = torch.rand(fft_image.shape, device=fft_image.device) > k

        # Do not mask zero frequency mode to retain majority of semantic information.
        # See Section 4 in https://arxiv.org/abs/2312.02205 for details.
        mask[:, 0, 0] = 1

        masked_image = fft_image * mask
        return masked_image
```

lightly/transforms/random_frequency_mask_transform.py
guarin, line 13:
```suggestion
    This transformation applies a binary mask on the fourier transform,
    across all channels. A proportion of k frequencies are set to 0 with this.
```

guarin (changes requested):
Thanks a lot for the PR! I left some comments, let me know what you think. The most important is to make `k` configurable.

lightly/transforms/random_frequency_mask_transform.py
payo101: What I understood from the paper is that the same binary mask needs to be applied on every channel of the RFFT of the Image. Based on that I think we can implement this. Please do tell if my understanding of the transform is wrong. Then your suggested change works best. 

```
k = np.random.uniform(low=self.k[0], high=self.k[1])

# Every mask for every channel will have same frequencies being turned off i.e. being set to zero
mask_type = torch.rand(fft_image.shape[1:], device=fft_image.device) > k # mask_type: (H, W)
mask = [] # Mask: (C, H, W)


for c in range(fft_image.size(dim=0)):
    mask.append(mask_type)
    
mask = torch.stack(mask)
# To retain majority of the semantic information. Please refer https://arxiv.org/abs/2312.02205
mask[:, 0, 0] = 1
masked_frequency_spectrum_image = fft_image * mask

return masked_frequency_spectrum_image
```

guarin: Ah yes you are right!

In that case we can do the masking like this:
```
        # Create mask with k entries set to zero.
        mask = torch.rand(fft_image.shape[1:], device=fft_image.device) > k
        # Do not mask zero frequency mode to retain majority of semantic information.
        # See Section 4 in https://arxiv.org/abs/2312.02205 for details.
        mask[0, 0] = 1
        mask = mask.unsqueeze(0) # Add channel dimension.
        masked_image = fft_image * mask
        return masked_image
```

payo101:
@guarin @philippmwirth I have made the changes, could you guys review the PR now?

lightly/transforms/random_frequency_mask_transform.py
```diff
@@ -19,20 +18,23 @@ class RFMTransform:
 
     """
 
-    def __call__(self, fft_image: Tensor) -> Tensor:
-        k = np.random.uniform(0.01, 0.1)
-        # Mask: (C, H, W)
-        mask = torch.ones_like(fft_image)
+    def __init__(self, k: Tuple[float, float] = (0.01, 0.2)) -> None:
```
guarin, line 21:
```suggestion
    def __init__(self, k: Tuple[float, float] = (0.01, 0.1)) -> None:
```

guarin (changes requested):
Thanks for the update! There is only one thing left, the default for k should be `(0.01, 0.1)` as in the paper.

payo101:
Done @guarin

guarin (approved):
LGTM! 🔥

PR #1681, opened 2024-10-08
Lambda functions removed from examples

### LINKED ISSUE #1512: Fix examples on Windows
We sometimes use lambdas in the examples when creating torch datasets: https://github.com/lightly-ai/lightly/blob/e7f13a53e68031ee0ae77124e76401b48381001f/examples/pytorch/mae.py#L87

This results in issues on Windows as lambdas cannot be pickled. We should update the examples and replace the lambdas with pickleable functions.

snehilchatterjee:
**Enhancement:**
- Replaced lambda functions with named functions in the examples.

**Fixes Issue:**
- Closes #1512.

**Approach:**
- I created a bash script to list all files containing lambda functions. After identifying the files, I manually replaced the lambda functions with named functions.

### DIFF
--- a/examples/pytorch/aim.py
+++ b/examples/pytorch/aim.py
@@ -67,7 +67,7 @@ transform = AIMTransform()
 # we ignore object detection annotations by setting target_transform to return 0
 
 
-def target_transform_function(t):
+def target_transform(t):
     return 0
 
 
@@ -75,7 +75,7 @@ dataset = torchvision.datasets.VOCDetection(
     "datasets/pascal_voc",
     download=True,
     transform=transform,
-    target_transform=target_transform_function,
+    target_transform=target_transform,
 )
 # or create a dataset from a folder containing images or videos:
 # dataset = LightlyDataset("path/to/folder")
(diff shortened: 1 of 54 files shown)


snehilchatterjee:
Hey @guarin any ideas what is causing the above error ?

examples/pytorch/aim.py
```diff
@@ -65,11 +65,17 @@ def forward(self, images):
 
 transform = AIMTransform()
 # we ignore object detection annotations by setting target_transform to return 0
+
+
+def target_transform_function(t):
```
guarin, line 70: Can we call this just `target_transform`? It keeps the examples a bit more concise and matches the `target_transform` argument name from the dataset.

guarin (commented):
Thanks a lot for the PR!

examples/pytorch/aim.py
snehilchatterjee: yeah sure I will update it right now

guarin (commented):
This looks great! Thanks so much.

guarin (approved):

snehilchatterjee:
> This looks great! Thanks so much.

Glad I could contribute :smile:

PR #1689, opened 2024-10-09
cleaning up docs in lightly/data subpackage

ChiragAgg5k:
Hey there, 

For now i have updated the docs for the helper functions only, and needed some review before i finish up with rest. Do let me know if you would like some modifications. I will finish up with rest very soon. Thanks :)

### DIFF
--- a/lightly/data/_helpers.py
+++ b/lightly/data/_helpers.py
@@ -4,7 +4,7 @@
 # All Rights Reserved
 
 import os
-from typing import Any, Callable, Dict, List, Optional, Set
+from typing import Any, Callable, Dict, Optional, Tuple
 
 from torchvision import datasets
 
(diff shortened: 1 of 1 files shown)


ChiragAgg5k:
Screenshot 2024-10-09 at 11 56 00 PM

i forgot to run the pre-commit, but now that i did, seems like everyone else also forgot to do so?? 😅

lightly/data/_helpers.py
```diff
@@ -34,104 +34,102 @@
...
 
     Returns:
-        Dataset consisting of images/videos in the root directory.
+        A dataset object (VideoDataset, ImageFolder, or DatasetFolder) based on the directory contents.
 
     Raises:
-        ValueError: If the specified dataset doesn't exist
-
+        ValueError: If the specified dataset directory doesn't exist or if videos are present
+                    but VideoDataset is not available.
```
guarin, line 116:
```suggestion
        ValueError: If the specified dataset directory doesn't exist or if videos are present
            but VideoDataset is not available.
```

We usually don't align indents but instead just indent one level more.

guarin (changes requested):
Thanks a lot for the PR, this looks very good! Just one small comment but otherwise this is ready to be merged.

Interesting that the pre-commit hooks try to format the `openapi_generated` files, those should be left untouched.

guarin:
The Code Format Check action is failing, could you run `make format` and update the PR?

guarin (approved):
LGTM thanks so much!

PR #1690, opened 2024-10-11
Create README.fr.md

kushal34712:
added french translation for readme.md file

PR #1693, opened 2024-10-12
Docs :update README_FR.md

Innovatorcloudy:
Added french to the document so that our french learners too can understand the value of lightly .


SauravMaheshkar:
We have another contributor suggesting the same (#1690). Once we merge that PR, you could add translations for other languages.

PR #1695, opened 2024-10-13
Readme: brought it to center

Hacker24a9:


guarin (commented):
Hi, thanks for the PR! After looking at it I believe it is better to keep the batches in markdown format and on the left.

PR #1706, opened 2024-10-22
Patch eval metrics to markdown

### LINKED ISSUE #1333: Report all metrics at end of benchmark
We should print all metrics at the end of a benchmark again to make them easier to extract from the logs.

TODO:
- [ ] Collect metrics over a whole benchmark (online, knn, linear, and finetune eval)
- [ ] Print metrics at end of benchmark script
- [ ] Optional: Print also as markdown table

EricLiclair:
- adds changes to return metrics from evals
- prints converted metrics to markdown

TODO's:
Collect metrics over the whole benchmark: 
- - [ ] online, 
- - [x] knn, 
- - [x] linear, and 
- - [x] finetune eval
- [x]  Print metrics at end of benchmark script (currently only printing the converted markdown)
- [x]  Optional: Print also as markdown table

---

prints the results like:
(values randomly generated and passed to the function)

```shell
| Eval Name | Metric Name | Value |
|:---------:|:-----------:|:-----:|
| knn       | val_top1    | 0.44  |
| knn       | val_top5    | 0.27  |
| linear    | val_top1    | 0.26  |
| linear    | val_top5    | 0.27  |
| finetune  | val_top1    | 0.68  |
| finetune  | val_top5    | 0.45  |
```
which prints markdown like,

| Eval Name | Metric Name | Value |
|:---------:|:-----------:|:-----:|
| knn       | val_top1    | 0.44  |
| knn       | val_top5    | 0.27  |
| linear    | val_top1    | 0.26  |
| linear    | val_top5    | 0.27  |
| finetune  | val_top1    | 0.68  |
| finetune  | val_top5    | 0.45  |

---
potentially fixes #1333


benchmarks/imagenet/vitb16/main.py
```diff
@@ -100,10 +108,11 @@ def main(
                 strategy=strategy,
             )
 
+        eval_metrics: EvalMetrics = []
```
guarin, line 111: What do you think about aggregating all the metrics in a simple dictionary?

```suggestion
        eval_metrics = {}
```

benchmarks/imagenet/vitb16/main.py
```diff
@@ -100,10 +108,11 @@ def main(
...
         if skip_knn_eval:
             print("Skipping KNN eval.")
         else:
-            knn_eval.knn_eval(
+            for k, v in knn_eval.knn_eval(
```
guarin, line 115:
```suggestion
            eval_metrics["knn"] = knn_eval.knn_eval(
```

benchmarks/metrics.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def eval_metrics_to_markdown(metrics: EvalMetrics) -> str:
+    lines = []
+    header = f"| Eval Name | Metric Name | Value |"
+    lines.append(header)
+    for metric in metrics:
+        line = f"| {metric.eval_name} | {metric.name} | {metric.value} |"
+        lines.append(line)
+    return "\n".join(lines)
```
guarin, line 39: Let's move this function to the `main.py` files for now even if we have to duplicate it. The goal for the benchmarks is that the code is as simple as possible to make them easy to understand and modify.

We might have to rework some of the benchmarking soon, there were some changes regarding metrics in PyTorch Lightning that could simplify our code as well. AFAIK the `trainer.validate()` functions now return the metrics directly which means we won't have to rely on the callback anymore.

benchmarks/metrics.py
guarin, line 1: It would be great if we could remove the need for this file completely to keep everything as simple as possible.

guarin (changes requested):
Thanks a lot for your PR! I left some comments on how I believe we could simplify it a bit but the code overall looks very good.

benchmarks/metrics.py
EricLiclair: sure. I was first planning to integrate the metric's classes in the lib itself and expose an api to convert the metrics to markdown which would've required scoping + understanding the complete metrics flow. But yes, i agree we can keep it simple. let me add a commit considering the suggestions.

guarin (approved):
Thanks so much, this looks great!

PR #1717, opened 2024-11-05
Allow v1/v2 independent import of torchvision.transforms

liopeer:
## Changes
- adds file `lightly/transforms/torchvision_transforms.py` with:
  - import v2 of `torchvision.transforms` if available, else keep using v1
  - add `ToTensor()` transform that is compatible with v2
- make transforms (irrespective of v1 or v2) available in module `lightly.transforms` as `lightly.transforms.torchvision_transforms`


guarin (approved):
Awesome 🎉

PR #1718, opened 2024-11-06
Add torchvision.transforms v1/v2 independent ToTensor() implementation

liopeer:
## Changes
- add `ToTensor()` transform that is independent of the imported `torchvision.transforms` version (v1 or v2)
- make it available for import from the `lightly.transforms` module
- add testing of output tensor:
  - correct dtype (from `np.uint8` to `torch.float32`)
  - correct shape (from channel last to channel first)
  - value range: from [0, 255] to [0, 1]


tests/transforms/test_torchvision_transforms.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def test_ToTensor() -> None:
+    img_np = np.random.randint(0, 255, (20, 30, 3), dtype=np.uint8)
+    img_pil = Image.fromarray(img_np)
+    img_tens = ToTensor()(img_pil)
+    assert isinstance(img_tens, torch.Tensor), "Expected torch tensor"
+    assert img_tens.shape == (3, 20, 30), "Wrong shape"
+    assert img_tens.dtype == torch.float32, "Wrong dtype"
+    assert img_tens.max() <= 1.0 and img_tens.min() >= 0.0, "Wrong range"
```
guarin, line 15: Two small things:
1. We usually import classes `from torch import Tensor` and then refer to them only via the class name without the module name. For functions we import the module name and the call `module.function()`
2. Don't overwrite the assertion messages (e.g with `"wrong shape"`). Pytest automatically generates detailed assertion messages that are lost if you overwrite them. For simple tests like this I would not mention the "wrong shape", "wrong dtype" etc.

lightly/transforms/torchvision_transforms.py
```diff
@@ -20,3 +21,15 @@
     from torchvision import transforms as torchvision_transforms
 
     _TRANSFORMS_V2 = False
+
+
+def ToTensor() -> Union[torchvision_transforms.Compose, ToTensorV1]:
```
guarin, line 26: Could you add a docstring?

guarin (changes requested):

liopeer:
## Additional Changes
 - removed assertion statements & imported class directly as requested
 - added missing docstring to the transform (tried to stay close to the docstrings of other transforms)
 - renamed the module in which the transform resides to `torchvision_v2compatibility` to resolve name-space issues

lightly/transforms/torchvision_v2compatibility.py
guarin, line 1: Optional: rename to `torchvision_v2_compatibility.py`

guarin (approved):
LGTM! Feel free to resolve and merge my comment.

PR #1720, opened 2024-11-06
BYOL: transition from torchvision.transforms v1 to version indep

liopeer:
## Changes
- use version independent `torchvision.transforms` instead of using only v1


guarin (approved):
LGTM ✅

PR #1721, opened 2024-11-06
DINO: transition from torchvision.transforms v1 to version indep

liopeer:
## Changes
- instead of defaulting to v1 of `torchvision.transforms`, allow using v2 if available


guarin (commented):
LGTM ✅

guarin (approved):

PR #1723, opened 2024-11-06
Remaining Transforms: Import version independent torchvision instead of defaulting to v1

liopeer:
## Changes
- change import from `import torchvision.transforms as T` to `from lightly.transforms.torchvision_v2_compatibility import torchvision_transforms` as T **in all files**


guarin (approved):
LGTM ✅

PR #1730, opened 2024-11-11
allow to check v2 transforms availability from anywhere

liopeer:
## Changes & Explanation
 - the availability of the `torchvision.transforms.v2` was so far only checked on module level
 - now the dependency check is implemented at the same level as other dependency checks and can also be accessed from other modules
 - change is necessary for future implementation of DetCon

### DIFF
--- a/lightly/utils/dependency.py
+++ b/lightly/utils/dependency.py
@@ -42,3 +42,19 @@ def timm_vit_available() -> bool:
     except ImportError:
         return False
     return True
+
+
+@functools.lru_cache(maxsize=1)
+def torchvision_transforms_v2_available() -> bool:
+    """Checks if torchvision supports the transforms.v2 API.
+
+    Returns:
+        True if transforms.v2 are available, False otherwise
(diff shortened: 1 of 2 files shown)


lightly/utils/dependency.py
```diff
@@ -42,3 +42,19 @@ def timm_vit_available() -> bool:
...
+def torchvision_transforms_v2_available() -> bool:
+    """Checks if torchvision supports the transforms.v2 API.
+
+    Returns:
+        True if transforms.v2 are available, False otherwise
+    """
+    try:
+        from torchvision.transforms import v2 as torchvision_transforms
+    except ImportError:
+        from torchvision import transforms as torchvision_transforms
```
guarin, line 57:
```suggestion
    try:
        from torchvision.transforms import v2
    except ImportError:
```

guarin (changes requested):

guarin (approved):
LGTM ✅

PR #1743, opened 2024-11-22
refactor: remove opencv dependencies partial

vectorvp:


vectorvp:
It looks like there is a problem with permissions:
```
RuntimeWarning: Failed to open container for /tmp/tmpl6i5ndzm/output-0.avi; Caught error: [Errno 13] Permission denied: '/tmp/tmpl6i5ndzm/output-0.avi'
```

vectorvp:
@guarin, do you know what could be the reason?

vectorvp:
@guarin, can I help somehow to resolve this?

guarin:
I am honestly not sure what the problem is and I sadly don't have time to reproduce it right now. At least in CI it is reproducible so there is something wrong. The only small thing I saw is that the `frames_over_videos ` are in `(T, W, H, C)` format instead of `(T, H, W, C)` but this shouldn't trigger the error.

vectorvp:
@guarin, I've resolved the issue, please check. The problem was with video format.

guarin (approved):
Thanks for taking the time and fixing this!

PR #1744, opened 2024-11-26
refactor: add typing to tests/models

vectorvp:


vectorvp:
@guarin, please take a look. I have added also typing for all 3 files in tests/models.

tests/models/modules/test_masked_autoencoder.py
```diff
@@ -14,13 +14,15 @@
     dependency.torchvision_vit_available(), "Torchvision ViT not available"
 )
 class TestMAEEncoder(unittest.TestCase):
-    def _vit(self):
+    def _vit(self) -> torchvision.models.vision_transformer.VisionTransformer:
```
guarin, line 17: I believe you have to add a `from __future__ import annotations` at the top of the file. Otherwise the annotation is evaluated when loading the file which breaks for old torchvision version.

You can also import the class only when type checking:
```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
     from torchvision.models.vision_transformer import VisionTransformer
```

If necessary you can also put it in the `if dependency.torchvision_vit_available()` block to only import if `VisionTransformer` is available.

guarin (changes requested):
This is great, thanks for the PR! I left a comment on how to fix the failing tests.

tests/models/modules/test_masked_autoencoder.py
vectorvp: Thank you, the issue is resolved.

guarin (approved):
Awesome, thanks!

PR #1745, opened 2024-11-27
replaces google tracked badges with shields.io

japrescott:
closes lig-5584
- replace google colab badges (which get blocked by most ad/tracking blockers) with img.shields

previously;
colab-badge (1)

after;
after-badge


guarin (approved):
LGTM

Issue #1462, opened 2023-12-20
New SSL algorithm called SparK: Sparse and Hierarchical masKed modeling


Djoels:
It would be great if this new MAE-style method called SparK was introduced to lightly.

Paper: https://arxiv.org/abs/2301.03580 featured in ICLR'23 Spotlight
Code: https://github.com/keyu-tian/SparK

It was successfully applied to medical image applications, as documented in this Nature paper: 
https://github.com/Wolfda95/SSL-MedicalImagining-CL-MAE/tree/main

philippmwirth:
Hey @Djoels , thanks for bringing this up! We will take a look shortly and add it to the model tracker if relevant 🙂

guarin:
This looks interesting indeed, thanks a lot for the issue! Added it to the methods tracker and will consider it for the paper session next week.

johnsutor:
I can take this issue.

guarin:
Thanks for looking into this @johnsutor! The original codebase implements the sparse net in a quite hacky way (see code [here](https://github.com/keyu-tian/SparK/blob/main/pretrain/encoder.py)) and I was wondering whether it would be possible to pass the masks explicitly to the forward function instead of assigning them to a global variable. Maybe this would be interesting to explore, wdyt?

johnsutor:
I'll investigate and get back to you!

johnsutor:
Seems fairly straightforward to achieve based on https://github.com/keyu-tian/SparK/tree/main/pretrain#regarding-sparse-convolution. I don't mind giving it a stab, my thoughts are to implement the ResNet within https://github.com/lightly-ai/lightly/tree/master/lightly/models, just naming the file something like `spark.py`, if this sounds good I'll give it a go.

guarin:
Sounds good! Thanks a lot for looking into it.

Maybe create a `lightly/models/sparse` subdirectory and put it there. You could even name the file `sparse_resnet.py`. And it would be create if you could keep the same structure as the original resnet in torchvision. Then it would be easy to convert from sparse resnet to dense resnet and vice-versa.

johnsutor:
I went ahead and implemented a resnet compatible with the standard `torchvision` library, so that we don't have to add `timm` as a dependency. 

Furthermore, I achieved passing the mask at runtime without setting a global variable using a pre-forward hook. This is how it looks so far:

```py
class SparseEncoder(nn.Module):
[...]

PR #1750, opened 2024-12-17
SparK Implementation

johnsutor:
This is a WIP implementation of the SparK algorithm, as requested by #1462

### DIFF
--- a/lightly/models/modules/sparse/spark.py
+++ b/lightly/models/modules/sparse/spark.py
@@ -1,7 +1,54 @@
-# Code adapted from https://github.com/keyu-tian/SparK/blob/main/pretrain/encoder.py
+# Code adapted from https://github.com/keyu-tian/SparK/blob/main/pretrain/encoder.py and https://github.com/huggingface/pytorch-image-models/blob/main/timm/layers/drop.py
+
+from typing import Literal, Optional
 
 import torch
 import torch.nn as nn
+
+
+def drop_path(x, drop_prob: float = 0.0, training: bool = False):
+    """Drop paths (Stochastic Depth) per sample (when applied in main path of residual blocks).
+
+    This is the same as the DropConnect impl I created for EfficientNet, etc networks, however,
+    the original name is misleading as 'Drop Connect' is a different form of dropout in a separate paper...
+    See discussion: https://github.com/tensorflow/tpu/issues/494#issuecomment-532968956 ... I've opted for
+    changing the layer and argument names to 'drop path' rather than mix DropConnect as a layer name and use
+    'survival rate' as the argument.
+
+    Attributes:
+        x:
+            The input tensor.
+        drop_prob:
+            The drop probability of the path. Default: 0.
+        training:
+            Whether the model is in training mode. Default: False
+    """
+    if drop_prob == 0.0 or not training:
+        return x
+    keep_prob = 1 - drop_prob
+    shape = (x.shape[0],) + (1,) * (
+        x.ndim - 1
+    )  
+    random_tensor = keep_prob + torch.rand(shape, dtype=x.dtype, device=x.device)
+    random_tensor.floor_()  
+    output = x.div(keep_prob) * random_tensor
+    return output
+
+
+class DropPath(nn.Module):
+    """Drop paths (Stochastic Depth) per sample  (when applied in main path of residual blocks).
+
+    Attributes:
+        drop_prob:
+            The drop probability of the path. Default: None.
+    """
+
+    def __init__(self, drop_prob: Optional[float] = None):
+        super(DropPath, self).__init__()
+        self.drop_prob = drop_prob
+
+    def forward(self, x):
+        return drop_path(x, self.drop_prob, self.training)
 
 
 def _get_active_ex_or_ii(
@@ -15,16 +62,15 @@ def _get_active_ex_or_ii(
         active_ex
         if returning_active_ex
         else active_ex.squeeze(1).nonzero(as_tuple=True)
-    )  # ii: bi, hi, wi
+    )  
 
 
-def sp_conv_forward(self, x: torch.Tensor):
+def sp_conv_forward(self, x: torch.Tensor) -> torch.Tensor:
     x = super(type(self), self).forward(x)
     x *= _get_active_ex_or_ii(
         self.sparse_mask.mask, H=x.shape[2], W=x.shape[3], returning_active_ex=True
-    )  # (BCHW) *= (B1HW), mask the output of conv
+    ) 
     return x
-
 
 def sp_bn_forward(self, x: torch.Tensor):
     ii = _get_active_ex_or_ii(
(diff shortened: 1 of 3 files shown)


guarin:
Thanks a lot for the PR! The review might take a while due to the upcoming holiday season.

lightly/models/sparse/encoder.py
```diff
@@ -0,0 +1,239 @@
...
+    Basic class to store the mask for the sparse model.
+    """
+
+    def __init__(self):
+        self.mask: Union[Tensor, None] = None
+
+
+class SparseEncoder(nn.Module):
+    def __init__(
+        self,
```
guarin, line 34: Looking at the code I don't think we need this class (yet). It also mixes sparsity logic with hierarchical features which will make the class hard to maintain. Hierarchical features are already natively supported by TIMM or can be handled with a hook in torchvision, I don't think we should reimplement the logic here.

IMO the sparsity logic should be directly added when one implements the final model. Let's say we add `SparK` to the benchmarks in `benchmarks/imagenet/resnet50`. Then I would imagine that the implementation looks like this:
```
class SparK(LightningModule):
    def __init__(...):
        self.backbone = resnet50()
        self.sparse_mask = SparseMask()
        self.sparse_backbone =  dense_model_to_sparse(self.backbone, self.sparse_mask)
    
    def training_step(...):
        self.sparse_mask.mask = random_mask()
        features = self.sparse_backbone(x)
        ...
```

If hierarchical features are needed then this logic should be handled by the non-sparse backbone.

lightly/models/spark.py
guarin, line 1: Could you move this file to `lightly/models/modules/spark.py`? The models in `lightly/models` like `lightly/models/simclr.py` etc. are deprecated and we don't want to add more models there.

lightly/models/spark.py
```diff
@@ -0,0 +1,280 @@
...
+import torch
+import torch.nn as nn
+from lightly.models.sparse.encoder import SparseResnet
+from lightly.models.modules.sparse.spark import SparseBatchNorm2d, SparseSyncBatchNorm2d, SparseConvNeXtLayerNorm
+
+def is_pow2n(n: int) -> bool:
+    return n > 0 and (n & (n - 1)) == 0
+
+
+class UNetBlock(nn.Module):
```
guarin, line 17:
```suggestion
class SparKUNetBlock(nn.Module):
```

lightly/models/spark.py
```diff
@@ -0,0 +1,280 @@
...
+        a UNet block with 2x up sampling
+        """
+        super().__init__()
+        self.up_sample = nn.ConvTranspose2d(
+            cin, cin, kernel_size=4, stride=2, padding=1, bias=True
+        )
+        self.conv = nn.Sequential(
+            nn.Conv2d(cin, cin, kernel_size=3, stride=1, padding=1, bias=False),
+            bn2d(cin),
+            nn.ReLU6(inplace=True),
```
guarin, line 29:
```suggestion
            nn.ReLU(inplace=True),
```

Typo?

lightly/models/spark.py
```diff
@@ -0,0 +1,280 @@
...
+            nn.Conv2d(cin, cout, kernel_size=3, stride=1, padding=1, bias=False),
+            bn2d(cout),
+        )
+
+    def forward(self, x):
+        x = self.up_sample(x)
+        return self.conv(x)
+
+
+class LightDecoder(nn.Module):
```
guarin, line 39:
```suggestion
class SparKLightDecoder(nn.Module):
```

lightly/models/spark.py
```diff
@@ -0,0 +1,280 @@
...
+    def __init__(self, up_sample_ratio, width=768, sync_batch_norm=True):
+        super().__init__()
```
guarin, line 41:
```suggestion
    def __init__(self, up_sample_ratio, width=768):
```

It shouldn't be necessary to add `sync_batch_norm` as argument as this can be handled outside the class. Torch has utility functions to automatically replace batch norms with sync batch norm.

lightly/models/spark.py
```diff
@@ -0,0 +1,280 @@
...
+        for i, d in enumerate(self.dec):
+            if i < len(to_dec) and to_dec[i] is not None:
+                x = x + to_dec[i]
+            x = self.dec[i](x)
+        return self.proj(x)
+
+    def extra_repr(self) -> str:
+        return f"width={self.width}"
+
+    def initialize(self):
```
guarin, line 68:
```suggestion
    def init_weights(self):
```

Make it clear that this only initializes the weights.

lightly/models/spark.py
```diff
@@ -0,0 +1,280 @@
...
+            return recon_loss
+        
+        masked_bchw = inp_bchw * active_b1hw
+        rec_bchw = self.unpatchify(rec * var + mean)
+        rec_or_inp = torch.where(active_b1hw, inp_bchw, rec_bchw)
+        return inp_bchw, masked_bchw, rec_or_inp
+
+    
+    def patchify(self, bchw):
+        p = self.downsample_ratio
```
guarin, line 229: Patchify and unpatchify already exist in `lightly/models/utils.py`

lightly/models/spark.py
```diff
@@ -0,0 +1,280 @@
...
+                kernel_size = 1 if i <= 0 else 3
+                densify_proj = nn.Conv2d(e_width, d_width, kernel_size=kernel_size, stride=1, padding=kernel_size // 2, bias=True)
+                print(f'[SparK.__init__, densify {i+1}/{self.hierarchy}]: densify_proj(ksz={kernel_size}, #para={sum(x.numel() for x in densify_proj.parameters()) / 1e6:.2f}M)')
+            self.densify_projs.append(densify_proj)
+            d_width //= 2
+        
+        print(f'[SparK.__init__] dims of mask_tokens={tuple(p.numel() for p in self.mask_tokens)}')
+        
+    
+    def mask(self, B: int, device, generator=None):
```
guarin, line 161: I am not sure how the masking works exactly but we already have a bunch of masking functions here https://github.com/lightly-ai/lightly/blob/356ae56fd8f6d6bbc548d4d189d9444f780a62ca/lightly/models/utils.py#L571-L838

If none of the existing functions are compatible then please add a new function there.

lightly/models/spark.py
```diff
@@ -0,0 +1,280 @@
...
+                nn.init.kaiming_normal_(m.weight, mode="fan_out", nonlinearity="relu")
+                if m.bias is not None:
+                    nn.init.constant_(m.bias, 0.0)
+            elif isinstance(
+                m, (nn.LayerNorm, nn.BatchNorm1d, nn.BatchNorm2d, nn.SyncBatchNorm)
+            ):
+                nn.init.constant_(m.bias, 0)
+                nn.init.constant_(m.weight, 1.0)
+
+class SparK(nn.Module):
```
guarin, line 88: We shouldn't add this class directly to the package. The goal of Lightly is to provide building blocks for models and not the full model implementation, we only add those in the benchmarks (e.g. `benchmarks/imagenet/resnet50/simclr.py`). So this class should eventually move to the benchmarks module. The different functionalities inside the class (masking, patchify, etc.) should move to `lightly/models/utils.py`.

lightly/models/sparse/encoder.py
```diff
@@ -0,0 +1,239 @@
...
+from torch import List, Optional, Tensor
+
+from lightly.models.modules.sparse.spark import (
+    SparseAvgPooling,
+    SparseBatchNorm2d,
+    SparseConv2d,
+    SparseMaxPooling,
+    SparseSyncBatchNorm2d,
+    SparseConvNeXtBlock,
+    SparseConvNeXtLayerNorm
```
guarin, line 15: `SparseConvNeXtLayerNorm ` is imported but not used

lightly/models/modules/sparse/spark.py
```diff
@@ -0,0 +1,258 @@
...
+                return super(SparseConvNeXtLayerNorm, self).forward(x)
+
+    def __repr__(self):
+        return (
+            super(SparseConvNeXtLayerNorm, self).__repr__()[:-1]
+            + f', ch={self.data_format.split("_")[-1]}, sp={self.sparse})'
+        )
+
+
+class SparseConvNeXtBlock(nn.Module):
```
guarin, line 189: What is the difference between this ConvNeXt implementation and the ones from torchvision and timm? Would it be possible to inherit from torchvisions `CNBlock` instead as it is used in `dense_model_to_sparse`? This way we wouldn't have to implement `DropPath` ourselves.

lightly/models/sparse/encoder.py
guarin, line 1: Let's move code in this file to `lightly/models/modules/spark_sparse.py`

lightly/models/modules/sparse/spark.py
guarin, line 1: Let's move the code in this file to `lighlty/models/modules/spark_sparse.py`

lightly/models/sparse/decoder.py
guarin, line 1: Let's not add this emtpy file

guarin (changes requested):
Thanks again for the PR! This will take a while to get merged as it contains many changes. I left a lot of comments/questions regarding the various parts. Let us know if you need help, we can also support you in making the changes or take over parts of the PR if that helps.

johnsutor:
Thanks for all the comments! I'll make some changes later today and I'll let you know how much help I'll need given the scope of this project.

guarin:
Sure, contributions are very welcome :)

gabrielfruet:
@guarin Starting new PR, for now, a draft. Fixing some things you pointed out. 

I will first fix those that do not imply in great modifications, then i will test and see if its working properly, model training and etc, then i will refactor it to adhere to better engineering patterns. Is that okay to you?

PR #1751, opened 2024-12-18
Type NTXEntLoss and tests

philippmwirth:
# Type NTXEntLoss and tests

* Use composition over inheritance for `MemoryBankModule`.
* Add typehints.
* Refactor and type tests.


guarin (approved):
LGTM

PR #1752, opened 2024-12-18
Pin Ubuntu version

philippmwirth:
# Pin Ubuntu version

GH started rolling out Ubuntu 24 as `ubuntu-latest` which kills some of our tests.


guarin (approved):
LGTM

PR #1753, opened 2024-12-18
Type VICRegLoss and tests

philippmwirth:
# Type VICRegLoss and tests

* Add typehints
* Refactor and add typehints to tests


tests/loss/test_vicreg_loss.py
```diff
@@ -38,10 +36,10 @@ def test_forward_pass(self):
             # symmetry
             l1 = loss(x0, x1)
             l2 = loss(x1, x0)
-            self.assertAlmostEqual((l1 - l2).pow(2).item(), 0.0)
+            assert (l1 - l2).pow(2).item() == pytest.approx(0.0)
```
MalteEbner, line 39: Why assert the equality so complicated? Same a few lines later.
```suggestion
            assert l1.item() == pytest.approx(l2.item())
```

MalteEbner (approved):
One suggestion for making the code easier, otherwise LGTM

PR #1756, opened 2024-12-23
Typecheck TiCo loss

philippmwirth:
# Typecheck TiCo loss


MalteEbner (approved):
Nice that you also refactored the tests!

PR #1757, opened 2024-12-24
Add files via upload

Abhishekjha3515:
Added MACL(Model Aware Contrastive learning )  Loss Function in Loss folder

### DIFF
--- a/lightly/loss/macl_loss.py
+++ b/lightly/loss/macl_loss.py
@@ -1,0 +1,58 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+class MACLLoss(nn.Module):
+    def __init__(self, temperature=0.1, alpha=0.5, A_0=0):
+        super().__init__()
+        self.t_0 = temperature
+        self.alpha = alpha
+        self.A_0 = A_0
+
+    def mask_correlated_samples(self, batch_size):
+        N = 2 * batch_size
+        mask = torch.ones((N, N), dtype=bool)
+        mask = mask.fill_diagonal_(0)
+        for i in range(batch_size):
+            mask[i, batch_size + i] = 0
+            mask[batch_size + i, i] = 0
+        return mask
+
+    def forward(self, z0, z1):
+        """
+        Args:
+            z0: first view embeddings
+            z1: second view embeddings
+        """
+        # Normalize embeddings
+        z0 = F.normalize(z0, dim=-1, p=2)
+        z1 = F.normalize(z1, dim=-1, p=2)
+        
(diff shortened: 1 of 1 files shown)


lightly/loss/macl_loss.py
```diff
@@ -0,0 +1,58 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+class MACLLoss(nn.Module):
+    def __init__(self, temperature=0.1, alpha=0.5, A_0=0):
+        super().__init__()
+        self.t_0 = temperature
+        self.alpha = alpha
+        self.A_0 = A_0
```
guarin, line 10:
```suggestion
    def __init__(self, temperature: float = 0.1, alpha: float = 0.5, A_0: float = 0):
        super().__init__()
        self.temperature = temperature
        self.alpha = alpha
        self.A_0 = A_0
```

Could you also add a short docstring that describes what the loss does, what the parameters are, and add a reference to the paper that introduced the loss? You can follow the example here: https://github.com/lightly-ai/lightly/blob/356ae56fd8f6d6bbc548d4d189d9444f780a62ca/lightly/loss/ntx_ent_loss.py#L17

guarin (changes requested):
Thanks for the contribution and sorry for the late review!

Could you add a minimal test in `tests/loss/test_macl_loss.py`? It is enough to pass some dummy values and check that the loss returns a valid value.

lightly/loss/macl_loss.py
```diff
@@ -0,0 +1,58 @@
...
+        # Concatenate embeddings
+        out = torch.cat([z0, z1], dim=0)
+        batch_size = z0.shape[0]
+        n_samples = len(out)
+        
+        # Compute similarity matrix
+        cov = out @ out.T
+        
+        # Get positive and negative pairs
+        mask = self.mask_correlated_samples(batch_size).to(out.device)
```
guarin, line 40:
```suggestion
        mask = cov.new_ones(cov.shape, dtype=bool)
        mask.diagonal()[:] = False
        mask.diagonal(batch_size)[:] = False
        mask.diagonal(-batch_size)[:] = False
```
This is a bit more concise and guarantees that the mask is on the same device as the input tensors.

Abhishekjha3515: thanks for the input , i was busy with my office work I will do all the changes u have  suggested by this weekend

lightly/loss/macl_loss.py
Abhishekjha3515: ok i will add the loss doc also

PR #1759, opened 2024-12-27
Typecheck SwAV loss

philippmwirth:
# Typecheck SwAV loss

### DIFF
--- a/lightly/loss/swav_loss.py
+++ b/lightly/loss/swav_loss.py
@@ -1,18 +1,19 @@
-from typing import List
+from typing import List, Union
 
 import torch
 import torch.distributed as dist
 import torch.nn as nn
 import torch.nn.functional as F
+from torch import Tensor
 
 
 @torch.no_grad()
 def sinkhorn(
-    out: torch.Tensor,
+    out: Tensor,
     iterations: int = 3,
     epsilon: float = 0.05,
     gather_distributed: bool = False,
-) -> torch.Tensor:
+) -> Tensor:
     """Distributed sinkhorn algorithm.
 
     As outlined in [0] and implemented in [1].
@@ -113,7 +114,7 @@ class SwaVLoss(nn.Module):
         self.sinkhorn_epsilon = sinkhorn_epsilon
         self.sinkhorn_gather_distributed = sinkhorn_gather_distributed
 
-    def subloss(self, z: torch.Tensor, q: torch.Tensor):
+    def subloss(self, z: Tensor, q: Tensor) -> Tensor:
         """Calculates the cross entropy for the SwaV prediction problem.
 
         Args:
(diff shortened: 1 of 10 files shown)


lightly/loss/swav_loss.py
```diff
@@ -131,10 +132,10 @@ def subloss(self, z: torch.Tensor, q: torch.Tensor):
 
     def forward(
         self,
-        high_resolution_outputs: List[torch.Tensor],
-        low_resolution_outputs: List[torch.Tensor],
-        queue_outputs: List[torch.Tensor] = None,
-    ):
+        high_resolution_outputs: List[Tensor],
+        low_resolution_outputs: List[Tensor],
+        queue_outputs: Union[List[Tensor], None] = None,
```
MalteEbner, line 137: We can also use the new syntax here, we already use it at other places: https://github.com/lightly-ai/lightly/blob/53d1af4c01d674e78d0ba00d38a8239d1c31b7e6/lightly/models/modules/ijepa_timm.py#L103

tests/loss/test_swav_loss.py
```diff
@@ -36,34 +31,25 @@ def test_forward_pass(self):
             for sinkhorn_iterations in range(3):
```
MalteEbner, line 31: This should be a parametrised test.

tests/loss/test_swav_loss.py
```diff
@@ -36,34 +31,25 @@ def test_forward_pass(self):
...
-        queue_length = 128
         queue = [torch.eye(128, 32) for i in range(n_high_res)]
 
         for n_low_res in range(6):
             for sinkhorn_iterations in range(3):
                 criterion = SwaVLoss(sinkhorn_iterations=sinkhorn_iterations)
                 low_res = [torch.eye(n, n) for i in range(n_low_res)]
+                loss = criterion(high_res, low_res, queue)
+                # loss should be almost zero for unit matrix
+                assert loss.cpu().numpy() < 0.5
```
MalteEbner, line 50: This should be a parametrised test.

tests/loss/test_swav_loss.py
```diff
@@ -87,16 +69,12 @@ def test_forward_pass_1d(self):
             for sinkhorn_iterations in range(3):
                 criterion = SwaVLoss(sinkhorn_iterations=sinkhorn_iterations)
                 low_res = [torch.eye(n, 1) for i in range(n_low_res)]
+                loss = criterion(high_res, low_res)
+                # loss should be almost zero for unit matrix
+                assert loss.cpu().numpy() < 0.5
```
MalteEbner, line 74: This should be a parametrised test.

tests/loss/test_swav_loss.py
```diff
@@ -24,10 +22,7 @@ def test__sinkhorn_gather_distributed_dist_not_available(
             SwaVLoss(sinkhorn_gather_distributed=True)
         mock_is_available.assert_called_once()
 
-
-class TestSwaVLossUnitTest(unittest.TestCase):
-    # Old tests in unittest style, please add new tests to TestSwavLoss using pytest.
-    def test_forward_pass(self):
+    def test_forward_pass(self) -> None:
```
MalteEbner, line 25: This should be a parametrised test instead of iterating with 2 for-loops. 
Same applies to the other tests.

MalteEbner (commented):
I would definitely use the new typing syntax wherever possible.

lightly/loss/swav_loss.py
philippmwirth: I'd like to keep consistent with the rest of the file. IMO it doesn't make sense to mix the styles. We should decide for one way and then pull it through the whole code base.

MalteEbner (approved):
LGTM

PR #1777, opened 2025-01-10
lionel-lig-5887 remove heuristics in masked pooling

liopeer:
## Changes
- removes automatic detection of `num_cls` and makes this argument non-optionable
- removes the test for automatic inference of `num_cls`
- fixes an issue with the argument `reduce`, which was not actually propagated
- brush up the docstring while at it

## Necessity
- using the heuristics was bad practice and could lead to issue when calling the function several times with different masks (e.g. different tensor output shapes from the two calls)
- user of function should be aware of the classes in the dataset anyway

### DIFF
--- a/lightly/models/utils.py
+++ b/lightly/models/utils.py
@@ -40,8 +40,8 @@ def pool_masked(
             integer indices.
         num_cls: The number of classes in the possible masks.
 
-    Returns: 
-        A tensor of shape :math:`(B, C, N)` or :math:`(C, N)` where :math:`N` 
+    Returns:
+        A tensor of shape :math:`(B, C, N)` or :math:`(C, N)` where :math:`N`
         corresponds to `num_cls`.
     """
     if source.dim() == 3:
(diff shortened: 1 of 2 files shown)


lightly/models/utils.py
```diff
@@ -28,23 +28,21 @@
...
+            reduced.
+        mask: Integer tensor of shape :math:`(B, H, W)` or :math:`(H, W)` containing the
+            integer indices.
+        num_cls: The number of classes in the possible masks.
 
     Returns:
-        A tensor of shape (B, C, N) or (C, N) where N is the number of unique elements
-        in `mask` or `num_cls` if specified.
+        A tensor of shape :math:`(B, C, N)` or :math:`(C, N)` where :math:`N`
+        corresponds to `num_cls`.
```
guarin, line 45: Minor: Might make it a bit easier to understand:

```suggestion
        A tensor of shape :math:`(B, C, num_cls)` or :math:`(C, num_cls)`.
```

guarin (approved):
LGTM!

PR #1778, opened 2025-01-10
Guarin lig 5347 add dino vit

guarin:


yutong-xiang-97:
Close this PR as it is addressed in #1802 and #1823

PR #1779, opened 2025-01-10
fix torchvision dependency test

liopeer:
## Changes
 - changes how we test if the `torchvision.transforms.v2` is available

## Necessity
 - Testing if we can import `torchvision.transforms.v2` is not enough since we need to make sure that we install a version that also has a stable version of the API. Trying to mport `Mask` should be safer.

### DIFF
--- a/lightly/utils/dependency.py
+++ b/lightly/utils/dependency.py
@@ -52,7 +52,7 @@ def torchvision_transforms_v2_available() -> bool:
         True if transforms.v2 are available, False otherwise
     """
     try:
-        from torchvision.transforms import v2
+        from torchvision.tv_tensors import Mask
     except ImportError:
         return False
     return True
--- a/lightly/utils/dependency.py
+++ b/lightly/utils/dependency.py
@@ -46,13 +46,16 @@ def timm_vit_available() -> bool:
 
 @functools.lru_cache(maxsize=1)
 def torchvision_transforms_v2_available() -> bool:
-    """Checks if torchvision supports the transforms.v2 API.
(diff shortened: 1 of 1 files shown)


lightly/utils/dependency.py
```diff
@@ -52,7 +52,7 @@ def torchvision_transforms_v2_available() -> bool:
         True if transforms.v2 are available, False otherwise
     """
     try:
-        from torchvision.transforms import v2
+        from torchvision.tv_tensors import Mask
```
guarin, line 55: Does this mean if `Mask` exists the whole v2 transforms API is stable? Or is this just for the masking part? Otherwise I would suggest to add  `torchvision_transforms_v2_mask_available` function instead.

Could you also add a comment explaining why we don't import `torchvision.transforms.v2` here?

liopeer: The issue is that `torchvision.transforms.v2` was available before the `tv_tensors` module was introduced, i.e. `transforms.v2` was introduced in [`v0.15.1`](https://github.com/pytorch/vision/releases/tag/v0.15.1), but the `tv_tensors` module only in [`v0.16.0`](https://github.com/pytorch/vision/releases/tag/v0.16.0):

> [transforms.v2] Renaming: torchvision.datapoints.Datapoint -> torchvision.tv_tensors.TVTensor (https://github.com/pytorch/vision/pull/7904, https://github.com/pytorch/vision/pull/7894)

I corrected it to just check if `tv_tensors` is available and added a comment.

guarin (approved):

PR #1784, opened 2025-01-27
Update MAE model docs

stegmuel:
# Changes

- Adds more info to the MAE model example here: https://docs.lightly.ai/self-supervised-learning/examples/mae.html
- Example of updated version:
Screenshot from 2025-01-28 08-39-00


IgorSusmelj (approved):
LGTM

PR #1785, opened 2025-01-28
Fix typo:  Dict() instead of dict() return class instanciation error

gatienc:
Dict is the type and dict is the class, python return an error when trying to create an instance of Dict()


guarin (approved):
LGTM! Thanks for the fix!

PR #1786, opened 2025-01-28
adjust pagination to be defensive at 2500 entries

japrescott:
closes lig-6096
- adjust pagination to be defensive 2500 entires to not go over 32MB limit.


MalteEbner (approved):
LGTM

PR #1787, opened 2025-01-28
add additional test to pooling

liopeer:
## Description
- added an additional test to the DetCon pooling operation, to ensure that this is not the problem


stegmuel (approved):
LGTM.

PR #1797, opened 2025-02-04
Remove-docs-build-warnings

liopeer:
## Description
- reduces warnings from 46 to 13
- all remaining warnings are in the `api` subpackage
- additionally adds another make command for a clean docs build – otherwise some warnings will not be shown due to cache-ing

### DIFF
--- a/docs/source/getting_started/benchmarks.rst
+++ b/docs/source/getting_started/benchmarks.rst
@@ -14,7 +14,7 @@ List of available benchmarks:
 ImageNet1k
 ----------
 
-- `Dataset <https://image-net.org/download.php>`_
+- `ImageNet1k Dataset <https://image-net.org/download.php>`_
 - `Code <https://github.com/lightly-ai/lightly/tree/master/benchmarks/imagenet/resnet50>`_
 
 The following experiments have been conducted on a system with 2x4090 GPUs.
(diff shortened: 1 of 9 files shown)


docs/source/getting_started/benchmarks.rst
JonasWurst, line 1: The changes in this file look fine, though I struggling to understand how they are connected to warning reduction.

docs/source/getting_started/benchmarks.rst
```diff
@@ -49,7 +49,7 @@ Found a missing model? Track the progress of our planned benchmarks on `GitHub <
 ImageNet100
 -----------
 
-- `Dataset <https://image-net.org/download.php>`_
+- `ImageNet100 Dataset <https://image-net.org/download.php>`_
 - :download:`Code <benchmarks/imagenet100_benchmark.py>`
```
JonasWurst, line 53: You also want to give more details on the code link?

docs/source/lightly.utils.rst
```diff
@@ -34,7 +34,7 @@ lightly.utils
    :members:
 
 .version_compare
----------------
+-----------------
```
JonasWurst, line 37: what was the warning here?

JonasWurst (approved):

docs/source/lightly.utils.rst
liopeer: Simply a too short underline. Sphinx is really allergic to cases where underlines are shorter than the title above – for whatever reason.

docs/source/getting_started/benchmarks.rst
liopeer: It's not actually about details, Sphinx throws warnings when using the same name `Dataset <some link>` several times in a single module/file. I think it's because you could use them for cross referencing. Why does it not throw warnings for the `Code <>` links? No idea tbh, though I suspect it has something to do with the trailing underscores – maybe I could have just left those away instead of changing the name of the link.

PR #1798, opened 2025-02-05
LIG-6047 Fix trainer strategy issues with mae examples

### LINKED ISSUE #1775: Exemplar MAE with DDP does not work
Hello all,

Thanks for the great open source package. I noticed the  example for MAE with pytorch lightning training DDP simply does not work? There is an  to be an issue with unused parameters. i.e. 

```
RuntimeError: It looks like your LightningModule has parameters that were not used in producing the loss returned by training_step. If this is intentional, you must enable the detection of unused parameters in DDP, either by setting the string value `strategy='ddp_find_unused_parameters_true'` or by setting the flag in the strategy with `strategy=DDPStrategy(find_unused_parameters=True)`.
    if torch.is_grad_enabled() and self.reducer._rebuild_buckets()
```

If you print which parameters are do not have a gradient, you'll see they are `vit.pos_embed`, `vit.head.weight`, `vit.head.bias`. The unused head parameter makes sense, but I don't see why the `vit.pos_embed` would be unused.

I'd really appreciate if could confirm this is an issue with the [example on main](https://github.com/lightly-ai/lightly/blob/master/examples/pytorch_lightning_distributed/mae.py) and if fixing this will be on the roadmap.

Thanks!

yutong-xiang-97:
Close #1775

### DIFF
--- a/examples/pytorch_lightning/mae.py
+++ b/examples/pytorch_lightning/mae.py
@@ -118,6 +118,5 @@ trainer = pl.Trainer(
     max_epochs=10,
     devices=1,
     accelerator=accelerator,
-    strategy="ddp_find_unused_parameters_true",
 )
 trainer.fit(model=model, train_dataloaders=dataloader)
--- a/examples/notebooks/pytorch_lightning/mae.ipynb
+++ b/examples/notebooks/pytorch_lightning/mae.ipynb
@@ -211,7 +211,6 @@
     "    max_epochs=10,\n",
     "    devices=1,\n",
     "    accelerator=accelerator,\n",
-    "    strategy=\"ddp_find_unused_parameters_true\",\n",
     ")\n",
     "trainer.fit(model=model, train_dataloaders=dataloader)"
    ]


examples/pytorch_lightning/mae.py
```diff
@@ -114,5 +114,10 @@ def target_transform(t):
 
 accelerator = "gpu" if torch.cuda.is_available() else "cpu"
 
-trainer = pl.Trainer(max_epochs=10, devices=1, accelerator=accelerator)
+trainer = pl.Trainer(
+    max_epochs=10,
+    devices=1,
+    accelerator=accelerator,
+    strategy="ddp_find_unused_parameters_true",
```
guarin, line 121: This example shouldn't be updated as it is for a single GPU and DDP doesn't make sense in that case. Could you revert this change?

guarin (changes requested):
Needs one small change otherwise LGTM! 🚀

guarin (approved):
LGTM ✅

PR #1803, opened 2025-02-13
Fix multiple issues with loss tests

yutong-xiang-97:
See https://linear.app/lightly/issue/LIG-6189/fix-multiple-issues-with-loss-tests


liopeer (approved):
LGTM!

PR #1810, opened 2025-03-13
test(BTLoss): use default values for `torch.allclose`

adosar:
Follow-up to #1806: see https://github.com/lightly-ai/lightly/pull/1806#discussion_r1970063586


liopeer (approved):
LGTM & thank you so much for the follow-up PR! I did not have the time yet to do it.

PR #1813, opened 2025-04-07
Remove `verbose` in `CosineWarmUpScheduler`

yutong-xiang-97:
The parent class of `CosineWarmUpScheduler`, `torch.optim.lr_scheduler.LambdaLR` has a `verbose` argument which is deprecated since version 2.2 ([explained here](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.LambdaLR.html#torch.optim.lr_scheduler.LambdaLR)). We removed it as it is not used anywhere in the codebase.


stegmuel (approved):
LGTM.

PR #1819, opened 2025-04-28
Add seed_everything import for reproducibility in benchmarks

yvesyue:
This PR imports `seed_everything` from PyTorch Lightning and invokes it at the start for 'main.py'  of both the ResNet50 and ViT-B/16 benchmark scripts.

### DIFF
--- a/benchmarks/imagenet/resnet50/main.py
+++ b/benchmarks/imagenet/resnet50/main.py
@@ -89,7 +89,7 @@ def main(
     ckpt_path: Union[Path, None],
     float32_matmul_precision: str,
     strategy: str,
-    seed: int | None = None
+    seed: int | None = None,
 ) -> None:
     print_rank_zero(f"Args: {locals()}")
     seed_everything(seed, workers=True, verbose=True)
--- a/benchmarks/imagenet/resnet50/main.py
+++ b/benchmarks/imagenet/resnet50/main.py
@@ -17,7 +17,7 @@ import swav
 import tico
 import torch
 import vicreg
-from pytorch_lightning import LightningModule, Trainer
+from pytorch_lightning import LightningModule, Trainer, seed_everything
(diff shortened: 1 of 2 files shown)


liopeer:
Hi @yvesyue and thank you for your contribution! We are currently reviewing your PR and will get back to you soon.

liopeer:
Alright. We quickly discussed this and always applying the (same) seed is not desirable for us for two reasons:

1. When using these scripts on small datasets we actually want variance in the results that stems from different initilializations.
2. Our reported benchmark results were produced without that seeding, so we don't want to add the impression now that they used that seed instead.

Suggested changes:

1. Move the seed to the arguments of the `main` function and default it to None
```python
def main(
    train_dir: Path,
    # ...
    seed: int | None = None
) -> None:
    print_rank_zero(f"Args: {locals()}")
    seed_everything(seed, workers=True, verbose=True)
```
2. Add a CLI argument in case you want to apply a `42` seed (or anything else):
```python
parser = ArgumentParser("ImageNet ResNet50 Benchmarks")
# ...
parser.add_argument("--seed", type=int, default=None)
```

yvesyue:
hi @liopeer , thanks for your reply, I am looking into it.

yvesyue:
Thanks so much for the feedback! 
I’ve made the seed an optional --seed argument (defaulting to None) and only call seed_everything(...) when the user provides it.
This preserves the original unseeded behavior by default while still allowing reproducible runs if desired.
Let me know if you’d like any tweaks!

liopeer (approved):
Thanks @yvesyue, LGTM! I took the liberty of fixing the failing CI tests by calling `make format`. This command is also documented in the [CONTRIBUTING.md](https://github.com/lightly-ai/lightly/blob/master/CONTRIBUTING.md) in case you did not have the chance yet to look at it.

PR #1820, opened 2025-04-30
Fix MyPy type-checking under NumPy ≥2.2

yvesyue:
This patch makes two updates so that MyPy passes with newer NumPy stubs:

- **PCA (`embeddings_2d.py`)**: Changed `mean` and `w` to `NDArray[Any]` (shape-only), while still casting to `float32` at runtime.
- **Jigsaw (`jigsaw.py`)**: Added a `# type: ignore[list-item]` on the permutation line to match its dynamic return type.

**Verification**  
- Ran under NumPy 2.0.1 and 2.2.5:  
  ```bash
  make install-dev
  make type-check

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
     entry: make test-fast
     language: system
     pass_filenames: false
-    stages: [push]
+    stages: [pre-push]
     # Avoid running tests if non-tested files have changed.
     # The regex follows the pattern in the docs: https://pre-commit.com/#regular-expressions
     exclude: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
     entry: make test-fast
     language: system
     pass_filenames: false
-    stages: [pre-push]
+    stages: [push]
     # Avoid running tests if non-tested files have changed.
     # The regex follows the pattern in the docs: https://pre-commit.com/#regular-expressions
     exclude: |
--- a/lightly/transforms/jigsaw.py
+++ b/lightly/transforms/jigsaw.py
@@ -84,5 +84,5 @@ class Jigsaw(object):
         crop_tensors: Tensor = torch.stack(
             [self.transform(crop) for crop in crop_images]
         )
-        permutation = [int(x) for x in np.random.permutation(self.n_grid**2)]
+        permutation: List[int] = np.random.permutation(self.n_grid**2).tolist()
         return crop_tensors[permutation]
--- a/lightly/transforms/jigsaw.py
+++ b/lightly/transforms/jigsaw.py
@@ -84,5 +84,6 @@ class Jigsaw(object):
         crop_tensors: Tensor = torch.stack(
             [self.transform(crop) for crop in crop_images]
         )
-        permutation: List[int] = np.random.permutation(self.n_grid**2).tolist()
+        # permutation: List[int] = np.random.permutation(self.n_grid**2).tolist()
+        permutation = np.random.permutation(self.n_grid**2).tolist()  # type: ignore[list-item]
         return crop_tensors[permutation]
--- a/lightly/transforms/jigsaw.py
+++ b/lightly/transforms/jigsaw.py
@@ -84,6 +84,5 @@ class Jigsaw(object):
         crop_tensors: Tensor = torch.stack(
             [self.transform(crop) for crop in crop_images]
         )
-        # permutation: List[int] = np.random.permutation(self.n_grid**2).tolist()
-        permutation = np.random.permutation(self.n_grid**2).tolist()  # type: ignore[list-item]
+        permutation = [int(x) for x in np.random.permutation(self.n_grid**2)]
         return crop_tensors[permutation]
--- a/lightly/utils/embeddings_2d.py
+++ b/lightly/utils/embeddings_2d.py
@@ -46,18 +46,16 @@ class PCA(object):
 
         """
         X = X.astype(np.float32)
-        # 1) mean() up-casts to float64 in numpy ≥2.2 (numpy#28805), See: https://github.com/numpy/numpy/issues/28805
-        self.mean = X.mean(axis=0).astype(np.float32)
+        self.mean = X.mean(axis=0)
         assert self.mean is not None
-        # 2) subtraction + Python float up-casts again → recast
-        X = (X - self.mean + self.eps).astype(np.float32)
-
+        X = (X - self.mean + self.eps).astype(
+            np.float32
+        )  # X is up-casted to float64 in numpy ≥2.2 (numpy#28805), See: https://github.com/numpy/numpy/issues/28805
         cov = np.cov(X.T) / X.shape[0]
         v, w = np.linalg.eig(cov)
         idx = v.argsort()[::-1]  # Sort eigenvalues in descending order
         v, w = v[idx], w[:, idx]
-        # 3) eig returns float64 eigenvectors → recast
-        self.w = w.astype(np.float32)
+        self.w = w
         return self
 
     def transform(self, X: NDArray[np.float32]) -> NDArray[np.float32]:
@@ -77,12 +75,13 @@ class PCA(object):
         """
         if self.mean is None or self.w is None:
             raise ValueError("PCA not fitted yet. Call fit() before transform().")
-
-        # subtraction & eps again → recast
-        X = (X.astype(np.float32) - self.mean + self.eps).astype(np.float32)
-
+        X = (X - self.mean + self.eps).astype(
+            np.float32
+        )  # X is up-casted to float64 in numpy ≥2.2 (numpy#28805), See: https://github.com/numpy/numpy/issues/28805
         transformed: NDArray[np.float32] = X.dot(self.w)[:, : self.n_components]
-        return transformed
+        return transformed.astype(
+            np.float32
+        )  # runtime here is float64, fixes test and matches annotation
 
 
 def fit_pca(
--- a/lightly/utils/embeddings_2d.py
+++ b/lightly/utils/embeddings_2d.py
@@ -10,6 +10,11 @@ from typing import TYPE_CHECKING, Any, Optional, cast
 import numpy as np
 from numpy.typing import NDArray
 
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
+else:
+    NDArray = Any
+
 
 class PCA(object):
     """Handmade PCA to bypass sklearn dependency.
--- a/lightly/utils/embeddings_2d.py
+++ b/lightly/utils/embeddings_2d.py
@@ -79,8 +79,8 @@ class PCA(object):
             np.float32
         )  # X is up-casted to float64 in numpy ≥2.2 (numpy#28805), See: https://github.com/numpy/numpy/issues/28805
         transformed: NDArray[np.float32] = X.dot(self.w)[:, : self.n_components]
-        return transformed.astype(
-            np.float32
+        return np.asarray(
+            transformed, dtype=np.float32
         )  # runtime here is float64, fixes test and matches annotation
 
 
--- a/lightly/utils/embeddings_2d.py
+++ b/lightly/utils/embeddings_2d.py
@@ -5,7 +5,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Any, Optional, cast
 
 import numpy as np
 
@@ -31,8 +31,9 @@ class PCA(object):
     def __init__(self, n_components: int = 2, eps: float = 1e-10):
         self.n_components = n_components
         self.eps = eps
-        self.mean: Optional[NDArray[np.float32]] = None
-        self.w: Optional[NDArray[np.float32]] = None
+        # We only care about array shape for typing, not dtype
+        self.mean: Optional[NDArray[Any]] = None
+        self.w: Optional[NDArray[Any]] = None
 
     def fit(self, X: NDArray[np.float32]) -> PCA:
         """Fits PCA to data in X.
@@ -53,10 +54,12 @@ class PCA(object):
         v, w = np.linalg.eig(cov)
         idx = v.argsort()[::-1]  # Sort eigenvalues in descending order
         v, w = v[idx], w[:, idx]
-        self.w = w
+        # At runtime we cast to float32 for ML performance,
+        # but MyPy only checks shape (dtype is `Any`).
+        self.w = w.astype(np.float32)
         return self
 
-    def transform(self, X: NDArray[np.float32]) -> NDArray[np.float32]:
+    def transform(self, X: NDArray[np.float32]) -> NDArray[Any]:
         """Uses PCA to transform data in X.
 
         Args:
@@ -76,8 +79,8 @@ class PCA(object):
 
         X = X.astype(np.float32)
         X = X - self.mean + self.eps
-        transformed: NDArray[np.float32] = X.dot(self.w)[:, : self.n_components]
-        return np.asarray(transformed)
+        result = X.dot(self.w)[:, : self.n_components].astype(np.float32)
+        return cast(NDArray[Any], result)
 
 
 def fit_pca(
--- a/lightly/utils/embeddings_2d.py
+++ b/lightly/utils/embeddings_2d.py
@@ -31,7 +31,6 @@ class PCA(object):
     def __init__(self, n_components: int = 2, eps: float = 1e-10):
         self.n_components = n_components
         self.eps = eps
-        # We only care about array shape for typing, not dtype
         self.mean: Optional[NDArray[np.float32]] = None
         self.w: Optional[NDArray[np.float32]] = None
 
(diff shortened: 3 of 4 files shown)


lightly/utils/embeddings_2d.py
```diff
@@ -5,12 +5,15 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Any, Optional, cast
```
yutong-xiang-97, line 8: Remove unused imports (which is also doable by `make format`)

lightly/utils/embeddings_2d.py
```diff
@@ -76,8 +82,8 @@ def transform(self, X: NDArray[np.float32]) -> NDArray[np.float32]:
 
         X = X.astype(np.float32)
         X = X - self.mean + self.eps
```
yutong-xiang-97, line 84: The error raises at this particular line because the results is still typed as `np.float64` even if all of the components used `np.float32` (see explanation [here](https://chatgpt.com/share/e/68186cf6-5d20-8002-b6d7-b229e572c9e1).)

So one possible solution is to cast `X = (X - self.mean + self.eps).astype(np.float32)` cuz every variable used in this computation should use `np.float32`.

lightly/utils/embeddings_2d.py
```diff
@@ -76,8 +82,8 @@ def transform(self, X: NDArray[np.float32]) -> NDArray[np.float32]:
...
-        transformed: NDArray[np.float32] = X.dot(self.w)[:, : self.n_components]
```
yutong-xiang-97, line 79: Also, don't change the variable name unless you have a reason to do so. What is your argument for changing `transformed` to `result`? I don't see it.

tests/utils/test_embeddings_2d.py
```diff
@@ -0,0 +1,48 @@
+import numpy as np
+import pytest
+
+from lightly.utils.embeddings_2d import PCA, fit_pca
+
+
+def test_pca_fit_transform_shapes_and_dtype() -> None:
```
yutong-xiang-97, line 7: Great that you are adding unit tests! I would recommend checking t[he tests of the original sklearn PCA](https://github.com/scikit-learn/scikit-learn/blob/36d056fc0b4ff84c6fd1f158b1b14798e9f75df2/sklearn/decomposition/tests/test_pca.py#L55) for inspiration. It doesn't need to be as complex as theirs, but should basically include:

- Correctness: use a hand-crafted matrix as input and check manually if the end result is correct. Think about it as a question in your homework for an introductory ML class covering PCA!
- Edge cases
- Shape and DType as you have done here.
- Error cases like done in `test_transform_without_fit_raises_value_error` but the case for incorrect `fraction` is missing!

tests/utils/test_embeddings_2d.py
```diff
@@ -0,0 +1,48 @@
...
+    assert pca.w.shape == (10, 10)
+
+    # Transform the data and check shape & dtype
+    X_transformed = pca.transform(X)
+    assert X_transformed.shape == (100, 2)
+    assert isinstance(X_transformed, np.ndarray)
+    assert X_transformed.dtype == np.float32  # type: ignore[comparison-overlap]
+
+
+def test_transform_without_fit_raises_value_error() -> None:
```
yutong-xiang-97, line 28: Great that you're aware of testing the API contracts in your first trial! Something that I could totally miss :)

.pre-commit-config.yaml
```diff
@@ -32,7 +32,7 @@ repos:
     entry: make test-fast
     language: system
     pass_filenames: false
-    stages: [push]
+    stages: [pre-push]
```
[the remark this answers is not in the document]
yvesyue: Sorry for the confusion—I wasn’t familiar enough with our pre-commit YAML and didn’t feel confident making that change. Although the hook passed locally, I forgot our convention to use stages: [push]. I did try your make install-pinned-extras suggestion but ran into errors on my setup. Then I ran:

`pip install --upgrade pip setuptools wheel
pip install -e ".[all-extras]"`
and make all-checks now completes smoothly. I’ve reverted the hook back to stages: [push]. Thanks  for catching this ! ;)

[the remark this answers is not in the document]
yvesyue: Hi @yutong-xiang-97,

About stages: [push]
I didn’t fully consider the difference—I wasn’t familiar with our pre-commit YAML, so I followed G’s suggestion and saw no errors locally, then forgot to revisit it. That oversight is on me—thank you for catching it and helping me understand the correct usage!
About make install-pinned-extras vs. local setup
Locally, make all-checks runs smoothly because I’m using newer dependency versions via
pip install --upgrade pip setuptools wheel  
pip install -e ".[all-extras]"
However, those versions don’t match CI exactly. When I tried `make install-pinned-extras`  it was very slow and many tests failed due to version incompatibilities. To address this, I’ll separate fast local checks from full tests by using a fresh virtual environment:

Local fast checks with my current setup
Full CI-pinned tests in a clean venv via make install-pinned-extras && make all-checks
Thanks again for your guidance!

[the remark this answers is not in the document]
yutong-xiang-97: In what hardware / OS / Python version config do your `make install-pinned-extras` fail? There are no issues with my MacBookPro M4, Sequoia 15.4.1, Python 3.12.8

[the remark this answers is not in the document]
yvesyue: Hi Yutong,
I'm currently running on an older MacBook with macOS Version: 11.7.10 (Big Sur)，
Python: 3.11.12 inside the lightly-dev Conda environment (base system Python is 3.9.18)，
Hardware: Intel MacBook Pro； the issue with `make install-pinned-extras` occurs inside the Conda environment.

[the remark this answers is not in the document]
yvesyue: Thanks for the tip, @yutong-xiang-97! Alright, I’ll give uv a try for venv management and look into how it integrates with our pyproject.toml workflow. Appreciate the guidance!

lightly/utils/embeddings_2d.py
```diff
@@ -5,12 +5,15 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Any, Optional, cast
 
 import numpy as np
+from numpy.typing import NDArray
```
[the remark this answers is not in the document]
yvesyue: that makes total sense. I’ve wrapped the from numpy.typing import NDArray in a if TYPE_CHECKING.

lightly/utils/embeddings_2d.py
```diff
@@ -5,12 +5,15 @@
...
 
 if TYPE_CHECKING:
     from numpy.typing import NDArray
+else:
```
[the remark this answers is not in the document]
yvesyue: Great point. I changed both self.mean and self.w to Optional[NDArray[np.float32]] so MyPy knows exactly which dtype they carry.

lightly/utils/embeddings_2d.py
```diff
@@ -31,8 +34,9 @@ class PCA(object):
     def __init__(self, n_components: int = 2, eps: float = 1e-10):
         self.n_components = n_components
         self.eps = eps
-        self.mean: Optional[NDArray[np.float32]] = None
-        self.w: Optional[NDArray[np.float32]] = None
+        # We only care about array shape for typing, not dtype
```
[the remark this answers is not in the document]
yvesyue: Absolutely—dtype precision is vital for both correctness and performance in ML workflows. Thanks for emphasizing this, @yutong-xiang-97! I’ve updated our annotations to NDArray[np.float32] throughout so that MyPy catches any dtype mismatches up front, and at runtime we remain in efficient float32.

lightly/utils/embeddings_2d.py
yvesyue: alright, all done !

lightly/utils/embeddings_2d.py
yvesyue: Good insight, I’ve now added explicit .astype(np.float32) after each of those steps so everything stays in float32, and MyPy will pass ( and the explanation link above shows no content )

lightly/utils/embeddings_2d.py
yvesyue: You’re right—I’ll keep the original transformed name. No need to rename it when its purpose is already clear.

tests/utils/test_embeddings_2d.py
yvesyue: Thank you for the suggestion—I just added a test_fit_pca_invalid_fraction_raises_value_error to cover fraction < 0 and > 1. Now both branches are tested.

lightly/transforms/jigsaw.py
```diff
@@ -84,5 +84,5 @@ def __call__(self, img: PILImage) -> Tensor:
         crop_tensors: Tensor = torch.stack(
             [self.transform(crop) for crop in crop_images]
         )
-        permutation: List[int] = np.random.permutation(self.n_grid**2).tolist()
+        permutation = [int(x) for x in np.random.permutation(self.n_grid**2)]
```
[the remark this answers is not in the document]
yvesyue: alright no problem, I leave it revert to upstream.

.pre-commit-config.yaml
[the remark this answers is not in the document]
yutong-xiang-97: Also don't forget to revert the changes to the pre-commit file :)

tests/utils/test_embeddings_2d.py
yutong-xiang-97: The testing for the correctness is still missing. You could either

1. compare with the results from PCA in sklearn
2. construct an easy example manually and compare the results by manual input and the one from your own PCA.

For the first option, one possible example by ChatGPT is:

```python
from sklearn.decomposition import PCA as SKPCA  
from numpy.testing import assert_allclose

def make_2d_line_dataset(n=100, noise=1e-6, seed=0):
    """Points roughly along the line y = 2x, plus tiny noise."""
    rng = np.random.RandomState(seed)
    x = rng.rand(n, 1) * 10
    data = np.hstack([x, 2 * x])
    data += rng.randn(n, 2) * noise
    return data.astype(np.float32)


def test_pca_aligns_with_sklearn():
    X = make_2d_line_dataset()
    
    # sklearn PCA for reference
    skpca = SKPCA(n_components=1, svd_solver='full').fit(X)
    sk_comp = skpca.components_[0]       # shape (2,)
    
    # our PCA
    pca = PCA(n_components=1).fit(X)
    ours = pca.w[:, 0]
    
    # eigenvector direction is only unique up to sign → compare absolute
    assert_allclose(np.abs(ours), np.abs(sk_comp), atol=1e-5)
```

tests/utils/test_embeddings_2d.py
```diff
@@ -0,0 +1,61 @@
...
+
+    out1 = helper.transform(X)
+    out2 = direct.transform(X)
+
+    # Both should produce shape (50, 1)
+    assert out1.shape == (50, 1)
+    assert out2.shape == (50, 1)
+
+
+def test_fit_pca_invalid_fraction_raises_value_error() -> None:
```
yutong-xiang-97, line 51: In this case, we also want to test whether it works for allowed fraction range.

One possible example is:

```python
@pytest.mark.parametrize(
    "frac",
    [
        0.5,  # 50% of data
        1.0,  # all data
    ]
)
def test_fit_pca_fraction_sample_size(frac: float, seed: int = 42) -> None:
    X = np.arange(40, dtype=np.float32).reshape(10, 4)
    n_sub = int(X.shape[0] * frac)

    np.random.seed(seed)
    perm = np.random.permutation(X.shape[0])
    expected_indices = perm[:n_sub]
    expected_mean = X[expected_indices].mean(axis=0)

    np.random.seed(seed)
    pca = fit_pca(X, n_components=2, fraction=frac)

    assert_allclose(pca.mean, expected_mean, rtol=1e-6)
```

and by testing with edge cases 0.0 and 1.0 we found out that it does not work with `fraction=0.0`! That's what testing is for. Please also fix the corresponding if-branch and docstring for `fit_pca`.

yutong-xiang-97 (commented):
Did another round of review focusing on static typing and tests.

Tip: If you use VSCode and wanna run individual tests easily you could try [their testing extension](https://code.visualstudio.com/docs/debugtest/testing)

lightly/utils/embeddings_2d.py
```diff
@@ -46,14 +47,18 @@ def fit(self, X: NDArray[np.float32]) -> PCA:
 
         """
         X = X.astype(np.float32)
-        self.mean = X.mean(axis=0)
+        self.mean = X.mean(axis=0).astype(
```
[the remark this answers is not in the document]
yvesyue: Thanks yutong for this , really helps -- I’ve removed the unnecessary cast and updated the inline comment to explain that the explicit float32 cast is a temporary fix for the upstream NumPy bug [numpy/numpy#28805], We’ll keep it until the issue is resolved, to maintain compatibility regardless of NumPy version.

lightly/utils/embeddings_2d.py
```diff
@@ -46,14 +46,18 @@ def fit(self, X: NDArray[np.float32]) -> PCA:
 
         """
         X = X.astype(np.float32)
-        self.mean = X.mean(axis=0)
+        # 1) mean() up-casts to float64 in numpy ≥2.2 (numpy#28805), See: https://github.com/numpy/numpy/issues/28805
+        self.mean = X.mean(axis=0).astype(np.float32)
```
yutong-xiang-97, line 50:
```suggestion
        self.mean = X.mean(axis=0)
```

lightly/utils/embeddings_2d.py
```diff
@@ -46,14 +46,18 @@ def fit(self, X: NDArray[np.float32]) -> PCA:
 
         """
         X = X.astype(np.float32)
-        self.mean = X.mean(axis=0)
+        # 1) mean() up-casts to float64 in numpy ≥2.2 (numpy#28805), See: https://github.com/numpy/numpy/issues/28805
```
yutong-xiang-97, line 49:
```suggestion

```

lightly/utils/embeddings_2d.py
```diff
@@ -46,14 +46,18 @@ def fit(self, X: NDArray[np.float32]) -> PCA:
...
+        self.mean = X.mean(axis=0).astype(np.float32)
         assert self.mean is not None
-        X = X - self.mean + self.eps
+        # 2) subtraction + Python float up-casts again → recast
```
yutong-xiang-97, line 52:
```suggestion
```

lightly/utils/embeddings_2d.py
```diff
@@ -46,14 +46,18 @@ def fit(self, X: NDArray[np.float32]) -> PCA:
...
+        X = (X - self.mean + self.eps).astype(np.float32)
```
yutong-xiang-97, line 53:
```suggestion
        X = (X - self.mean + self.eps).astype(np.float32)  # X is up-casted to float64 in numpy ≥2.2 (numpy#28805), See: https://github.com/numpy/numpy/issues/28805
```

lightly/utils/embeddings_2d.py
```diff
@@ -46,14 +46,18 @@ def fit(self, X: NDArray[np.float32]) -> PCA:
...
+
         cov = np.cov(X.T) / X.shape[0]
         v, w = np.linalg.eig(cov)
         idx = v.argsort()[::-1]  # Sort eigenvalues in descending order
         v, w = v[idx], w[:, idx]
-        self.w = w
+        # 3) eig returns float64 eigenvectors → recast
```
yutong-xiang-97, line 59:
```suggestion
```

lightly/utils/embeddings_2d.py
```diff
@@ -46,14 +46,18 @@ def fit(self, X: NDArray[np.float32]) -> PCA:
...
+        self.w = w.astype(np.float32)
```
yutong-xiang-97, line 60:
```suggestion
        self.w = w
```

lightly/utils/embeddings_2d.py
```diff
@@ -74,10 +78,11 @@ def transform(self, X: NDArray[np.float32]) -> NDArray[np.float32]:
         if self.mean is None or self.w is None:
             raise ValueError("PCA not fitted yet. Call fit() before transform().")
 
-        X = X.astype(np.float32)
-        X = X - self.mean + self.eps
+        # subtraction & eps again → recast
+        X = (X.astype(np.float32) - self.mean + self.eps).astype(np.float32)
```
yutong-xiang-97, line 82:
```suggestion
        X = (X - self.mean + self.eps).astype(np.float32)  # X is up-casted to float64 in numpy ≥2.2 (numpy#28805), See: https://github.com/numpy/numpy/issues/28805
```

yutong-xiang-97 (commented):
A lot better! Fix the remaining open suggestions and you're good to go. I will approve and merge the PR once it is ready.

yvesyue:
Hi Yutong,

I encountered a version compatibility issue when CI minimal dependency tests 3.7. 

I initially used from numpy.typing import NDArray to type the return of make_2d_line_dataset, which passed pytest locally.
However, the CI failed with:
ModuleNotFoundError: No module named 'numpy.typing'
May I refer to this is because of  numpy.typing.NDArray was introduced in numpy 1.20, and the CI uses Python 3.7 with an older numpy version, right ? 
My Attempts:
-I first guarded the import using TYPE_CHECKING, but the CI still need the module for run time. 
-I replaced NDArray[np.float32] with np.ndarray, which fixed the CI import error. However, this caused a mypy failure:
 Missing type parameters for generic type "ndarray"
-Current Solution:
To balance CI compatibility (Python 3.7 + old numpy) and static checking via mypy, I updated the return type of make_2d_line_dataset to:

`from typing import Any

def make_2d_line_dataset(...) -> np.ndarray[Any, Any]: `

Would this solution sound OK ? 

Thanks!

yutong-xiang-97 (approved):

PR #1823, opened 2025-05-06
DINOv2 Implementation

yutong-xiang-97:
DINOv2 implementation derived from https://github.com/lightly-ai/lightly/pull/1778

Major adaptation include:

- Computing `teacher_temp` externally and use the same `teacher_temp` for DINOLoss and IBOTLoss by passing it as a parameter, thus removing the internal `teacher_temp` schedules in the DINOLoss.
- Fixing the `dino.py`s accordingly
- Improving default params and docstrings of IBOTLoss
- Adding DINOv2 to `main.py`
- Fixing wrong std when initializing the input tensor with a truncated normal distribution
- Adding `num_sanity_val_steps` and corresponding notes.
- Fixing tests accordingly
- Using Python 3.10 typing convention for added/modified scripts/modules

### DIFF
--- a/benchmarks/imagenet/resnet50/finetune_eval.py
+++ b/benchmarks/imagenet/resnet50/finetune_eval.py
@@ -96,7 +96,7 @@ def finetune_eval(
         logger=TensorBoardLogger(save_dir=str(log_dir), name="finetune_eval"),
         precision=precision,
         strategy=strategy,
-        num_sanity_val_steps=0,  # NOTE: prevent problems from warmup schedule or validation metrics
+        num_sanity_val_steps=0,  # NOTE: save shared memory usage
     )
     classifier = FinetuneClassifier(
         model=model,
--- a/lightly/loss/dino_loss.py
+++ b/lightly/loss/dino_loss.py
@@ -1,5 +1,3 @@
-from typing import List
-
 import torch
 import torch.nn.functional as F
 from torch import Tensor
@@ -24,17 +22,12 @@ class DINOLoss(Module):
     Attributes:
         output_dim:
             Dimension of the model output.
-        warmup_teacher_temp:
-            Initial value of the teacher temperature. Should be decreased if the
-            training loss does not decrease.
         teacher_temp:
-            Final value of the teacher temperature after linear warmup. Values
-            above 0.07 result in unstable behavior in most cases. Can be
-            slightly increased to improve performance during fine-tuning.
-        warmup_teacher_temp_epochs:
-            Number of epochs for the teacher temperature warmup.
+            Temperature parameter for the teacher network.
         student_temp:
-            Temperature of the student.
+            Temperature parameter for the student network.
+        center_mode:
+            Mode for center calculation. Only 'mean' is supported.
         center_momentum:
             Momentum term for the center calculation.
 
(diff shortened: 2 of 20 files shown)


lightly/models/modules/heads.py
```diff
@@ -722,7 +722,7 @@ def _init_weights(self, module: nn.Module) -> None:
             utils._no_grad_trunc_normal(
                 module.weight,
                 mean=0,
-                std=0.2,
+                std=0.04,
```
guarin, line 725: The original code uses 0.02:
* https://github.com/facebookresearch/dino/blob/7c446df5b9f45747937fb0d72314eb9f7b66930a/vision_transformer.py#L283
* https://github.com/facebookresearch/dinov2/blob/e1277af2ba9496fbadf7aec6eba56e8d882d1e35/dinov2/layers/dino_head.py#L32

guarin (commented):
Left some minor comments but otherwise LGTM!

lightly/models/modules/heads.py
yutong-xiang-97: Yes but this would cause the test to fail unfortunately: 
```python
    def test_dino_projection_head_freeze_last_layer(self, seed: int = 0) -> None:
        """Test if freeze last layer cancels backprop."""
        torch.manual_seed(seed)
        for norm_last_layer in [False, True]:
            for freeze_last_layer in range(-1, 3):
                head = DINOProjectionHead(
                    input_dim=4,
                    hidden_dim=4,
                    output_dim=4,
                    bottleneck_dim=4,
                    freeze_last_layer=freeze_last_layer,
                    norm_last_layer=norm_last_layer,
                )
                optimizer = torch.optim.SGD(head.parameters(), lr=1)
                criterion = DINOLoss(output_dim=4)
                # Store initial weights of last layer
                initial_data = [
                    param.data.detach().clone()
                    for param in head.last_layer.parameters()
                ]
                for epoch in range(5):
                    with self.subTest(
                        f"norm_last_layer={norm_last_layer}, "
                        f"freeze_last_layer={freeze_last_layer}, "
                        f"epoch={epoch}"
                    ):
                        views = [torch.rand((3, 4)) for _ in range(2)]
                        teacher_out = [head(view) for view in views]
                        student_out = [head(view) for view in views]
                        loss = criterion(teacher_out, student_out)
                        optimizer.zero_grad()
                        loss.backward()
                        head.cancel_last_layer_gradients(current_epoch=epoch)
                        optimizer.step()
                        params = head.last_layer.parameters()
                        # Verify that weights have (not) changed depending on epoch.
                        for param, init_data in zip(params, initial_data):
                            if param.requires_grad:
                                are_same = torch.allclose(param.data, init_data)
                                if epoch >= freeze_last_layer:
>                                   self.assertFalse(are_same)
E                                   AssertionError: True is not false

tests/models/test_ProjectionHeads.py:227: AssertionError
```

in any case I will add a note for this.

benchmarks/imagenet/resnet50/finetune_eval.py
```diff
@@ -96,7 +96,7 @@ def finetune_eval(
         logger=TensorBoardLogger(save_dir=str(log_dir), name="finetune_eval"),
         precision=precision,
         strategy=strategy,
-        num_sanity_val_steps=0,  # NOTE: prevent problems from warmup schedule or validation metrics
+        num_sanity_val_steps=0,  # NOTE: save shared memory usage
```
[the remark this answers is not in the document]
guarin: Ah then the comment was wrong, thanks for explaining!

lightly/models/modules/heads.py
guarin: In that we have to update the test

yutong-xiang-97: I tried to adjust the error bar with `torch.allclose()` but in no avail :(.

lightly/loss/dino_loss.py
```diff
@@ -50,56 +45,40 @@ class DINOLoss(Module):
...
     """
 
     def __init__(
         self,
         output_dim: int = 65536,
-        warmup_teacher_temp: float = 0.04,
         teacher_temp: float = 0.04,
-        warmup_teacher_temp_epochs: int = 30,
         student_temp: float = 0.1,
-        center_momentum: float = 0.9,
```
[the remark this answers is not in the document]
yutong-xiang-97: I pushed a modification with backward compatibility.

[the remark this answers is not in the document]
guarin: I don't see the change, the arguments of the `__init__` method must remain unchanged.

guarin (changes requested):

guarin (approved):

PR #1837, opened 2025-06-02
ApiWorkflowClient: Split _ListingMixin apart from _DatasourcesMixin

MalteEbner:
## Description

Splits the `ApiWorkflowClient`: `_ListingMixin` apart from `_DatasourcesMixin`. This prevents the 900-lines file `api_workflow_datasources.py` into two files, making working with them easier.

Added the new `api_workflow_listing.py` to the mypy-ignore for the moment to make this PR only about the split. The typing of it belongs to a follow-up PR.

## Tests

Also split the existing tests into two files.

### DIFF
--- a/lightly/api/api_workflow_client.py
+++ b/lightly/api/api_workflow_client.py
@@ -14,10 +14,10 @@ from lightly.api.api_workflow_artifacts import _ArtifactsMixin
 from lightly.api.api_workflow_collaboration import _CollaborationMixin
 from lightly.api.api_workflow_compute_worker import _ComputeWorkerMixin
 from lightly.api.api_workflow_datasets import _DatasetsMixin
+from lightly.api.api_workflow_datasource_listing import _DatasourceListingMixin
 from lightly.api.api_workflow_datasources import _DatasourcesMixin
 from lightly.api.api_workflow_download_dataset import _DownloadDatasetMixin
 from lightly.api.api_workflow_export import _ExportDatasetMixin
-from lightly.api.api_workflow_listing import _ListingMixin
 from lightly.api.api_workflow_predictions import _PredictionsMixin
 from lightly.api.api_workflow_selection import _SelectionMixin
 from lightly.api.api_workflow_tags import _TagsMixin
@@ -57,7 +57,7 @@ class ApiWorkflowClient(
     _UploadCustomMetadataMixin,
     _TagsMixin,
     _DatasourcesMixin,
-    _ListingMixin,
+    _DatasourceListingMixin,
     _ComputeWorkerMixin,
     _CollaborationMixin,
     _PredictionsMixin,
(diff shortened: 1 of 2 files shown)


lightly/api/api_workflow_listing.py
```diff
@@ -0,0 +1,627 @@
...
+    DatasourceProcessedUntilTimestampResponse,
+    DatasourceRawSamplesData,
+)
+from lightly.openapi_generated.swagger_client.models.datasource_raw_samples_data_row import (
+    DatasourceRawSamplesDataRow,
+)
+
+
+class _ListingMixin:
+    def download_raw_samples(
```
japrescott, line 20: can we deprecate this one as there is a `download_new_raw_samples`?

lightly/api/api_workflow_listing.py
```diff
@@ -0,0 +1,627 @@
...
+            exist.
+
+        :meta private:  # Skip docstring generation
+        """
+        return self._datasources_api.get_custom_embedding_file_read_url_from_datasource_by_dataset_id(
+            dataset_id=self.dataset_id,
+            file_name=filename,
+        )
+
+    def list_datasource_permissions(
```
japrescott, line 479: don't think this fn should be moved?

lightly/api/api_workflow_listing.py
```diff
@@ -0,0 +1,627 @@
...
+    DatasourceProcessedUntilTimestampRequest,
+    DatasourceProcessedUntilTimestampResponse,
+    DatasourceRawSamplesData,
+)
+from lightly.openapi_generated.swagger_client.models.datasource_raw_samples_data_row import (
+    DatasourceRawSamplesDataRow,
+)
+
+
+class _ListingMixin:
```
japrescott, line 19: can we call it `_DatasourceListingMixin` or `ListingDatasoruceMixin` as it does not list locally. Wdyt?

japrescott (approved):
thanks Malte. Left some suggestions but LGTM!

lightly/api/api_workflow_listing.py
MalteEbner: No, we cannot, as we use the download_raw_samples in the worker. See https://github.com/lightly-ai/lightly-core/blob/33852d81379364794e4218e2c77e1e993edd5a90/onprem-docker/lightly_worker/src/lightly_worker/datasources/cloud_input_datasource.py#L94

However, we could delete/deprecate the `download_new_raw_samples`, as that one is not used in the worker. Will have a closer look at it separately.

lightly/api/api_workflow_listing.py
MalteEbner: Yes, I renamed it to `_DatasourceListingMixin`, also the files.

PR #1842, opened 2025-06-06
Fix cosine scheduler warmup epochs

guarin:
## Changes

* Fix corner case where `step == warmup_steps == max_steps`


lightly/utils/scheduler.py
```diff
@@ -49,10 +49,10 @@ def cosine_schedule(
             end_value
             - (end_value - start_value) * (np.cos(2 * np.pi * step / period) + 1) / 2
```
yutong-xiang-97, line 50: In the comments, when `period=None` it should be that by default the 1/2 cosine cycle is `max_step - warmup_step` instead of the full cycle. For specified period it is indeed for full cycle.

Cosine Decay Schedule

yutong-xiang-97 (approved):
Approved with minor comments.

PR #1844, opened 2025-06-11
Fix DINOv2 Bugs

yutong-xiang-97:
1. EMA update should happen in on_train_batch_end 
2. Freeze the teacher and put it to eval mode
3. Add wrappers to ensure the param groups are named differently for the DINO and iBOT heads
4. Fix bugs in learning rate schedule


stegmuel (approved):
LGTM.

PR #1854, opened 2025-07-03
iBOT Implementation

yutong-xiang-97:
Implement iBOT in benchmarks

Difference from DINOv2:

- Teacher and student heads are always shared. The only difference between teacher and student heads is `norm_last_layer`
- Head output_dim is 8192
- Final teacher temp for both the cls(DINO) loss and patch (IBOTPatch) loss is 0.07
- init lr 0.0005 with linear scaling
- No layerwise lr decay
- momentum start value is 0.996

### DIFF
--- a/benchmarks/imagenet/vitb16/ibot.py
+++ b/benchmarks/imagenet/vitb16/ibot.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import copy
+from functools import partial
 
 import torch
 from pytorch_lightning import LightningModule
@@ -43,6 +44,7 @@ class IBOT(LightningModule):
         super().__init__()
         self.save_hyperparameters()
         self.batch_size_per_device = batch_size_per_device
+        self.output_dim = 8192
 
         # Backbones
         vit_teacher = vit_small_patch16_224(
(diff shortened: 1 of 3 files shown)


benchmarks/imagenet/vitb16/ibot.py
```diff
@@ -0,0 +1,293 @@
...
+            (cls_token.detach(), targets), batch_idx
+        )
+        self.log_dict(cls_log, prog_bar=True, sync_dist=True, batch_size=len(targets))
+        return cls_loss
+
+    def configure_optimizers(self):
+        # Don't use weight decay for batch norm, bias parameters, and classification
+        # head to improve performance.
+        min_lr = 1e-6
+        lr_scale = self.batch_size_per_device * self.trainer.world_size / 1024
```
stegmuel, line 208: In [iBOT ](https://github.com/bytedance/ibot/blob/da316d82636a7a7356835ef224b13d5f3ace0489/main_ibot.py#L306)they scale the lr using 256 as reference batchsie

stegmuel (changes requested):
One small change, otherwise all good.

stegmuel (approved):
LGTM!

PR #1855, opened 2025-07-04
Add iBOT to Docs with Example

yutong-xiang-97:
Add iBOT examples to docs.


stegmuel (changes requested):

benchmarks/imagenet/vitb16/ibot.py
```diff
@@ -258,9 +258,9 @@ def configure_gradient_clipping(
     def on_before_optimizer_step(self, optimizer: AdamW, *args) -> None:
         # Optionally zero out the learning rate of the last layer.
         if self.current_epoch < 1:
-            for param_group in optimizer.param_groups:
-                if "last_layer" in param_group["name"]:
-                    param_group["lr"] = 0.0
+            for name, param in self.student_head.named_parameters():
```
stegmuel, line 261: The projection head already has the [logic](https://github.com/lightly-ai/lightly/blob/5d63696b509981d8074588061c0f9523a3d009f8/lightly/models/modules/heads.py#L704) implemented.

stegmuel (changes requested):
You can re-use the logic already implemented in the head to cancel the gradients.

PR #1859, opened 2025-07-10
Fix OnlineClassifier Issues

yutong-xiang-97:
Add cls_loss from OnlineClassifier to BP.

Note that the input of the online classifier is the first cls_token instead of the average of cls_tokens from the last 4 blocks as in the original iBOT codebase.


guarin (approved):
LGTM

PR #1861, opened 2025-07-15
Support Tensor Input for GaussianBlur

### LINKED ISSUE #1860: SimCLRTransform claims to support Tensor input, but GaussianBlur does not accept Tensors
The documentation for SimCLRTransform and SimCLRViewTransform states that input can be either a PIL Image or a Tensor. However, the default transform pipeline includes GaussianBlur, which supports only PIL Images. Passing a Tensor as input results in an error at this stage.

Error raised: 
```
    return sample.filter(ImageFilter.GaussianBlur(radius=sigma))
           ^^^^^^^^^^^^^
AttributeError: 'Tensor' object has no attribute 'filter'

```

Best regards.

yutong-xiang-97:
Close #1860

### DIFF
--- a/lightly/transforms/gaussian_blur.py
+++ b/lightly/transforms/gaussian_blur.py
@@ -8,7 +8,7 @@ import numpy as np
 from PIL import ImageFilter
 from PIL.Image import Image
 from torch import Tensor
-from torchvision.transforms.functional import to_pil_image
+from torchvision.transforms.functional import to_pil_image, to_tensor
 
 
 class GaussianBlur:
@@ -50,7 +50,7 @@ class GaussianBlur:
         self.prob = prob
         self.sigmas = sigmas
 
-    def __call__(self, sample: Union[Tensor, Image]) -> Image:
+    def __call__(self, sample: Union[Tensor, Image]) -> Union[Tensor, Image]:
         """Blurs the image with a given probability.
 
         Args:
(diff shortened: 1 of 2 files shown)


tests/transforms/test_gaussian_blur.py
```diff
@@ -13,6 +14,13 @@ def test_on_pil_image(self) -> None:
                 sample = Image.new("RGB", (w, h))
                 gaussian_blur(sample)
 
+    def test_on_tensor(self) -> None:
+        for w in range(1, 100):
+            for h in range(1, 100):
```
guarin, line 19: Doesn't have to be fixed now but the loop with two ranges runs 10'000 tests which seems excessive. Same for the test above. 

```suggestion
        for w in [1, 10, 100]:
            for h in [1, 10, 100]:
```

lightly/transforms/gaussian_blur.py
```diff
@@ -59,12 +61,20 @@ def __call__(self, sample: Image) -> Image:
...
             # PIL GaussianBlur https://github.com/python-pillow/Pillow/blob/76478c6865c78af10bf48868345db2af92f86166/src/PIL/ImageFilter.py#L154 label the
             # sigma parameter of the gaussian filter as radius. Before, the radius of the patch was passed as the argument.
             # The issue was addressed here https://github.com/lightly-ai/lightly/issues/1051 and solved by AurelienGauffre.
-            return sample.filter(ImageFilter.GaussianBlur(radius=sigma))
-        # return original image
-        return sample
+            return sample_pil.filter(ImageFilter.GaussianBlur(radius=sigma))
+
+        # return pil image
+        return sample_pil
```
guarin, line 80: I would expect a tensor output if the input is a tensor.

PR #1862, opened 2025-07-16
Use ToTensor helper function

ajtritt:
This PR implements the use of `lightly.transforms.torchvision_v2_compatibility.ToTensor` in place of torchvision `ToTensor`

`torchvision.transforms.v2.ToTensor` [is deprected](https://docs.pytorch.org/vision/0.20/generated/torchvision.transforms.v2.ToTensor.html?highlight=totensor#torchvision.transforms.v2.ToTensor). The `lightly.transforms.torchvision_v2_compatibility.ToTensor` function was added in #1718 to address this, but the function was never used anywhere. 

The `transform` argument in the `Jigsaw` class currently uses the torchvision v2 `ToTensor`, which means any time that class gets interpreted, such as when you import `LightlyDataset`, a warning is emitted. 

I'm not sure why the helper function was never used anywhere, but it would be nice if, at the very least, the `Jigsaw` default was addressed in future releases.

### DIFF
--- a/lightly/data/collate.py
+++ b/lightly/data/collate.py
@@ -15,6 +15,7 @@ import torchvision.transforms as T
 from PIL import Image
 
 from lightly.transforms import GaussianBlur, Jigsaw, RandomSolarization
+from lightly.transforms.torchvision_v2_compatibility import ToTensor
 from lightly.transforms.random_crop_and_flip_with_grid import RandomResizedCropAndFlip
 from lightly.transforms.rotation import random_rotation_transform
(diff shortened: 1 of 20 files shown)


yutong-xiang-97 (changes requested):
Hi @ajtritt , thank you so much for your support for LightlySSL! The Lightly team really appreciates it :)

Your PR is great, and yes, we absolutely need this change to appease the warnings. We've made a short review for your PR to fix some remaining issues. Please have a look and then we can proceed to merge it.

PR #1865, opened 2025-07-19
Move rotation after flip for random transforms (#1863)

### LINKED ISSUE #1863: Make Order of Random Rotation Transforms Consistent
The order of the random rotation and random flip transforms in LightlySSL is inconsistent. Sometimes we do the random rotation before the flip and vice versa. The goal of this issue is to make this consistent by always applying the flip before the random rotation.

- [ ] Go through the [transforms in LightlySSL](https://github.com/lightly-ai/lightly/tree/master/lightly/transforms) and make sure all random flip transforms come before the random rotation transforms

KylevdLangemheen:
This PR moved `random_rotation_transform` after any `RandomHorizontalFlip` and `RandomVerticalFlip`, thus making the order of application of these [transforms in LightlySSL](https://github.com/lightly-ai/lightly/tree/master/lightly/transforms) consistent. Resolves #1863 

- Also updated one docstring where this was appropriate
- Also removed unused `random_rotation_transform` import


yutong-xiang-97 (commented):
Awesome work! Now you've taken your first steps towards contributing to LightlySSL :).

The PR does exactly what is expected, with only one part left: for the [image collate function](https://github.com/lightly-ai/lightly/blob/a3c2fd5c04921c4bcd891f4118ac982560d3b8ab/lightly/data/collate.py#L175) here you should do the same. Please fix this and you are good to go!

KylevdLangemheen:
You're right! I think I looked only at the transforms. Thanks for pointing it out, collate has been edited.

Issue #1864, opened 2025-07-18
Fix Deprecation Warnings of ToTensor


yutong-xiang-97:
See the issue proposed in https://github.com/lightly-ai/lightly/pull/1862. This also happens with the benchmarking code and in some tests.

- [ ] Apply the solution to wherever needed

PR #1866, opened 2025-07-21
Fix deprecation warning totensor (#1864)

KylevdLangemheen:
This pull request:
- Introduces a shim in `torchvision_v2_compatibility.py` to replace / add our own functions to `torchvision_transforms`
- Using the shim, consistency in torchvision.transforms based transforms is restored: `ToTensor()` is back to `T.ToTensor()`
- Usage of ToTensor in `docs`, `benchmarks`, and `tests` is moved to the compatibility layer as discussed in #1862
- Usage of `torchvision.transforms` beyond ToTensor has also been moved to the compatibility layer

This PR does _not_:
- Change docstring usage of `torchvision.transforms`
- Replace instances where a specific `v1` or `v2`  transform was imported, e.g. https://github.com/lightly-ai/lightly/blob/2e884a04c3fec0a6d016f8e32013b39eb137459c/lightly/transforms/add_grid_transform.py#L8
- Move instances of functional transforms to a compatibility layer


KylevdLangemheen:
It seems `numpy.typing` is from a too recent version of numpy. I will try `numpy.ndarray[Any]` instead.

PR #1868, opened 2025-07-22
Add iBOT Benchmarks

yutong-xiang-97:
Add iBOT benchmark results. The docs pages will be available in the next release.

It also contains 

1. a fix to ibot masking with `min_num_masks_per_block` from the original codebase: https://github.com/bytedance/ibot/blob/da316d82636a7a7356835ef224b13d5f3ace0489/loader.py#L77
2. a fix to the fine-tuning eval which should unfreeze the teacher backbone.
3. clarification of DINO re. the use of cls_tokens


stegmuel (approved):
Approved.

PR #1869, opened 2025-07-22
Improve pre-commit hooks for dev env

yutong-xiang-97:
The contributing guide for setting up the dev env re. the pre-commit hooks is confusing. Also, the hook overlap with the Makefile checks and the scope of the mypy check in the hook is not correct.

This PR propose an improvement in the dev env set up. Such that:

- Pre-commit hooks is always installed
- The hooks only contains large file check and necessary tools for Jupyter notebooks. Formatting, checks, and tests should always be done with `make format && make all-checks`
- Make instructions in the contributing guide clearer.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,7 +84,8 @@ Follow these steps to start contributing:
 4. Set up a development environment by running the following command in a virtual environment:
 
    ```bash
-   pip install -e ".[dev]"
+   pip install -e . --all-extras --requirement pyproject.toml
+   pre-commit install
    ```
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -84,7 +84,8 @@ Follow these steps to start contributing:
 4. Set up a development environment by running the following command in a virtual environment:
 
    ```bash
-   pip install -e ".[dev]"
+   pip install -e . --all-extras --requirement pyproject.toml
+   pre-commit install
```
guarin, line 88: Why not `make install-dev`?

guarin (approved):
LGTM! Left one question

PR #1872, opened 2025-07-24
Fix the banner and tab name of LightlySSL Docs

yutong-xiang-97:
Fix the banner and tab name of LightlySSL Docs

Preview: 
[Documentation — LightlySSL 1.5.22 documentation.pdf](https://github.com/user-attachments/files/21411558/Documentation.LightlySSL.1.5.22.documentation.pdf)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -21,7 +21,7 @@ import lightly
 
 # -- Project information -----------------------------------------------------
 
-project = "lightly"
+project = "LightlySSL"
 copyright_year = "2020-<script>document.write((new Date()).getFullYear())</script>"
 copyright = "Lightly AG"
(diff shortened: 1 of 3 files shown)


docs/source/conf.py
```diff
@@ -21,7 +21,7 @@
 
 # -- Project information -----------------------------------------------------
 
-project = "lightly"
+project = "LightlySSL"
```
guarin, line 24: What are the implications of this change?

yutong-xiang-97: AFAIK this is simply a variable injected into the HTML title, no? I don't see anything breaking the docs when doing `make serve-local`, nor do I see it referenced in the doc release workflow file.

guarin (approved):

PR #963, opened 2022-10-20
Added directclr loss

Atharva-Phatak:
Addded Implementation of DirectCLR loss as proposed in #781 

@philippmwirth I need some help on writing tests. If you could guide me that would be amazing.

Issue #781, opened 2022-04-26
DirectCLR: Understanding Dimensional Collapse in Contrastive Self-supervised Learning


guarin:
## DirectCLR: Understanding Dimensional Collapse in Contrastive Self-supervised Learning

18.10.2021
https://arxiv.org/abs/2110.09348
https://github.com/facebookresearch/directclr

Similar model as SimCLR but does not require a projection head and instead calculates the loss only on a subset of the embedding dimensions. The main focus of the paper is on dimensional collapse and it uses SVD to monitor collapse during training. Outperforms SimCLR with a single layer projection head.

![Screenshot 2022-04-26 at 09 31 55](https://user-images.githubusercontent.com/43336610/165246733-a82dd87e-7bd0-41b7-bf3c-b6320c011857.png)

Estimated effort to implement in Lightly: Low
* Add DirectCLR loss
* Add SVD to monitor dimensional collapse

Atharva-Phatak:
I can take a look at this and let you know. Are you guys hiring new grads remotely? I would be excited to work with you guys.

philippmwirth:
Hi @Atharva-Phatak thank you for your valuable contributions and your interest in working with us 🙂 Please make sure to check our jobs page (https://www.lightly.ai/jobs) for information about openings at Lightly.

We can split this issue into separate pull-requests to manage the work load. One for the DirectCLR loss and one for monitoring dimensional collapse with SVD. What do you think?

Atharva-Phatak:
@philippmwirth That sounds good. Let me take a look at DIRECTCLR paper and the implementation that they have released. 
I took a look at job opening, you are only hiring senior engineers unfortunately 😭 .

Atharva-Phatak:
@philippmwirth Quick question, will this PR also require implementing DIRECT-CLR head ? Also do we have a working implementation of InfoNCELoss ??

yutong-xiang-97:
Closed by #1874

PR #1874, opened 2025-07-31
Add DirectCLR (#781)

KylevdLangemheen:
Issue #781 proposes adding DirectCLR to Lightly. In PR #963 a loss was proposed but never merged.

As the paper uses a version of [infoNCE that is identical to NTX-ent](https://discordapp.com/channels/@me/1399964647523876874/1400461672883159160), we can use our existing implementation of NTX-ent instead. This way, implementing DirectCLR becomes a simple matter of taking the first _d_ values of the embedding and applying the NTX-ent loss to that instead (if my understanding is correct :smile:  ).

As DirectCLR does away with a projection head, I did not see an obvious way to implement it under low level module blocks. The proposed implementation is added as a deprecated model, but perhaps it does not even need to get its own model file and can just live as an example.

This PR:
- Adds DirectCLR as deprecated model
- Adds example for DirectCLR
- Adds test for DirectCLR

This PR _does not_:
- Add DirectCLR to lightly cli
- Implement benchmarks for DirectCLR

Let me know if you would prefer to see a variant outside of the deprecated high-level models. I could make a "dummy" projection head which just slices the embedding for example.

### DIFF
--- a/examples/pytorch/directclr.py
+++ b/examples/pytorch/directclr.py
@@ -1,0 +1,56 @@
+# This example requires the following dependencies to be installed:
+# pip install lightly
+
+# Note: The model and training settings do not follow the reference settings
+# from the paper. The settings are chosen such that the example can easily be
+# run on a small dataset with a single GPU.
+
+import torch
+import torchvision
+from torch import nn
+
+from lightly.loss import NTXentLoss
+from lightly.models import DirectCLR
+from lightly.transforms.simclr_transform import SimCLRTransform
+
+resnet = torchvision.models.resnet18()
+backbone = nn.Sequential(*list(resnet.children())[:-1])
+model = DirectCLR(backbone)
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+model.to(device)
+
+transform = SimCLRTransform(input_size=32, gaussian_blur=0.0)
+dataset = torchvision.datasets.CIFAR10(
+    "datasets/cifar10", download=True, transform=transform
+)
+# or create a dataset from a folder containing images or videos:
+# dataset = LightlyDataset("path/to/folder", transform=transform)
+
+dataloader = torch.utils.data.DataLoader(
+    dataset,
+    batch_size=256,
+    shuffle=True,
+    drop_last=True,
+    num_workers=8,
+)
+
+criterion = NTXentLoss()
+optimizer = torch.optim.SGD(model.parameters(), lr=0.06)
+
+print("Starting Training")
+for epoch in range(10):
+    total_loss = 0
+    for batch in dataloader:
+        x0, x1 = batch[0]
+        x0 = x0.to(device)
+        x1 = x1.to(device)
+        z0 = model(x0)
+        z1 = model(x1)
+        loss = criterion(z0, z1)
+        total_loss += loss.detach()
+        loss.backward()
+        optimizer.step()
+        optimizer.zero_grad()
+    avg_loss = total_loss / len(dataloader)
+    print(f"epoch: {epoch:>02}, loss: {avg_loss:.5f}")
--- a/lightly/models/directclr.py
+++ b/lightly/models/directclr.py
@@ -1,0 +1,106 @@
+""" DirectCLR Model """
+
+# Copyright (c) 2025. Lightly AG and its affiliates.
+# All Rights Reserved
+
+import warnings
+from typing import Optional, Tuple
+
+import torch
+import torch.nn as nn
+from torch import Tensor
+
+
+class DirectCLR(nn.Module):
+    """Implementation of the DirectCLR architecture
+
+    Recommended loss: :py:class:`lightly.loss.ntx_ent_loss.NTXentLoss`
+
+    [0] DirectCLR, 2021, https://arxiv.org/abs/2110.09348
+
+    Attributes:
+        backbone:
+            Backbone model to extract features from images.
+        dim:
+            Length of the subvector of the feature vector on which to apply the loss.
+
+    """
+
+    def __init__(self, backbone: nn.Module, dim: int = 32) -> None:
+        super(DirectCLR, self).__init__()
+
+        self.backbone = backbone
+        self.dim = dim
+
+        warnings.warn(
+            Warning(
+                "The high-level building block DirectCLR will be deprecated in version 1.3.0. "
+                + "Use low-level building blocks instead. "
+                + "See https://docs.lightly.ai/self-supervised-learning/lightly.models.html for more information"
+            ),
+            DeprecationWarning,
+        )
+
+    def forward(
+        self, x0: Tensor, x1: Optional[Tensor] = None, return_features: bool = False
+    ) -> (
+        Tensor
+        | Tuple[Tensor, Tensor]
+        | Tuple[Tuple[Tensor, Tensor], Tuple[Tensor, Tensor]]
+    ):
+        """Embeds and projects the input images.
+
+        Extracts features with the backbone and takes the first self.dim
+        values as subvector of the output space. If both x0 and x1 are not None,
+        both will be passed through the backbone and sliced. If x1 is None, only
+        x0 will be forwarded.
+
+        Args:
+            x0:
+                Tensor of shape bsz x channels x W x H.
+            x1:
+                Tensor of shape bsz x channels x W x H.
+            return_features:
+                Whether or not to return the intermediate features backbone(x).
+
+        Returns:
+            The subvector output of the embedding of x0 and (if x1 is not None)
+            the subvector output of x1. If return_features is True, the output for
+            each x is a tuple (out, f) where f are the features before slicing.
+
+        Examples:
+            >>> # single input, single output
+            >>> out = model(x)
+            >>>
+            >>> # single input with return_features=True
+            >>> out, f = model(x, return_features=True)
+            >>>
+            >>> # two inputs, two outputs
+            >>> out0, out1 = model(x0, x1)
+            >>>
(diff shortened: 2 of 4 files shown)


examples/pytorch/directclr.py
```diff
@@ -0,0 +1,56 @@
...
+
+import torch
+import torchvision
+from torch import nn
+
+from lightly.loss import NTXentLoss
+from lightly.models import DirectCLR
+from lightly.transforms.simclr_transform import SimCLRTransform
+
+resnet = torchvision.models.resnet18()
```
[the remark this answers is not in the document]
KylevdLangemheen: Changed it, also for `torchvision.datasets`

tests/models/test_ModelsDirectCLR.py
```diff
@@ -0,0 +1,101 @@
...
+    backbone = nn.Sequential(
+        lightly.models.batchnorm.get_norm_layer(3, 0),
+        *list(resnet.children())[:-1],
+        nn.Conv2d(last_conv_channels, num_ftrs, 1),
+        nn.AdaptiveAvgPool2d(1),
+    )
+    return backbone
+
+
+class TestModelsDirectCLR(unittest.TestCase):
```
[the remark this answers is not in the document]
KylevdLangemheen: In the current set-up I've made DirectCLR loss inherit from the NTXent loss. As such, many tests become redundant. I currently test whether it can take various loss_dims, whether some of the calculations (that use NTXent) are done correctly, and whether the parent parameter (temperature) is set correctly. Let me know if some tests are missing or redundant still!

lightly/models/directclr.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # return out0 if x1 is None
+        if x1 is None:
+            # return features if requested
+            if return_features:
+                return out0, f0
+            return out0
+
+        # forward pass of second input x1
+        f1: Tensor = self.backbone(x1).flatten(start_dim=1)
```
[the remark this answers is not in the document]
KylevdLangemheen: Now that it is no longer a model but an example, I have included this with the example. Let me know if it is better to keep it simple in the example instead.

examples/pytorch/directclr.py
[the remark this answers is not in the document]
KylevdLangemheen: I don´t know exactly where the line is drawn (i.e. nn.Sequential). There's some variance in the codebase I think?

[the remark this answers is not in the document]
liopeer: It's actually a hard line haha. :) Unfortunately a lot of the code is not up-to-date.

lightly/loss/directclr_loss.py
```diff
@@ -0,0 +1,91 @@
...
+    head. Set loss_dim to the desired truncated representation length.
+    DirectCLRLoss inherits from NTXentLoss, its parameters can be set after
+    setting loss_dim.
+
+    - [0] DirectCLR, 2021, https://arxiv.org/abs/2110.09348
+
+    Attributes:
+        loss_dim:
+            Computes the loss only on the first loss_dim values of the encoding.
+        *args:
```
liopeer, line 27: Since we have full control over the `NTXentLoss` and its arguments it's usually preferred to be verbose and copy the arguments from the NTXentLoss over here. I would only use `*args` and `**kwargs` when the subsequently instantiated class comes from a third-party package and the interface might change unexpectedly – and even in that case I would only allow passing keyword arguments and not positional arguments.

examples/pytorch/directclr.py
```diff
@@ -0,0 +1,53 @@
...
+from torch.optim import SGD
+from torch.utils.data import DataLoader
+from torchvision import models
+from torchvision.datasets import CIFAR10
+
+from lightly.loss import DirectCLRLoss
+from lightly.transforms.simclr_transform import SimCLRTransform
+
+resnet = models.resnet18()
+model = nn.Sequential(*list(resnet.children())[:-1])
```
liopeer, line 19: Nit:
```suggestion
model = Sequential(*list(resnet.children())[:-1])
```

lightly/loss/directclr_loss.py
```diff
@@ -0,0 +1,91 @@
...
+from torch import Tensor
+
+from lightly.loss import NTXentLoss
+
+
+class DirectCLRLoss(NTXentLoss):
+    """Implementation of the NT-Xent based DirectCLR Loss.
+
+    Following the DirectCRL[0] paper, this loss should be used without projection
+    head. Set loss_dim to the desired truncated representation length.
```
liopeer, line 18: Nit:
```suggestion
    head. Set `loss_dim` to the desired truncated representation length.
```
This will render nicely once we add it to the docs. Same thing further down.

lightly/loss/directclr_loss.py
```diff
@@ -0,0 +1,91 @@
...
+                Shape: (batch_size, embedding_size)
+            out1:
+                Output projections of the second set of transformed images.
+                Shape: (batch_size, embedding_size)
+
+        Returns:
+            DirectCLR Loss value.
+        """
+
+        out0 = out0.flatten(start_dim=1)[:, : self.loss_dim]
```
liopeer, line 86: Flattening would not really be necessary since you have a really nice docstring that tells the user how it should be used. But you can leave it as is.

tests/loss/test_directclr_loss.py
```diff
@@ -0,0 +1,107 @@
...
+    @pytest.mark.parametrize("n_samples", [1, 2, 4])
+    @pytest.mark.parametrize("dimension", [1, 2, 16, 64])
+    @pytest.mark.parametrize("loss_dim", [1, 2, 32])
+    def test_with_values(
+        self,
+        n_samples: int,
+        dimension: int,
+        loss_dim: int,
+    ) -> None:
+        out0 = torch.tensor(
```
liopeer, line 25: You could also directly use `torch.randn`, but it's of course totally okay this way as well.

tests/loss/test_directclr_loss.py
```diff
@@ -0,0 +1,107 @@
...
+
+class TestDirectCLRLoss:
+    def test_set_parent_temperature(self) -> None:
+        loss_function = DirectCLRLoss(loss_dim=32, temperature=0.5)
+        assert (
+            loss_function.temperature == 0.5
+        ), f"Expected temperature to be 0.5, but got {loss_function.temperature}"
+
+    @pytest.mark.parametrize("n_samples", [1, 2, 4])
+    @pytest.mark.parametrize("dimension", [1, 2, 16, 64])
```
liopeer, line 17: I would use smaller dimensions here. We have a lot of unit tests in LightlySSL, so we should always try to use the fastest possible unit tests – ofc without compromising correctness.

tests/loss/test_directclr_loss.py
```diff
@@ -0,0 +1,107 @@
...
+        l2_manual = _calc_directclr_loss_manual(
+            out0, out1, loss_dim=loss_dim, temperature=0.5
+        )
+        assert l1 == pytest.approx(l2, abs=1e-5)
+        assert l1 == pytest.approx(l1_manual, abs=1e-5)
+        assert l2 == pytest.approx(l2_manual, abs=1e-5)
+
+    def test_forward_pass(self) -> None:
+        loss = DirectCLRLoss(loss_dim=32, temperature=0.5, memory_bank_size=0)
+        for bsz in range(1, 20):
```
liopeer, line 47: Also here we should speed things up a notch. I would test max 2 cases: Batch size 1 and something else. And I would parametrize the test, so that you get a dedicated test result for each case. Parametrizing would also allow to use e.g. `pytest-xdist` to run the tests in parallel.

tests/loss/test_directclr_loss.py
```diff
@@ -0,0 +1,107 @@
...
+            batch_2 = torch.randn((bsz, 64))
+
+            # symmetry
+            l1 = loss(batch_1, batch_2)
+            l2 = loss(batch_2, batch_1)
+            assert (l1 - l2).pow(2).item() == pytest.approx(0.0)
+
+    def test_forward_pass_1d(self) -> None:
+        loss = DirectCLRLoss(loss_dim=32, temperature=0.5, memory_bank_size=0)
+        for bsz in range(1, 20):
```
liopeer, line 58: Same thing here.

liopeer (changes requested):
Awesome work! It's looking really nice and many of my comments are really just nitpicking and beautification.

The two things we should really change however before merging: The `*args, **kwargs` story, and the *slightly too comprehensive* 🙃 unit tests. After that we're ready to merge.

yutong-xiang-97:
Quick note: the `examples/` are for the supported models / SSL methods in the docs (like shown [here](https://docs.lightly.ai/self-supervised-learning/examples/simclr.html)) but not for the losses. We could have a follow-up issue implementing the whole DirectCLR at `benchmarks/`. Then we can add examples and docs for it.

yutong-xiang-97 (commented):
Added my review as well.

KylevdLangemheen:
I have:
- Generated the example notebooks with `make generate-example-notebooks`
- Reduced dims in tests
- Fixed typo
- Added formatting for loss_dim
- Directly import functions/classes at the module level
- Use `torch.randn` consistently
- Added NTXentLoss arguments directly

Is there a standard way to ensure or even automate that the parameters relating to NTXentLoss always match? This way the defaults can get out of sync pretty easily :(

KylevdLangemheen:
Resolved the temperature parameter and default parameter comments.

liopeer (approved):
Everything addressed from my side as well. Amazing work @KylevdLangemheen !

PR #1876, opened 2025-08-06
Fixed SMoG examples

thomasmarchioro3:
This PR addresses some minor fixes in the SMoG example scripts and notebook (`examples/notebooks/pytorch/smog.ipynb`, `examples/notebooks/pytorch_lightning/smog.ipynb`, `examples/pytorch/smog.py`, `examples/pytorch_lightning/smog.py`):

OS and enviroment:
- Ubuntu 24.04.2 LTS
- Python 3.12.7
- lightly==1.5.22
- scikit-learn==1.7.1
- torch==2.7.1
- pytorch-lightning==2.5.2

---

#### Fix 1: Missing `scikit-learn` requirement

- **Issue**: `scikit-learn` was not listed in the example dependencies, leading to `ImportError` when running the code.
- **Fix**: Added `scikit-learn>=1.7.1` to the example installation instructions.
  _Note_: Consider also adding this to the default dependencies in `pyproject.toml`.

---

#### Fix 2: KMeans `ValueError` due to incorrect tensor shape

- **Issue**: The KMeans call fails with:
```
kmeans = KMeans(self.n_groups).fit(features) raises
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 ValueError: n_samples=128 should be >= n_clusters=300.
```
caused by an incorrect array shape — features were shaped `(num_features, memory_size)` instead of `(memory_size, num_features)`.

- **Fix**: Corrected array shape by transposing the feature array:
```python
features = features.cpu().numpy().T
```

---

#### Fix 3: Incorrect import in pytorch_lightning examples

- **Issue**: Outdated MemoryBankModule + missing feature shape
```python
self.memory_bank = loss.memory_bank.MemoryBankModule(size=memory_bank_size)
```

- **Fix**: Used correct module (same as pytorch examples)

---

#### Additional minor fixes
- Added `#type: ignore` when converting `features` from tensor to numpy array to avoid warnings from the language server


liopeer:
Hey @thomasmarchioro3 and thank you for the contribution! We will have a look at your PR in the next few days :)

yutong-xiang-97:
Hi @thomasmarchioro3, thank you for your contribution! 

Quick note: in `example/` the notebooks are auto-generated from the example scripts for both `pytorch`, `pytorch-lightning`, and `pytorch-lightning-distributed`. So whenever you add or modify examples, make sure that:

1. you add or update the **script** examples for all three of them, then
2. generate corresponding notebooks with `make generate-example-notebooks` following our contributing guide: https://github.com/lightly-ai/lightly/blob/master/CONTRIBUTING.md

Then the "check example notebooks" test will pass.

thomasmarchioro3:
Thanks for the suggestions! I hadn't notice that the feature tensor was being transposed in `_reset_group_features`.

I also have removed the `# type: ignore`. It causes some warnings from LSPs (like Pylance or Pyright) but that's not an actual issue, since examples are not type checked.

PR #1878, opened 2025-08-07
Added type annotations and fixes for Bitmask #1635

thomasmarchioro3:
This PR addresses issue [#1635](https://github.com/lightly-ai/lightly/issues/1635) for `lightly/api/bitmask.py` and `tests/api/test_BitMask.py`. It also adds some fixes for the `BitMask` class and related tests (based on comments found in the code).

---

#### Changes to `lightly/api/bitmask.py`

- Added full type annotations (passes `mypy` without issues).
- Updated `BitMask.invert()` and `BitMask.complement()`. Made the `total_size` parameter optional: if `total_size` is not provided, it is now inferred as `len(bin(int_value))-2`.

---

#### Changes to `tests/api/test_BitMask.py`

- Added type annotations (passes `mypy` without issues).
- Added a test for `BitMask.invert()` to ensure correct behavior.
- Removed the unused `setup(self)` method:
  - It was not overriding the correct `unittest.TestCase` method (`setUp(self)`), and was unused in practice.

---

Sorry for the two PRs in a row, I was going through the codebase and fixing issues as I ran into them :)

### DIFF
--- a/lightly/api/bitmask.py
+++ b/lightly/api/bitmask.py
@@ -1,9 +1,9 @@
-""" Module to work with Lightly BitMasks """
+"""Module to work with Lightly BitMasks"""
 
 # Copyright (c) 2020. Lightly AG and its affiliates.
 # All Rights Reserved
 import copy
-from typing import List
+from typing import Any, List, Optional
 
 
 def _hex_to_int(hexstring: str) -> int:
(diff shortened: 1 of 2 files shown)


yutong-xiang-97:
Hi @thomasmarchioro3, unfortunately the `lightly.api` module is already stable and we have no plans in modifying it in any case, nor is it open for contributions from external. 

I will close this PR now. Thank you all the same for your PR though!

Issue #1888, opened 2025-11-12
LeJEPA


MohamedFarag21:
Hi all,

I have found this today, and I believe it will be a very strong addition to the library!

Looks amazing from my perspective

[leJEPA](https://arxiv.org/pdf/2511.08544)

yutong-xiang-97:
Hi @MohamedFarag21,

Thank you so much for sharing the paper! We will see if we can put this on our agenda :)

IgorSusmelj:
Remaining tasks to finish LeJepa. I suggest to do the changes in 3 pull requests if possible to make reviews easier. It should not be much work :)

## 1) Core LeJEPA Building Blocks

  - [ ] Add lightly/models/modules/lejepa.py with reusable LeJEPA components:
      - [ ] LeJEPAEncoder (wrapper around backbone + projector forward)
      - [ ] lejepa_invariance_loss(proj: Tensor) -> Tensor implementing (proj.mean(0) - proj).square().mean()
      - [x] LeJEPALoss (combines SIGReg + invariance with lambda)
  - [ ] Add a dedicated projector head in lightly/models/modules/heads.py:
      - [ ] LeJEPAProjectionHead (MLP style consistent with other methods, e.g. BN/ReLU blocks)
  - [ ] Export new classes/functions:
      - [ ] Update lightly/models/modules/__init__.py to expose LeJEPA module + projector
  - [ ] Add/extend tests together with this PR:
      - [ ] projector forward/backward shape test
      - [ ] combined LeJEPA loss wiring test (SIGReg + inv + lambda)


  ## 2) LeJEPA Transforms & Data Integration

  - [ ] Add lightly/transforms/lejepa_transform.py:
      - [ ] LeJEPATransform returning 2 global + 6 local views
      - [ ] Parameters aligned to LeJEPA defaults (global/local crop sizes and scales configurable)
      - [ ] Reuse existing transform primitives (prefer DINOViewTransform / MultiViewTransform)
  - [ ] Export transform:
      - [ ] Update lightly/transforms/__init__.py to expose LeJEPATransform
  - [ ] Ensure integration behavior matches other transform building blocks:
      - [ ] Compatible output format for existing dataset/dataloader flows
      - [ ] No new collator required for LeJEPA (multi-view transform path is sufficient)
  - [ ] Tests policy for this PR:
      - [ ] Add tests only if new logic is introduced
      - [ ] If transform is thin reuse of existing DINO-style components, no extra test required


[...]

PR #1892, opened 2026-02-05
SparK Implementation

gabrielfruet:
Addresses #1462 

Continues #1750

Inspired by: https://github.com/keyu-tian/SparK

https://arxiv.org/pdf/2301.03580

PR #1893, opened 2026-02-12
LeJEPA SIGReg Loss implementation

gabrielfruet:
Implemented the SIGReg loss from the [LeJEPA](https://arxiv.org/pdf/2301.08243) paper.

Started addressing issue #1888


liopeer:
Quick heads-up: Will review this once I am don with #1892

PR #1897, opened 2026-02-26
Resnet50 knn evaluation each epoch

Quentincami:
Modified the knn_eval to create a class allowing to perform knn evaluation during the pretraining after each epoch.
Modified the main to takes this modification into account and log every epochs evaluation.

### DIFF
--- a/benchmarks/imagenet/resnet50/knn_eval.py
+++ b/benchmarks/imagenet/resnet50/knn_eval.py
@@ -33,7 +33,7 @@ class KNNEvalCallback(Callback):
         strategy: str,
         num_classes: int, 
         knn_k: int, 
-        knn_t: int):
+        knn_t: float):
         """
         Args:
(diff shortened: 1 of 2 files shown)


benchmarks/imagenet/resnet50/knn_eval.py
```diff
@@ -13,93 +13,128 @@
...
-        batch_size=batch_size_per_device,
-        shuffle=False,
-        num_workers=num_workers,
-    )
+        classifier = KNNClassifier(
+        model=pl_module,
+        num_classes=self.num_classes,
+        knn_k=self.knn_k,
+        knn_t=self.knn_t,
+        )
```
liopeer:
```suggestion
        classifier = KNNClassifier(
            model=pl_module,
            num_classes=self.num_classes,
            knn_k=self.knn_k,
            knn_t=self.knn_t,
        )
```

Issue #1564, opened 2024-06-25
style: migrate to `ruff`


SauravMaheshkar:
[astral-sh/ruff](https://github.com/astral-sh/ruff) has emerged as a competitor to the existing linters. It's considerably faster than the current linter `pylint` viz.

![](https://user-images.githubusercontent.com/1309177/232603514-c95e9b0f-6b31-43de-9a80-9e844173fd6a.svg)

_Source: [ruff repository](https://github.com/astral-sh/ruff)_

Switching to `ruff` would enable us to switch to a single tool, as it has drop-in parity with `black`, `isort` and support for `pylint` rules.

Happy to open a PR if interested 😄 

## References:

* https://github.com/astral-sh/ruff/issues/689

philippmwirth:
Great idea, we already use `ruff` internally. We'd welcome a PR.

TakshPanchal:
I would like to contribute.

SauravMaheshkar:
@TakshPanchal sounds good, LMK if you need any help with the PR

TakshPanchal:
Currently, `isort` and `black` is being used right? So I have to `pylintrc` as reference config. From docs, 
> Ruff can be configured through a `pyproject.toml`, `ruff.toml`, or `.ruff.toml` file. 

So majorly I have to convert `pylintrc` with one of these file right?

SauravMaheshkar:
> Currently, `isort` and `black` is being used right? So I have to `pylintrc` as reference config. From docs,
> 
> > Ruff can be configured through a `pyproject.toml`, `ruff.toml`, or `.ruff.toml` file.
> 
> So majorly I have to convert `pylintrc` with one of these file right?

Yes, and let's stick with one single configuration file, so convert `pylintrc` to a ruff configuration within `pyproject.toml`.

Ideally the changes lead to no style changes, i.e. running `make lint` with ruff leads to no changes in code files.

Abhrajitdas02:
can I work on this issue?

SauravMaheshkar:
Sure, we only assign one issue at a time. So, I'll assign this issue for now. If you want to work on some other issue first, let me know.

SauravMaheshkar:
@Abhrajitdas02 would you like any help with this issue?

SauravMaheshkar:
@guarin I was working on this and noticed that we actually don't enforce `pylint` anywhere.

Source: https://github.com/search?q=repo%3Alightly-ai%2Flightly%20pylint&type=code

[...]

PR #1909, opened 2026-04-07
isort and black for python pre-commit hook

### LINKED ISSUE #1908: pre-commit does not format and lint python files.
The current pre-commit configuration is limited to Jupyter Notebooks, requiring us to manually run `make all-checks` and `make format-check`. 

I would like to extend this configuration to include python files. I often forget to run the checks and formats and this causes the CI to fail. 

https://github.com/lightly-ai/lightly/blob/904c50a7cc8bff9dbc0863a99bc71c84422322a8/.pre-commit-config.yaml#L10-L13

If this change is desirable, I'm happy to open a PR! If I overlooked something, please let me know!

gabrielfruet:
closes #1908

## Description

Included the necessary pre-commit hooks for running *isort* and *black* on python files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,3 +11,14 @@ repos:
     - id: nbqa-black
     - id: nbqa-isort
       args: ["--profile=black"]
+- repo: https://github.com/psf/black
+  rev: 23.1.0
+  hooks:
+    - id: black
+      exclude: '^lightly/openapi_generated/'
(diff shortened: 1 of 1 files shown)


liopeer (approved):
Thanks, LGTM! I agree that this makes the dev workflow nicer. I would prefer to have a single source of truth about exclusion/inclusion of directories, but for now it's good enough – and the goal is to anyway migrate to `ruff` in the near to mid-term (see #1564).

PR #1914, opened 2026-04-11
[MNT-style]: Migrate from black, isort, and pylint to ruff

RecreationalMath:
## Summary

Migrate from `black` + `isort` + `pylint` to `ruff` as a single formatting and linting tool.

Closes #1564

## Behaviour equivalence

This is a **like-for-like replacement** (Option A from the issue discussion. Refer [this comment](https://github.com/lightly-ai/lightly/issues/1564#issuecomment-4216227695)). No new enforcement is added. 

| What | Old | New | Enforced? |
|---|---|---|---|
| Formatting | `black` | `ruff format` | Yes (CI: `make format-check`) |
| Import sorting | `isort` | `ruff check --select I` | Yes (CI: `make format-check`) |
| Lint | `pylint` via `make lint` | `ruff check` via `make lint` | No (unchanged) |
| Type checking | `mypy` | `mypy` (unchanged) | Yes (CI: `make type-check`) |
| Pre-commit: format | `black` hook | `ruff-format` hook | Local only |
| Pre-commit: imports | `isort` hook | `ruff --fix --select I` hook | Local only |
| Pre-commit: lint | *(none)* | *(none)* | N/A |
| Notebooks | `nbqa-black` + `nbqa-isort` | ruff native `.ipynb` support | Local only |

## Commit structure: tooling changes vs formatting normalization

- Commits 1-5: Tooling migration - config, deps, Makefile targets, pylintrc removal, docs. Small, easy to review manually.
- Commit 6: One-time formatting normalization (~96 files). ruff's formatter has [documented deviations from black](https://docs.astral.sh/ruff/formatter/black/). All changes are formatting-only with zero semantic impact. Reviewable with `git diff --ignore-all-space`.
- Commit 7: `.git-blame-ignore-revs` so `git blame` skips the formatting commit (per @guarin's [suggestion](https://github.com/lightly-ai/lightly/issues/1564#issuecomment-4224799936)).

## What comes next

After this PR merges, follow-up work per discussion with the maintainers in [this thread](https://github.com/lightly-ai/lightly/issues/1564#issuecomment-4216227695):
1. Progressively enable broader lint rules in `[tool.ruff.lint]`
2. Enable `D` rules (pydocstyle) to enforce Google-style docstring format #1217 
3. Remove `--select I` from the pre-commit hook to enable full lint on commit
4. Add `make lint` to CI workflow (`test_code_format.yml`)
5. Fix latent lint issues module by module

## Developer note

After pulling this change, run:
```bash
[...]

PR #1918, opened 2026-04-27
[MNT-style]: Add ruff pydocstyle for Google-style docstrings

### LINKED ISSUE #1217: Figure out if we can use tool to verify docstring format
## Requirements

* Follows the [google python docstring format](https://google.github.io/styleguide/pyguide.html)
* It must be possible to ignore formatting on a per function basis

## Tasks

- [ ] Figure out if such a tool exists
- [ ] Add it to PIP and document how to use it. Best is to add a new command to the makefile

RecreationalMath:
## Summary

Adopts `ruff`'s pydocstyle plugin to verify Google-style docstrings as part of the existing toolchain. Scoped to tooling only, not docstring content cleanup as per [discussion in this thread](https://github.com/lightly-ai/lightly/issues/1217#issuecomment-4280774122).

Closes #1217

## What's enabled

| Surface | Behaviour |
|---|---|
| `[tool.ruff.lint] select` | extends to seven safely auto-fixable D rules: `D202, D209, D212, D214, D410, D411, D412` |
| `[tool.ruff.lint.pydocstyle]` | `convention = "google"` (applies the standard Google-convention exclusions) |
| `[tool.ruff.lint.per-file-ignores]` | mirrors the explicitly-deprecated paths from `[tool.mypy] exclude` |
| `[tool.ruff.format]` | `docstring-code-format = true`, normalizes Python code blocks inside docstrings |
| `.pre-commit-config.yaml` | drops `--select I` from `ruff-check` so the new D rules trigger locally |
| `CONTRIBUTING.md` | one-line note that docstrings follow the Google convention and are checked by ruff |

The selected D rules are exactly those marked `[*]` (safely auto-fixable) by `ruff check --select D --statistics` under Google convention. Manual-fix and unsafe-fix D rules (`D100-D107` undocumented-*, `D205, D401, D415, ...`) are unselected.

## Commit structure: tooling changes vs formatting normalization

Three commits, designed so the second can be skipped during review:

1. Commit 1 (config + docs): `pyproject.toml`, `.pre-commit-config.yaml`, `CONTRIBUTING.md`. Small, easy to review manually.
2. Commit 2: One-time normalisation (101 files). Pure tool output from `ruff check --fix` and `ruff format` on top of commit 1, no manual edits.
3. Commit 3: `git-blame-ignore-revs` so git blame skips the formatting commit (per @guarin's [suggestion](https://github.com/lightly-ai/lightly/issues/1564#issuecomment-4224799936))
## Notes for reviewers

- `per-file-ignores` mirrors only the explicitly deprecated portions of `[tool.mypy] exclude` (the three sub-lists with "deprecated" in their header comments). The provisionally-untyped portion (`# TODO: Remove these one by one`) is not mirrored, since the selected D rules are auto-fixable and impose no manual effort.
- Two deprecated files appear in commit 2's diff (`lightly/active_learning/config/selection_config.py`, `lightly/models/nnclr.py`). These come from `docstring-code-format = true` reformatting Python inside `>>>` blocks, not from D rules. `per-file-ignores` excludes those paths from D rule lint, not from formatter normalization. Matches #1914's accepted behaviour of running `ruff format` on all files.

## Developer note

After pulling this change, run:

```bash
pre-commit clean
pre-commit install
```

to pick up the widened pre-commit hook (D rules now run on commit, not just `make format-check`).

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,5 @@
 # Ruff formatting normalization (lightly-ai/lightly#1564)
 e7047dec9ef72f1d2d718e4915328021be661865
+
+# Ruff docstring normalization (lightly-ai/lightly#1217)
+926eadbaaed579b464303152792f02e763563279
--- a/lightly/cli/version_cli.py
+++ b/lightly/cli/version_cli.py
@@ -3,7 +3,7 @@
 
 Example:
     >>> # show the version of the installed package
-    >>> lightly-version
+    >>> lightly - version
 """
 
 # Copyright (c) 2021. Lightly AG and its affiliates.
--- a/lightly/cli/version_cli.py
+++ b/lightly/cli/version_cli.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 """**Lightly Version:** Show the version of the installed package.
 
-Example:
-    >>> # show the version of the installed package
-    >>> lightly - version
+Example::
+
+    # show the version of the installed package
+    $ lightly-version
 """
 
 # Copyright (c) 2021. Lightly AG and its affiliates.
--- a/lightly/data/_video.py
+++ b/lightly/data/_video.py
@@ -128,10 +128,10 @@ class VideoLoader(threading.local):
         >>> from torchvision import io
         >>>
         >>> # get timestamps
-        >>> ts, fps = io.read_video_timestamps('myvideo.mp4', pts_unit = 'sec')
+        >>> ts, fps = io.read_video_timestamps("myvideo.mp4", pts_unit="sec")
         >>>
         >>> # create a VideoLoader
-        >>> video_loader = VideoLoader('myvideo.mp4', ts)
+        >>> video_loader = VideoLoader("myvideo.mp4", ts)
         >>>
         >>> # get frame at specific timestamp
         >>> frame = video_loader.read_frame(ts[21])
@@ -504,7 +504,7 @@ class VideoDataset(datasets.VisionDataset):
         E.g. when retrieving a sample from the video
         `my_video.mp4` at frame 153, the filename will be:
 
-        >>> my_video-153-mp4.png
+        >>> my_video - 153 - mp4.png
 
         Args:
             index:
--- a/lightly/data/_video.py
+++ b/lightly/data/_video.py
@@ -501,10 +501,8 @@ class VideoDataset(datasets.VisionDataset):
         The filename is created from the video filename, the frame number, and
         the video format. The frame number will be zero padded to make sure
         all filenames have the same length and can easily be sorted.
-        E.g. when retrieving a sample from the video
-        `my_video.mp4` at frame 153, the filename will be:
-
-        >>> my_video - 153 - mp4.png
+        For example, when retrieving a sample from the video ``my_video.mp4``
+        at frame 153, the filename will be ``my_video-153-mp4.png``.
 
         Args:
             index:
--- a/lightly/models/_momentum.py
+++ b/lightly/models/_momentum.py
@@ -44,15 +44,15 @@ class _MomentumEncoderMixin:
     >>>         self.backbone = backbone
     >>>         self.projection_head = get_projection_head()
     >>>
-    >>>         # initialize momentum_backbone and momentum_projection_head
+    >>> # initialize momentum_backbone and momentum_projection_head
     >>>         self._init_momentum_encoder()
     >>>
     >>>     def forward(self, x: Tensor):
     >>>
-    >>>         # do the momentum update
+    >>> # do the momentum update
     >>>         self._momentum_update(0.999)
     >>>
-    >>>         # use momentum backbone
+    >>> # use momentum backbone
     >>>         y = self.momentum_backbone(x)
     >>>         y = self.momentum_projection_head(y)
 
--- a/lightly/models/_momentum.py
+++ b/lightly/models/_momentum.py
@@ -44,16 +44,12 @@ class _MomentumEncoderMixin:
     >>>         self.backbone = backbone
     >>>         self.projection_head = get_projection_head()
     >>>
-    >>> # initialize momentum_backbone and momentum_projection_head
-    >>>         self._init_momentum_encoder()
+    >>>         self._init_momentum_encoder()  # initialize momentum_backbone and momentum_projection_head
     >>>
     >>>     def forward(self, x: Tensor):
+    >>>         self._momentum_update(0.999)  # do the momentum update
     >>>
-    >>> # do the momentum update
-    >>>         self._momentum_update(0.999)
-    >>>
-    >>> # use momentum backbone
-    >>>         y = self.momentum_backbone(x)
+    >>>         y = self.momentum_backbone(x)  # use momentum backbone
     >>>         y = self.momentum_projection_head(y)
 
     """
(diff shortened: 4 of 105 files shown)


RecreationalMath:
Force-pushed to clean up commit messages: backticks in the messages had been shell-expanded during commit invocation, splicing the output of `make format-check` and `make lint` into the message text. 

Tree contents are unchanged across the three commits, except `.git-blame-ignore-revs` which now references the new SHA of the formatter commit.

lightly/cli/version_cli.py
```diff
@@ -3,7 +3,7 @@
 
 Example:
     >>> # show the version of the installed package
-    >>> lightly-version
+    >>> lightly - version
```
liopeer: This is wrong. Seems to get interpreted as Python code. Consider moving to just normal markdown/reST code fences instead.

lightly/data/_video.py
```diff
@@ -504,7 +504,7 @@ def get_filename(self, index):
         E.g. when retrieving a sample from the video
         `my_video.mp4` at frame 153, the filename will be:
 
-        >>> my_video-153-mp4.png
+        >>> my_video - 153 - mp4.png
```
liopeer: Same issue here.

lightly/models/modules/memory_bank.py
```diff
@@ -49,9 +49,9 @@ class MemoryBankModule(Module):
         >>>         output, negatives = super().forward(output)
         >>>
         >>>         if negatives is not None:
-        >>>             # evaluate loss with negative samples
+        >>> # evaluate loss with negative samples
```
liopeer: We should probably add `pass` here and then put the comments behind those.

lightly/models/_momentum.py
```diff
@@ -44,15 +44,15 @@ class _MomentumEncoderMixin:
     >>>         self.backbone = backbone
     >>>         self.projection_head = get_projection_head()
     >>>
-    >>>         # initialize momentum_backbone and momentum_projection_head
+    >>> # initialize momentum_backbone and momentum_projection_head
```
liopeer: Here we should also move the commend to behind the code (idk why it formats it this way).

lightly/transforms/irfft2d_transform.py
```diff
@@ -18,9 +18,8 @@ class IRFFT2DTransform:
     """
 
     def __init__(self, shape: Tuple[int, int]):
-        """
-        Args:
-            shape: The desired output shape (H, W) after applying the inverse FFT
+        """Args:
```
liopeer: Here we should add a proper description in the first line. After that the formatting will be applied correctly.

lightly/transforms/random_crop_and_flip_with_grid.py
```diff
@@ -30,15 +30,12 @@ class Location:
...
-    """
-    Do a random resized crop and return both the resulting image and the location. See base class.
-
-    """
+    """Do a random resized crop and return both the resulting image and the location. See base class."""
 
     def forward(self, img: Image.Image) -> Tuple[Image.Image, Location]:
-        """
-        Args:
+        """Args:
```
liopeer: Here again. Add quick one-liner explaining the logic, then the formatting will work afterwards.

lightly/utils/benchmarking/knn_classifier.py
```diff
@@ -66,7 +66,7 @@ def __init__(
             >>>         self.projection_head = SimCLRProjectionHead(512, 512, 128)
             >>>
             >>>     def forward(self, x):
-            >>>         # Forward must return image features.
+            >>> # Forward must return image features.
```
liopeer: Move to the end of a code line.

lightly/utils/benchmarking/linear_classifier.py
```diff
@@ -56,7 +55,7 @@ def __init__(
             >>>         self.projection_head = SimCLRProjectionHead(512, 512, 128)
             >>>
             >>>     def forward(self, x):
-            >>>         # Forward must return image features.
+            >>> # Forward must return image features.
```
liopeer: Move to the end of a code line.

tests/loss/test_dcl_loss.py
```diff
@@ -127,9 +125,8 @@ def test_dclloss_matches_reference(
 def dcl_reference(
     x1: torch.Tensor, x2: torch.Tensor, temperature: float
 ) -> torch.Tensor:
-    """
-    Loop-based reference:
-      L = avg_i[ -sim_pos(i) + log(sum_j exp(sim_neg(i,j))) ]
+    """Loop-based reference:
```
liopeer: Here we should probably also use an introductory line (also fine here to just ignore the formatting rules, since it's in a test).

.git-blame-ignore-revs
```diff
@@ -1,2 +1,5 @@
 # Ruff formatting normalization (lightly-ai/lightly#1564)
 e7047dec9ef72f1d2d718e4915328021be661865
+
+# Ruff docstring normalization (lightly-ai/lightly#1217)
+926eadbaaed579b464303152792f02e763563279
```
liopeer, line 5: ❤️

liopeer (changes requested):
Overall very nice! There are a few quirky reformats that are due to 1. incorrect usage of docstring REPL code, or 2. incorrect writing of the docstrings.

Should be very easy to fix by an agent. Do you want to fix it yourself? We can also take it over from our side.

RecreationalMath:
> Overall very nice! There are a few quirky reformats that are due to 1. incorrect usage of docstring REPL code, or 2. incorrect writing of the docstrings.
> 
> Should be very easy to fix by an agent. Do you want to fix it yourself? We can also take it over from our side.

Thanks for the review. 
I'll look into the comments and update the PR.

lightly/cli/version_cli.py
RecreationalMath: Moving it to reST format.

lightly/models/_momentum.py
RecreationalMath: > (idk why it formats it this way).

I looked into this a bit, and it seems the all `>>>` doctest input is not what `ruff`'s formatter is designed around. Per [ruff/#8908](https://github.com/astral-sh/ruff/issues/8908), it expects PS1 (`>>>`) for statement starts and PS2 (`...`) for continuations. When a codebase (such is the case for `lightly`) uses `>>>` for everything, `ruff` makes a best-effort interpretation, but the result for edge cases (like comments inside multi-statement classes with blank `>>>` lines) is undefined and can produce surprising output.

> Here we should also move the commend to behind the code

As per your suggestion to move the comment after the code, the line becomes too long.

```python
    >>>         self._init_momentum_encoder()  # initialize momentum_backbone and momentum_projection_head
```

I personally find using continuation `...` to be cleaner as compared to lengthy lines (better readability), but I guess it would be going away from the package-wide convention.

lightly/utils/benchmarking/knn_classifier.py
RecreationalMath: same comment as one of the above. Doing so would make the line a bit too long than standard 88 chars. But doing it nonetheless.

tests/loss/test_dcl_loss.py
RecreationalMath: Added an introductory line. Sticking to the convention seems like a better choice than to add a `file-ignore`.

RecreationalMath:
Addressed all the issues flagged. I also ran a sweep across all the 101 files reformatted by the formatting commit, looking for the same four patterns (non-Python content in `>>>`, dedented standalone comments, empty `if`/`else` bodies, missing summary lines). Found nothing else.

Please take another pass when you have a moment. Happy to iterate further if anything else comes up.

liopeer (approved):
LGTM! Thanks a lot!

Issue #1894, opened 2026-02-13
PIXIO Implementation Plan


gabrielfruet:
PIXIO differs from standard MAE in three key differences:

1. Deeper decoder
2. More mask granularity
3. Increase of CLS tokens

Research repo is hosted here:

[facebookresearch/pixio](https://github.com/facebookresearch/pixio)

## 1. Deeper decoder

MAE Decoder generally has 8 blocks due to their philosophy of letting the encoder do the hard work. Pixio drifts from this by explictly stating that the encoder is responsible for encoding semantics and the decoder is responsible for caring about the pixel-level detail. By separating this they create a much richer latent space for the encoder.

[Decoder forward of Pixio](https://github.com/facebookresearch/pixio/blob/b83555d6581f4906f6f1fc1ae0abb04f622db32f/pretraining/models_pixio.py#L222-L256)

[Our MAEDecoder](https://github.com/lightly-ai/lightly/blob/ee30cd481d68862c80de4ef45920cfe1ab1f67b1/lightly/models/modules/masked_autoencoder.py#L380C1-L506C32)

This could be a module, just like we have the MAEDecoder. We could likely subclass from it also.

## 2. More mask granularity

They use a mask that is not simply randomly dropping points, they mask by blocks.

![](https://github.com/facebookresearch/pixio/raw/main/assets/pixio.png)

We also already have that implemented

[Block Masking](https://github.com/lightly-ai/lightly/blob/ee30cd481d68862c80de4ef45920cfe1ab1f67b1/lightly/models/utils.py#L654-L747)

## 3. Increase of CLS tokens

An example for `timm` models is to increase the number of `reg_tokens`

By default it has only 1 CLS token, but we can add more via REG tokens.

[VisionTransformer from timm](https://github.com/huggingface/pytorch-image-models/blob/56467ad5f2b443d454df1a5bb8b23c04697823c7/timm/models/vision_transformer.py#L807-L808)

-------

Let me know your thoughts.

PR #1919, opened 2026-04-27
Add random block wise mask from Pixio

gabrielfruet:
#1894 

This is part of a series of PR's to implement the Pixio paper. 

- Tests added
- Documentation referencing the author
- Using same API as the other masking functions

The only difference is the `num_prefix_tokens` vs `mask_class_token`. Since Pixio relies on register tokens, the `mask_class_token` was not suitable for choosing whether to mask or not the REG/CLS tokens.

### DIFF
--- a/lightly/models/utils.py
+++ b/lightly/models/utils.py
@@ -614,6 +614,121 @@ def random_token_mask(
     return idx_keep, idx_mask
 
 
+def random_block_wise_mask(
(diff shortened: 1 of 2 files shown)


lorinczszabolcs:
Likely this can be closed in favor of #1965

gabrielfruet:
True! We can close this one. Thank you!

PR #1922, opened 2026-05-07
[BugFix] Skip TestGatherLayer_Losses on macOS

### LINKED ISSUE #1917: [BUG] macOS: `make test-fast` hangs indefinitely
## Description

Running `make test-fast` on macOS hangs indefinitely during `tests/utils/test_dist__gather__losses.py`. The pytest process enters a sleeping state and never makes further progress. I left it running for over 70 minutes and observed only about 1 minute of cumulative CPU time.

There is no exception and no traceback, so nothing is printed when the hang occurs. Nothing in the terminal indicates which file is responsible.

## Reproduction

On macOS:

```bash
git clone https://github.com/lightly-ai/lightly.git
cd lightly
uv venv
source .venv/bin/activate
make install-dev
make test-fast
```

Or isolated to the offending file:

```bash
pytest tests/utils/test_dist__gather__losses.py -v
```

**Expected**: the suite completes in a few minutes.

**Actual**: hangs after 1 to 2 minutes of apparent progress, at or during `tests/utils/test_dist__gather__losses.py`. The hang is presistant on macOS across repeated runs.

## Orphaned child processes after kill

If the hung pytest is terminated (`Ctrl-C` or `kill`), the spawned child process is frequently orphaned. The orphan remains in a sleeping state with roughly 500 MB to 1 GB resident memory and does not exit on its own, it must be cleaned up manually with `kill -9 <pid>`. Observed repeatedly during investigation.

## Not observed in CI

The four tests in `TestGatherLayer_Losses` are guarded by a class-level `pytest.mark.skipif` decorator gated on the `GITHUB_ACTIONS` environment variable, so they are always skipped on GitHub Actions runners. The existing skip reason attributes the skip to a codecov compatibility concern, not a platform compatibility concern. Local macOS contributors have no equivalent skip applied.

## Environment

- OS: macOS 26.4.1 (arm64)
- Python: 3.12.13
- torch: 2.11.0
- pytorch-lightning: 2.6.1
- Installed via: `make install-dev` at commit `c792812` on `master`
- Invocation: `make test-fast`

## Impact

- Affected: contributors running the full test suite on macOS locally.
- Not affected: GitHub Actions CI and Linux contributors.
- Severity: high friction for new and returning macOS contributors. There is no error, no progress indication, and the hang is easy to mistake for a slow test. Cleanup after a kill also requires manual intervention to remove orphaned child processes.

RecreationalMath:
closes #1917

## Description

`TestGatherLayer_Losses` uses `DDPStrategy(devices=2)` under pytest. On macOS the spawn start method causes the DDP worker child to re-import the test module and block on pytest's fixture machinery, while the parent waits for the child to join the DDP process group. The suite hangs indefinitely with no error.

The same class is already skipped on GitHub Actions via `pytest.mark.skipif` for codecov reasons, so this PR adds `sys.platform == "darwin"` to that condition. Empirical reproduction (~40 runs) and root-cause analysis are in the issue thread.

Manual verification on macOS 26.4.1 (arm64):
- `pytest tests/utils/test_dist__gather__losses.py -v`: 4 skipped, no hang (5.8s).
- `make test-fast`: 1560 passed, 258 skipped, no hang (3:30).
- `make format`: ruff clean.

No documentation changes required: this is a test-file skip-condition adjustment with no public API or docs surface.

## Tests
- [ ] I need help on it.


RecreationalMath:
@liopeer, could you please initiate the test run and review this? Thanks!

gabrielfruet:
Thank you for the PR! Sorry for the delay. I'll try to review it this week!

RecreationalMath:
The failed CI job died during the dependency install step, not a code/test failure.
Re-running the job should suffice.

PR #1926, opened 2026-05-18
refactor: simplify LeJEPA loss and module surface

gabrielfruet:
related to https://github.com/lightly-ai/lightly/issues/1888

## Description
- [x] My change is breaking

This PR starts addressing the LeJEPA follow-up work in #1888.

Changes:
- aligned LeJEPA defaults and behavior more closely with the paper
- updated the loss to support separate local views and global views
- removed the unnecessary `LeJEPAEncoder` wrapper to reduce coupling and keep the module surface focused on reusable building blocks

## Tests

## Documentation

## Improvements put into another issue:
- #1888

## Issues covering the breaking change:
- https://github.com/lightly-ai/lightly/issues/1888


lightly/loss/lejepa_loss.py
```diff
@@ -8,27 +8,57 @@
 from lightly.utils import dist as lightly_dist
 
 
-def lejepa_invariance_loss(proj: Tensor) -> Tensor:
+def lejepa_invariance_loss(*, local_proj: Tensor, global_proj: Tensor) -> Tensor:
```
liopeer, line 11: Thanks for the kwarg-only implementation! ❤️ This is always so much easier to maintain in the future.

lightly/loss/lejepa_loss.py
```diff
@@ -38,7 +68,7 @@ def __init__(
         self,
         knots: int = 17,
```
liopeer, line 70: Btw, here it's also always helpful to only allow kwargs.

lightly/loss/lejepa_loss.py
```diff
@@ -236,15 +268,19 @@ def __init__(
...
-            proj: Projected embeddings of shape ``(V, N, D)``.
+            local_proj: Local-view projected embeddings of shape ``(Vl, N, D)``.
+            global_proj: Global-view projected embeddings of shape ``(Vg, N, D)``.
         """
-        sigreg_loss = self.sigreg(proj)
-        inv_loss = lejepa_invariance_loss(proj)
-        loss: torch.Tensor = (
+        _validate_projection_shapes(local_proj=local_proj, global_proj=global_proj)
+        sigreg_loss = self.sigreg(local_proj)
+        inv_loss = lejepa_invariance_loss(
```
liopeer, line 282: Now you're validating the projection shapes twice which is a bit overkill.

tests/loss/test_lejepa_loss.py
```diff
@@ -1,3 +1,5 @@
+from typing import Tuple
```
liopeer: Prefer using `tuple` and `from __future__ import annotations`.

liopeer (approved):
LGTM with minor comments. None of them absolutely have to be addressed. Feel free to address the ones you find valid and resolve yourself!

lightly/loss/lejepa_loss.py
gabrielfruet: When I was writing the example I managed to swap both 🤣. So for no one to fall for the same mistake, I did the kwargs only

PR #1929, opened 2026-05-24
feat: Implement DSE (Dense representation Structure Estimator)

IgorSusmelj:
## Motivation

This is complementary (partially overlapping) with gram anchoring from DINOv3. There is a [NeurIPS 2025 paper](https://arxiv.org/abs/2510.17299) that looks more carefully in degradation for dense tasks in SSL pretraining and proposes a more general regulariser (not a post training fix like gram anchoring. And it should work for all SSL methods:
- works for conv and transfomer backbones!

## Changes
-  Adds DSERegularizer, a dense-representation regularizer based on the Dense Structure Estimator (DSE)
- Accepts (B, N, D) dense representations (transformer patch tokens or flattened conv features) and returns a scalar loss term to add to any existing SSL loss
- Implements k-means-based structure estimation: local intra-image clustering, global inter/intra-batch clustering, and effective dimensionality via spectral entropy
- Uses k-means++ initialization with squared-distance weighting, correcting a bug in the reference implementation
- Includes full test coverage: forward/backward pass, weight scaling, input validation, and numerical comparison of the deterministic components against the reference formulas
- Registered in lightly.loss.regularizer and documented in the API reference

### DIFF
--- a/lightly/loss/regularizer/dse.py
+++ b/lightly/loss/regularizer/dse.py
@@ -1,0 +1,253 @@
+"""DSE regularizer."""
+
+import math
+from typing import Tuple
+
+import torch
+import torch.nn.functional as F
+from torch import Tensor
+from torch.nn import Module
+
+
+class DSERegularizer(Module):
+    """Dense Structure Estimator regularizer for dense representations.
+
+    This experimental regularizer is based on the training regularizer variant from "Exploring
+    Structural Degradation in Dense Representations for Self-supervised
+    Learning", 2025, https://arxiv.org/abs/2510.17299.
+
+    The regularizer expects dense representations of shape ``(B, N, D)``, where
+    ``B`` is the batch size, ``N`` is the number of dense locations such as
+    patches or spatial positions, and ``D`` is the feature dimension. Higher DSE
+    values indicate better dense structure, so this module returns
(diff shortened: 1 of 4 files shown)


lightly/loss/regularizer/dse.py
```diff
@@ -0,0 +1,253 @@
...
+        batch_size = representations.shape[0]
+        terms = []
+        for batch_index in range(batch_size):
+            image_representations = representations[batch_index]
+            _, labels = self._kmeans(
+                image_representations.detach(), self.local_clusters
+            )
+            ranks = []
+            for cluster_index in torch.unique(labels):
+                cluster = image_representations[labels == cluster_index]
```
gabrielfruet, line 153: Why we don't check this? Is cluster already guaranteed to have at least 1 positive? 

https://github.com/EldercatSAM/SSL-Degradation/blob/b75fd515c1fe90ab7b85798a2b6e6f8ed380d78c/DSE_regularizer.py#L219-L227

gabrielfruet (approved):

lightly/loss/regularizer/dse.py
IgorSusmelj: We have torch.unique which only returns non empty clusters and then we check for the shape. It should be equivalent.

PR #1938, opened 2026-06-11
fix(loss): raise clear error for single-view DINOLoss instead of silent NaN

devteamaegis:
## What's broken

`DINOLoss` returns a silent `NaN` when called with a single teacher view and single student view — which is exactly the usage shown in the class docstring (`loss_fn([teacher_out], [student_out])`). NaN silently poisons training rather than surfacing an error.

```python
loss_fn = DINOLoss(output_dim=8)
loss_fn([torch.randn(4,8)], [torch.randn(4,8)]).item()   # nan
```

## Why it happens

With one view on each side the loss matrix is 1×1; `fill_diagonal_(0)` zeros its only entry, so `n_terms = numel - diagonal = 0`, and `loss.sum() / (n_terms * batch_size) = 0/0 = NaN`. Single-view DINO is genuinely degenerate (no cross-view distillation), but the code produces NaN instead of a clear error, and the docstring advertises this exact call.

## Fix

Raise a clear `ValueError` when there are no cross-view terms (matching the codebase's convention of raising on degenerate loss inputs), and correct the docstring example to use two views.

## Test

`test_single_view_raises` asserts the single-view call raises `ValueError`; the rest of the DINO loss suite is unchanged.

### DIFF
--- a/lightly/loss/dino_loss.py
+++ b/lightly/loss/dino_loss.py
@@ -44,15 +44,15 @@ class DINOLoss(Module):
         >>> # initialize loss function
         >>> loss_fn = DINOLoss(128)
         >>>
-        >>> # generate a view of the images with a random transform
-        >>> view = transform(images)
+        >>> # generate two views of the images with a random transform
+        >>> view0, view1 = transform(images), transform(images)
         >>>
-        >>> # embed the view with a student and teacher model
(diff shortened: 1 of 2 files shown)


liopeer:
@codex review

gabrielfruet (approved):

PR #1941, opened 2026-06-12
fix: return plain Python float from cosine_schedule

### LINKED ISSUE #1902: `cosine_schedule` returns `np.float64` instead of `float`, breaking PyTorch checkpoint loading
In [lightly/utils/scheduler.py line 61](https://github.com/lightly-ai/lightly/blob/12da2867f59ef3eff7ca1e3eef3aad68df30e531/lightly/utils/scheduler.py#L61), the cosine_schedule function returns a value of type np.float64 when step < max_step, but returns a plain Python float in the else-branch. This inconsistency causes a downstream issue when the returned decay value is saved into a PyTorch checkpoint.

PyTorch introduced stricter safety mechanisms for checkpoint loading (via torch.load with weights_only=True). When a checkpoint contains a np.float64 value, loading it requires explicitly allowlisting NumPy types as safe globals, e.g.:
```
torch.serialization.add_safe_globals([numpy.float64])
```

# Fix
Simply typecasting to `float(...)` will fix this.

saud5150:
closes #1902

## Description
- [ ] My change is breaking

`cosine_schedule` returns `np.float64` instead of `float` when the value goes through `np.cos` (both the cosine decay branch and the `period` branch), while the end-value branches return a plain `float`. When the value is stored in a checkpoint (e.g. as an EMA momentum or weight decay), loading it with `torch.load(weights_only=True)` — the default since PyTorch 2.6 — fails with an `UnpicklingError` unless numpy types are allowlisted via `torch.serialization.add_safe_globals`.

This PR casts the value to `float` at the single return point, so all branches return a plain Python float as the signature and docstring already promise. This also covers the end-value branches, which previously returned `end_value` unchanged (an `int` when called like `cosine_schedule(epoch, 10, 0.996, 1)` as in the examples). `cosine_warmup_schedule` and `CosineWarmupScheduler` are fixed transitively. Values are unchanged; only the type is normalized.

## Tests

Added `test_cosine_schedule__returns_python_float` in `tests/utils/test_scheduler.py`, parametrized over all return branches (cosine decay, `max_steps <= 1`, `step >= max_steps - 1`, int `end_value`, and `period`). The test asserts `type(decay) is float` — `isinstance` would not catch the bug since `np.float64` subclasses `float`. The test fails on master for the two `np.cos` branches and passes with the fix.

Manually verified the reported scenario: before the fix, `torch.save({"momentum": cosine_schedule(1, 10, 1.0, 0.0)})` followed by `torch.load(buf, weights_only=True)` raises `UnpicklingError`; after the fix it loads fine. `make static-checks` passes (ruff format/lint clean, mypy clean) and the full test suite passes apart from 4 pre-existing macOS/Python 3.9 environment failures that also occur on a clean master checkout and are unrelated to this change.

## Documentation

No API change — the signature already declares `-> float`; the implementation now matches it. No docstring or `.rst` changes needed.

A possible follow-up: `np.cos`/`np.pi` in `cosine_schedule` could be replaced with `math.cos`/`math.pi`, removing the module's numpy dependency entirely. Left out to keep this diff minimal.


gabrielfruet:
Thank you for your contribution and feel welcome as a new LightlySSL contributor :) 

Can you add a test for that? A roundtrip style test, to avoid someone to break the weight loading in the future.

saud5150:
Thanks for the warm welcome, happy to contribute to LightlySSL!

gabrielfruet:
Apparently you used weights_only, but float is not a weight/Tensor. That's why tests are faillling.

gabrielfruet:
I'd say that skipping the test entirely if weight loading for float is not available is too broad. You can use `pytest.skip` based on the torch version. 

This is probably due to the support for python 3.7 :( . I'm going to raise an issue for us to update the version. 

If finding the version that this breaks (appears to be pytorch 2.6) is too hard, let me know and we can figure out a better way or use this one Claude proposed.

saud5150:
Thanks for the feedback. You're right, the version guard is cleaner and more explicit than the probe approach.

gabrielfruet (approved):

Issue #1942, opened 2026-06-12
Add missing docstrings to public ViewTransform classes


saud5150:
14 public view-transform classes have no docstring:

- `SimCLRViewTransform`
- `BYOLView1Transform`, `BYOLView2Transform`
- `DINOViewTransform`
- `IBOTViewTransform`
- `MSNViewTransform`
- `SimSiamViewTransform`
- `SwaVViewTransform`
- `SmoGViewTransform`
- `VICRegViewTransform`, `VICRegLViewTransform`
- `FDAView1Transform`, `FDAView2Transform`
- `DetConSViewTransform`

All are exported from `lightly.transforms` and rendered by autodoc in `docs/source/lightly.transforms.rst`, so they currently show up in the docs without any description.

I'd like to add Google-style docstrings mirroring each parent transform's docstring (same `Attributes:` section, since the parameters are identical), split into ~4 PRs grouped by method family so reviews stay small. First PR (SimCLR only) is ready and can serve as the format pilot — happy to adjust the format there before doing the remaining classes.

liopeer:
Perfect, we're very happy to receive those!

hiSandog:
A docs fix here would be more useful if it includes a minimal, copy-pasteable example for “Add missing docstrings to public ViewTransform classes” rather than only describing the behavior. A short verification step would also help users tell whether `SimCLRViewTransform` is configured correctly or they are hitting a separate environment problem.

liopeer:
Hi I agree @hiSandog! However, this issue was actually completely worked through by @saud5150 already anyway, so I will skip adding it now and close it. 🙂

PR #1946, opened 2026-06-17
docs: add docstrings to BYOLView1Transform and BYOLView2Transform

saud5150:
Part of #1942

## Description
- [ ] My change is breaking

Follow-up to the review on this PR (raised by @liopeer): align the BYOL
view-transform docstrings with the Google style we want to standardise on.

- Move the constructor parameter docs out of the class-level `Attributes:`
  section into an `Args:` section inside each `__init__` docstring.
  `Attributes:` is for public instance variables, which these classes don't
  expose (only the internal `self.transform`), so the parameters belong under
  `__init__`.
- Use compact Google style: argument name and description on the same line
  (also applied to the two `__call__` docstrings so the file is internally
  consistent).
- Register `lightly.transforms.byol_transform` in
  `docs/source/lightly.transforms.rst` — it was missing entirely — with
  `:special-members: __call__, __init__` so the `__init__` docstrings render
  under autodoc.

Docs only — no runtime behaviour changes.

## Tests

No code change, so existing tests apply. Verified:
- `make format-check` — clean
- `make lint` — clean
- `make type-check` — clean (533 source files)
- `python -m pytest tests/transforms` — 60 passed
- `make html-noplot` (in `docs/`) — build succeeded; the 17 warnings are
  pre-existing on master and unrelated to this file. Confirmed the rendered
  `build/html/lightly.transforms.html` now shows the `BYOLView1Transform` and
  `BYOLView2Transform` `__init__` parameters.

## Documentation
- [x] My change requires a change to the documentation (`.rst` files).

- The remaining undocumented `*ViewTransform` classes (#1942) follow in separate
  PRs grouped by method family, using this same format.
- `SimCLRViewTransform` (#1943) merged with the old `Attributes:` style; I'll
  update it to match in a follow-up, as agreed in the review thread.

### DIFF
--- a/docs/source/lightly.transforms.rst
+++ b/docs/source/lightly.transforms.rst
@@ -7,6 +7,10 @@ lightly.transforms
    :members:
    :special-members: __call__
 
+.. automodule:: lightly.transforms.byol_transform
+   :members:
+   :special-members: __call__, __init__
+
 .. automodule:: lightly.transforms.densecl_transform
    :members:
    :special-members: __call__
--- a/lightly/transforms/byol_transform.py
+++ b/lightly/transforms/byol_transform.py
@@ -31,52 +31,6 @@ class BYOLView1Transform:
 
     - [0]: Bootstrap Your Own Latent, 2020, https://arxiv.org/pdf/2006.07733.pdf
 
-    Attributes:
-        input_size:
-            Size of the input image in pixels.
-        cj_prob:
-            Probability that color jitter is applied.
-        cj_strength:
-            Strength of the color jitter. `cj_bright`, `cj_contrast`, `cj_sat`, and
-            `cj_hue` are multiplied by this value. For datasets with small images,
-            such as CIFAR, it is recommended to set `cj_strength` to 0.5.
-        cj_bright:
-            How much to jitter brightness.
-        cj_contrast:
-            How much to jitter contrast.
-        cj_sat:
-            How much to jitter saturation.
-        cj_hue:
-            How much to jitter hue.
-        min_scale:
-            Minimum size of the randomized crop relative to the input_size.
-        random_gray_scale:
-            Probability of conversion to grayscale.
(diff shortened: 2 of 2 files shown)


liopeer:
Hey @saud5150, we had a quick internal discussion and aligned that since we're already adding these docstrings we should actually properly follow the Google docstring format (which we're supposed to follow, but we do it inconsistently).

The issues currently are:
- arguments and argument descriptions are not on the same line
- we only have `Attributes`, which is supposed to document "publicly accessible" attributes of the class, not the arguments to the constructor
- there are two ways of documenting the arguments of a constructor: either in the `__init__` method or inside the class docstring; now we do neither. I personally think that the `__init__` docstring is preferrable.

For these reasons I would prefer if we stick to the following guidelines for these PRs:

## Docstring Guidelines

### 1. Attributes vs. Args

* **`Attributes`**: Use only in the **class docstring** to document public, stable class variables.
* **`Args`**: Use inside the **`__init__` docstring** (preferred) to document initialization parameters.

### 2. Google-Style Formatting

Keep descriptions compact. The argument name and its explanation must start on the **same line**:

```python
class SimCLRViewTransform:
    """Transforms an image into a single view.
    
    Attributes:
        public_attribute: A stable, public class variable.
    """

    def __init__(self, input_size: int = 224):
        """Initializes the transform.
        
        Args:
            input_size: Size of the input image in pixels.
        """

```

### 3. Sphinx Documentation Config

To ensure `__init__` docstrings actually show up in the generated HTML documentation, add the `:special-members:` flag to the class directive:

```rst
.. autoclass:: lightly.path.to.ClassName
   :members:
   :special-members: __init__

```
We can later activate this globally.

saud5150:
Thanks — I agree on all three points. I’m happy to update SimCLRViewTransform from #1943 to match as well

liopeer (approved):
LGTM!

PR #1947, opened 2026-06-17
chore: add mypy to pre-commit

### LINKED ISSUE #1912: add mypy to pre-commit
The code base currently has some `mypy` errors. Therefore, we can't simply add the `mypy` to the pre-commit hooks, because we wouldn't be able to commit due to the errors. 

I'll open some PRs to:

1. Fix `mypy` errors
2. Add `mypy` to pre-commit

`mypy` errors:

```
#mypy lightly tests
lightly/models/modules/masked_causal_vision_transformer.py: note: In member "forward" of class "MaskedCausalAttention":
lightly/models/modules/masked_causal_vision_transformer.py:21: error: Signature of "forward" incompatible with supertype "Attention"  [override]
lightly/models/modules/masked_causal_vision_transformer.py:21: note:      Superclass:
lightly/models/modules/masked_causal_vision_transformer.py:21: note:          def forward(self, x: Tensor, attn_mask: Tensor | None = ..., is_causal: bool = ...) -> Tensor
lightly/models/modules/masked_causal_vision_transformer.py:21: note:      Subclass:
lightly/models/modules/masked_causal_vision_transformer.py:21: note:          def forward(self, x: Tensor, mask: Tensor | None = ...) -> Tensor
lightly/models/modules/masked_causal_vision_transformer.py: note: In member "forward" of class "MaskedCausalBlock":
lightly/models/modules/masked_causal_vision_transformer.py:126: error: Signature of "forward" incompatible with supertype "Block"  [override]
lightly/models/modules/masked_causal_vision_transformer.py:126: note:      Superclass:
lightly/models/modules/masked_causal_vision_transformer.py:126: note:          def forward(self, x: Tensor, attn_mask: Tensor | None = ..., is_causal: bool = ...) -> Tensor
lightly/models/modules/masked_causal_vision_transformer.py:126: note:      Subclass:
lightly/models/modules/masked_causal_vision_transformer.py:126: note:          def forward(self, x: Tensor, mask: Tensor | None = ...) -> Tensor
lightly/models/modules/masked_causal_vision_transformer.py: note: In member "forward_features" of class "MaskedCausalVisionTransformer":
lightly/models/modules/masked_causal_vision_transformer.py:170: error: Signature of "forward_features" incompatible with supertype "VisionTransformer"  [override]
lightly/models/modules/masked_causal_vision_transformer.py:170: note:      Superclass:
lightly/models/modules/masked_causal_vision_transformer.py:170: note:          def forward_features(self, x: Tensor, attn_mask: Tensor | None = ..., is_causal: bool = ...) -> Tensor
lightly/models/modules/masked_causal_vision_transformer.py:170: note:      Subclass:
lightly/models/modules/masked_causal_vision_transformer.py:170: note:          def forward_features(self, x: Tensor, mask: Tensor | None = ...) -> Tensor
tests/loss/test_detcon_loss.py: note: In function "byol_nce_detcon":
tests/loss/test_detcon_loss.py:223: error: Incompatible types in assignment (expression has type "ndarray[tuple[Any, ...], dtype[float64]]", variable has type "ndarray[tuple[int, int], dtype[float64]]")  [assignment]
Found 4 errors in 2 files (checked 530 source files)

```

saud5150:
closes #1912

## Description

mypy was missing from `.pre-commit-config.yaml` despite `pyproject.toml`
already pinning `mypy==1.4.1` with a comment saying the versions should
stay in sync. `mirrors-mypy` is not viable here because `torch` cannot be
listed as an `additional_dependency` — the hook's isolated environment has
no access to torch stubs, causing hundreds of cascading false errors. A
`repo: local` hook with `language: system` runs mypy inside the project's
venv instead, matching exactly what `make type-check` already does.

Changes:
- Add mypy hook to `.pre-commit-config.yaml` with `language: system`,
  `pass_filenames: false`, and `args: [lightly, tests]`
- Add `types-PyYAML` to `[dev]` deps in `pyproject.toml` (required because
  the new test imports `yaml`)

## Documentation
- [x] I have updated the documentation.

## Tests

Added `test_pre_commit_config__mypy_hook_config` to
`tests/test_requirements.py` — asserts the hook is present, uses
`language: system`, sets `pass_filenames: false`, and passes
`lightly`/`tests` as targets.

Verified on latest upstream master (`fd1eee44`):
```
make format-check   # clean
make type-check     # no issues found in 533 source files
pytest tests/test_requirements.py -v  # 3 passed
pre-commit run mypy --all-files        # passed
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,3 +11,12 @@ repos:
     - id: ruff-check
       args: [--fix]
     - id: ruff-format
+- repo: local
+  hooks:
+  - id: mypy
+    name: mypy
+    entry: mypy
+    language: system
+    types: [python]
+    pass_filenames: false
+    args: [lightly, tests]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+default_install_hook_types: [pre-commit, pre-push] # install both git hooks (mypy runs on pre-push)
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v4.6.0
@@ -19,4 +20,5 @@ repos:
     language: system
     types: [python]
     pass_filenames: false
+    stages: [pre-push] # a full mypy run is too slow for every commit
     args: [lightly, tests]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,7 @@ dev = [
   "types-python-dateutil",
   "types-toml",
   "types-requests",
+  "types-PyYAML",
   "nbformat",
   "jupytext"
 ]
--- a/tests/test_requirements.py
+++ b/tests/test_requirements.py
@@ -25,8 +25,9 @@ def test_requirements_base__openapi() -> None:
 
 
 def test_pre_commit_config__mypy_hook_config() -> None:
-    """Check mypy hook uses the project's installed environment and runs on
-    the full project rather than individual staged files.
+    """Check mypy hook uses the project's installed environment, runs on the
+    full project rather than individual staged files, and is scoped to the
+    pre-push stage (a full mypy run is too slow for every commit).
     """
     hook = _pre_commit_mypy_hook()
     assert hook is not None, "mypy hook not found in .pre-commit-config.yaml"
@@ -42,6 +43,10 @@ def test_pre_commit_config__mypy_hook_config() -> None:
     assert "lightly" in args and "tests" in args, (
         "mypy hook must pass 'lightly' and 'tests' as args — without them mypy "
         "runs with no targets and checks nothing"
+    )
+    assert "pre-push" in hook.get("stages", []), (
+        "mypy hook must be scoped to the pre-push stage so a full mypy run "
+        "does not slow down every commit"
     )
 
 
--- a/tests/test_requirements.py
+++ b/tests/test_requirements.py
@@ -1,7 +1,8 @@
 from pathlib import Path
-from typing import List, Set
+from typing import Any, Dict, List, Optional, Set, cast
 
 import toml
+import yaml
 
 
 def test_requirements_base__pyproject() -> None:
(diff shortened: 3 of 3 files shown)


gabrielfruet:
Thank you for the nice work!

I tested on my machine and it worked perfectly! The only problem is that `mypy` is so painfully slow haha.

I was thinking that maybe hooking into `pre-push` instead of `pre-commit` can be a better dev experience. Do you have any opinion?

saud5150:
Thanks for testing it out and for the feedback!
Hooking it into `pre-push` sounds like a good compromise: we still enforce type checks before code is pushed, but keep the inner dev loop fast.

Borda:
Would it be better to enable the pre-commot bot so it is tested on each PR?

saud5150:
Good question! A couple of things:

mypy already runs on every PR — test_code_format.yml runs make type-check (Python 3.7 + 3.12, full env), so PRs are covered.

And pre-commit.ci can't actually run this hook: it needs the project env (mypy + torch), which isn't available in the bot's sandbox — same reason mirrors-mypy didn't work here. Plus it's a pre-push hook, which the bot doesn't run. So the local hook stays a fast convenience while CI does the enforcing.

gabrielfruet:
Hi @Borda,

Thanks for the suggestion, that’s a good point!

In this case, I don’t think `pre-commit.ci` would help much with this particular hook. The `mypy` check needs to run inside the project environment because of dependencies like `torch`, which is why we use a
local `language: system` hook instead of `mirrors-mypy`.

Also, since we moved this to `pre-push` to avoid slowing down every commit, it is mainly intended as a local developer check before pushing.

The actual enforcement still happens in CI via `make type-check`. The local hook is mostly there to give contributors faster feedback before pushing or updating a PR, especially when workflows from external
contributors need approval before they can run.

gabrielfruet (approved):

gabrielfruet:
Thank you for the contribution @saud5150  !!

PR #1948, opened 2026-06-18
docs: align SimCLR transform docstrings with Google style

saud5150:
Part of #1942

## Description
- [ ] My change is breaking

Follow-up to the maintainer feedback in #1946: move constructor parameter
documentation from the class-level `Attributes:` section into compact same-line
Google-style `Args:` sections in each `__init__` docstring, matching the format
that landed in #1946 for the BYOL transforms.

Covers both classes in `simclr_transform.py`:
- `SimCLRViewTransform` — deletes `Attributes:`, adds `__init__` `Args:`
- `SimCLRTransform` — same treatment; also removes the duplicate Input/Output
  block that was left in the class docstring

Adds `:special-members: __init__` to the `simclr_transform` autodoc entry so
constructor parameters render in the Sphinx docs. Docs only — no runtime changes.

## Tests

```
make format-check          # All checks passed
make lint                  # All checks passed
make type-check            # Success: no issues found in 533 source files
python -m pytest tests/transforms  # 60 passed
```

## Documentation
- [x] My change requires a change to the documentation ( `.rst` files).

Constructor parameters now render under each class's `__init__` heading in
Sphinx, matching the BYOL transform pages from #1946.

- The remaining undocumented `*ViewTransform` classes from #1942 can follow the
  same `__init__` Args pattern once this format is confirmed.

### DIFF
--- a/docs/source/lightly.transforms.rst
+++ b/docs/source/lightly.transforms.rst
@@ -77,7 +77,7 @@ lightly.transforms
 
 .. automodule:: lightly.transforms.simclr_transform
    :members:
-   :special-members: __call__
+   :special-members: __call__, __init__
 
 .. automodule:: lightly.transforms.simsiam_transform
    :members:
--- a/lightly/transforms/simclr_transform.py
+++ b/lightly/transforms/simclr_transform.py
@@ -32,56 +32,6 @@ class SimCLRTransform(MultiViewTransform):
     - [0]: SimCLR v1, 2020, https://arxiv.org/abs/2002.05709
     - [1]: SimCLR v2, 2020, https://arxiv.org/abs/2006.10029
 
-    Input to this transform:
-        PIL Image or Tensor.
-
-    Output of this transform:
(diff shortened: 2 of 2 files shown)


liopeer (approved):
LGTM, thanks a lot! 

One side note (this is just in case it appears in some of the follow-up PRs, but was not the case here; also important for @gabrielfruet ): If in another class we do actually add constructor arguments as class attributes, we will need to keep the "Attributes" section for those to not introduce breaking changes for the users, i.e. if we start with the following class
```python
class SomeTransform:
    """SomeClass description.

    Attributes:
        arg1: some description
        arg2: some other description
    """
    def __init__(self, arg1, arg2):
        self.arg1 = args1
        self.transform = Compose([Resize(arg1), Normalize(arg2)])
```
we won't be able to remove the "Attributes" section entirely. We'll need to keep the `arg1` in there since someone might be accessing it in their code.

PR #1950, opened 2026-06-22
docs: add docstrings to DINOViewTransform and IBOTViewTransform

saud5150:
Part of #1942

This PR:
- Adds Google-style docstrings to `DINOViewTransform` and `IBOTViewTransform`,
  following the format agreed in #1946: an `Args:` block in `__init__` (name and
  description on the same line) rather than an `Attributes:` block on the class
- Updates `docs/source/lightly.transforms.rst` to render the new `__init__`
  docstrings: adds `__init__` to `:special-members:` for `dino_transform`, and
  adds a new autodoc entry for `ibot_transform`, which was previously missing
  from the docs entirely

This PR does not:
- Change any runtime behavior (docs only)
- Touch the parent `DINOTransform` / `IBOTTransform` docstrings, which still use
  the older `Attributes:` style — those are out of scope for #1942

Two side effects of the rst changes, worth flagging:
- The `ibot_transform` module (both `IBOTTransform` and `IBOTViewTransform`) now
  appears in the rendered docs for the first time
- `:special-members: __init__` applies to all classes in each module, so the
  parent transforms' `__init__` signatures now render too; their parameters are
  still documented in their class-level `Attributes:` blocks

Tested with:
```
make format-check
make lint
make type-check
python -m pytest tests/transforms
```
All pass. Docstring augmentation lists were checked against each transform's
actual augmentation pipeline.

### DIFF
--- a/docs/source/lightly.transforms.rst
+++ b/docs/source/lightly.transforms.rst
@@ -17,7 +17,7 @@ lightly.transforms
 
 .. automodule:: lightly.transforms.dino_transform
    :members:
-   :special-members: __call__
+   :special-members: __call__, __init__
 
 .. automodule:: lightly.transforms.fast_siam_transform
    :members:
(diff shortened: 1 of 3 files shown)


docs/source/lightly.transforms.rst
```diff
@@ -31,6 +31,10 @@ lightly.transforms
    :members:
    :special-members: __call__
 
+.. automodule:: lightly.transforms.ibot_transform
```
liopeer, line 34: ❤️

liopeer (approved):
LGTM!

PR #1951, opened 2026-06-22
docs: add docstrings to MSN, SimSiam, VICReg, and VICRegL view transforms

saud5150:
Part of #1942

This PR:
- Adds Google-style docstrings to `MSNViewTransform`, `SimSiamViewTransform`,
  `VICRegViewTransform`, and `VICRegLViewTransform`, following the format agreed
  in #1946: an `Args:` block in `__init__` (name and description on the same line)
  rather than an `Attributes:` block on the class
- Adds `__init__` to `:special-members:` for these four modules in
  `docs/source/lightly.transforms.rst` so the new docstrings render in HTML

This PR does not:
- Change any runtime behavior (docs only)
- Touch the parent `MSNTransform` / `SimSiamTransform` / `VICRegTransform` /
  `VICRegLTransform` docstrings, which still use the older `Attributes:` style —
  those are out of scope for #1942

Notes:
- Each view docstring mirrors its parent transform's reference: `MSNViewTransform`
  and `VICRegLViewTransform` cite the method paper; `SimSiamViewTransform` and
  `VICRegViewTransform` do not, since their parents don't either
- `VICRegLViewTransform` only applies the photometric augmentations (color jitter,
  grayscale, blur, solarization, normalize) — the crop and flip are applied
  separately by `VICRegLTransform`, which the docstring calls out explicitly
- `:special-members: __init__` applies to all classes in each module, so the
  parent transforms' `__init__` signatures now render too; their parameters are
  still documented in their class-level `Attributes:` blocks

Tested with:
```
make format-check
make lint
make type-check
python -m pytest tests/transforms
```
All pass. Docstring augmentation lists were checked against each transform's
actual augmentation pipeline.

### DIFF
--- a/docs/source/lightly.transforms.rst
+++ b/docs/source/lightly.transforms.rst
@@ -49,7 +49,7 @@ lightly.transforms
 
 .. automodule:: lightly.transforms.msn_transform
    :members:
-   :special-members: __call__
+   :special-members: __call__, __init__
 
 .. automodule:: lightly.transforms.multi_crop_transform
    :members:
@@ -81,7 +81,7 @@ lightly.transforms
 
 .. automodule:: lightly.transforms.simsiam_transform
    :members:
-   :special-members: __call__
+   :special-members: __call__, __init__
 
 .. automodule:: lightly.transforms.smog_transform
    :members:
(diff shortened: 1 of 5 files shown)


liopeer (approved):
Thanks a lot! ❤️

PR #1953, opened 2026-06-28
Add PatchKernelAlignmentLoss (PaKa paper)

IgorSusmelj:
## Changes

 - This PR adds a new loss for dense representations based on [Patch-Level Kernel Alignment for Dense Self-Supervised Learning](https://arxiv.org/abs/2509.05606)

### DIFF
--- a/lightly/loss/dense_relational_utils.py
+++ b/lightly/loss/dense_relational_utils.py
@@ -1,0 +1,231 @@
+"""Helpers for the dense relational PatchKernelAlignmentLoss.
+
+These utilities back :class:`~lightly.loss.PatchKernelAlignmentLoss`. The loss
+compares the pairwise similarity (kernel) structure of student and teacher dense
+features of shape ``(B, N, D)``. The helpers here cover:
+
+- validating dense-feature inputs,
+- optionally subsampling tokens to bound the ``O(N^2)`` kernel cost,
+- mask-aware double-centering of a kernel matrix.
+
+All math is local to each image, so the helpers are AMP-, CUDA- and DDP-safe
+without any cross-rank communication.
+"""
+
+from __future__ import annotations
+
+import torch
+from torch import Tensor
+
+
+def _validate_dense_inputs(
+    student_features: Tensor,
+    teacher_features: Tensor,
+    mask: Tensor | None,
+) -> None:
+    """Validates the shapes and dtypes of dense relational loss inputs.
+
+    Raises:
+        ValueError: If any input has an invalid shape or dtype.
+    """
+    if student_features.dim() != 3:
+        raise ValueError(
+            f"student_features must have shape (B, N, D) but has shape "
+            f"{tuple(student_features.shape)}."
+        )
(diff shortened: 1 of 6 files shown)


lightly/loss/dense_relational_utils.py
```diff
@@ -0,0 +1,231 @@
...
+        out_h: Output grid height.
+        out_w: Output grid width.
+
+    Returns:
+        Resampled features ``(B, out_h * out_w, C)`` in row-major token order.
+
+    Raises:
+        ValueError: If shapes are invalid.
+    """
+    from torchvision.ops import roi_align
```
liopeer: Would move to module level.

lightly/loss/dense_relational_utils.py
```diff
@@ -0,0 +1,231 @@
...
+    """
+    student = student_features
+    teacher = teacher_features.detach()
+
+    batch_size, n_tokens, _ = student.shape
+    if mask is None:
+        valid = student.new_ones(batch_size, n_tokens)
+    else:
+        # mask=True marks tokens to ignore, so valid tokens are ~mask.
+        valid = (~mask).to(student.dtype)
```
liopeer: Consider also moving to the device of student (since `.new_ones` will also do that).

lightly/loss/patch_kernel_alignment_loss.py
```diff
@@ -0,0 +1,126 @@
...
+)
+
+# Clamp before sqrt in the CKA denominator so the gradient stays finite when an
+# image's kernel is (near-)constant (HSIC -> 0); for non-degenerate images this
+# is a no-op.
+_EPS = 1e-6
+
+
+class PatchKernelAlignmentLoss(Module):
+    """Patch kernel alignment loss for dense self-supervised learning.
```
liopeer, line 21: Is there some paper we can cite?

lightly/loss/dense_relational_utils.py
```diff
@@ -0,0 +1,231 @@
...
+        )
+        teacher = torch.gather(
+            teacher, 1, indices.unsqueeze(-1).expand(-1, -1, teacher.shape[-1])
+        )
+        valid = torch.gather(valid, 1, indices)
+
+    return student, teacher, valid
+
+
+def roi_resample_to_grid(
```
liopeer, line 115: Am I understanding that this is often necessary to be called before the actual loss? If yes, should we not properly document how to use it? Especially with a cross-view example.

lightly/loss/patch_kernel_alignment_loss.py
IgorSusmelj: Added

lightly/loss/dense_relational_utils.py
IgorSusmelj: Yes, we need to align the feature space and therefore resample. I updated the docstring with more infos and added an example to it as well.

liopeer (approved):
LGTM!

PR #1958, opened 2026-07-02
feat: add BEiT support

aymen-000:
## Description

This PR adds support for **BEiT (BERT Pre-Training of Image Transformers)** for masked image modeling to Lightly.

### What was added

- Implemented the BEiT masked image modeling architecture.
- Added the BEiT image tokenizer.
- Added the BEiT masking transform.
- Added the masked image modeling prediction head.
- Added training examples (single-GPU and distributed).
- Added notebooks demonstrating BEiT training.
- Added documentation and API references.
- Added unit tests covering the new components.

### Relation to the existing MAE implementation

While both **MAE** and **BEiT** are masked image modeling methods, they differ in their reconstruction targets:

- **MAE** reconstructs the raw pixel values of masked image patches.
- **BEiT** predicts discrete visual tokens generated by a pretrained image tokenizer instead of reconstructing pixels.

### References

- **BEiT: BERT Pre-Training of Image Transformers**
  https://arxiv.org/abs/2106.08254

- **Original implementation**
 https://github.com/microsoft/unilm/tree/master/beit

## Documentation

- [x] I have updated the documentation.

## Tests


liopeer:
Thanks a lot @aymen-000, what a cool contribution! This will probably take a bit of time to review, but we will try to do it as quickly as possible.

examples/pytorch/beit.py
```diff
@@ -0,0 +1,174 @@
...
+#     optimizer = torch.optim.Adam(tokenizer.parameters(), lr=1e-3)
+#
+#     for images in dataloader:
+#         logits, recon = tokenizer(images)
+#         loss = F.mse_loss(recon, images)  # Reconstruction loss
+#         loss.backward()
+#         optimizer.step()
+#         optimizer.zero_grad()
+#
+#     torch.save(tokenizer.state_dict(), "tokenizer.pth")
```
liopeer: I doubt that it's quite that easy (but unfortunately DALL-E does not have a code release and neither has Microsoft for the tokenize pretraining). Since I think that this is unlikely to yield any good results, I would rather remove it.

lightly/models/beit.py
```diff
@@ -0,0 +1,187 @@
+"""BEiT: BERT Pre-Training of Image Transformers.
```
liopeer: We don't actually want these `model` modules anymore. Instead we want the full implementations to be inside the notebooks/examples. Sorry if this is not clear from the contributing instructions, will try to make this better in the future.

docs/source/examples/beit.rst
```diff
@@ -0,0 +1,74 @@
+.. _beit:
+
+BEiT
+====
+
+implementation of the BEiT (BERT Pre-Training of Image Transformers)
```
liopeer:
```suggestion
Implementation of the BEiT (BERT Pre-Training of Image Transformers) 
```

lightly/models/modules/__init__.py
```diff
@@ -8,6 +8,7 @@
 # Copyright (c) 2021. Lightly AG and its affiliates.
 # All Rights Reserved
 
+from lightly.models.modules.beit_tokenizer import ImageTokenizer
```
liopeer: I would call it BEiTImageTokenizer, so that it's clear to which pipeline it belongs. And I mean not only the import but the actual class.

lightly/models/modules/__init__.py
```diff
@@ -16,6 +17,7 @@
     DINOProjectionHead,
     DINOv2ProjectionHead,
     LeJEPAProjectionHead,
+    MIMHead,
```
liopeer: Same here: BEiTMIMHead.

examples/pytorch/beit.py
liopeer: Note: I found that BEiT seems to use a smooth L1 loss https://github.com/microsoft/unilm/blob/master/beit/modeling_discrete_vae.py#L124

But we don't have a training recipe, so even with that piece of information it's unlikely to get anything good.

examples/pytorch/beit.py
```diff
@@ -0,0 +1,174 @@
...
+dataloader = torch.utils.data.DataLoader(
+    dataset=dataset,
+    batch_size=16,
+    shuffle=True,
+    drop_last=True,
+    num_workers=2,
+)
+
+criterion = MaskedImageModelingLoss()
+optimizer = torch.optim.AdamW(
```
liopeer, line 128: Since you mention in the comment at the top that the parameters should not require grad, I would apply that here as well. I am aware that the tokenizer uses the `@no_grad` decorator, but better to completely deactivate the gradients. You can also do it above in the BEIT class.

examples/pytorch/beit.py
```diff
@@ -0,0 +1,174 @@
...
+import torchvision
+from torch import nn
+from torch.utils.data import Subset
+
+from lightly.loss import MaskedImageModelingLoss
+from lightly.models.modules import BEITEncoder, ImageTokenizer, MIMHead
+from lightly.transforms import BEITTransform
+
+
+class BEIT(nn.Module):
```
liopeer: Would rename to `BEiT` in order to match the paper's name better.

liopeer (commented):
This is only an initial review with already some remarks. Will continue over the next few days. 🙂

aymen-000:
Hi @liopeer thanks for the detailed review! I've addressed the comments:

* Removed the tokenizer pretraining recipe.
* Moved the BEiT implementation into the example.
* Renamed the tokenizer, MIM head, and model class for consistency.
* Froze the tokenizer parameters and set the tokenizer to evaluation mode.
* Updated the documentation, including the suggested capitalization fix.

Thanks again! I'll keep an eye out for any additional comments from the rest of the review.

Issue #1957, opened 2026-07-01
Add DetConS examples


saud5150:
Add training examples for DetConS based on [Efficient Visual Pretraining with Contrastive Detection](https://arxiv.org/abs/2103.10957).

The transform (`DetConSTransform`), loss (`DetConSLoss`), and a masked pooling utility (`pool_masked` in `lightly/models/utils.py`) are already implemented. The forward pass retains spatial backbone feature maps, pools them per mask region, and passes the resulting `(B, M, D)` embeddings to `DetConSLoss`. For the example, the built-in spatial grid heuristic (`DetConSTransform(grid_size=(5, 5))`) avoids any external segmentation dependency — the paper shows a 5×5 grid is optimal.

### Tasks

- [ ] Add `examples/pytorch/detcon.py`
- [ ] Add `examples/pytorch_lightning/detcon.py`
- [ ] Add `examples/pytorch_lightning_distributed/detcon.py`

Follow `examples/pytorch/densecl.py` as the structural template. I'd like to take this on.

liopeer:
Note: You probably will need some additional dependency like scikit-image or so in order to generate the pseudo-masks!

PR #1961, opened 2026-07-03
Add DetConS pytorch training example

saud5150:
Part of #1957

## Summary

- Adds `examples/pytorch/detcon.py` for DetConS (SimCLR-based contrastive detection).
- Uses `DetConSTransform(grid_size=(5, 5))` with the built-in `AddGridTransform` — no
  external segmentation dependency (no scikit-image needed).

## Testing

- `make format`
- `make generate-example-notebooks`
- `make all-checks` *(fails locally in `tests/utils/test_dist__gather__losses.py::TestGatherLayer_Losses::test_loss_dcl` and 3 others; all pre-exist on upstream master)*
- Trained the example for 5 epochs on CIFAR-10; loss went from 18.5 to 15.3.

### DIFF
--- a/examples/notebooks/pytorch/detcon.ipynb
+++ b/examples/notebooks/pytorch/detcon.ipynb
@@ -1,0 +1,224 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "0",
+   "metadata": {},
+   "source": [
+    "This example requires the following dependencies to be installed:\n",
+    "pip install lightly\n",
+    "Note: This example requires torchvision >= 0.17 for tv_tensors support."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip install lightly"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2",
+   "metadata": {},
+   "source": [
+    "Note: The model and training settings do not follow the reference settings\n",
+    "from the paper. The settings are chosen such that the example can easily be\n",
+    "run on a small dataset with a single GPU."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import torch\n",
+    "import torch.nn.functional as F\n",
+    "import torchvision\n",
+    "from torch import nn\n",
+    "from torchvision.transforms.v2 import ToImage\n",
+    "from torchvision.tv_tensors import Mask"
+   ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


gabrielfruet (approved):
LGTM, Thank you for your nice contribution!

liopeer:
Thanks a lot @saud5150! I saw that you use grid masks in your examples. Would it be possible to extend the example unsupervised masks like in the paper? We can also add an optional dependency like scikit-image to make this work. I would imagine something like the following:

```python
# in order to run unsupervised mask segmentation with this example make sure to also install
# scikit-image, i.e. run `pip install lightly scikit-image`

try:
    from skimage.segmentation import felzenszwalb
    SCIKIT_IMAGE_INSTALLED = True
except ImportError:
    print("Scikit-image is not installed, running the example with grid masks.")
    SCIKIT_IMAGE_INSTALLED = False

# ...
if SCIKIT_IMAGE_INSTALLED:
    _detcons_transform = DetConSTransform(input_size=96) # no `grid_size`
else:
    # existing transform

# in the training loop, calculate the mask
```
It would be better if the DetConS transform actually support passing a `callable` to generate the mask, so that it could be done in the dataloader workers, but I would leave that for another day.

saud5150:
> Thanks a lot @saud5150! I saw that you use grid masks in your examples. Would it be possible to extend the example unsupervised masks like in the paper? We can also add an optional dependency like scikit-image to make this work. I would imagine something like the following:
> 
> ```python
> # in order to run unsupervised mask segmentation with this example make sure to also install
> # scikit-image, i.e. run `pip install lightly scikit-image`
> 
> try:
>     from skimage.segmentation import felzenszwalb
>     SCIKIT_IMAGE_INSTALLED = True
> except ImportError:
>     print("Scikit-image is not installed, running the example with grid masks.")
>     SCIKIT_IMAGE_INSTALLED = False
> 
> # ...
> if SCIKIT_IMAGE_INSTALLED:
>     _detcons_transform = DetConSTransform(input_size=96) # no `grid_size`
> else:
>     # existing transform
> 
> # in the training loop, calculate the mask
> ```
> 
> It would be better if the DetConS transform actually support passing a `callable` to generate the mask, so that it could be done in the dataloader workers, but I would leave that for another day.

Implemented try/except with Felzenszwalb as primary path and the grid as fallback. Used scale=100, sigma=0.5, min_size=20 which gives ~10–15 segments on 32×32 CIFAR images; clipped output to num_cls=25 so model architecture stays the same either way. Happy to tune the parameters.

liopeer:
I think that sounds reasonable, thanks a lot! Will you post the implementation together with the PRs for the Lightning examples? @saud5150

saud5150:
Yes, I'll open a follow-up PR for the scikit-image path in the pytorch example (since #1961 is already merged), and the Lightning examples will include it from the start. Will have both up shortly.

Issue #1900, opened 2026-03-04
Implement: Stochastic positional embeddings improve masked image modeling


IgorSusmelj:
From this paper: https://arxiv.org/pdf/2308.00566

AFAIK we could use that in combination with many other building blocks. Could be interesting to see. E.g. with LeJEPA :)

JiwaniZakir:
The paper (SoPE, Bao et al. 2023) proposes replacing fixed sinusoidal or learned positional embeddings with stochastic ones sampled at training time — effectively adding noise to token positions — which forces the model to rely less on absolute position and learn more semantic patch representations during MIM pretraining. In lightly, the natural integration point would be the positional embedding modules inside `lightly/models/modules/` (e.g., wherever `VisionTransformer` or patch embedding layers inject position encodings), since the stochastic sampling only needs to apply during the forward pass at training time, not at inference. The LeJEPA connection makes sense given its I-JEPA heritage — the context/target block sampling already introduces spatial stochasticity, and SoPE could compound that by also randomizing the positional signal fed to the encoder. Worth noting that SoPE is reported to be largely orthogonal to masking strategy, so it should compose cleanly with existing MIM heads in the codebase without requiring architectural changes beyond the embedding layer.

kir-7:
Hi,

I am new to this community and am interested in this feature. Since this issue has been quiet for a few months, I wanted to check if there is still interest in implementing this? If no one else is working on it, can I work on it.

gabrielfruet:
Hi @kir-7,

We are still interested in it! Please feel free to open a PR.

PR #1963, opened 2026-07-05
feat: adding stochastic positional embeddings

kir-7:
Part of #1900 

## Summary
* Adds `add_stochastic_positional_noise` as a utility function to `lightly/models/utils.py`.
* Adds optional use of StoP to: 
      * `lightly/models/modules/ijepa.py` and 
      * `lightly/models/modules/ijepa_timm.py`
* Keeps StoP disabled by default to preserve existing behaviour.
* Adds a new test `tests/utils/test_stochastic_positional_embedding.py` to test the behaviour of the StoP function while disabled and enabled.

## Additional details:
* StoP can be used for masked image modelling (MIM) models and this PR uses it for IJEPA. It can be used for MAE as well, but currently MAEDecoder in `lightly/models/modules`  does not provide functionality to modify the masks directly, these are modified while creating the MAE class (`examples/pytorch/mae.py`). These examples can be modified to evaluate StoP.


kir-7:
@gabrielfruet  The failing check seems to be regarding test coverage for ijepa.py for from_vit_encoder, should I add tests for this? And is there anything else I need to modify?

gabrielfruet:
@kir-7 Thank you for your contribution! No need to worry, we are planning to remove it #1966 . Either me or Lionel are going to review your PR as soon as we can : )

lightly/models/modules/ijepa_timm.py
```diff
@@ -97,6 +99,9 @@ def __init__(
             ]
         )
 
+        self.use_stop = use_stop
```
liopeer: I saw that the original implementation uses `use_stop`, but imo it's not very descriptive, so I would recommend a more descriptive name and also make sure to update the docstring accordingly. We can also say that we don't use a boolean value at all, but instead only use the `noise_std`, but it's 0 by default.

lightly/models/utils.py
```diff
@@ -1316,3 +1316,46 @@ def apply_masks(x: Tensor, masks: Tensor | list[Tensor]) -> Tensor:
...
+            ``(batch_size, num_tokens, predictor_embed_dim)``.
+        projection:
+            Matrix A used to project gaussian noise to the pos_embedding
+            dimension.
+        noise_dim:
+            Dimension of the sampled gaussian noise before projection.
+        noise_std:
+            Standard deviation of the gaussian noise.
+        enabled:
+            If False, returns ``pos_embeddings`` unchanged.
```
liopeer: We should use the Google Docstring format and put arguments+description on the same line.

liopeer (changes requested):

lightly/models/modules/ijepa_timm.py
kir-7: Appreciate the review. I have removed `use_stop` and now use `noise_std=0.0` by default to disable StoP. I also updated the docstrings accordingly.

lightly/models/utils.py
kir-7: Thanks for the feedback. I updated the docstrings accordingly.

lightly/models/utils.py
```diff
@@ -1316,3 +1316,40 @@ def apply_masks(x: Tensor, masks: Tensor | list[Tensor]) -> Tensor:
...
+
+    noise = torch.normal(
+        mean=0.0,
+        std=noise_std,
+        size=(pos_embeddings.shape[0], pos_embeddings.shape[1], noise_dim),
+        device=pos_embeddings.device,
+        dtype=pos_embeddings.dtype,
+    )
+
+    return pos_embeddings + projection(noise)
```
liopeer: I actually think we want the projection without the bias term here, i.e. only the matrix multiplication.

liopeer (commented):
One last thing to address, otherwise looking good!

lightly/models/utils.py
kir-7: Thanks for the review. The function now directly accepts the projection weight matrix and uses `nn.functional.linear(noise, projection_weight, bias=None)` to project the noise without the bias term.

kir-7:
Hi @liopeer, a quick reminder for this PR. I have finished resolving the previous comments. When you have a chance, could you please take another look and let me know if any further changes are needed? Thanks!

liopeer (changes requested):
It looks mostly correct now, sorry that it took a long time to review. One thing that we need to address though: Adding noise should only happen if the module is in training mode!

liopeer (approved):
LGTM, thanks a lot!

kir-7:
I also noticed that `IJEPAPredictor.from_vit_encoder` does not properly initial `noise_std`, so model created through this will default `noise_std=0.0`. Should we add an optional `noise_std: float = 0.0` argument to `from_vit_encoder` and pass it to cls(...)?

liopeer:
I think it's fine to leave it as it is, no need to cover this imo.

kir-7:
Sorry for the delay. I merged the latest `master`, resolved the conflicts in `test_ijepa_timm.py`, and updated the tests to cover `noise_std`. CI is currently pending approval to run.

PR #1965, opened 2026-07-07
Add Pixio

### LINKED ISSUE #1894: PIXIO Implementation Plan

lorinczszabolcs:
This PR adds **Pixio** to lightly.

Pixio is a masked-autoencoder method: it builds on MAE with three changes:
- a deeper decoder
- coarser block (grid) masking
- multiple class tokens whose mean is the global representation. This is a reimplementation from the paper; the reference repository is under the FAIR Noncommercial license, so no code or weights are taken.

- Paper: [In Pursuit of Pixel Supervision for Visual Pre-training, CVPR 2026](https://arxiv.org/abs/2512.15715)
- Reference implementation: https://github.com/facebookresearch/pixio

Closes #1894.

## What's included

- `lightly/models/utils.py` — `random_grid_token_mask`, block/grid masking returning lightly's `(idx_keep, idx_mask)` convention, with input validation (positive patch count, perfect-square grid, block divisibility). Also generalizes the 2D sincos positional-embedding helpers to N prefix tokens (fixing a latent bug where >1 prefix token was mis-sized).
- `lightly/models/modules/masked_autoencoder_timm.py` — `PixioDecoderTIMM` (a thin `MAEDecoderTIMM` subclass, default decoder depth 32) plus a `num_prefix_tokens` parameter on `MAEDecoderTIMM` (default 1, backward-compatible).
- `benchmarks/imagenet/vitb16/pixio.py` — ViT-B/16 benchmark (256px, 4×4 grid mask, 8 class tokens via `reg_tokens=7`, 512×32 decoder, normalized-pixel loss), registered in `main.py`.
- PyTorch / PyTorch-Lightning / Lightning-Distributed examples (+ generated notebooks) and an example docs page, added to the models list. PDF attached.  [pixio.pdf](https://github.com/user-attachments/files/29821006/pixio.pdf)

- Tests in `tests/models/test_ModelUtils.py` and `tests/models/modules/test_masked_autoencoder_timm.py`.

## Testing

- New tests cover the mask util (keep/mask partition, whole-cell granularity, keep-count, `mask_ratio` 0.0/1.0 extremes, `grid_size=4`, error paths, device) and the decoders (CPU + CUDA forward, output shapes, finite values).
- `make format`, `mypy`, ruff, and the Pixio tests pass locally. `make all-checks` is green except three pre-existing, environment-only failures unrelated to this change (a distributed `test_loss_dcl` and two `test_version_checking` network-timeout tests).
- Ran a short training of the model on FashionMNIST with a shallower network (decoder depth=2) due to hardware constraints: the loss decreased monotonically. A full ImageNet benchmark isn't possible on my side, the PR just contributes the method, not a SOTA reproduction (the paper's numbers rely on web-scale data + distillation).

## Notes

- 256px / patch-16 so the 16×16 patch grid divides evenly into 4×4 mask blocks; docs note the 2×2-grid / 4-class-token configuration reported as dense-optimal in the paper's ablations.
- **Relationship to #1919:** that PR independently adds a block-wise mask for this issue. Its current implementation lifts `noise[:, 0] = -1` from `random_token_mask`,  where that line protects the CLS token at index 0, but in block space index 0 is a spatial block, so it deterministically keeps the top-left block in every sample (when `num_prefix_tokens > 0` and ≥1 block is kept), biasing the masking. I confirmed this by running that function: the top-left block is kept in **100%** of samples, vs the uniform **25%** ours produces on the same config; That PRs own tests don't catch it because they only assert the prefix token at index 0 is kept. The current `random_grid_token_mask` protects prefix tokens structurally (unconditional prepend), so image blocks are masked uniformly at random.
- Unrelated fix: seeded the pre-existing `test_normalize_mean_var` and loosened its variance `atol` to `1e-4`. It was latently flaky, unseeded `torch.rand` plus a tolerance too tight for `normalize_mean_var`'s eps regularization (output variance is `1 - eps/var`, not exactly 1), and this PR's new tests shifted global RNG ordering, which surfaced it.


gabrielfruet:
Thank you so much for the nice contribution! I'm going to review it ASAP.

lorinczszabolcs:
Regenerated the notebooks in the meantime, as that test has failed.

benchmarks/imagenet/vitb16/pixio.py
```diff
@@ -0,0 +1,166 @@
...
+from torch.optim import AdamW
+
+from lightly.models import utils
+from lightly.models.modules import MaskedVisionTransformerTIMM, PixioDecoderTIMM
+from lightly.transforms import MAETransform
+from lightly.utils.benchmarking import OnlineLinearClassifier
+from lightly.utils.scheduler import CosineWarmupScheduler
+
+
+class PIXIO(LightningModule):
```
gabrielfruet: Let's rename it to "Pixio" instead of "PIXIO"

docs/source/examples/pixio.rst
```diff
@@ -0,0 +1,90 @@
+.. _pixio:
+
+PIXIO
```
gabrielfruet:
```suggestion
Pixio
```

docs/source/examples/pixio.rst
```diff
@@ -0,0 +1,90 @@
...
+=====
+
+Example implementation of the PIXIO method. PIXIO builds on the `Masked Autoencoder
```
gabrielfruet: Yeah, let's rename in all places from PIXIO to Pixio

lightly/models/modules/masked_autoencoder_timm.py
```diff
@@ -176,6 +182,26 @@ def _initialize_weights(self) -> None:
...
+    decoder. This is an :class:`MAEDecoderTIMM` with a deeper default decoder;
+    ``decoder_depth`` defaults to 32 following PIXIO. See :class:`MAEDecoderTIMM` for
+    all parameters. Implemented from the paper; not derived from the reference code.
+
+    - [0]: In Pursuit of Pixel Supervision for Visual Pre-training, 2025,
+      https://arxiv.org/abs/2512.15715
+
+    """
+
+    def __init__(self, *, decoder_depth: int = 32, **kwargs: Any):
```
gabrielfruet: Prefer explicit args instead of kwargs, and document them on the docstring also

docs/source/examples/pixio.rst
```diff
@@ -0,0 +1,90 @@
...
+(MAE) <https://arxiv.org/abs/2111.06377>`_ and adapts it for dense prediction through
+three changes: a much deeper decoder (32 blocks) that takes over pixel-level detail
+modeling, a larger masking granularity that masks whole blocks of patches on a regular
+grid instead of individual patches, and multiple class tokens whose mean is used as the
+global image representation.
+
+Key Components
+--------------
+
```
gabrielfruet, line 15: (Optional) would be nice if you could attach a PDF of how it looks after rendering. 

Documentation for that is here: https://github.com/lightly-ai/lightly/tree/master/docs

examples/pytorch/pixio.py
```diff
@@ -0,0 +1,132 @@
...
+        self.num_prefix_tokens = vit.num_prefix_tokens
+
+        self.backbone = MaskedVisionTransformerTIMM(vit=vit)
+        self.sequence_length = self.backbone.sequence_length
+        self.decoder = PixioDecoderTIMM(
+            num_patches=vit.patch_embed.num_patches,
+            patch_size=self.patch_size,
+            embed_dim=vit.embed_dim,
+            decoder_embed_dim=decoder_dim,
+            decoder_depth=2,  # kept small so the example runs quickly; PIXIO uses 32
```
gabrielfruet: You can use 32 here IMHO

examples/pytorch/pixio.py
```diff
@@ -0,0 +1,132 @@
...
+model = PIXIO(vit)
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+model.to(device)
+
+transform = MAETransform(input_size=256)
+# we ignore object detection annotations by setting target_transform to return 0
+
+
+def target_transform(t):
```
gabrielfruet, line 94: We have lots of notebooks with this function, but it doesn't seem to be necessary anymore with the torchvision datasets.

examples/pytorch_lightning/pixio.py
```diff
@@ -0,0 +1,128 @@
...
+        self.patch_size = vit.patch_embed.patch_size[0]
+        self.num_prefix_tokens = vit.num_prefix_tokens
+        self.backbone = MaskedVisionTransformerTIMM(vit=vit)
+        self.sequence_length = self.backbone.sequence_length
+        self.decoder = PixioDecoderTIMM(
+            num_patches=vit.patch_embed.num_patches,
+            patch_size=self.patch_size,
+            embed_dim=vit.embed_dim,
+            decoder_embed_dim=decoder_dim,
+            decoder_depth=2,  # kept small so the example runs quickly; PIXIO uses 32
```
gabrielfruet: Same thing here, you can use 32.

examples/pytorch_lightning_distributed/pixio.py
```diff
@@ -0,0 +1,130 @@
...
+        self.patch_size = vit.patch_embed.patch_size[0]
+        self.num_prefix_tokens = vit.num_prefix_tokens
+        self.backbone = MaskedVisionTransformerTIMM(vit=vit)
+        self.sequence_length = self.backbone.sequence_length
+        self.decoder = PixioDecoderTIMM(
+            num_patches=vit.patch_embed.num_patches,
+            patch_size=self.patch_size,
+            embed_dim=vit.embed_dim,
+            decoder_embed_dim=decoder_dim,
+            decoder_depth=2,  # kept small so the example runs quickly; PIXIO uses 32
```
gabrielfruet: Same here, you can use 32

examples/pytorch_lightning_distributed/pixio.py
```diff
@@ -0,0 +1,130 @@
...
+        optim = torch.optim.AdamW(self.parameters(), lr=1.5e-4)
+        return optim
+
+
+model = PIXIO()
+
+transform = MAETransform(input_size=256)
+
+
+# we ignore object detection annotations by setting target_transform to return 0
```
gabrielfruet, line 99: You can remove this one here also

examples/pytorch_lightning/pixio.py
```diff
@@ -0,0 +1,128 @@
...
+        return optim
+
+
+model = PIXIO()
+
+transform = MAETransform(input_size=256)
+
+
+# we ignore object detection annotations by setting target_transform to return 0
+def target_transform(t):
```
gabrielfruet, line 100: You can remove this as explained

examples/pytorch_lightning_distributed/pixio.py
```diff
@@ -0,0 +1,130 @@
...
+from lightly.models.modules import MaskedVisionTransformerTIMM, PixioDecoderTIMM
+from lightly.transforms import MAETransform
+
+
+class PIXIO(pl.LightningModule):
+    def __init__(self):
+        super().__init__()
+
+        decoder_dim = 512
+        vit = vit_base_patch16_224(img_size=256, reg_tokens=7)
```
gabrielfruet: Prefer a smaller vit like vit_small

examples/pytorch_lightning/pixio.py
```diff
@@ -0,0 +1,128 @@
...
+from lightly.models.modules import MaskedVisionTransformerTIMM, PixioDecoderTIMM
+from lightly.transforms import MAETransform
+
+
+class PIXIO(pl.LightningModule):
+    def __init__(self):
+        super().__init__()
+
+        decoder_dim = 512
+        vit = vit_base_patch16_224(img_size=256, reg_tokens=7)
```
gabrielfruet: Prefer a smaller vit like vit_small

examples/pytorch/pixio.py
```diff
@@ -0,0 +1,132 @@
...
+        # get image patches for masked tokens
+        patches = utils.patchify(images, self.patch_size)
+        # must adjust idx_mask for the prefix tokens
+        target = utils.get_at_index(patches, idx_mask - self.num_prefix_tokens)
+        target = utils.normalize_mean_var(target)
+        return x_pred, target
+
+
+# vit-b/16 at 256px with 8 prefix tokens (1 class token + 7 register tokens).
+vit = vit_base_patch16_224(img_size=256, reg_tokens=7)
```
gabrielfruet: Prefer a smaller vit like vit_small

I personally prefer that this goes inside the class, as you did in the lightning ones.

`self.vit = ...`

tests/models/modules/test_masked_autoencoder_timm.py
```diff
@@ -63,3 +63,87 @@ def test_forward(self) -> None:
...
+
+    def test_forward__num_prefix_tokens(self) -> None:
+        self._test_forward__num_prefix_tokens(torch.device("cpu"))
+
+    @unittest.skipUnless(torch.cuda.is_available(), "Cuda not available.")
+    def test_forward__num_prefix_tokens_cuda(self) -> None:
+        self._test_forward__num_prefix_tokens(torch.device("cuda"))
+
+
+class TestPixioDecoderTIMM(unittest.TestCase):
```
gabrielfruet: I think you don't need to inherit from `unitest.TestCase`. 

The why we do it is more historically. I think we didn't used pytest back then.

tests/models/modules/test_masked_autoencoder_timm.py
```diff
@@ -63,3 +63,87 @@ def test_forward(self) -> None:
...
+    def _test_forward(self, device: torch.device) -> None:
+        torch.manual_seed(0)
+        num_patches, num_prefix_tokens, patch_size = 64, 8, 16
+        seq_length = num_patches + num_prefix_tokens
+        decoder = PixioDecoderTIMM(
+            num_patches=num_patches,
+            patch_size=patch_size,
+            embed_dim=128,
+            decoder_embed_dim=64,
+            decoder_depth=2,  # keep the test cheap
```
gabrielfruet, line 133: That's good :)

benchmarks/imagenet/vitb16/pixio.py
```diff
@@ -0,0 +1,166 @@
...
+        self.mask_ratio = 0.75
+        self.grid_size = 4
+        # vit-b/16 at 256px with 8 prefix tokens (1 cls + 7 reg). dynamic_img_size
+        # lets downstream evaluation run at other resolutions via pos-embed resampling.
+        vit = vit_base_patch16_224(img_size=256, reg_tokens=7, dynamic_img_size=True)
+        self.num_prefix_tokens = vit.num_prefix_tokens
+        self.patch_size = vit.patch_embed.patch_size[0]
+        self.sequence_length = vit.patch_embed.num_patches + vit.num_prefix_tokens
+        mask_token = Parameter(torch.zeros(1, 1, decoder_dim))
+        torch.nn.init.normal_(mask_token, std=0.02)
```
gabrielfruet: We already initialize it in the decoder, I think this line is not needed

lightly/models/utils.py
```diff
@@ -612,6 +612,101 @@ def random_token_mask(
...
+            (and width) must be divisible by grid_size.
+        num_prefix_tokens:
+            Number of prefix tokens (e.g. class or register tokens) that precede the
+            patch tokens. They are never masked and are always returned first in
+            idx_keep.
+        device:
+            Device on which to create the index masks.
+
+    Returns:
+        An (idx_keep, idx_mask) tuple. idx_keep contains the prefix token indices
```
gabrielfruet: I'm personally much in favor of showing the expected shapes of the tensors. It's really hard to me to read and understand the shapes and what data it contains.

I know that in the other functions we don't do this, but I think we can start it now.

gabrielfruet (changes requested):
LGTM! Very nice work, just small comments and we are good to go!

examples/pytorch/pixio.py
lorinczszabolcs: Good call to question it, turns out it's still needed here because these examples use VOCDetection, whose target is a variable-length annotation dict (one entry per object). Deleting it crashes the DataLoader on the very first batch:
```
RuntimeError: each element in list of batch should be of equal size
```
I verified on real VOC2007, object counts vary per image, so `default_collate` can't stack them, `target_transform=lambda t: 0` is what makes the batch collatable. It is unnecessary for simple-label datasets (FashionMNIST / CIFAR / ImageFolder).

examples/pytorch/pixio.py
lorinczszabolcs: Moved it as a local `vit` rather than `self.vit`,  nothing outside `__init__` uses it and the backbone already owns the encoder (`self.backbone.vit`), so `self.vit` would just duplicate it in the `state_dict`.

docs/source/examples/pixio.rst
lorinczszabolcs: PDF was attached (penultimate bullet of the `What's included` section). Re-exported it after the latest changes, so it reflects the PIXIO→Pixio rename and the updated examples.

examples/pytorch_lightning/pixio.py
lorinczszabolcs: Same explanation as earlier

examples/pytorch_lightning_distributed/pixio.py
lorinczszabolcs: Same explanation as earlier

gabrielfruet (approved):
Thank you for the really nice first contribution!! Welcome to the open-source world :) 

I'll try to accomodate some time to run benchmarks for Pixio

gabrielfruet:
I missed that we should also update the README.md. If you wish, you can open a follow-up PR for that. I would just say to wait for #1975 to get merged before opening the PR or branching from it, since it changes the order of the papers.

lorinczszabolcs:
> I missed that we should also update the README.md. If you wish, you can open a follow-up PR for that. I would just say to wait for #1975 to get merged before opening the PR or branching from it, since it changes the order of the papers.

@gabrielfruet Updated the README in #1986

PR #1969, opened 2026-07-08
examples: add scikit-image segmentation path to DetConS pytorch example

saud5150:
## Summary

Follow-up to #1961 — adds Felzenszwalb unsupervised segmentation as the primary mask path, as requested in the review. Falls back gracefully to the 5×5 grid when scikit-image is not installed. Both paths use `num_cls=25` for consistent model architecture.

## Test plan

- [x] Tested with scikit-image installed: Felzenszwalb produces 10–15 segments on 32×32 CIFAR images as expected
- [x] Tested fallback path: grid mask generation works when scikit-image unavailable
- [x] Format, lint, and type-check pass

Part of #1957.

### DIFF
--- a/examples/notebooks/pytorch/detcon.ipynb
+++ b/examples/notebooks/pytorch/detcon.ipynb
@@ -7,7 +7,9 @@
    "source": [
     "This example requires the following dependencies to be installed:\n",
     "pip install lightly\n",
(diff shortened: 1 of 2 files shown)


liopeer (approved):
Awesome stuff!

PR #1975, opened 2026-07-09
docs: sort model table by release date instead of alphabetically

### LINKED ISSUE #1974: Sort implementations by year instead of alphabetical.

Example:

```
 ```

Image

minagayid:
Closes #1974.

The Models table in `README.md` was ordered alphabetically. This sorts it **chronologically by each paper's arXiv v1 submission date** (oldest → newest) and appends a `<!-- DD/MM/YYYY -->` comment to every row, exactly as the issue suggested, so the ordering is explicit and easy to maintain when new methods are added.

### How the dates were determined
Every date is the **v1 `published` date from the arXiv API** (not a guess from the arXiv ID) — e.g. MoCo `13/11/2019` … LeJEPA `11/11/2025`. DINO and NNCLR were both submitted on 29/04/2021; they're tie-broken by arXiv ID.

### Year column
To keep the `Year` column monotonic with the new ordering, I aligned four entries to their **arXiv release year** (they previously showed the conference/publication year):

| Model | Was | Now |
|-------|-----|-----|
| DenseCL | 2021 | 2020 |
| SimSiam | 2021 | 2020 |
| SimMIM  | 2022 | 2021 |
| FroSSL  | 2024 | 2023 |

If you'd rather keep those as **publication** years, I'm happy to revert just those four — the sort itself uses the exact date comments, so it stays correct either way.

### Notes
- Diff is a pure reorder: 20 insertions / 20 deletions, all paper/docs/Colab links preserved verbatim.
- No sort direction was specified; I went oldest-first (reads as an evolution timeline). Easy to flip to newest-first if you prefer.


gabrielfruet (approved):

gabrielfruet:
Could you address the merge conflict? A new method was added to the docs.

PR #1976, opened 2026-07-09
docs: sort model table by release date

### LINKED ISSUE #1974: Sort implementations by year instead of alphabetical.

iamdhrv:
## Summary
- sort the Supported Models table by arXiv release date, newest first
- add hidden DD/MM/YYYY date comments to preserve the sort source for future edits
- keep the existing Year column values unchanged

## Validation
- verified 20 model rows have date comments
- verified date comments are sorted newest first

Fixes #1974


gabrielfruet:
Thank you for your PR! but it's a duplicate

PR #1978, opened 2026-07-11
Add MaskedVisionTransformerDecoder

### LINKED ISSUE #1615: Add MaskedVisionTransformerDecoder
We already have `MaskedVisionTransformer` classes for TIMM and torchvision that take images as input and output a token for every image patch. These classes are versatile as they can be used for many different methods (I-JEPA, DINOv2, MAE, etc.). Some of these methods also require a ViT decoder (I-JEPA and MAE) which we currently implement independently for every method (we have `IJEPAPredictor` and `MAEDecoder`). These decoders have very similar structures and I think we could cover them in a single `MaskedVisionTransformerDecoder` class.

The class would have the following interface:
```
class MaskedVisionTransformer(Module):
    def __init__(
        self,
        embed_dim: int,
        depth: int,
        # ... other args from VisionTransformer
    )
        self.mask_token = ...
        self.pos_embed = ...
        self.blocks = ...
        self.norm = ...

    def forward(
        self,
        x: Tensor,
        idx_keep: Tensor,
        idx_mask: Tensor,
        mask: Tensor,
    ) -> Tensor:
        x = self.preprocess(x, idx_keep, idx_mask, mask)
        x = self.blocks(x)
        x = self.norm(x)
        return x
    
    def preprocess(
        self,
        x: Tensor,
        idx_keep: Tensor,
        idx_mask: Tensor,
        mask: Tensor,
    ) -> Tensor:
        # Same as in MaskedVisionTransformer.preprocess. Add positional embedding
        # and apply the masking.
```

This would allow us to implement MAE like this:
```

class MAE(Module)
    def __init__(self):
        self.encoder = MaskedVisionTransformerTIMM(...)
        self.embed = Linear(...)
        self.decoder = MaskedVisionTransformerDecoderTIMM(...)
        self.prediction_head = Linear(...)

    def forward(self, images: Tensor):
        idx_keep, idx_mask = utils.random_token_mask(...)
        x = self.encoder(images, idx_keep)
        x = self.embed(x)

        x_decode = x.new_zeros(x.size())
        x_decode = utils.set_at_index(x_decode, idx_keep, x)
        x_decode = self.decoder(x_decode, idx_mask)

        x_pred = utils.get_at_index(x_decode, idx_mask)
        x_pred = self.prediction_head(x_pred)
        return x_pred
```

And I-JEPA like this:
```
class IJEPA(Module):
    def __init__(self):
        self.encoder = MaskedVisionTransformerTIMM(...)
        self.embed = Linear(...)
        self.decoder = MaskedVisionTransformerDecoderTIMM(...)
        self.prediction_head = Linear(...)
        self.target_encoder = copy.deepcopy(self.encoder)


    def forward_context(self, images: Tensor, mask_enc: Tensor, mask_pred: Tensor):
        x = self.encoder(images, idx_keep=mask_enc)
        x = self.embed(x)

        x_decode = x.new_zeros(...)
        x_decode = utils.set_at_index(x_decode, mask_enc, x)
        x_decode = x_decode.repeat_interleave(len(mask_pred), dim=0)
        x_decode = self.decoder(x_decode, idx_mask=mask_pred, idx_keep=mask_enc | mask_pred)
        x_decode = ... # select only mask_pred tokens
        x_pred = self.prediction_head(x_decode)
        return x_pred
    

    @torch.no_grad()
    def forward_target(self, images: Tensor, mask_enc: Tensor, mask_pred: Tensor):
        x = self.target_encoder(images)
        x = x.repeat_interleave(len(mask_pred), dim=0)
        x = utils.get_at_index(x, mask_pred)
        return x
```

By sharing the `MaskedVisionTransformerDecoder` class we can deduplicate a lot of the code around positional embeddings, transformer blocks, and masking. And by moving the `embed` and `prediction_head` layers out of the decoder, the decoder class becomes more modular and easier to reuse. We cannot re-use `MaskedVisionTransformer` directly because it expects images instead of tokens as input.

lorinczszabolcs:
This PR adds a reusable `MaskedVisionTransformerDecoder` abstraction, resolving #1615.

The ViT-decoder logic was implemented independently for every method (`MAEDecoderTIMM`, `PixioDecoderTIMM`, `IJEPAPredictorTIMM`), even though these decoders share the same structure: a mask token, a fixed 2D sin-cos positional embedding, a stack of transformer blocks, and a final norm. This PR consolidates that structure into a single `MaskedVisionTransformerDecoder` class, mirroring the existing `MaskedVisionTransformer` encoder abstraction. Unlike the encoder it takes patch tokens (not images) as input, and the input embedding and prediction head are kept in the model so the decoder stays modular and reusable.

Closes #1615.

## What's included

- `lightly/models/modules/masked_vision_transformer_decoder.py` — new abstract `MaskedVisionTransformerDecoder` (mirrors `MaskedVisionTransformer`, but for tokens): a concrete `preprocess` that adds the positional embedding and applies
  masking via the `idx_mask` / `idx_keep` / `mask` convention, plus abstract `forward` / `add_pos_embed` / `decode` / `sequence_length`.
- `lightly/models/modules/masked_vision_transformer_decoder_timm.py` — `MaskedVisionTransformerDecoderTIMM`, the TIMM backend (timm `Block`s, fixed 2D sin-cos positional embedding, mask token, final norm), following the interface
  proposed in the issue.
- `lightly/models/modules/ijepa_timm.py` — `IJEPAPredictorTIMM` now builds on `MaskedVisionTransformerDecoderTIMM`, reusing the shared positional embedding, transformer blocks, norm, and mask token, and keeping its own input embedding,
  prediction projection, and multi-mask `forward`. The public constructor and `forward(x, masks_x, masks)` signatures are unchanged and the outputs are numerically identical.
- `lightly/models/modules/__init__.py` — exports `MaskedVisionTransformerDecoderTIMM`.
- `benchmarks/imagenet/vitb16/{mae,pixio}.py` and `examples/{pytorch,pytorch_lightning,pytorch_lightning_distributed}/{mae,pixio}.py` (+ regenerated notebooks) — migrated to the new decoder using the `encoder → embed → decoder → prediction_head` pattern from the issue. The change is behaviour-preserving.
- `tests/models/modules/test_masked_vision_transformer_decoder_timm.py` — new tests for the decoder.

## Testing

- New tests cover: forward output shapes (with and without prefix tokens, CPU + CUDA), the masking semantics (`idx_mask`, `mask`, `idx_keep`), the frozen 2D sin-cos positional embedding, the `idx_mask` + `mask` error path, equivalence of the decoder core to `MAEDecoderTIMM.decode`, and a full-flow test proving the MAE migration is bit-identical to the previous flow.
- The existing MAE / Pixio / I-JEPA decoder tests remain green (the migration is non-breaking).
- `make format`, `mypy`, and `ruff` pass locally, `make generate-example-notebooks` produces no diff, and the test suite passes.

## Notes

- **Backward compatibility:** `MAEDecoder`, `MAEDecoderTIMM`, `PixioDecoderTIMM`, `IJEPAPredictor`, and `IJEPAPredictorTIMM` remain exported and usable. First-party code (benchmarks and examples) now uses the new abstraction, mirroring how `MaskedVisionTransformer` was introduced alongside `MAEEncoder` / `MAEDecoder`.
- **I-JEPA `state_dict`:** the refactor renames `IJEPAPredictorTIMM`'s internal submodules (`predictor_blocks` → `blocks`, `predictor_norm` → `norm`, `predictor_pos_embed` → `pos_embed`). The public API and numerics are unchanged, but existing I-JEPA checkpoints no longer load with `strict=True`. This is within `IJEPAPredictorTIMM`'s documented experimental status.
- **I-JEPA masking:** as anticipated in the issue, I-JEPA's multi-target masking does not generalize cleanly into the shared `idx_mask` / `idx_keep` `preprocess`, so `IJEPAPredictorTIMM` keeps its own `forward` and reuses the shared building blocks (positional embedding, transformer blocks, norm) via `decode()`.
- **Backends:** the issue and the shared building blocks are TIMM-based, so only a TIMM backend ships. The abstract base leaves the door open for a torchvision backend once a method needs one.


lorinczszabolcs:
@gabrielfruet this is for #1615 (open since 2024). Came up during the Pixio work: it added a third decoder with the same structure as MAEDecoderTIMM and IJEPAPredictorTIMM, so this consolidates them into one MaskedVisionTransformerDecoder. MAE/Pixio migrated, I-JEPA reuses it, non-breaking.

gabrielfruet:
Hey! I'm trying to allocate some time to review this. I'll need some more attention since we are adding new abstractions.

lightly/models/modules/masked_vision_transformer_decoder.py
```diff
@@ -0,0 +1,152 @@
...
+        x: Tensor,
+        idx_mask: Optional[Tensor] = None,
+        idx_keep: Optional[Tensor] = None,
+        mask: Optional[Tensor] = None,
+    ) -> Tensor:
+        """Decodes a batch of token sequences.
+
+        Args:
+            x:
+                Tensor with shape (batch_size, sequence_length, embed_dim) containing
```
gabrielfruet, line 47: Nit: state that `sequence_length` == `num_tokens_to_mask + num_tokens_to_keep`.

lightly/models/modules/masked_vision_transformer_decoder_timm.py
```diff
@@ -0,0 +1,143 @@
...
+from torch.nn import LayerNorm, Parameter, Sequential
+
+from lightly.models import utils
+from lightly.models.modules.masked_vision_transformer_decoder import (
+    MaskedVisionTransformerDecoder,
+)
+from lightly.models.modules.masked_vision_transformer_timm import init_weights
+
+
+class MaskedVisionTransformerDecoderTIMM(MaskedVisionTransformerDecoder):
```
gabrielfruet, line 17: Please add docstrings to the public methods here.

lightly/models/modules/ijepa_timm.py
```diff
@@ -63,41 +67,31 @@ def __init__(
...
-                    qkv_bias=qkv_bias,
-                    drop_path=drop_path_rate,
-                    proj_drop=proj_drop_rate,
-                    attn_drop=attn_drop_rate,
-                    # timm's type hints for norm_layer vary between versions.
-                    norm_layer=norm_layer,  # type: ignore[arg-type]
-                )
-                for _ in range(depth)
-            ]
+        utils.initialize_2d_sine_cosine_positional_embedding(
```
gabrielfruet, line 100: Since `forward` is overridden anyway, composition reads cleaner than inheritance here — inject the abstract `MaskedVisionTransformerDecoder` and drive it directly:

```python
with torch.no_grad():
    self.decoder.mask_token.zero_()

utils.initialize_2d_sine_cosine_positional_embedding(
    pos_embedding=self.decoder.pos_embed, num_prefix_tokens=0
)
```

No need for an `_initialize_weights` override then.

examples/pytorch_lightning_distributed/pixio.py
```diff
@@ -27,38 +30,42 @@ def __init__(self):
         self.num_prefix_tokens = vit.num_prefix_tokens
         self.backbone = MaskedVisionTransformerTIMM(vit=vit)
         self.sequence_length = self.backbone.sequence_length
-        self.decoder = PixioDecoderTIMM(
+        self.decoder_embed = nn.Linear(vit.embed_dim, decoder_dim)
+        # deep (32-block) decoder following Pixio
+        self.decoder = MaskedVisionTransformerDecoderTIMM(
```
gabrielfruet, line 35: Should we deprecate `PixioDecoderTIMM`/`MAEDecoderTIMM` next, now that this covers both?

lightly/models/modules/masked_vision_transformer_decoder_timm.py
```diff
@@ -0,0 +1,143 @@
...
+        x = self.preprocess(x, idx_mask=idx_mask, idx_keep=idx_keep, mask=mask)
+        x = self.decode(x)
+        return x
+
+    def add_pos_embed(self, x: Tensor) -> Tensor:
+        out: Tensor = x + self.pos_embed
+        return out
+
+    def decode(self, x: Tensor) -> Tensor:
+        out: Tensor = self.blocks(x)
```
gabrielfruet, line 141: typing like this is nice

examples/pytorch_lightning_distributed/pixio.py
```diff
@@ -27,38 +30,42 @@ def __init__(self):
...
         return self.backbone.encode(images=images, idx_keep=idx_keep)
 
     def forward_decoder(self, x_encoded, idx_keep, idx_mask):
-        # build decoder input
+        # embed encoded tokens into the decoder dimension
         batch_size = x_encoded.shape[0]
-        x_decode = self.decoder.embed(x_encoded)
-        x_masked = utils.repeat_token(
-            self.decoder.mask_token, (batch_size, self.sequence_length)
+        x_decode = self.decoder_embed(x_encoded)
```
gabrielfruet, line 54: Nit, not important here since this is an example — but in the library itself, worth casting these for typing where it is not too complex, e.g. `x_decode: Tensor = self.decoder_embed(x_encoded)`.

lightly/models/modules/ijepa_timm.py
```diff
@@ -63,41 +67,31 @@ def __init__(
         norm_layer: Callable[..., nn.Module] = partial(nn.LayerNorm, eps=1e-6),
     ):
         """Initializes the IJEPAPredictorTIMM with the specified dimensions."""
-        super().__init__()
-
```
gabrielfruet, line 67: Renaming `predictor_pos_embed`/`predictor_blocks`/`predictor_norm` to `pos_embed`/`blocks`/`norm` breaks strict-loading old I-JEPA checkpoints. Worth a warning plus a docs note, or a migration test like the MAE one in `test_masked_vision_transformer_decoder_timm.py`.

gabrielfruet (changes requested):
Overall work was pretty good! just some small alignments and then we are good to go!

lightly/models/modules/masked_vision_transformer_decoder.py
lorinczszabolcs: Added to the forward docstring.

lightly/models/modules/masked_vision_transformer_decoder_timm.py
lorinczszabolcs: Added docstrings to `sequence_length`, `forward`, `add_pos_embed`, and `decode`.

lightly/models/modules/ijepa_timm.py
lorinczszabolcs: `IJEPAPredictorTIMM` now holds a `MaskedVisionTransformerDecoderTIMM` as `self.decoder` and drives it directly, same as the MAE/Pixio models. Dropped the inheritance and the `# type: ignore[override]` on `forward`.

examples/pytorch_lightning_distributed/pixio.py
lorinczszabolcs: Agree it makes sense. They are still public API and used by their own tests plus the new equivalence tests, so I would rather deprecate them in a separate follow-up PR.

examples/pytorch_lightning_distributed/pixio.py
lorinczszabolcs: In the library these are already cast, `decode` and `add_pos_embed` return `out: Tensor`, and the only `Linear` calls reassign an already-typed `Tensor`, so nothing stays Any. Left the example as it is, since you flagged it's not important here.

lightly/models/modules/ijepa_timm.py
lorinczszabolcs: `IJEPAPredictorTIMM` now registers a load hook that remaps the old keys (`predictor_pos_embed`/`predictor_blocks`/`predictor_norm`/`mask_token `to their `decoder.*` names) and warns with a `DeprecationWarning`, so old checkpoints keep loading. Added a docstring note plus a migration test that loads a legacy-named checkpoint through the hook and checks it reproduces the pre-refactor predictor's output.

lightly/models/modules/ijepa_timm.py
lorinczszabolcs: I kept `initialize_weights=False` on purpose, otherwise the decoder's `_initialize_weights` also runs `apply(init_weights)` on the blocks, and `I-JEPA` relies on timm's default block init there.

gabrielfruet (approved):
Pretty nice work!! Thank you so much.

PR #1979, opened 2026-07-11
docs: sort implementations table by arXiv submission date

### LINKED ISSUE #1974: Sort implementations by year instead of alphabetical.

DresdenGman:
Sort the model implementations table in README by arXiv ID (year+month) to give a chronological view of the field's development, rather than alphabetical.

The arXiv ID prefix (YYMM) serves as a deterministic sort key that respects the actual publication timeline — papers submitted earlier appear first, and within the same month they retain their original relative order.

Closes #1974.


gabrielfruet:
Thank you for your PR! but it's a duplicate

PR #1980, opened 2026-07-13
fix(loss): use autograd-aware all_reduce in BarlowTwinsLoss

### LINKED ISSUE #1977: [BUG] BarlowTwinsLoss with `gather_distributed=True` produces gradient which is off by `1/world_size` under DDP
## Summary

`BarlowTwinsLoss` supports a `gather_distributed=True` mode that reduces the
cross-correlation matrix `c` across ranks before computing the loss. The forward
pass is correct — the loss value at every rank equals the global Barlow Twins
loss on the concatenated batch. But the backward pass is not. The gradient is off
by a factor of `1/world_size`. Under DDP gradient averaging, the parameter
gradient through the loss is exactly `1/world_size` of the true gradient that
single-GPU Barlow Twins on the global batch would produce.

The practical effect is silent: any multi-GPU run with `gather_distributed=True`
trains the backbone with a gradient scaled down by the number of GPUs (½ on 2
GPUs, ¼ on 4, …), with no error and a normal-looking loss curve.

I came across this while auditing the loss subpackage for the same
non-autograd-aware `all_reduce` pattern that was just fixed for SIGReg in #1920 /
#1923. `BarlowTwinsLoss` is the one remaining loss with it.

## Cause

https://github.com/lightly-ai/lightly/blob/f7af25e8ca01edf0931585bf9d0adb9d1420e185/lightly/loss/barlow_twins_loss.py#L76-L80

```python
if self.gather_distributed and dist.is_initialized():
    world_size = dist.get_world_size()
    if world_size > 1:
        c = c / world_size
        dist.all_reduce(c)
```

The low-level `torch.distributed.all_reduce` is not autograd-aware. It overwrites
`c`'s `.data` in place with the global sum but does not register the cross-rank
operation in the autograd graph. The forward value at each rank correctly becomes
the global average, but autograd at each rank still treats `c` as a function of
the local `c / world_size` only.

Each rank's autograd therefore computes the partial gradient from its local
samples with a spurious extra `1/world_size` factor. DDP then averages across
ranks, yielding `(1/world_size) * (1/world_size) * Σ_r partial_r` = `1/world_size`
of the true gradient.

## Reference comparison

This is the exact bug, and the exact fix, that #1920 identified for SIGReg and
#1923 landed. That fix swapped the raw collective for the autograd-aware
`torch.distributed.nn.all_reduce`, whose backward is itself an all_reduce and so
scales the upstream gradient back up by `world_size`:

https://github.com/lightly-ai/lightly/blob/f7af25e8ca01edf0931585bf9d0adb9d1420e185/lightly/loss/lejepa_loss.py#L197-L200

`BarlowTwinsLoss` still uses the un-fixed raw collective.

## Numerical reproduction

<details>
<summary>Single-process two-rank simulation (mocks torch.distributed.all_reduce by overwriting <i>.data</i> in place, exactly as the real collective does)</summary>

```python
from unittest import mock

import torch
from torch import distributed as dist

from lightly.loss.barlow_twins_loss import BarlowTwinsLoss

WORLD_SIZE = 2


def param_grad(loss_fn, x_a, x_b, weight):
    """Grad of loss_fn(x_a @ W, x_b @ W) w.r.t. a shared weight matrix W."""
    w = weight.clone().requires_grad_(True)
    loss_fn(x_a @ w, x_b @ w).backward()
    return w.grad.clone()


torch.manual_seed(0)
weight = torch.randn(128, 128)                          # stand-in "backbone" params
x_a_local, x_b_local = torch.randn(16, 128), torch.randn(16, 128)
x_a_global = torch.cat([x_a_local, x_a_local], dim=0)   # global batch == 2 ranks
x_b_global = torch.cat([x_b_local, x_b_local], dim=0)

# Truth: single-rank gradient on the full global batch.
truth = param_grad(BarlowTwinsLoss(gather_distributed=False), x_a_global, x_b_global, weight)

# Lightly's gather_distributed=True mode. all_reduce sums across ranks; with
# identical ranks that sum is 2x the local tensor, so we mock it as such.
with mock.patch.object(dist, "is_initialized", return_value=True), \
     mock.patch.object(dist, "get_world_size", return_value=WORLD_SIZE), \
     mock.patch.object(dist, "all_reduce", side_effect=lambda t, *a, **k: t.data.mul_(WORLD_SIZE)):
    buggy = param_grad(BarlowTwinsLoss(gather_distributed=True), x_a_local, x_b_local, weight)

print(f"truth={truth.norm():.6f}  lightly_ddp={buggy.norm():.6f}  ratio={buggy.norm()/truth.norm():.4f}")
```

Output:
```
truth=84.970490  lightly_ddp=42.485245  ratio=0.5000
```

For world_size=2 the ratio is `0.5000`, exactly `1/2`. Forward loss values at
every rank equal the single-rank truth (the bug is gradient-only).

</details>

<details>
<summary>Real DDP (2 processes, real <code>gloo</code> process group, no mocks) — same result</summary>

Two real processes, a real process group, the real cross-rank `all_reduce` inside
the unedited loss forward, gradients averaged across ranks exactly as DDP does:

```
ratio DDP/truth   = 0.5000   (correct = 1.0)
cosine similarity = 1.0000   (right direction, half the magnitude)
```

Applying the fix below and re-running the same script gives `ratio = 1.0000`.

</details>

## Proposed fix

Direct analog of #1923 — swap the raw collective for the autograd-aware
`torch.distributed.nn.all_reduce`:

```python
from torch.distributed import nn as dist_nn
...
    if world_size > 1:
        c = dist_nn.all_reduce(c / world_size)
```

Its backward is itself an all_reduce, which scales the upstream gradient by
`world_size` and cancels the `1/world_size` under-scaling. Verified: with this
change both reproductions above print `ratio = 1.0000`.

The existing `test__gather_distributed` only checks instantiation; there is no
gradient test today. Happy to send a PR with the fix plus distributed-vs-non-
distributed forward and gradient tests, modeled on the ones #1923 added in
`tests/loss/test_lejepa_loss.py`.

---

Not environment-specific and produces no traceback — the forward succeeds and the
loss value is correct; only the backward gradient is wrong. Present since
multi-GPU support was added in #666 (2021).

saud5150:
closes #1977

## Description
- [ ] My change is breaking

`BarlowTwinsLoss(gather_distributed=True)` reduced the cross-correlation matrix
with the raw `torch.distributed.all_reduce`, which isn't autograd-aware. The
forward value was correct, but the backward pass scaled the gradient down by
`1/world_size` under DDP — silently. Swapped it for
`torch.distributed.nn.all_reduce`, whose backward is itself an all_reduce, so the
gradient is correct and the forward value is unchanged. Same fix #1923 landed for
SIGReg; Barlow Twins was the last loss with the pattern.

## Tests

Added two tests (world_size=2, two identical ranks): forward matches the
non-distributed loss on the global batch, and the gradient matches single-GPU
Barlow Twins on the global batch. The gradient test fails on master and passes
with the fix. `make static-checks` and the loss suite pass.

## Documentation

Internal implementation only — no API, docstring, or `.rst` changes.

None. Public API, forward loss value, and single-GPU behavior unchanged; only the
multi-GPU gradient is corrected.


gabrielfruet (commented):
/review

tests/loss/test_barlow_twins_loss.py
```diff
@@ -92,3 +93,76 @@ def test__loss_is_affine_invariant(self) -> None:
 
         # Loss should be invariant to affine transformations.
         assert torch.allclose(loss(x, x), loss(x, 2 * x + 4))
+
+    def test__gather_distributed_forward_matches_non_distributed(
```
gabrielfruet: I think these monkeypatched tests generally don't add a lot of value. Since we need to replicate the logic and they may get outdated. I think for now it's better to not add them. I'll also ask a colleague his opinion about this.

lightly/loss/barlow_twins_loss.py
```diff
@@ -76,8 +77,10 @@ def forward(self, z_a: torch.Tensor, z_b: torch.Tensor) -> torch.Tensor:
         if self.gather_distributed and dist.is_initialized():
             world_size = dist.get_world_size()
             if world_size > 1:
-                c = c / world_size
-                dist.all_reduce(c)
+                # Use the autograd-aware torch.distributed.nn.all_reduce so the
+                # backward pass scales gradients by world_size, matching
+                # single-GPU behaviour. ref #1977
+                c = dist_nn.all_reduce(c / world_size)
```
gabrielfruet, line 83: I took a look, and you can also give your opinions. But, using ReduceOp.AVG without the world_size division seems to be the ideal :)

gabrielfruet (changes requested):

lightly/loss/barlow_twins_loss.py
liopeer: I actually prefer the manual division. `ReduceOp.AVG` might be a tiny bit faster with NCCL, but e.g. the `gloo` backend (CPU) does not support `ReduceOp.AVG` – unfortunately this is not really nicely documented anywhere.

tests/loss/test_barlow_twins_loss.py
liopeer: Fully agree, they are not very useful. I'll create an issue for expanding distributed test coverage. There are nice ways of doing it, even on CPU, and without having to start/teardown the distributed context for every single test (which would be incredibly slow).

saud5150:
Kept the manual division — agreed, gloo not supporting `ReduceOp.AVG` makes it the safer choice for the CPU path. Removed the two mocked tests; real distributed coverage will come through #1982 (happy to add a BarlowTwins case once that context lands).

gabrielfruet (approved):
LGTM!! Thank you for the nice work

PR #1971, opened 2026-07-08
examples: add DetConS training examples (pytorch_lightning, pytorch_lightning_distributed)

saud5150:
Part of #1957

## Description
- [ ] My change is breaking

Adds `examples/pytorch_lightning/detcon.py` and
`examples/pytorch_lightning_distributed/detcon.py` following their respective
`densecl.py` templates. Uses `DetConSTransform` with Felzenszwalb segmentation
(scikit-image) as the primary mask path and the 5×5 grid as fallback — same
mask generation pattern as #1969. `gather_distributed=True` is set for the
distributed variant to sync embeddings across GPUs.

## Tests
- [x] My change is covered by existing tests.
- [ ] My change needs new tests.
- [ ] I have added/adapted the tests accordingly.
- [x] I have manually tested the change.

```
make format-check  # All checks passed
make lint          # All checks passed
make generate-example-notebooks
```

## Documentation
- [ ] I have added docstrings to all public functions/methods.
- [ ] My change requires a change to the documentation (`.rst` files).
- [ ] I have updated the documentation accordingly.
- [x] The autodocs update the documentation accordingly.

PR #1981, opened 2026-07-14
feat(transforms): add mask_fn to DetConSTransform for callable segmentation

### LINKED ISSUE #1972: DetCon Transform: Allow passing `Callable` for unsupervised segmentation
Currently, the use of unsupervised masks with the DetConS (or DetConB) transforms requires constructing a helper function as the actual transform, see #1969 and #1971 . It would be much more elegant if the transforms could accept an optional `Callable[[PILImage | Tensor], PILImage | Tensor]` that constructs the mask.

The priority would be compatibility with unsupervised segmentation algorithms from scikit-image or cv2, such as Felzenszwalb, SLIC, Watershed etc.

saud5150:
Closes #1972.

## Description
- [ ] My change is breaking

Using unsupervised masks with DetConS previously required a helper function that ran the segmenter, wrapped the output into a `Mask`, and threaded it through the transform by hand (see #1969, #1971). This adds an optional `mask_fn` that maps an image to segmentation labels, so scikit-image / cv2 segmenters work via `functools.partial`:

```python
DetConSTransform(mask_fn=partial(felzenszwalb, scale=100, sigma=0.5, min_size=20))
```

The mask is generated once per image inside the transform — so it runs in the dataloader workers and both views share the same segmentation — and is wrapped into a `torchvision.tv_tensors.Mask` internally, removing all user-side `Mask(...)` boilerplate. Typed `Callable[[Image | Tensor], Image | Tensor]` per the torchvision v2 conventions. Purely additive: the existing `grid_size` and pre-segmented-`Mask` paths are untouched; `grid_size` and `mask_fn` are mutually exclusive (both → `ValueError`). Scope is `DetConSTransform`, the shared DetCon augmentation transform — there is no separate `DetConBTransform`.

On the "override the forward method" note: `DetConSTransform` extends `MultiViewTransformV2`, which dispatches via `__call__` (there is no `forward` in the hierarchy), so I overrode `__call__`. I generate the mask there — once per image, before the view split — so both views share one segmentation and the segmenter is not run twice. Happy to move it into a v2 `Transform` if you'd prefer, but that path runs `mask_fn` per-view.

## Tests

Added `test_mask_fn` and `test_grid_size_and_mask_fn_are_mutually_exclusive` in `tests/transforms/test_detcon_transform.py` (patterns from `test_generate_grid_mask` / `test_given_masks`). Manually verified end-to-end with real `felzenszwalb` on a PIL image: `transform(pil_img)` returns two `(3, 96, 96)` image views and matching `(1, 96, 96)` int64 masks, with labels preserved as integers across randomized crop/rotate/resize.

```
make format        # All checks passed
make lint          # All checks passed
make type-check    # Success: no issues found in 541 source files
python -m pytest tests/transforms/test_detcon_transform.py   # 5 passed
make generate-example-notebooks
```

Before → after in `examples/pytorch/detcon.py`:

```python
# before: hand-rolled helper threads the mask through the transform
def transform(pil_img):
    tv_img = _to_image(pil_img)
    segments = felzenszwalb(np.array(pil_img), scale=100, sigma=0.5, min_size=20).astype(np.int64)
    segments = np.clip(segments, 0, num_cls - 1)
    mask = Mask(torch.from_numpy(segments).unsqueeze(0))
    return _detcons_transform(tv_img, mask)

# after: the transform builds the mask itself
transform = DetConSTransform(mask_fn=felzenszwalb_mask, input_size=96)
```

## Documentation

`mask_fn` is documented in the `DetConSTransform` docstring (three mask-sourcing modes, plus a note that it must be picklable for `DataLoader` workers). The `examples/pytorch/detcon.py` change flows into the rendered example page via its `literalinclude`; the notebook is regenerated.

None. The public API, the `grid_size` path, and the pre-segmented-mask path are all unchanged; `mask_fn` is additive.

### DIFF
--- a/lightly/transforms/detcon_transform.py
+++ b/lightly/transforms/detcon_transform.py
@@ -1,4 +1,8 @@
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+
+from PIL.Image import Image
+from torch import Tensor
+from torchvision.tv_tensors import Mask
 
 from lightly.transforms.add_grid_transform import AddGridTransform
 from lightly.transforms.multi_view_transform_v2 import MultiViewTransformV2
@@ -28,8 +32,16 @@ class DetConSTransform(MultiViewTransformV2):
         - RandomGrayscale
         - GaussianBlur (only for the first view)
 
-    Can additionally apply a segmentation of the image into a regular grid if not provided
-    with a pre-segmented image.
+    The segmentation mask can be provided in one of three ways:
+        - Passing ``grid_size`` segments the image into a regular grid.
+        - Passing ``mask_fn`` derives the mask from each image on the fly, for example
+          with an unsupervised segmentation algorithm such as
+          ``skimage.segmentation.felzenszwalb``. The mask is generated once per image
+          and shared by both views, so their region correspondence is preserved.
+        - Passing neither expects the caller to supply a pre-segmented mask alongside
+          the image.
+
+    ``grid_size`` and ``mask_fn`` are mutually exclusive.
 
     References:
         - [0] DetCon, 2021, https://arxiv.org/abs/2103.10957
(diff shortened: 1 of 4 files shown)


lightly/transforms/detcon_transform.py
```diff
@@ -38,7 +50,15 @@ class DetConSTransform(MultiViewTransformV2):
...
     Attributes:
         grid_size: Size of the grid segmentation as a tuple (num_rows, num_cols), or None
-            if the segmentation mask is to be provided by the user.
+            if the segmentation mask is provided by the user or by ``mask_fn``.
+        mask_fn: Optional callable that maps an image to a segmentation mask. It receives
+            the image passed to the transform and returns the integer segmentation labels
+            as a ``Tensor`` or PIL image, for example the output of an unsupervised
+            segmenter such as ``skimage.segmentation.felzenszwalb`` wrapped as a tensor.
+            The transform wraps the result into a ``torchvision.tv_tensors.Mask``
+            internally. It must be picklable to run in ``DataLoader`` worker processes,
```
liopeer, line 59: Thanks, very good point! ❤️

liopeer (approved):
LGTM!

Issue #1982, opened 2026-07-14
Proper distributed tests


liopeer:
We do some proper distributed tests in `tests/utils/test_dist__gather__losses.py`, however those are separate from the test modules of the individual losses so adding new losses always gets forgotten or we try to mock the distributed behaviour, which often does not work very well, e.g. in #1980 .

Instead, it would be nice to have a shared distributed context that we could submit distributed tests to. This should run entirely on CPU (gloo backend for PyTorch). A really nice example for this is [`torchmetrics`](https://github.com/Lightning-AI/torchmetrics/blob/master/tests/unittests/conftest.py).

Let me know if you'd like to work on this and I will give some more proper guidance and detail on how this can be done.

Following things to be done:
- [ ] add distributed pool to the test setup following torchmetrics (test with a single loss)
- [ ] add to CI
- [ ] abstract it nicely
- [ ] expand coverage to other losses/use cases
- [ ] get rid of `tests/utils/test_dist__gather__losses.py`

saud5150:
I'd like to take this on — I ran into exactly this in #1980.

I already have a small CPU test for that bug using 2 real gloo ranks, and it works (shows the gradient is half what it should be on the old code, correct after the fix). But mine spins up the processes fresh for each test, which is the slow thing you're trying to avoid. The torchmetrics conftest you linked handles that better — start the gloo pool once at the beginning of the session and reuse it — so that's the direction I'd go.

Before I start, I'd love your take on how you want tests to actually submit work to the pool, so it fits how you'd like the loss tests written.

liopeer:
I'd really follow the torchmetrics implementation:
1. Copy the session start/teardown hooks `pytest_sessionstart` and `pytest_sessionfinish`. Make sure that `USE_PYTEST_POOL` by default resolves to `False` for now (we can later see how well it works together with CI etc. and maybe change that). Overall, you can probably re-use most things from their `conftest.py` file (make sure to give credit!).
2. For submitting certain tests/test cases to the pool, I think you can just use `@pytest.mark.DDP` and `@pytest.mark.skipif(not USE_PYTEST_POOL, reason="DDP pool is not available")`.

I would start with with that basic setup, not activate anything for CI yet and just change a single test of some loss function to use that created pool. You can also already ping me if you only have a draft and then we can iterate on it together!

saud5150:
[...]

PR #1983, opened 2026-07-16
test: add shared gloo process pool for distributed loss tests

saud5150:
Part of #1982 (PR 1 — draft for early feedback)

> Stacked on #1980: the gradient regression test needs that fix (master still has
> the raw `dist.all_reduce`), so #1980's two commits show in this diff until it
> merges — I'll rebase onto `master` then. Review only the four files listed below.

## Description
- [ ] My change is breaking

Distributed loss behaviour is currently tested either in a separate module
(`tests/utils/test_dist__gather__losses.py`) disconnected from each loss's own
tests, or by mocking `torch.distributed` — which only exercises the mock, not the
real collective (exactly the gap that hid #1977 until the mock tests were reverted
in #1980).

This adds a torchmetrics-style session-scoped **gloo (CPU) process pool**, started
once per session and reused, so a test can submit real multi-rank work cheaply.
Following @liopeer's guidance in #1982:

- `tests/ddp_helpers.py` (new): `USE_PYTEST_POOL` flag (**default off**, folds in the
  gloo/non-Windows checks) and `setup_ddp`. Adapted from and credited to
  [torchmetrics' `conftest.py`](https://github.com/Lightning-AI/torchmetrics/blob/master/tests/unittests/conftest.py).
- `tests/conftest.py`: `pytest_sessionstart`/`pytest_sessionfinish` start/stop the
  pool; registers the `DDP` marker. **Nothing is enabled in CI.**
- `tests/loss/test_barlow_twins_loss.py`: first consumer — two real 2-rank tests
  (`@pytest.mark.DDP` + `skipif(not USE_PYTEST_POOL)`). The gradient test is a
  regression test for #1977 and fails on the raw `dist.all_reduce`.
- `pyproject.toml`: `--import-mode=importlib` in `addopts`. Required because the
  pool uses `spawn`, whose workers re-import the test module and deadlock under
  pytest's default `prepend` mode. It is a no-op for the rest of the suite.

Three things worth your review, since you asked to iterate: the submission API
(`pytest.pool.starmap(worker, ...)`, workers return results to the parent to
assert); `USE_PYTEST_POOL` folding the backend check into the flag rather than a
separate skip condition; and the Barlow tests using **identical data per rank**
(catches the `1/world_size` scaling bug but not heterogeneous-rank bugs).

## Tests
- [ ] My change is covered by existing tests

Two new tests in `tests/loss/test_barlow_twins_loss.py` run on the pool; the rest
of the suite is unaffected (flag off → they skip). Verified on macOS/Apple Silicon,
Python 3.9, torch 2.8:

```bash
make static-checks                                   # format-check + type-check pass
pytest tests/loss/test_barlow_twins_loss.py          # flag off: 4 passed, 2 skipped
USE_PYTEST_POOL=1 pytest tests/loss/test_barlow_twins_loss.py   # 6 passed (pool runs)
```

Before/after on the regression test (`USE_PYTEST_POOL=1`, gradient test only):

```
raw dist.all_reduce (master):   GRAD MATCH: False   # gradient scaled by 1/world_size
dist_nn.all_reduce  (fixed):    GRAD MATCH: True     # forward is identical either way
```

Full `make test-fast` under the new import mode: `4 failed, 1687 passed` — the 4 are
the known pre-existing macOS/Python-3.9 env failures (PyAV pickling, a py3.9
`assertWarns` bug, the existing gloo DCL mismatch), unchanged by this PR and not hit
by CI (Ubuntu 3.7/3.12).

## Documentation

Test-only change; no public API or `.rst` docs — no public functions added (the
pool module, hooks, and `setup_ddp` are internal; the worker helpers use inline
comments per the surrounding test file's style).

- Follow-ups within #1982 (deliberately out of scope here to keep PR 1 small):
  migrate the existing losses in `test_dist__gather__losses.py` onto the pool;
  add a heterogeneous-data-per-rank case; enable the pool in CI once it's proven
  green on Linux; revisit `pytest.pool` (global attr) vs a fixture.
- Not yet confirmed under pytest on Linux CI — the `spawn`/`importlib` mechanism is
  platform-independent, but a maintainer sanity-check there would be valuable before
  any CI-enablement PR. Happy to iterate.


tests/ddp_helpers.py
```diff
@@ -0,0 +1,35 @@
...
+    os.getenv("USE_PYTEST_POOL", "0") == "1"
+    and torch.distributed.is_available()
+    and sys.platform not in ("win32", "cygwin")
+)
+NUM_PROCESSES = 2
+# All ranks must rendezvous on the same port, so a single fixed port is used.
+MASTER_PORT = "8088"
+
+
+def setup_ddp(rank: int, world_size: int) -> None:
```
liopeer, line 31: I think it would be great to additionally have a teardown function that calls `dist.destroy_process_group()`, which we can use during `pytest_sessionfinish` with another starmap call. Makes it again a tiny bit safer.

tests/ddp_helpers.py
```diff
@@ -0,0 +1,35 @@
...
+once per session and reuses it for all distributed tests instead of spawning
+fresh processes per test:
+https://github.com/Lightning-AI/torchmetrics/blob/master/tests/unittests/conftest.py
+
+Disabled by default; set ``USE_PYTEST_POOL=1`` to enable. The gloo backend and
+non-Windows checks are folded into the flag so that a test's
+``skipif(not USE_PYTEST_POOL)`` also covers platforms without a usable backend.
+
+The pool assumes a single test session; running under pytest-xdist (``-n``)
+would start one pool per worker, all contending for the same port.
```
liopeer, line 13: That's fine, but thanks for the comment!

tests/loss/test_barlow_twins_loss.py
```diff
@@ -92,3 +107,40 @@ def test__loss_is_affine_invariant(self) -> None:
 
         # Loss should be invariant to affine transformations.
         assert torch.allclose(loss(x, x), loss(x, 2 * x + 4))
+
+    @pytest.mark.DDP
+    @pytest.mark.skipif(not USE_PYTEST_POOL, reason="DDP pool is not available")
+    def test__gather_distributed_forward_matches_non_distributed(self) -> None:
```
liopeer, line 113: We definitely will want to abstract this somehow, so that it's easier for others to contribute.

tests/loss/test_barlow_twins_loss.py
```diff
@@ -92,3 +107,40 @@ def test__loss_is_affine_invariant(self) -> None:
...
+        z_a = torch.randn(16, 64)
+        z_b = torch.randn(16, 64)
+
+        losses = pytest.pool.starmap(  # type: ignore[attr-defined]
+            _forward_worker,
+            [(rank, NUM_PROCESSES, z_a, z_b) for rank in range(NUM_PROCESSES)],
+        )
+        loss_truth = BarlowTwinsLoss(gather_distributed=False)(z_a, z_b)
+
+        assert torch.allclose(losses[0], loss_truth, atol=1e-5)
```
liopeer: Nit: I guess we could/should test both loss values.

liopeer (commented):
Overall looking pretty solid. I would address the small things I mentioned and then we merge.

After this is merged:
- [ ] add to CI
- [ ] abstractions for the test cases that need to run distributed
- [ ] expand to more tests

tests/loss/test_barlow_twins_loss.py
saud5150: Agreed, keeping this PR small; I'll take the abstraction as the #1982 follow-up once this lands.

tests/ddp_helpers.py
saud5150: Done, `teardown_ddp` runs from `sessionfinish` via starmap, guarded with `is_initialized()`. Setup is safe because `init_process_group` blocks and spreads the calls across workers; teardown doesn't, so the pool can run both on one worker where a second `destroy_process_group()` raises. The guard makes the repeat a no-op.

liopeer (approved):
LGTM!

liopeer:
Next thing would be to add it to CI. Sorry that it currently takes a bit longer to review things!

Issue #1987, opened 2026-07-22
Migrate unittest.TestCase to pytest-style test classes


gabrielfruet:
Several test files still use `unittest.TestCase` from before we adopted pytest. New tests are plain pytest-style classes (see `tests/loss/test_ibot_loss.py` for the pattern: no `TestCase` base, `assert` instead of `self.assertEqual`/`self.assertTrue`, `pytest.mark.parametrize` instead of manual test variants, `pytest.mark.skipif` instead of `unittest.skipUnless`). This migrates the rest.

**Excluded from this issue:** files being deleted by #1939 (API client removal) — `tests/api/test_BitMask.py`, `tests/api/test_rest_parser.py`, `tests/api/test_utils.py`, `tests/api_workflow/mocked_api_workflow_client.py`, `tests/api_workflow/test_api_workflow_client.py`. No point migrating tests that are about to disappear.

Split into one PR per group below — comment to claim a group, don't mix groups in one PR.

- [ ] **Core** — `tests/core/test_Core.py`
- [ ] **Data** — `tests/data/test_LightlyDataset.py`, `tests/data/test_VideoDataset.py`, `tests/data/test_data_collate.py`
- [ ] **Embedding** — `tests/embedding/test_embedding.py`
- [ ] **Models** — `tests/models/test_ModelsBYOL.py`, `tests/models/test_ModelsMoCo.py`, `tests/models/test_ModelsNNCLR.py`, `tests/models/test_ModelsSimCLR.py`, `tests/models/test_ModelsSimSiam.py`, `tests/models/test_ModelUtils.py`, `tests/models/test_ProjectionHeads.py`
- [ ] **Model modules** — `tests/models/modules/test_ijepa_timm.py`, `tests/models/modules/test_masked_autoencoder.py`, `tests/models/modules/test_masked_autoencoder_timm.py`, `tests/models/modules/test_memory_bank.py`
- [ ] **Transforms** — `tests/transforms/test_gaussian_blur.py`, `tests/transforms/test_jigsaw.py`, `tests/transforms/test_Solarize.py`
- [ ] **Utils** — `tests/utils/test_debug.py`, `tests/utils/test_dist.py`, `tests/utils/test_io.py`, `tests/utils/test_version_compare.py`

Mechanical conversion per file:
- `class Foo(unittest.TestCase):` → `class Foo:`
- `setUp`/`tearDown` → pytest fixture
[...]

PR #1992, opened 2026-07-23
Migrate data tests to pytest style

Jayashanker-Padishala:
Part of #1987 — **Data** group (claimed in the issue comments). Not using `closes` since other groups are still open.

## Description
- [ ] My change is breaking

Converts `tests/data/test_LightlyDataset.py`, `test_VideoDataset.py` and `test_data_collate.py` from `unittest.TestCase` to plain pytest-style classes per the pattern in `tests/loss/test_ibot_loss.py`:
- Dropped `TestCase` bases and `unittest` imports; `self.assert*` → plain `assert`
- `TestVideoDataset`'s class-level `skipUnless`/`skipIf` stack → equivalent `pytest.mark.skipif` stack; its `tearDown` → autouse yield fixture restoring the default video backend
- `test_multi_crop_collate`'s 2×6 `subTest` loop → stacked `pytest.mark.parametrize` (12 cases). Order-dependent `subTest` blocks (progressive chmod in `test_dataset_with_subdirs`/`test_dataset_plain`/`test_video_dataset_no_read_rights`) stay sequential with labels preserved as comments — parametrizing them would change behavior.
- Note: 9 occurrences of `assertEqual(len(a), len(b), len(c))` in `test_data_collate.py` — the third argument was unittest's `msg` parameter, not a comparand — became `assert len(a) == len(b)`.
- Added `-> None` annotations to test methods, which removes 45 pre-existing mypy errors under explicit invocation (52 → 7, all remaining ones pre-date this PR). These files are currently in the mypy `exclude` list; they could be removed from it once the small helpers get annotations, if you want that in a follow-up.

## Tests
- [ ] My change needs new tests
- [ ] I have added/adapted tests accordingly.

```bash
pytest tests/data/test_LightlyDataset.py tests/data/test_VideoDataset.py tests/data/test_data_collate.py -v
# before: 36 passed, 15 skipped (+17 subtests) — after: 47 passed, 15 skipped
# (47 = 36 − 1 + 12: the subTest loop became 12 parametrized tests; same skips; no coverage lost)
ruff check --select I <files> && ruff format --check <files>  # clean
```

## Documentation

Test-only change; no docs impact.


gabrielfruet:
Thank you for your contribution. @Jayashanker-Padishala 

Some tests are failling

Jayashanker-Padishala:
> Thank you for your contribution. @Jayashanker-Padishala
> 
> Some tests are failling

Thanks for flagging this, @gabrielfruet! I reproduced the failures locally and pushed a fix.

Root cause: the migration switched `TestLightlyDataset` from a `unittest.TestCase` to a plain class. `tests/data/test_LightlySubset.py` subclasses `TestLightlyDataset` and still relied on unittest's lifecycle — its `setUp()` set `self.input_dir` and its video tests used `@unittest.skipUnless`. pytest doesn't call `setUp()` or honor unittest skip markers on a plain-class subclass, so the subset tests failed with `AttributeError: 'TestLightlySubset' object has no attribute 'input_dir'` (on both 3.7 and 3.12), and the video subset test ran instead of skipping on the minimal/no-video job.

Fix: migrated `test_LightlySubset.py` to match its now-pytest parent — `setUp` → `setup_method` and `@unittest.skipUnless(...) → @pytest.mark.skipif(not ...)`. No assertions changed, coverage is identical.

Verified locally with the CI-pinned deps: `tests/data/` now passes fully (93 passed, 10 skipped, was 3 failed), and `ruff format`/import-sort and `mypy` are clean. Could you approve the workflow run when you get a chance? Thanks!

gabrielfruet (approved):
LGTM! Thank you for your contribution and for fixing the stuff!

Issue #1985, opened 2026-07-22
New SSL method: CAPI (Cluster and Predict Latent Patches)


lorinczszabolcs:
It would be great to add CAPI to lightly. It's a masked-image-modeling method from the DINOv2 authors that closes most of the gap to DINOv2 with a much simpler, single-loss recipe, and it slots neatly between the MAE and DINO/iBOT methods lightly already has.

- Paper: https://arxiv.org/abs/2502.08769 (Darcet, Baldassarre, Oquab, Mairal, Bojanowski, TMLR 2025)
- Code + weights: https://github.com/facebookresearch/capi

## Description

Instead of reconstructing pixels (MAE) or regressing latent features (I-JEPA), CAPI predicts the *cluster assignment* of each masked patch:

1. Mask most patches of the image.
2. An EMA teacher encodes the full image and produces per-patch representations.
3. An online clustering step (Sinkhorn-Knopp) turns the teacher's masked-patch representations into soft cluster assignments. These are the targets.
4. The student sees only the visible patches and predicts each masked patch's cluster assignment through a small predictor.
5. Loss is cross-entropy between predicted and target assignments. Pure MIM, no pixel targets, no contrastive pairs.

## Why it's a good fit
- Fills the gap between MAE/SimMIM (pixel-space MIM) and DINO/iBOT (clustering) methods with a stable, clustering-based MIM objective.
- Strong results for the simplicity: ViT-L reaches 83.8% ImageNet linear probe and 32.1% ADE20K linear mIoU, beating prior MIM methods and approaching DINOv2.
- Most of the building blocks already live in lightly, so this is largely an assembly of existing components rather than a from-scratch implementation.

## Implementation sketch
Reuse:
- Masked ViT encoder: `lightly/models/modules/masked_vision_transformer*.py`
- Predictor/decoder base: `MAEDecoderTIMM`, or the new `MaskedVisionTransformerDecoder` (#1978), depending on if it will get merged or not.
- Online clustering / centering: the Sinkhorn and centering logic already in the SwAV and DINO/iBOT losses

New:
- `CAPILoss` (cross-entropy on predicted vs teacher cluster assignments, with the online clustering + EMA target head)
- A `capi` module wiring encoder + predictor + EMA teacher + clustering head
- `CAPITransform` (masking / views)
- Example under `examples/` and an entry in the methods tracker (#1172)

## Contribution
[...]

PR #1993, opened 2026-07-23
feat: add CAPI loss

lorinczszabolcs:
Adds `CAPILoss` and a positionwise `sinkhorn_knopp` helper for CAPI (#1985).

The teacher's cluster logits become soft targets via a positionwise Sinkhorn-Knopp normalization, and the student learns to predict them with cross-entropy. The same loss trains the clustering prototypes when the teacher logits are passed as the student logits, matching the reference.

`forward` also takes an optional `teacher_index`: the Sinkhorn is computed over all teacher tokens and the predicted positions are gathered afterward, so the balancing spans the full grid rather than only the predicted subset.

CAPI is Apache-2.0, so this is a clean reimplementation with the paper and repo cited in the docstring.

First PR in the CAPI stack (loss → transform → modules → examples → benchmark).


gabrielfruet (approved):
LGTM!! Thank you for the contribution

PR #1994, opened 2026-07-23
feat: add CAPI transform

lorinczszabolcs:
Adds `CAPITransform` for CAPI (#1985): random resized crop (scale 0.6 to 1.0), horizontal flip, and normalization, matching the reference augmentation.

Follow up of #1993.

### DIFF
--- a/lightly/transforms/capi_transform.py
+++ b/lightly/transforms/capi_transform.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+from PIL.Image import Image
+from torch import Tensor
+from torchvision.transforms import InterpolationMode
+
+from lightly.transforms.torchvision_v2_compatibility import torchvision_transforms as T
+from lightly.transforms.utils import IMAGENET_NORMALIZE
+
+
+class CAPITransform:
+    """Implements the view augmentation for CAPI [0].
+
+    Input to this transform:
+        PIL Image or Tensor.
+
+    Output of this transform:
+        List of Tensor of length 1.
+
+    Applies the following augmentations by default:
+        - Random resized crop
+        - Random horizontal flip
+
+    - [0]: CAPI: Cluster and Predict Latent Patches for Improved Masked Image Modeling, 2025, https://arxiv.org/abs/2502.08769
+
+    Attributes:
+        input_size:
+            Size of the input image in pixels.
+        min_scale:
+            Minimum size of the randomized crop relative to the input_size.
+        normalize:
+            Dictionary with 'mean' and 'std' for torchvision.transforms.Normalize.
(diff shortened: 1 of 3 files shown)


lightly/transforms/capi_transform.py
```diff
@@ -0,0 +1,64 @@
...
+        normalize:
+            Dictionary with 'mean' and 'std' for torchvision.transforms.Normalize.
+
+    """
+
+    def __init__(
+        self,
+        input_size: int | tuple[int, int] = 224,
+        min_scale: float = 0.6,
+        normalize: dict[str, list[float]] = IMAGENET_NORMALIZE,
```
gabrielfruet: Should be `dict[str, list[float]] | None`, since you are validating 

```python
if normalize:
   ...
```

lightly/transforms/capi_transform.py
```diff
@@ -0,0 +1,64 @@
...
+    ):
+        transforms = [
+            T.RandomResizedCrop(
+                input_size, scale=(min_scale, 1.0), interpolation=3
+            ),  # 3 is bicubic
```
gabrielfruet: I think we should use the torchvision Enum for interpolation instead of a magic number

lightly/transforms/capi_transform.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __call__(self, image: Tensor | Image) -> list[Tensor]:
+        """Applies the transforms to the input image.
+
+        Args:
+            image:
+                The input image to apply the transforms to.
+
+        Returns:
+            The transformed image.
```
gabrielfruet: Should be something like "A list containing the transformed image"

tests/transforms/test_capi_transform.py
```diff
@@ -0,0 +1,18 @@
...
+    transform = CAPITransform(input_size=32)
+    output = transform(Image.new("RGB", (100, 100)))
+    assert len(output) == 1
+    assert output[0].shape == (3, 32, 32)
+
+
+def test_default_input_size() -> None:
+    transform = CAPITransform()
+    output = transform(Image.new("RGB", (256, 256)))
+    assert output[0].shape == (3, 224, 224)
```
gabrielfruet, line 20: Let's add a test for the normalize branch

gabrielfruet (changes requested):
Nice work!! I've made some comments with some small stuff to address.

lorinczszabolcs:
@gabrielfruet rebased and addressed the change requests. Let me know if there are further changes needed.

gabrielfruet (approved):
LGTM! Thank you for the contribution

Issue #1881, opened 2025-09-11
DinoV3 support?


201power:
Hi there, can I use lightly with DINOv3?
want to train yolo11 backbone with DINOv3 head and SSL

MosGeo:
> Hey [@201power](https://github.com/201power) !
> 
> We are not offering DINOv3 pretraining at this point, but the pretrained checkpoints from Meta are already super performant.
> 
> Just out of curiosity: Why would you like to integrate the DINOv3 ViT in a YOLO model? Don't you think you might get better performance/compatibility with a DETR variant instead?

I have not tested the dinov3 pretrained model but would love to be able to have the ability to pretrained it on custom specialized scientific datasets.

liopeer:
> I have not tested the dinov3 pretrained model but would love to be able to have the ability to pretrained it on custom specialized scientific datasets.

Got it! We definitely plan on supporting DINOv3 pretraining (or at least parts of the training stages) in the mid-term future. But I think we'll only get to work on it towards the end of the year.

Alighieri1231:
Hello, I am reading about dinov3 and besides the scale would only adding gram anchoring and rope give us the same blocks as Dinov2? Or am I missing something? Thanks

liopeer:
Hi @Alighieri1231 , you're pretty close yes. Adding it here in LightlySSL should be not crazy much work, it's just unfortunately low on our priority list. But if you'd like to contribute I am happy to provide guidance! 😄

saud5150:
Hi @liopeer, I'd like to take you up on the guidance offer and contribute this. @Alighieri1231 — you scoped it first, so happy to collaborate if you're interested.

**Current state / delta**

The DINOv2 stack is already in place (`DINOv2ProjectionHead`, `IBOTPatchLoss`, `KoLeoLoss`, the `MaskedVisionTransformer` classes, plus DINOv2 example and ViT-B/16 benchmark). Reading the DINOv3 paper (Siméoni et al., 2025, arXiv:2508.10104), the missing building blocks are:

[...]

PR #1995, opened 2026-07-23
feat: add CAPI clustering head and cross-attention predictor

lorinczszabolcs:
Adds the CAPI building blocks for #1985: `CAPIProjectionHead` in `heads.py` (L2-normalizes the input, then a linear layer to `num_clusters` prototypes; a `weight_norm` option gives the student head unit-norm prototypes, while the teacher clustering head leaves it off and uses a bias), `CAPIPredictorTIMM`, a cross-attention decoder with rotary position embeddings via `timm.layers.RotaryEmbeddingCat`, gated behind `timm_vit_available()` like the other `*_timm` modules, and `random_inverse_block_mask` in `models/utils.py` (inverse-block masking: a contiguous block is kept visible, the rest masked, with optional roll).

Head and predictor put in the library since it matches the existing pattern (`heads.py` already holds the method heads, and `MaskedVisionTransformerDecoder` is a library-level predictor), and the RoPE predictor is reusable rather than CAPI-only wiring. The masking util sits with the existing masking utilities, which all live in `models/utils.py`.

Follow up of #1994.

### DIFF
--- a/lightly/models/modules/__init__.py
+++ b/lightly/models/modules/__init__.py
@@ -12,6 +12,7 @@ from lightly.models.modules.heads import (
     BarlowTwinsProjectionHead,
     BYOLPredictionHead,
     BYOLProjectionHead,
+    CAPIProjectionHead,
     DenseCLProjectionHead,
     DINOProjectionHead,
     DINOv2ProjectionHead,
@@ -45,6 +46,7 @@ if _dependency.torchvision_vit_available():
 
 if _dependency.timm_vit_available():
     # Requires timm >= 0.9.9
+    from lightly.models.modules.capi_predictor_timm import CAPIPredictorTIMM
     from lightly.models.modules.heads_timm import AIMPredictionHead
     from lightly.models.modules.ijepa_timm import IJEPAPredictorTIMM
     from lightly.models.modules.masked_autoencoder_timm import (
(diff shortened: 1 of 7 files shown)


lorinczszabolcs:
@gabrielfruet Small deviation from what we agreed on in the issue: we agreed only loss + transform in the lib, and wiring goes in the examples.

This PR also puts the clustering head and predictor in the library, since `heads.py` already holds every method's head and `MAEDecoderTIMM` is a library-level predictor, so they felt like reusable blocks. I can move them into the example instead if you wish.

lorinczszabolcs:
On a related note, CAPI puts RoPE inside the encoder and runs it over the masked sequence, so a faithful version needs a masking-aware RoPE encoder, which would be a separate component (I think DINOv3 #1881, would need it as well). For now I've kept the encoder as a standard masked ViT with RoPE only in the predictor.

gabrielfruet:
Tip: When you're stacking PRs you should change the target merge branch to the one you stacked on top. It's easier for me to review :) 

* I took a look and that's not possible, unfortunately :(

lorinczszabolcs:
> Tip: When you're stacking PRs you should change the target merge branch to the one you stacked on top. It's easier for me to review :)
> 
> * I took a look and that's not possible, unfortunately :(

Yes, unfortunately that's the cross-fork limitation I flagged in #1985. Let me rebase so it's easier to review

gabrielfruet (commented):
/codex review

gabrielfruet (approved):
LGTM !! Thank you for the nice owrk

PR #1968, opened 2026-07-07
Add VISRegLoss (VISReg paper)

saud5150:
Related to #1960 (PR 1 of 2).

## Description
- [ ] My change is breaking

Adds `lightly/loss/visreg_loss.py` implementing [VISReg](https://arxiv.org/abs/2606.02572) (Wu et al., 2026) per Algorithm 1, structured like `lejepa_loss.py`. Covers the open points from #1960:

1. **Distributed**: per-device slices, no `broadcast`/`all_reduce` (Section 3.2, avoiding #1920). `gather_distributed=True` optionally gathers embeddings for the statistics via the same autograd-aware path as `VICRegLoss`.
2. **Component losses**: `visreg_center/scale/shape_loss` helpers plus `forward_components()` returning a `(total, pred, scale, shape, center)` NamedTuple (per #1422). `forward()` returns a scalar.
3. **No new head**: reuses `lejepa_invariance_loss`; intended for `LeJEPAProjectionHead`.
4. **Hyperparameters**: paper's ImageNet-1K defaults (`lambda_param=0.9`, `num_slices=4096`, equal weights).

Implemented from the paper's math only (official repo is CC BY-NC). Three choices worth review: `forward(*, local_proj, global_proj)` rather than the issue's `forward(z)` sketch (the prediction term needs the global views); prediction and regularization run over **all** views per Eq. 8 / Section 3.1 (differs from `LeJEPALoss`'s local-only, pinned by a test); and Algorithm 1's `(N,K)-(N,)` broadcasting bug is fixed with an `unsqueeze`.

## Tests
- [ ] My change is covered by existing tests.
- [x] My change needs new tests.
- [x] I have added/adapted the tests accordingly.
- [x] I have manually tested the change. if_yes_describe_how

37 tests in `tests/loss/test_visreg_loss.py` (patterns from `test_lejepa_loss.py`): shape loss pinned against a naive Algorithm 1 transcription, non-vanishing gradient under collapse, stop-gradient behavior, all-views regularization scope, no-collectives on the default distributed path, and gather matching the global-batch loss. `make format`, `make static-checks`, and `make test` pass (1693 passed; 4 failures are pre-existing macOS/Python-3.9 env issues, not hit by CI).

## Documentation
- [x] I have added docstrings to all public functions/methods.
- [x] My change requires a change to the documentation ( `.rst` files).
- [x] I have updated the documentation accordingly.
- [x] The autodocs update the documentation accordingly.

[...]

Issue #1960, opened 2026-07-02
Implement VISReg (Variance-Invariance-Sketching Regularization) loss + examples


gabrielfruet:
## Add VISReg loss + examples

Paper: https://arxiv.org/abs/2606.02572 (Wu, Balestriero, Levine, Jun 2026)
Official repo: https://github.com/HaiyuWu/visreg

Sits between `VICRegLoss` and `SIGReg`/`LeJEPALoss`: keeps the variance term from VICReg, replaces covariance with a sliced-Wasserstein sketching term against an isotropic Gaussian. O(NDK), and unlike SIGReg the gradient doesn't vanish under collapse (their Fig. 2).

Note: the official repo is CC BY-NC 4.0, non-commercial. Fine as a reference for the math (Algorithm 1 is unambiguous), but we shouldn't port code directly given our license.

Two PRs:

### PR 1: Loss

- `lightly/loss/visreg_loss.py`, structured like `LeJEPALoss`
   - Don't forget to reference the paper and authors.
- register in `lightly/loss/__init__.py`
- update `docs/source/lightly.loss.rst` (autodoc entry, same pattern as `VICRegLoss`/`LeJEPALoss`)

```python
class VISRegLoss(nn.Module):
    def __init__(
        self,
        lambda_param: float = 0.9,
        num_slices: int = 4096,
        lambda_scale: float = 1.0,
        lambda_shape: float = 1.0,
        lambda_center: float = 1.0,
        gather_distributed: bool = False,
        eps: float = 1e-4,
    ):
        ...

    def forward(self, z: Tensor) -> Tensor:
        ...
```

Open points:
1. Distributed: paper generates K slices per GPU, no batch gather (Sec 3.2, Fig. 6). Use that instead of the SIGReg `all_reduce` pattern from #1920.
2. Return component losses (scale/shape/center/pred) separately — same as #1422 for VICReg. Table 12 shows shape-loss weight is the main tuning lever for long-tailed/low-rank data.
3. Reuse `LeJEPAProjectionHead`, no new head needed (no changes to `lightly/models/modules/heads.py` expected).
4. Use the best hyperparameters combinations defined in the paper. 

Tests:
- Try to match same testing patterns for the other losses.

### PR 2: Examples (depends on PR 1)

- `examples/pytorch/visreg.py`
- `examples/pytorch_lightning/visreg.py`
- `examples/pytorch_lightning_distributed/visreg.py`
[...]

PR #1997, opened 2026-07-23
feat(benchmarks): add VISReg to the ImageNet vitb16 benchmark

saud5150:
Related to #1960 — adds the ViT benchmark requested in [this maintainer comment](https://github.com/lightly-ai/lightly/issues/1960#issuecomment-5056968762), building on the `VISRegLoss` from #1968.

## Description
- [ ] My change is breaking

Registers VISReg as a selectable method in the `imagenet/vitb16` runner: a new `benchmarks/imagenet/vitb16/visreg.py` (VISReg `LightningModule` + multi-crop `transform`) plus a two-line entry in `main.py`'s `METHODS` dict. The module mirrors `lejepa.py` so the two run head-to-head on the same backbone, optimizer, and eval, differing only where VISReg's paper says they should.

Per the maintainer note ("use the best setting in the paper as much as possible, like LeJEPA"), the settings are VISReg's own paper-best, each cross-checked against the paper **and** the official repo (not just the paper prose, which diverges from the code in two places):

| Setting | Value | Source |
|---|---|---|
| backbone | `vit_small_patch16_224` | ViT-S/16 — smaller than the paper's ViT-B/16 per the maintainer ("we train smaller models") |
| views | 4 global + 6 local | paper App. A.1 / repo `n_global: 4` (needs a custom `MultiViewTransform`; `DINOTransform` is fixed at 2 global) |
| augmentation | one uniform recipe on every crop (blur 0.5, solarize 0.2) | reference `_shared_aug` |
| lr | `9e-4` | paper Table 3 (9e-4 > 5e-4 at this scale); repo `lr: 9e-4` |
| projector | 3-layer MLP `embed→2048→2048→256`, BatchNorm + **GELU**, last layer bare | reference `ViTEncoder.proj` (`proj_gelu: true`); projects the **last-layer** CLS only |
| loss | `VISRegLoss(num_slices=2048, gather_distributed=True)` | K per paper App. A.1; `gather_distributed=True` matches the reference's autograd-aware `all_gather` before the reg statistics, and the sibling LeJEPA |
| λ / weight decay | `0.9` / `5e-2` | paper + repo |

Two spots where the reference code overrides the paper's own text, worth noting since they look like bugs otherwise: the projector runs on the **last-layer CLS token only** (`self.proj(last_cls)`), not the "concatenated last-two-layer CLS" the paper describes — that concat is the eval-probe embedding — so `input_dim=embed_dim`, not `2×`; and the augmentation is a single distribution shared by all crops, with no DINO-style global asymmetry.

Purely additive: no `lightly/` source is touched, `VISRegLoss` is only imported, and the new `METHODS["visreg"]` key affects a run only when `--methods visreg` is passed.

## Tests

Benchmark method modules aren't unit-tested in this repo (same as `dino.py`, `lejepa.py`, etc.) and sit outside the `pytest`/`mypy` targets, so I verified with a smoke test plus the standard checks. **Before:** `--methods visreg` is not a registered method. **After:** it constructs and trains a step end-to-end.

# smoke test: transform emits 4 global + 6 local, projector is the paper's GELU
# 384->2048->2048->256, and one training step runs end-to-end
python - <<'PY'
import sys; sys.path.insert(0, "benchmarks/imagenet/vitb16")
import torch, visreg
from PIL import Image
v = visreg.transform(Image.new("RGB", (256, 256)))
print(len(v), "views;", sum(x.shape[-1] == 224 for x in v), "global,", sum(x.shape[-1] == 96 for x in v), "local")
m = visreg.VISReg(batch_size_per_device=4, num_classes=10)
m.log = m.log_dict = lambda *a, **k: None
b = [torch.randn(4, 3, 224, 224)] * 4 + [torch.randn(4, 3, 96, 96)] * 6
print("loss", float(m.training_step((b, torch.randint(0, 10, (4,)), [""] * 4), 0).detach()))
PY
# -> 10 views; 4 global, 6 local
# -> loss ~3 (finite scalar; exact value varies with random init)

make format-check   # All checks passed!
make lint           # All checks passed!
make type-check     # Success: no issues found in 542 source files
make test-fast      # 1687 passed, 249 skipped; 4 failures are pre-existing macOS/Py3.9 env issues, not hit by CI and unrelated to this diff
ruff check benchmarks/imagenet/vitb16/visreg.py benchmarks/imagenet/vitb16/main.py   # All checks passed!
```

The full ImageNet run (ViT-S/16, `--methods visreg`) that produces the comparison numbers + checkpoint is left to the maintainers' cluster.

## Documentation
- [ ] I have added docstrings to all changed/added public functions/methods.

Benchmark method modules carry no docstrings by convention (matching the sibling files); the non-obvious paper deviations are noted in short module comments. The benchmark results table needs real ImageNet numbers from a cluster run, so the `docs/.../benchmarks.rst` row is a follow-up once those exist — no `.rst` change here.

- `benchmarks/imagenet/resnet50/visreg.py` (the maintainer's stated follow-up) and the docs results row are separate PRs.
- The lightly runner imposes a couple of schedule differences from the paper the maintainer already OK'd ("changing the schedule and batch size is ok"): ~1-epoch warmup (vs 5) and no grad clipping (vs the repo's `clip_grad_norm: 1.0`). The cosine floor already matches (`end_value=0.001` == repo `final_lr_div: 1000`).
- Only integration surface is the `METHODS` registry; existing methods are unaffected.


gabrielfruet (commented):
LGTM! Thank you

gabrielfruet (approved):

PR #1998, opened 2026-07-23
feat: add CAPI imagenet benchmark

### LINKED ISSUE #1985: New SSL method: CAPI (Cluster and Predict Latent Patches)

lorinczszabolcs:
Adds the ImageNet benchmark for CAPI (#1985): `benchmarks/imagenet/vitb16/capi.py` and `capi` registered in `main.py`'s `METHODS`. Uses a ViT-B/16 backbone with full-scale settings (K=16384, a 12-block predictor) so it produces real benchmark numbers in the vitb16 harness, plus an `OnlineLinearClassifier`, AdamW with cosine warmup, and the EMA momentum schedule.

Follows the reference recipe where it matters: 16 register tokens, inverse-block masking (`random_inverse_block_mask`), a weight-normed student head and a bias teacher head, prediction subsampling (0.05), stochastic depth (0.2) on the student backbone, the clustering head trained at half the backbone learning rate (its own param group), and the predictor mask token excluded from weight decay. LR, weight decay, and momentum follow the shared vitb16 harness rather than CAPI's exact values, for comparability with the other methods.

Known deviation: the encoder is a standard masked ViT. The reference also applies RoPE inside the encoder, whereas here RoPE is only in the predictor, keeping the standard building-block encoder, as discussed. The deviation is also documented for the CAPI examples in #1996.

Script only. I can't run the ImageNet benchmark on my side, so no numbers are filled in yet. Verified that it imports and static checks pass.

Follow up of #1996. Closes #1985.


lorinczszabolcs:
@gabrielfruet you might want to swap to `ViT-S` similarly to Pixio.

gabrielfruet:
@lorinczszabolcs Thank you!

gabrielfruet:
@lorinczszabolcs Would be interesting if you could add the ROPE to the encoder. I'm trying to figure out the best way to do this. I took a look into a "hacky" way of using `attn_layer` in the timms vit but is too fragile. Let me know if you have any ideas. 

Apparently there is also this: [Eva](https://github.com/huggingface/pytorch-image-models/blob/main/timm/models/eva.py)

lorinczszabolcs:
> @lorinczszabolcs Would be interesting if you could add the ROPE to the encoder. I'm trying to figure out the best way to do this. I took a look into a "hacky" way of using `attn_layer` in the timms vit but is too fragile. Let me know if you have any ideas.
> 
> Apparently there is also this: [Eva](https://github.com/huggingface/pytorch-image-models/blob/main/timm/models/eva.py)

Sure, will take a look!

lorinczszabolcs:
@gabrielfruet I think the Eva model you linked is the way to go, not patching timm's attention. Eva already has the rotary embedding inside its attention, so we don't override anything, we just wrap an Eva ViT and give it the positions of the visible patches. I tried it on a ViT-S/16 and it runs. This also overlaps what @saud5150 plans for DINOv3 (#1881): both need the wrapper to handle Eva backbones, so worth coordinating so we don't build it twice.

This would be a change to the backbone wrapper (`MaskedVisionTransformerTIMM` learning to drive an Eva model), which is reusable library code, not benchmark code. So I think it would make sense as its own PR, with while this one could stay as the benchmark plus the encoder note that's already in the docs.

One deviation is that Eva numbers patch positions as grid indices (0, 1, 2, ...) while CAPI uses fractions of the image (0 to 1), so the encoder would match what the predictor already does, not the exact numbers from the paper. Getting the paper's exact numbers would mean a small custom rotation instead of Eva.

saud5150:
> @gabrielfruet I think the Eva model you linked is the way to go, not patching timm's attention. Eva already has the rotary embedding inside its attention, so we don't override anything, we just wrap an Eva ViT and give it the positions of the visible patches. I tried it on a ViT-S/16 and it runs. This also overlaps what @saud5150 plans for DINOv3 (#1881): both need the wrapper to handle Eva backbones, so worth coordinating so we don't build it twice.
> 
> This would be a change to the backbone wrapper (`MaskedVisionTransformerTIMM` learning to drive an Eva model), which is reusable library code, not benchmark code. So I think it would make sense as its own PR, with while this one could stay as the benchmark plus the encoder note that's already in the docs.
> 
> One deviation is that Eva numbers patch positions as grid indices (0, 1, 2, ...) while CAPI uses fractions of the image (0 to 1), so the encoder would match what the predictor already does, not the exact numbers from the paper. Getting the paper's exact numbers would mean a small custom rotation instead of Eva.

Nice, agreed on both — Eva rather than patching timm's attention, and keeping it out of the benchmark PR. Happy to pick the wrapper up since DINOv3 needs the same thing, but just shout if you'd rather own it, no strong feelings either way.

Might be worth pinning down the interface while we're here, not just who writes it. Right now `MaskedVisionTransformerTIMM` is typed against `VisionTransformer`, it calls `self.vit.blocks(tokens)` with nowhere to pass `rope=`, and `add_pos_embed` reaches for `self.vit.pos_embed`, which Eva leaves as `None`. So it'd need to accept `Eva`, skip the absolute pos-embed, and run the blocks with `rope=` (`EvaBlock.forward(x, rope=..., attn_mask=...)`). The bit I'd rather not guess at is rope under masking — the rotation has to be indexed down to the visible patches, and that feels like the spot where our two use cases could quietly drift apart if we each solve it on our own. How are you handling it for CAPI?

On the coordinate thing you mentioned — this might be good news: `timm.create_model("vit_small_patch16_dinov3")` gives you an `Eva` whose rope is `RotaryEmbeddingDinoV3`, and that class takes `normalize_coords` (`"separate"` for these models) along with `shift`, `jitter` and `rescale_coords`. So fractional coordinates are configurable there at least. Could be worth a look at whichever rope class CAPI ends up with before falling back to a custom rotation.

Last thing, and it'll catch us either way: `RotaryEmbeddingDinoV3` only landed in timm 1.0.20, and we're still on `timm>=0.9.9` with a minimal-deps CI leg. That bump will need to ride along with the wrapper PR.

lorinczszabolcs:
@saud5150

Sounds good, feel free to go ahead and lead the wrapper. I'll just use it for `CAPI` when done.

`CAPI` drops the masked patches, so the encoder only sees the visible ones and rope has to line up with that. I build the full-grid rope once, gather the rows for the kept patch indices with `apply_keep_indices_nlc`, and pass that into the blocks. Tried it on an `Eva ViT-S/16` and it runs. So as long as the wrapper takes an `idx_keep` per forward and gathers the rope before the blocks, that's all `CAPI` needs.

Good call on `RotaryEmbeddingDinoV3` for the coords. It gives continuous positions but centers them to [-1, 1], and CAPI uses [0, 1], so we should check `normalize_coords` gets us close enough before writing a custom rotation.

And yeah, the wrapper PR needs to bump the timm floor from 0.9.9 for that class.

lorinczszabolcs:
@gabrielfruet would you want the encoder rope in this PR, or is it fine to merge this as the benchmark and do the rope separately?

gabrielfruet:
> would you want the encoder rope in this PR, or is it fine to merge this as the benchmark and do the rope separately?

I would prefer the Rope in this PR if it's not so much work.

If something is still unclear about the Rope, please let me know and we can brainstorm together :)

lorinczszabolcs:
Hey @gabrielfruet ,

Took a bit more time to check the DINOv3 and CAPI code properly.

- CAPI's RoPE is its own non-standard, coupled variant (`rotate_half` spans the full head dim, it uses learnable freqs, [0,1] coords, schedule is 7→7e-4, and there is no box-jitter), so timm's/DINOv3's RoPE can't reproduce it.
- To be as close to the reference implementation as possible CAPI's encoder would need a ViT reimplemented from scratch (coupled RoPE per block, RMSNorm, bias-free, edge-coordinate registers, no cls tokens) that doesn't fit `MaskedVisionTransformerTIMM`,  this was the "separate masking-aware building block" I flagged as the known deviation in #1996.
- I prototyped a bit-exact port (~325 LoC, rope+encoder+decoder), the predictor also differs from CAPI beyond RoPE. Given your [point ](https://github.com/lightly-ai/lightly/issues/1985#issuecomment-5050873335) that the lib is building blocks not full implementations, should we keep the standard encoder with the documented deviation, or add the faithful CAPI encoder/decoder as a proper port, and if so, should it be ported into the lib or just put together here in the benchmark?

gabrielfruet:
> Took a bit more time to check the DINOv3 and CAPI code properly.

No worries 🙂

Since the goal of LightlySSL is providing building blocks, I think it makes more sense to use the standard encoder, and we properly document the deviation. It's fine.

Thank you for investing time into investigating this

lorinczszabolcs:
Added the deviation note in the file + PR body, from my part it's ready to be reviewed and merged.

benchmarks/imagenet/vitb16/capi.py
```diff
@@ -0,0 +1,209 @@
...
+        num_clusters: int = 16384,
+    ) -> None:
+        super().__init__()
+        self.save_hyperparameters()
+        self.batch_size_per_device = batch_size_per_device
+        self.mask_ratio = 0.65
+        self.prediction_subsampling = 0.05
+
+        # Deviation from the reference: standard masked ViT encoder; RoPE is only in the predictor, not the encoder.
+        vit = vit_base_patch16_224(img_size=256, reg_tokens=16, dynamic_img_size=True)
```
gabrielfruet: Let's use vit small here, explicit pretrained=False

benchmarks/imagenet/vitb16/capi.py
```diff
@@ -0,0 +1,209 @@
...
+        self.sequence_length = self.student_backbone.sequence_length
+        self.num_prefix_tokens = vit.num_prefix_tokens
+        self.embed_dim = vit.embed_dim
+        grid_size = vit.patch_embed.grid_size[0]
+
+        self.student_head = CAPIProjectionHead(
+            input_dim=self.embed_dim, num_clusters=num_clusters, weight_norm=True
+        )
+        self.predictor = CAPIPredictorTIMM(
+            embed_dim=self.embed_dim, grid_size=grid_size, depth=12, num_heads=12
```
gabrielfruet: decoder depth should be half of the encoder depth (adapt to the vit-small case)

num_heads=6
depth=6 

from the section of the paper: 4.1 model architecture 


_When using a different encoder size, we align the embedding dimension, MLP ratio, and number of attention heads of the predictor to those of the encoder, and use a predictor depth equal to half that of the encoder._

gabrielfruet (changes requested):
just small alignments regarding the backbone :)

gabrielfruet (approved):
LGTM! thank you :)

gabrielfruet:
@lorinczszabolcs I'll schedule this benchmark :) Probably the results are coming in the next few weeks

Issue #1197, opened 2023-05-10
New Benchmarks


guarin:
We are in the process of updating our benchmark code and adding benchmarks on the ImageNet dataset. The goal of this issue is to coordinate and track progress of the update.

The new benchmark code lives in [lightly/benchmarks](https://github.com/lightly-ai/lightly/tree/master/benchmarks/imagenet/resnet50) and initial focus is on reference implementations for the various SSL methods following the settings from the respective papers. We'll limit training to 100 epochs and small batch sizes (2x128) for now due to limited compute resources.

> **Note**: The benchmarking code is very much work in progress, there will be frequent and breaking changes.

## Implemented Methods

Methods are roughly ordered by importance (most important first).

- [x] SimCLR #1162 
- [ ] SwaV
- [ ] MAE
- [ ] DINO
- [ ] BYOL
- [ ] MoCov2
- [ ] BarlowTwins
- [ ] MSN
- [ ] PMSN
- [ ] DCL & DCLW
- [ ] SimMIM
- [ ] TiCo
- [ ] VICReg
- [ ] VICRegL
- [ ] NNCLR
- [ ] SMoG

## Contribute

Contributions are very welcome! You can help us with the following:
* Add one of the existing models to the new benchmarks. For this you can follow the SimCLR example.
* Let us know if you spot any mistake in the models or benchmarking code.
* Run benchmarks for different epochs / batch size.

Let us know if you are working on a model and we can help you out with reviews/compute.

ibro45:
Will you be creating a tensorboard with training losses of all these networks? May be useful when a user wants to get an idea of how the loss behaves during training.

guarin:
Hi, yes this is the goal :)

PR #2004, opened 2026-07-27
feat: add SimMIM imagenet benchmark

lorinczszabolcs:
Adds a SimMIM benchmark under `benchmarks/imagenet/vitb16/` and registers it in `main.py`, completing the unchecked "SimMIM ImageNet Benchmark" item in #1197.

The benchmark reuses the shared vitb16 utilities (timm ViT-B/16 `MaskedVisionTransformerTIMM`, `OnlineLinearClassifier`, AdamW + cosine warmup) and follows the SimMIM paper (https://arxiv.org/abs/2111.09886) settings.

Like the other benchmark contributions, this adds the script only. The full ImageNet run, results row, and released checkpoints are left out from this PR.

Two notes for review:
- The 2x2 block masking follows the paper in contrast with `examples/.../simmim.py`, which masks single tokens.
- The base-lr rewrite to 1e-4 per 256 batch size assumes linear scaling from the paper's 8e-4 at batch 2048.

### DIFF
--- a/benchmarks/imagenet/vitb16/main.py
+++ b/benchmarks/imagenet/vitb16/main.py
@@ -13,6 +13,7 @@ import lejepa
 import linear_eval
 import mae
 import pixio
+import simmim
 import torch
 from pytorch_lightning import LightningModule, Trainer, seed_everything
 from pytorch_lightning.callbacks import (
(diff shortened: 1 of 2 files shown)


lorinczszabolcs:
This is an older method, so feel free to ignore it if the benchmark gaps for these models no longer matter.

PR #2006, opened 2026-07-28
Deprecate mae pixio decoder timm

lorinczszabolcs:
Follow-up to #1978. Now that MaskedVisionTransformerDecoderTIMM covers both, this deprecates the old TIMM decoders, as discussed in the review (https://github.com/lightly-ai/lightly/pull/1978#discussion_r3660592649).

- MAEDecoderTIMM and PixioDecoderTIMM emit a DeprecationWarning on init pointing to MaskedVisionTransformerDecoderTIMM, and get a docstring note. Both stay functional and exported.
- Migrated the last first-party consumer (the imagenette benchmark) off MAEDecoderTIMM.
- Added deprecation tests; marked the existing decoder and equivalence tests to ignore the warning.


lorinczszabolcs:
@gabrielfruet this is the follow-up PR to deprecate MAEDecoderTIMM and PixioDecoderTIMM.

gabrielfruet (commented):
The migration is correct. I ran the MAE flow end to end and the mask semantics match the old code. Tests pass: 18 passed, 4 skipped.

Two blockers. `PixioDecoderTIMM` users never see the deprecation warning, and the benchmark lost the decoder weight init. The other comments are smaller.

---

One more that will not anchor, because `lightly/models/modules/__init__.py` is not in the diff.

**`lightly/models/modules/__init__.py`, the `MAEDecoderTIMM, PixioDecoderTIMM` import**

This warns nobody who imports the name and never makes an instance. After the removal, those users get an `ImportError` with no hint.

A module `__getattr__` covers both cases. It needs Python 3.7. Our CI floor is 3.7.

```python
def __getattr__(name: str) -> Any:
    if name in ("MAEDecoderTIMM", "PixioDecoderTIMM"):
        from lightly.models.modules import masked_autoencoder_timm

        warn_deprecated(name, "MaskedVisionTransformerDecoderTIMM", removed_in="1.7.0")
        return getattr(masked_autoencoder_timm, name)
    raise AttributeError(name)
```

Remove the eager import. If the attribute already exists, `__getattr__` never runs. Keep the `timm_vit_available()` guard.

Keep the warning in the constructor too. A direct import from `lightly.models.modules.masked_autoencoder_timm` does not pass through here.

---

Filed #2019 for the repo-wide part, which is out of scope here.

lightly/models/modules/masked_autoencoder_timm.py
```diff
@@ -72,6 +76,13 @@ def __init__(
     ):
         """Initializes the MAEDecoderTIMM with the specified parameters."""
         super().__init__()
+        warnings.warn(
```
gabrielfruet: `PixioDecoderTIMM` users never see this warning.

Python ignores `DeprecationWarning` by default. It shows the warning only when the warning points at `__main__`:

```text
('default', None, DeprecationWarning, '__main__', 0)
('ignore',  None, DeprecationWarning, None,       0)
```

For a direct `MAEDecoderTIMM` user, `stacklevel=2` points at the caller. For Pixio it points at `PixioDecoderTIMM.__init__`, which is a lightly module. The second filter then drops it. I made both classes in a subprocess. Only the MAE call printed a warning.

Python does not filter `FutureWarning` by module. One helper keeps this in a single place:

```python
# lightly/utils/deprecation.py
def warn_deprecated(
    name: str, alternative: str, removed_in: str, stacklevel: int = 3
) -> None:
    warnings.warn(
        f"{name} is deprecated and will be removed in lightly {removed_in}. "
        f"Use {alternative} instead.",
        FutureWarning,
        stacklevel=stacklevel,
    )
```

Call the helper from each class. Do not call it once in the base class. The inherited call is the cause of this bug, and the next subclass repeats it.

Also give the removal version. `byol`, `simclr`, `moco`, `nnclr`, `simsiam`, and `barlowtwins` still say "will be deprecated in version 1.3.0". We are on 1.5.26. I suggest 1.7.0.

lightly/models/modules/masked_autoencoder_timm.py
```diff
@@ -72,6 +76,13 @@ def __init__(
...
+            f"{type(self).__name__} is deprecated in favor of "
+            "MaskedVisionTransformerDecoderTIMM; see the MAE and Pixio examples "
+            "under examples/ for the migration pattern.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
 
         self.num_prefix_tokens = num_prefix_tokens
```
gabrielfruet, line 101: This class repeats about 120 lines that `MaskedVisionTransformerDecoderTIMM` now does. You can compose instead:

```python
self.decoder = MaskedVisionTransformerDecoderTIMM(
    num_patches=num_patches,
    embed_dim=decoder_embed_dim,
    depth=decoder_depth,
    num_heads=decoder_num_heads,
    num_prefix_tokens=num_prefix_tokens,
    mlp_ratio=mlp_ratio,
    proj_drop_rate=proj_drop_rate,
    attn_drop_rate=attn_drop_rate,
    norm_layer=norm_layer,
    mask_token=mask_token,
)
self.decoder_embed = nn.Linear(embed_dim, decoder_embed_dim, bias=True)
self.decoder_pred = nn.Linear(decoder_embed_dim, patch_size**2 * in_chans, bias=True)
```

Then `embed`, `decode`, and `predict` become thin forwards. The old `decode` (pos_embed, blocks, norm) equals the new `forward` with no mask arguments. The behaviour does not change.

Two things to keep:

- Apply `init_weights` to `decoder_embed` and `decoder_pred`. The inner decoder initializes its own submodules only.
- The state_dict keys move: `decoder_pos_embed` to `decoder.pos_embed`, `decoder_blocks.*` to `decoder.blocks.*`, `decoder_norm.*` to `decoder.norm.*`, and `mask_token` to `decoder.mask_token`. You already handle this rename set in `IJEPAPredictorTIMM._migrate_legacy_state_dict`. That hook transfers almost as-is.

The tests read `decoder.decoder_pos_embed` (line 88) and `decoder.decoder_blocks` (line 143). Add properties for them, or update the tests.

tests/models/modules/test_masked_autoencoder_timm.py
```diff
@@ -12,7 +12,19 @@
 from lightly.models.modules import MAEDecoderTIMM, PixioDecoderTIMM
 
 
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
 class TestMAEDecoderTIMM:
+    def test_init__is_deprecated(self) -> None:
+        with pytest.warns(DeprecationWarning):
```
gabrielfruet: `pytest.warns` installs `simplefilter("always")`, so this passes whether or not a real user ever sees the warning. That is why the Pixio bug got through.

`FutureWarning` removes the visibility problem, so the test only needs to pin the category:

```python
with pytest.warns(FutureWarning):
    PixioDecoderTIMM(...)
```

`DeprecationWarning` is not a subclass of `FutureWarning`, so this fails if the category ever moves back. The class-level `filterwarnings` marks need the new category too.

docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -785,38 +785,41 @@ def __init__(self, dataloader_kNN, num_classes):
         self.patch_size = vit.patch_embed.patch_size[0]
         self.sequence_length = vit.patch_embed.num_patches + 1
         self.backbone = MaskedVisionTransformerTIMM(vit=vit)
-        self.decoder = MAEDecoderTIMM(
+        self.decoder_embed = nn.Linear(vit.embed_dim, decoder_dim)
```
gabrielfruet, line 789: `decoder_embed` and `prediction_head` are plain `nn.Linear` layers outside the decoder. Nothing initializes them now. Before, `_initialize_weights` reached them through `self.apply(init_weights)`: xavier uniform weights, zero bias.

At 768 to 512, the `decoder_embed` weight std moves from 0.0395 to 0.0208, and the max absolute bias moves from 0.0 to 0.0359. This changes what the benchmark trains.

```python
self.decoder_embed.apply(init_weights)
self.prediction_head.apply(init_weights)
```

`test_matches_mae_decoder_full_flow` cannot catch this. It copies the weights before it compares. `examples/pytorch/mae.py` and the other examples have the same gap. That matters more, because people copy the examples.

docs/source/getting_started/benchmarks/imagenette_benchmark.py
```diff
@@ -785,38 +785,41 @@ def __init__(self, dataloader_kNN, num_classes):
         self.patch_size = vit.patch_embed.patch_size[0]
         self.sequence_length = vit.patch_embed.num_patches + 1
```
gabrielfruet: `self.sequence_length` uses a fixed `+ 1`. The decoder two lines below gets `num_prefix_tokens=vit.num_prefix_tokens`. The two agree for a plain ViT. For a ViT with register tokens they disagree, and `x + self.pos_embed` fails.

`examples/pytorch/mae.py:29` uses `self.backbone.sequence_length`.

lightly/models/modules/masked_autoencoder_timm.py
lorinczszabolcs: Thanks, this was a really useful review!

Added a `warn_deprecated` helper in `lightly/utils/deprecation.py` using `FutureWarning` with `removed_in='1.7.0'`, called from each class. The base only warns for a direct `MAEDecoderTIMM` (guarded on `type(self))`, so Pixio no longer inherits the mis-stacked call. Verified under the default filter, one warning each, pointing at the caller.

lightly/models/modules/masked_autoencoder_timm.py
lorinczszabolcs: Held off on this one for now. The `_migrate_legacy_state_dict` hook from `IJEPAPredictorTIMM` handles the key renames so it's not much work, but it still adds a migration hook and a couple of properties (`decoder_pos_embed`, `decoder_blocks`) to a class we're removing in 1.7.0 anyway. If you'd prefer to still add it, let me know.

tests/models/modules/test_masked_autoencoder_timm.py
lorinczszabolcs: Done. Tests pin `pytest.warns(FutureWarning)` and the class-level filterwarnings marks use `FutureWarning`. I also import the deprecated names from the submodule in the non-deprecation tests, so the warning only fires where it is asserted.

docs/source/getting_started/benchmarks/imagenette_benchmark.py
lorinczszabolcs: Great catch! Restored, `decoder_embed` and `prediction_head` now get `init_weights` (xavier uniform, zero bias), matching the old decoder init. On the examples having the same gap, they are outside this PR's diff, so I would suggest a small follow-up (or folding it into #2019).

docs/source/getting_started/benchmarks/imagenette_benchmark.py
lorinczszabolcs: Fixed, it uses `self.backbone.sequence_length` now instead of the hardcoded + 1, matching `examples/pytorch/mae.py`.

lorinczszabolcs:
@gabrielfruet Thanks for the review! Addressed most of it, left a few conversations open with some comments, let me know your thoughts how to proceed.

lightly/models/modules/masked_autoencoder_timm.py
gabrielfruet: Makes sense to me also!

lightly/models/modules/__init__.py
```diff
@@ -62,3 +60,17 @@
     from lightly.models.modules.masked_vision_transformer_timm import (
         MaskedVisionTransformerTIMM,
     )
+
+
+def __getattr__(name: str) -> Any:
```
gabrielfruet: Scratch this.

It warns twice per import. `_handle_fromlist` probes with `hasattr`, then `IMPORT_FROM` fetches it for real, and both land in `__getattr__`:

```text
#1  __getattr__('masked_autoencoder_timm')  via importlib._bootstrap  -> AttributeError
#2  __getattr__('MAEDecoderTIMM')           via importlib._bootstrap  -> returned
#3  __getattr__('MAEDecoderTIMM')           via user code             -> returned
```

And a module-level `__getattr__` returning `Any` means mypy accepts anything off this module, so `modules.MaskedVisionTransformerTIM` now type-checks clean.

The only thing it covers beyond the constructor warning is importing the name and never instantiating, which isn't a use the removal breaks. The `AttributeError` on L76 carries no hint either.

Put the eager import back and keep the constructor warning. This file reverts to master exactly, plus the two `test_package_getattr__*` tests go.

lightly/models/modules/masked_autoencoder_timm.py
```diff
@@ -72,6 +77,13 @@ def __init__(
     ):
         """Initializes the MAEDecoderTIMM with the specified parameters."""
         super().__init__()
+        # Warn per class so subclasses do not inherit a mis-stacked base-class call.
+        if type(self) is MAEDecoderTIMM:
```
gabrielfruet, line 94: One gap left: subclasses get no warning, since the `type(self)` guard skips them. `__init_subclass__` covers it and points at the `class` statement, which is the line they'd have to change:

```python
def __init_subclass__(cls, **kwargs: Any) -> None:
    super().__init_subclass__(**kwargs)
    # lightly's own subclasses warn from their __init__.
    if cls.__module__ != __name__:
        warn_deprecated(
            cls.__bases__[0].__name__,
            "MaskedVisionTransformerDecoderTIMM",
            removed_in="1.7.0",
        )
```

Skip it if subclassing these is rare enough not to bother.

gabrielfruet (commented):
Everything from the last round looks right, thanks. One reversal from me below — the `__getattr__` I suggested isn't worth it, sorry for the churn.

lightly/models/modules/masked_autoencoder_timm.py
```diff
@@ -266,3 +295,8 @@ def __init__(
             initialize_weights=initialize_weights,
             mask_token=mask_token,
         )
+        warn_deprecated(
```
gabrielfruet: `PixioDecoderTIMM` warns twice for user subclasses; `MAEDecoderTIMM` warns once.

`__init_subclass__` fires when the subclass is defined, then this `__init__` warns again on every construction — it's missing the `type(self) is` guard that `MAEDecoderTIMM.__init__` has on line 94.

```python
if type(self) is PixioDecoderTIMM:
    warn_deprecated(...)
```

lorinczszabolcs: Ah, my bad! Sorry for all this back and forth.

gabrielfruet: No worries, I gave some bad calls also 😅

gabrielfruet (approved):
LGTM ! thank you

PR #2008, opened 2026-07-30
docs: interpret custom augmentation embeddings

### LINKED ISSUE #581: tutorial_custom_augmentations: make evaluation of embeddings more intuitive
## Description:
The tutorial tries to evaluate whether the embeddings are good by looking at it's neighbours:  It computes for each example image the class distribution of neighbour images, which is some kind of kNN validation.

However there are two disadvantages of this kind of evaluation, as already discussed:
https://github.com/lightly-ai/lightly/pull/577#discussion_r754065240

## Task:
- Add the interpretation of the plots as text answering the following questions: Are the embeddings good or bad? How do you see that in the plots?

iamdhrv:
## Change

- explain how to compare the example labels with their nearest-neighbor label
  distributions
- interpret all three rendered neighborhoods
- distinguish the strong normal-image grouping from the weak multi-label
  abnormal match
- state the limitation of drawing an accuracy conclusion from three examples

## Why

The tutorial says the plots evaluate representation quality but previously
stopped immediately after rendering them. The added interpretation answers
whether the displayed embeddings are good or bad, points readers to the
evidence in each plot, and frames this as a qualitative check rather than a
diagnostic metric.

## Checks

- `uvx ruff check docs/source/tutorials_source/package/tutorial_custom_augmentations.py`
- `git diff --check`

Fixes #581

### DIFF
--- a/docs/source/tutorials_source/package/tutorial_custom_augmentations.py
+++ b/docs/source/tutorials_source/package/tutorial_custom_augmentations.py
@@ -439,11 +439,6 @@ plot_knn_multilabels(embeddings, multilabels, [4111, 3340, 1796], fnames, n_neig
 #
 # The first two examples have no critical finding, and ``No finding`` also
 # dominates their neighborhoods. The embeddings therefore group these normal
-# X-rays well. The third example has several findings, but most of its neighbors
-# have no finding and only small fractions share its annotated findings. This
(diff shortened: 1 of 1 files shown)


liopeer (approved):
Valid addition, thanks!

PR #2014, opened 2026-07-31
fix(benchmarks): use ViT-S/16 for Pixio benchmark

gabrielfruet:
Pixio's ImageNet ViT benchmark was hardcoded to ViT-B/16. Switching to ViT-S/16 for a lighter benchmark run.


IgorSusmelj (approved):
LGTM

PR #2016, opened 2026-08-04
Migrate development workflow to uv lockfile

### LINKED ISSUE #2015: Mypy and numpy incompatible through `make install-dev`
I had the issue that `make install-dev` installed a numpy 2.x.x version, which was incompatible with our pinned mypy version in `pyproject.toml`. We should move to a lockfile approach like in [LightlyTrain](https://github.com/lightly-ai/lightly-train) and clean up the dependency handling overall while we're at it.

- [ ] unpin the mypy version, and instead use `>=` operator
- [ ] create `make lock` command in the Makefile with a pinned `EXCLUDE_NEWER_DATE` inside the Makefile, so that the lock does not constantly change
- [ ] use `uv run --frozen` in all the make commands instead of `python`
- [ ] move `dev` optional dependencies to a `uv` group instead (`dependency-groups`), so that the dev group is not actually distributed via PyPI
- [ ] update the `CONTRIBUTING.md` file to reflect the workflow changes (no need to activate the virtual env anymore)
- [ ] update any agent files (`CLAUDE.md` etc.) to instruct the agent to always use `uv run --frozen` instead of `python`

If anyone would like to work on this (or parts of it; we can also split this into sub-issues), let me know!

wmcv:
Fixes #2015

## Summary

- Move development dependencies to a uv dependency group
- Add a reproducible `uv.lock` with the configured cutoff date
- Add `make lock`
- Run development commands through `uv run --frozen`
- Update `CONTRIBUTING.md` and `CLAUDE.md` for the new workflow

## Testing

- `make lock` passes
- `make format-check` passes
- `make type-check` now completes and reports five existing type-checking errors in unrelated files
- `make install-minimal` currently fails due to an existing NumPy 1.18.1/Python 3.12 compatibility issue

## Summary

- Migrated development workflows to a reproducible `uv.lock` setup.
- Updated Makefile targets, CI workflows, and documentation to use frozen `uv` commands.
- Raised the minimum Python version to 3.8 and moved development dependencies into `uv` groups.
- Added type casts and typed test helpers for static-check compatibility.

## Validation

- `make lock` and `make format-check` pass.
- `make type-check` reports five existing unrelated errors.
- `make install-minimal` remains blocked by NumPy 1.18.1 incompatibility with Python 3.12.


CONTRIBUTING.md
```diff
@@ -82,19 +82,21 @@ Follow these steps to start contributing:
...
+   This synchronizes the project environment from the lockfile and installs the
+   pre-commit hooks. You do not need to activate the virtual environment manually.
+
 5. Develop the features on your branch.
 
-   As you work on the features, you should make sure that the code is formatted and the
-   test suite passes:
+   As you work on the features, you should make sure that the code is formatted and
+   the test suite passes. The Makefile runs development tools through the locked `uv`
+   environment, so manual virtual environment activation is not required:
```
liopeer, line 102: Seems like there is currently no section on the unit tests. Would add explanation on running `make test` and on running singular test directories/modules/tests with `uv run --frozen pytest ...`.

Makefile
```diff
@@ -100,16 +100,13 @@ install-uv:
 
 .PHONY: reset-venv
 reset-venv:
-	deactivate || true
 	rm -rf .venv
-	uv venv .venv
+	uv sync --frozen --all-extras
```
liopeer: This is not the same logic anymore. Previously it was:
1. delete the venv
2. create a new venv
Now we're also installing all the dependencies (including extras). So this needs to be addressed.

Makefile
```diff
@@ -158,17 +160,17 @@ install-api-only:
 # --reinstall: Reinstall dependencies to make sure they satisfy the constraints.
 .PHONY: install-minimal
 install-minimal:
-	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} ${EDITABLE} ".[dev]"
+	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} --group dev ${EDITABLE} .
 	uv pip install --resolution=lowest-direct --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall ${EDITABLE} ".[minimal]"
 	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall "setuptools<50"
```
liopeer: The goal here is now that we only need a singular command.

Makefile
```diff
@@ -158,17 +160,17 @@ install-api-only:
...
-# We do not use --all-extras because it includes the dev dependencies for which we don't
-# want to install the minimal versions.
+# Install selected extras separately so their minimal versions can be tested.
+# Development dependencies are installed from the dev dependency group.
 .PHONY: install-minimal-extras
 install-minimal-extras:
-	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} ${EDITABLE} ".[dev]"
+	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} --group dev ${EDITABLE} .
 	uv pip install --resolution=lowest-direct --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall ${EDITABLE} ".[matplotlib,minimal,timm,video]" --requirement pyproject.toml
 	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall "setuptools<50"
```
liopeer: Here we also just want a singular command.

Makefile
```diff
@@ -189,7 +191,8 @@ install-pinned-extras:
 # available at EXCLUDE_NEWER_DATE. This excludes video dependencies.
 .PHONY: install-pinned-extras-no-video
 install-pinned-extras-no-video:
-	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall ${EDITABLE} ".[dev,matplotlib,minimal,timm]" --requirement pyproject.toml
+	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall --group dev
+	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall ${EDITABLE} ".[matplotlib,minimal,timm]" --requirement pyproject.toml
```
liopeer: Same here.

liopeer (changes requested):

Makefile
```diff
@@ -75,7 +75,7 @@ all-checks: static-checks test
 
 ## build source and wheel package
 dist: clean
-	python -m build
+	uv build
```
liopeer: Need to check if that actually gives the same result.

Makefile
```diff
@@ -100,34 +100,41 @@ install-uv:
 
 .PHONY: reset-venv
 reset-venv:
-	deactivate || true
```
liopeer, line 109: Make it retain the same Python version.

Makefile
```diff
@@ -100,34 +100,41 @@ install-uv:
...
-# When running these commands locally, it is recommended to first reset the environment
-# with: `make reset-venv && source .venv/bin/activate`
-# Otherwise old dependencies might linger around.
+# Project commands use uv and do not require manually activating the virtual environment.
 
-# Set EDITABLE to -e to install the package in editable mode outside of CI. This is
-# useful for local development.
+# Install the package in non-editable mode in CI.
 ifdef CI
 EDITABLE=
```
liopeer: I think EDITABLE is not used anymore. Only NO_EDITABLE.

Makefile
```diff
@@ -155,21 +159,18 @@ install-api-only:
...
 # See install-minimal for explanation of flags.
-# We do not use --all-extras because it includes the dev dependencies for which we don't
-# want to install the minimal versions.
+# Install selected extras separately so their minimal versions can be tested.
+# Development dependencies are installed from the dev dependency group.
 .PHONY: install-minimal-extras
 install-minimal-extras:
-	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} ${EDITABLE} ".[dev]"
-	uv pip install --resolution=lowest-direct --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall ${EDITABLE} ".[matplotlib,minimal,timm,video]" --requirement pyproject.toml
+	uv sync --python=${MINIMAL_PYTHON_VERSION} --resolution=lowest-direct --exclude-newer ${EXCLUDE_NEWER_DATE} ${NO_EDITABLE} --group dev --extra matplotlib --extra minimal --extra timm --extra video --upgrade-group dev
```
liopeer: This currently does not properly work since we're not setting lower bounds for many packages.

liopeer (commented):
Thanks for the work! I think I will have to take this over, since we will need to also fix a few things in the overall dependency handling. Might touch a bit too many things in the code base for a first contribution (but don't worry, once we merge the commits will still be associated with you as well!).

.github/workflows/release_pypi.yml
```diff
@@ -21,18 +21,25 @@ jobs:
...
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USER_NAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
       run: |
-        pip3 install wheel
-        pip3 install twine
-        pip3 install build
-        make dist
-        twine upload -u ${{ secrets.PYPI_USER_NAME }} -p ${{ secrets.PYPI_PASSWORD }} dist/*
+        uv build
```
liopeer, line 38: We'll only see if this works once we release.

gabrielfruet:
Deferred review follow-up: the type-check job installs no extras, so mypy checks timm/matplotlib-using code as `Any`. Switching it to `install-maximal-extras` closes that gap but may surface new mypy errors, so it's better as its own PR.

gabrielfruet (approved):

gabrielfruet (approved):

PR #2017, opened 2026-08-05
build: add make target to run distributed tests locally

saud5150:
Follow-up to #2011, requested by @liopeer in https://github.com/lightly-ai/lightly/pull/2011#issuecomment-5176019537.

## Description
- [ ] My change is breaking

Adds `make test-distributed` to run the `@pytest.mark.DDP` tests on the shared gloo pool locally, mirroring the CI `distributed-tests` job (`USE_PYTEST_POOL=1 ... --runslow -m DDP`) so the invocation doesn't have to be reconstructed by hand.

One non-obvious detail, documented as a comment on the target: it must use `python -m pytest` rather than plain `pytest` like the neighbouring targets. `python -m` puts the repo root on `sys.path`, which the spawned pool workers inherit; with the bare `pytest` entry point the workers crash on their first task with `ModuleNotFoundError: No module named 'tests'` and the session hangs silently at the pool's `starmap`. (CI never hits this because it already invokes `python -m pytest`.)

`--runslow` is kept for parity with the CI job even though neither current DDP test is slow-marked, so a future slow+DDP test behaves the same locally and in CI. On platforms without a usable backend (e.g. Windows), `USE_PYTEST_POOL` folds in the platform checks, so the target just reports the DDP tests as skipped.

## Tests

macOS arm64, Python 3.12, fresh venv (`.[dev,matplotlib,minimal,timm,video]`):

```bash
make test-distributed
# => 2 passed, 1 skipped, 2002 deselected in 15.70s   # BarlowTwins DDP tests run on the pool

# control — same selection with the flag off:
python -m pytest tests --runslow -m DDP
# => 3 skipped, 2002 deselected                       # DDP tests correctly skip

# regression check for the sys.path detail — bare pytest instead of python -m pytest:
USE_PYTEST_POOL=1 pytest tests --runslow -m DDP
# => workers die with ModuleNotFoundError: No module named 'tests'; session hangs
```

## Documentation

Makefile-only change; the `python -m pytest` requirement is documented on the target itself.

- `.PHONY` is omitted to match the neighbouring test targets (`test`, `test-fast`, `lint`, ...), none of which declare it. Happy to add it here (or as a repo-wide `.PHONY` cleanup in a separate PR) if you prefer.


gabrielfruet (approved):
LGTM!! Thank you

liopeer:
Very confused that I somehow can't run the workflows here 🤔

PR #2020, opened 2026-08-11
Move IJEPAMaskCollator to its own module, keep the old import working

gabrielfruet:
`IJEPAMaskCollator` sat in `lightly/data/collate.py` alongside the deprecated collate functions, so it was in line to be deleted with them despite never being deprecated itself. Import it from `lightly.data` now.

## Changes

- The class moved to `lightly/data/ijepa_collate.py`, body unchanged apart from type annotations the new module needs: `collate.py` is on the mypy exclude list and it isn't.
- `collate.py` re-exports the name through a module-level `__getattr__` with a `FutureWarning`, and that path goes away in 1.7.0. It sits in the `else` branch of a `TYPE_CHECKING` guard, keeping the name bound for type checkers so unknown attributes still error instead of becoming `Any`.
- First tests the class has ever had, in `tests/data/test_ijepa_collate.py`. The two that hit `default_collate` skip on torch<1.11, which doesn't have it.
- Docs and `examples/pytorch/ijepa.py` plus its generated notebook use the new path.

Module-level `__getattr__` is PEP 562, which needs Python 3.7, and `pyproject.toml` still declares `requires-python = ">=3.6"`; on 3.6 the old import path raises ImportError instead of warning. CI's lowest version is 3.7.


liopeer (changes requested):
Two things:
- We should not use `Optional` and `Union` in the type hints, instead we should use `|` with `None` and `from __future__ import annotations`.
- This PR brings a breaking change without any deprecation warning. To me it looks like something we should not merge before we have not set up the proper process for planning 2.x.x release.

gabrielfruet:
Both fixed — `|` unions throughout, and `lightly.data.collate` re-exports the class with a `FutureWarning` until 1.7.0, so it's no longer a breaking change.

liopeer (approved):

PR #2023, opened 2026-08-12
fix(loss): include all views in LeJEPA SIGReg and invariance loss (#2002)

### LINKED ISSUE #2002: LeJEPA loss missing contributions from global views?
I appreciate the addition of LeJEPA to LightlySSL, but I was confused by one aspect of the implementation as I worked through adding it to one of my projects.

In the [original paper](https://arxiv.org/abs/2511.08544), I read the loss function as described in pseudocode on p. 11, or with standard math notation on p. 12, as a sum of contributions from all of the image views, local as well as global.

In the [LightlySSL implementation of the LeJEPA loss](https://github.com/lightly-ai/lightly/blob/master/lightly/loss/lejepa_loss.py), however, we seem to only include the contributions from the local views.

I suspect this is a bug since it seems odd to only constrain the embeddings associated with the local views during training. The original authors also describe e.g. training ResNets using only global views, and that doesn't make sense as the loss is implemented here.

Assuming my understanding is correct, I believe the fix is simple: https://github.com/lightly-ai/lightly/blob/master/lightly/loss/lejepa_loss.py#L327-L330 should pass all views, local and global, where it currently passes only local_proj, to the underlying SIGReg and invariance losses.

harmehak0173:
### Summary

Fixes #2002.

This PR updates `LeJEPALoss` to include all views (local + global) in both `SIGReg` regularization and `lejepa_invariance_loss` calculation, following the paper specification:
- Concatenates `local_proj` and `global_proj` into `all_proj` in `LeJEPALoss.forward()`.
- Passes `all_proj` to `self.sigreg(all_proj)` and `lejepa_invariance_loss(local_proj=all_proj, global_proj=global_proj)`.
- Updated docstrings and unit tests in `tests/loss/test_lejepa_loss.py`.

### Testing
- `pytest tests/loss/test_lejepa_loss.py`
- `ruff check`


lightly/loss/lejepa_loss.py
```diff
@@ -324,9 +324,10 @@ def forward(self, *, local_proj: Tensor, global_proj: Tensor) -> Tensor:
             local_proj: Local-view projected embeddings of shape ``(Vl, N, D)``.
             global_proj: Global-view projected embeddings of shape ``(Vg, N, D)``.
         """
-        sigreg_loss = self.sigreg(local_proj)
+        all_proj = torch.cat([local_proj, global_proj], dim=0)
+        sigreg_loss = self.sigreg(all_proj)
         inv_loss = lejepa_invariance_loss(
-            local_proj=local_proj, global_proj=global_proj
+            local_proj=all_proj, global_proj=global_proj
```
gabrielfruet: `ruff format` joins this into one line now (87 < 88), so `make format-check` fails.

```python
inv_loss = lejepa_invariance_loss(all_proj=all_proj, global_proj=global_proj)
```

lightly/loss/lejepa_loss.py
```diff
@@ -324,9 +324,10 @@ def forward(self, *, local_proj: Tensor, global_proj: Tensor) -> Tensor:
             local_proj: Local-view projected embeddings of shape ``(Vl, N, D)``.
             global_proj: Global-view projected embeddings of shape ``(Vg, N, D)``.
         """
-        sigreg_loss = self.sigreg(local_proj)
+        all_proj = torch.cat([local_proj, global_proj], dim=0)
```
gabrielfruet: `local_proj` now receives all views — let's rename it here rather than in a follow-up. It isn't exported from `lightly.loss`, so it's cheap.

```python
def lejepa_invariance_loss(*, all_proj: Tensor, global_proj: Tensor) -> Tensor:
```

Other callers: `visreg_loss.py:328` and both test files. `_validate_projection_shapes` and its error strings too.

Nit: VISReg cats global first. Same result, worth matching.

gabrielfruet (changes requested):
Change is right, and it lines LeJEPA up with VISReg (`visreg_loss.py:325`).

Two things inline. Also `docs/source/examples/lejepa.rst:20` still says "Each local view's projection is pulled toward the centroid" — same edit as the class docstring.

harmehak0173:
Thanks @gabrielfruet! All requested updates have been addressed and pushed:

1. Renamed the `local_proj` parameter to `all_proj` in `lejepa_invariance_loss` (and updated docstrings, `_validate_projection_shapes`, `visreg_loss.py`, and test files).
2. Updated `docs/source/examples/lejepa.rst` line 20 to match the class docstring ("Every view's projection is pulled toward the centroid...").
3. Matched `torch.cat([global_proj, local_proj])` view concatenation order in `LeJEPALoss.forward` and formatted with `ruff`.

lightly/loss/lejepa_loss.py
```diff
@@ -241,13 +239,13 @@ class LeJEPALoss(nn.Module):
 
     The loss is a convex combination of two terms:
 
-    - ``SIGReg(local_proj)`` regularizes local projections toward an
+    - ``SIGReg(all_proj)`` regularizes all projected embeddings (local and global) toward an
       isotropic Gaussian distribution.
-    - ``lejepa_invariance_loss(local_proj=local_proj, global_proj=global_proj)``
-      pulls each local view toward the mean of the global views.
+    - ``lejepa_invariance_loss(local_proj=all_proj, global_proj=global_proj)``
```
gabrielfruet: Rename missed this one — copy-pasting it now raises `TypeError`.

```
lejepa_invariance_loss(all_proj=all_proj, global_proj=global_proj)
```

lightly/loss/lejepa_loss.py
```diff
@@ -8,21 +8,21 @@
...
 
     Reference:
         LeJEPA, 2025, https://arxiv.org/abs/2511.08544
 
     Args:
-        local_proj:
-            Projected embeddings of shape ``(Vl, N, D)`` where ``Vl`` is the
-            number of local views, ``N`` is the batch size, and ``D`` is the
+        all_proj:
+            Projected embeddings of shape ``(Val, N, D)`` where ``Val`` is the
```
gabrielfruet: `Val` is one char off `Vl`, which still means local views at line 322. The error string at line 43 uses `V_all`. Pick one spelling. I prefer `V_all`

gabrielfruet (approved):
All addressed, and format-check is green now. Two doc nits inline, neither blocking. I would suggest that you fix those :)

harmehak0173:
Thanks @gabrielfruet! Addressed both doc nits in `lejepa_loss.py` (`V_all` spelling consistency and `all_proj=` in example snippet).

gabrielfruet:
Sorry for dragging this out, pushed two small commits to your branch instead of another round.

First moves the shape check ahead of `torch.cat`, so mismatched views raise `ValueError` again. Second footnotes the LeJEPA benchmark numbers, which predate this fix.

Thanks again for picking this up.

harmehak0173:
Thanks @gabrielfruet! Appreciate the polish commits. Happy to help out!

PR #2024, opened 2026-08-12
added types for online_linear_classifier.py

rayyanshuda:
Part of#1635

Added type annotations to lightly/utils/benchmarking/online_linear_classifier.py and its test file, and removes both from the mypy exclude list in pyproject.toml. Type-checked files go from 551 to 553. I checked with another file in utils/benchmarking for confirmation (not an open issue file).

I left lightly.utils.benchmarking.* in the follow_imports = "skip" list, since benchmark_module.py and knn_classifier.py in that package are still excluded.

make format-check, make type-check, and pytest tests/utils/benchmarking/test_online_linear_classifier.py all pass locally.


gabrielfruet (approved):
Thank you!! these improvements in typing for the library are really valuable :)

PR #2026, opened 2026-08-12
fix: examples +  benchmarks: weight init in mae+pixio

lorinczszabolcs:
Follow-up to the review on (https://github.com/lightly-ai/lightly/pull/2006#discussion_r3747867713).

After the MAE/Pixio decoders moved to `MaskedVisionTransformerDecoderTIMM`, `decoder_embed` and `prediction_head` are plain `nn.Linear` layers outside the decoder. The decoder only initializes its own submodules, so those two layers were left at PyTorch's default init instead of the xavier-uniform weight / zero bias that `MAEDecoderTIMM` applied through `init_weights`. #2006 fixed this in the imagenette benchmark; this applies the same fix to the examples and the vitb16 benchmark scripts.

- Examples: `mae.py` and `pixio.py` under `pytorch`, `pytorch_lightning`, and `pytorch_lightning_distributed` now apply `init_weights` to `decoder_embed` and `prediction_head`. Regenerated the notebooks.
- Benchmarks: same fix in `benchmarks/imagenet/vitb16/{mae,pixio}.py`. This changes their initialization, so the MAE/Pixio benchmark numbers would need a re-run.

Left `IJEPAPredictorTIMM` unchanged: it builds the decoder with `initialize_weights=False` and relies on the timm default init for its predictor layers on purpose, so it isn't the same case.

Verified `make format-check` passes, the notebooks are up to date, and `decoder_embed`/`prediction_head` now initialize with zero bias.

### DIFF
--- a/benchmarks/imagenet/vitb16/mae.py
+++ b/benchmarks/imagenet/vitb16/mae.py
@@ -11,6 +11,7 @@ from lightly.models.modules import (
     MaskedVisionTransformerDecoderTIMM,
     MaskedVisionTransformerTIMM,
 )
+from lightly.models.modules.masked_vision_transformer_timm import init_weights
 from lightly.transforms import MAETransform
 from lightly.utils.benchmarking import OnlineLinearClassifier
 from lightly.utils.scheduler import CosineWarmupScheduler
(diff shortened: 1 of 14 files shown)


lorinczszabolcs:
[pixio.pdf](https://github.com/user-attachments/files/30989370/pixio.pdf)
[mae.pdf](https://github.com/user-attachments/files/30989379/mae.pdf)

gabrielfruet (approved):
Small and self contained! thank you a lot Lorincz :)

PR #2027, opened 2026-08-12
test: finish the unittest→pytest migration of model-module tests (#1987)

lorinczszabolcs:
Follows #1987 (migrate `unittest.TestCase` tests to pytest style).

Two commits:

1. **Migrate the file #1987 missed.** `test_masked_vision_transformer_decoder_timm.py` wasn't in any of #1987's groups — it was added later in #1978, after the groups were claimed, so it was never migrated. It's the only remaining non-API `unittest.TestCase` file; this converts it to the style in `tests/loss/test_ibot_loss.py` (no `TestCase` base, `assert`, `pytest.raises`, `parametrize("device", ...)`).

2. **Parametrize the cpu/cuda variants** in the already-migrated model-module tests (`test_masked_autoencoder_timm.py`, `test_ijepa_timm.py`, `test_masked_autoencoder.py`), replacing the `_test_forward(device)` helper + `test_forward`/`test_forward_cuda` pairs with `parametrize("device", ...)`, matching the pattern #1987 describes. `test_memory_bank.py` is left as-is, its cuda test is a standalone body, not a variant of a shared helper.


gabrielfruet (approved):
LGTM!! thank you

PR #1939, opened 2026-06-11
Remove API client from LightlySSL

gabrielfruet:
## Summary

Removes the LightlyOne/API client surface from LightlySSL.

This includes:
- Removing `lightly.api`, `lightly.openapi_generated`, and `lightly.active_learning`
- Removing the `lightly-download` and `lightly-serve` CLI commands
- Dropping API-only dependencies, tooling config, workflows, tests, mocks, and docs
- Keeping the package version unchanged at `1.5.24`

## Breaking Changes

- `lightly.api` is no longer available
- `lightly.openapi_generated` is no longer available
- `lightly.active_learning` is no longer available
- `lightly-download` and `lightly-serve` CLI commands are removed

Users on the old `ApiWorkflowClient` workflow should pin to `pip install "lightly<1.16"` (see the deprecation `ImportError` raised by the package).

## Verification

Before commit:
- `make format` passed
- `ruff format-check`: 446 files already formatted
- Smoke test: `lightly` version `1.5.24`, all public subpackages import OK
- `pytest`: 1347 passed, 3 failed
  - Failures are pre-existing:
    - `VideoLoader` pickle x2
    - `torch.allclose` numerical failure
- `mypy`: pre-existing INTERNAL ERROR in third-party `transformers` on Python 3.13, unrelated to these changes
- Deprecated import shim tested: `import lightly.api`, `from lightly import api`, `from lightly.api import ApiWorkflowClient`, `importlib.import_module("lightly.api")` all raise the expected `ImportError` (`tests/test_deprecated_api_import.py`)

## Follow-ups (intentionally deferred to separate PRs)

- **Version bump/fix:** the package still reads `1.5.24` while the latest tag is `v1.15.24`. Fixing that (and coordinating the release notes) will be its own PR so this one stays scoped to the API removal.
- **Docs (external repo):** companion PR [lightly-ai/readme-docs#14](https://github.com/lightly-ai/readme-docs/pull/14) pins the install command to `"lightly<1.16"` in the three tutorials/recipes that still use `ApiWorkflowClient` (active-learning-for-waste-sorting, active-learning-yolov7, recipes/create-dataset).
- **README studio links:** addressed in `f06dd3e` per the liopeer review (now points to `https://docs.lightly.ai/studio/` instead of the renamed "Lightly Platform" strings).

## Notes

No release/version bump and no customer communications work included.

PR #2030, opened 2026-08-14
chore: standardise deprecations using the warn_deprecated helper

### LINKED ISSUE #2019: Standardise deprecations on a single helper
Deprecations in this repo use three shapes:

- A raw `warnings.warn` in the constructor (`models/byol.py:67`, copied into simclr, moco, nnclr, simsiam, and barlowtwins).
- A named helper (`data/collate.py:1514`, `active_learning/__init__.py:4`).
- A docstring note with no runtime warning (`loss/msn_loss.py:124`).

Six of them say "will be deprecated in version 1.3.0". We are on 1.5.26 and all six are still here.

Proposal: one helper in `lightly/utils/deprecation.py`.

```python
def warn_deprecated(
    name: str, alternative: str, removed_in: str, stacklevel: int = 3
) -> None:
```

It raises `FutureWarning`, not `DeprecationWarning`. Python ignores `DeprecationWarning` unless the warning points at `__main__`. A user who calls lightly from inside their own package sees nothing.

Work:

- Add the helper. #2006 uses it for the two decoder classes.
- Move the six model classes, `data/collate.py`, and `active_learning/__init__.py` onto the helper. Give each one a real removal version.
- Add `filterwarnings = ["error::FutureWarning"]` to the pytest configuration. The repo has no `filterwarnings` configuration today. That is why the benchmark kept calling a deprecated API and nobody saw it.

lorinczszabolcs:
Closes #2019. Follows up on #2006, which added `warn_deprecated`.

The high-level model classes (BarlowTwins, BYOL, MoCo, NNCLR, SimCLR, SimSiam), the collate functions, and the active-learning entry point now warn through `warn_deprecated`. Each raises a `FutureWarning` and names 1.7.0 as the removal version, replacing the stale 1.3.0 and 1.4.0 strings. Each migrated symbol has a deprecation test.

The pytest configuration sets `filterwarnings = ["error::FutureWarning", ...]`, so a deprecated API used in the tests fails the suite. Tests that use one on purpose opt out with a filter mark. Torch's own `weight_norm` and `_pytree` warnings are allowlisted because they are not ours to fix.

Questions:
- The collate deprecation is allowlisted because the train CLI still builds a collate function internally (`lightly/cli/train_cli.py`), which trips the guard. With the guard on, `lightly-train` now shows this `FutureWarning` to users at runtime. Do you want to keep the allowlist and migrate the CLI to transforms in a follow-up, or handle it differently in this PR?
- The removal version is 1.7.0 to match the decoders deprecated in #2006.

### DIFF
--- a/lightly/active_learning/__init__.py
+++ b/lightly/active_learning/__init__.py
@@ -1,11 +1,9 @@
-import warnings
+from lightly.utils.deprecation import warn_deprecated
 
 
-def raise_active_learning_deprecation_warning():
-    warnings.warn(
-        "Using active learning via the lightly package is deprecated and will be removed soon. "
-        "Please use the Lightly Solution instead. "
-        "See https://docs.lightly.ai for more information and tutorials on doing active learning.",
-        DeprecationWarning,
-        stacklevel=2,
+def raise_active_learning_deprecation_warning() -> None:
+    warn_deprecated(
+        "Active learning via the lightly package",
+        "the Lightly Solution",
+        removed_in="1.7.0",
     )
--- a/lightly/data/collate.py
+++ b/lightly/data/collate.py
@@ -4,7 +4,6 @@
 # All Rights Reserved
 
 from typing import TYPE_CHECKING, Any, List, Optional, Tuple, Union
-from warnings import warn
 
 import torch
 import torch.nn as nn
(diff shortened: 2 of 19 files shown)


lorinczszabolcs:
Closing this. I opened it before realizing the same thing is being worked on in #2021 and #1939

PR #2031, opened 2026-08-15
(prototype) SimCLR v2 refactor

gabrielfruet:
SimCLR existed four times here and nothing compared the copies: the two shipped heads differed by 4,922,112 parameters, split and fused forwards gave gradients at cosine similarity 0.0837, and both examples took `NTXentLoss()`'s default 0.5 while the benchmark passed 0.1.

First vertical slice of the 2.0 design, SimCLR only. The other 29 methods are untouched.

- `lightly/data/sample.py`: a batch is a `Sample` of typed `View`s, one generic collate, plus `legacy_collate`.
- `lightly/backbones/`: `Backbone` and `DenseBackbone` protocols, a torchvision ResNet adapter, a conformance suite.
- `lightly/nn/axes.py`: `encode(..., group_by="shape")`, so BatchNorm sees 2N.
- `lightly/functional/ntxent.py` and `lightly/optim/`: the equation and the optimiser pieces get homes; old paths still import.
- `examples/simclr.py` and `benchmarks/simclr/` replace the four old copies.

`tests/test_simclr_agrees.py` compares the method, not the run: blocks, view contract, fused forward, head width, and that each side states its temperature. Batch size, backbone and temperature values are free to differ, and do.

Breaking: `SimCLRTransform`, `MoCoV1/V2Transform` and `DenseCLTransform` return `list[View]`. Read `views[0].data`, or pass `collate_fn=legacy_collate`.

Testing: format-check, mypy and pytest clean bar one pre-existing DCL distributed failure; the example runs an epoch; `--fast-dev-run` logs `train_loss` and both probes; both gate mutations fail as intended.


gabrielfruet:
Split into a stack of seven, one thing each, since this was too big to read as one diff:

1. #2033 the `View`/`Sample` batch contract
2. #2034 backbone protocols and the torchvision ResNet adapter
3. #2035 the NT-Xent equation into `lightly.functional`
4. #2036 `lightly.optim`
5. #2037 SimCLR-family transforms return views
6. #2038 the SimCLR example, benchmark and gate
7. #2039 delete the four old copies

The `encode(..., group_by="shape")` helper that was in this branch is gone from the stack. For two same-shape views it was `cat` and `chunk` behind a name, and the example exists to be read. It comes back when DINO's 224/96 multi-crop makes `cat` impossible.

PR #2032, opened 2026-08-15
(draft) feat: add LeWM, an action-conditioned latent world model trained with SIGReg

gabrielfruet:
LeWM is a latent world model: it predicts the next frame's embedding from past embeddings and the action taken, never reconstructing pixels. SIGReg, already here for LeJEPA, prevents collapse, so there is no teacher, no EMA and no stop-gradient, and the encoder trains from pixels alongside the predictor. Paper: https://arxiv.org/abs/2603.19312

- `lightly/models/modules/world_model/`: LatentDynamicsPredictor, a causal transformer with AdaLN-Zero action conditioning and a `rollout()` that feeds predictions back over a sliding window, plus an ActionEncoder MLP.
- `lightly/loss/`: LeWMLoss (prediction MSE plus `lambda_param` * SIGReg, default 0.1) and `latent_distance`, a new l1/l2 helper with optional layer norm for later methods.
- A PyTorch example, docs, and ~500 lines of tests.
- `scaled_dot_product_attention` goes through `getattr`, with a manual fallback, so type checking passes on torch 1.10.

The API covers only what LeWM needs, the first of a planned sequence; later methods add keyword arguments defaulting to this behavior. Missing: the Lightning example variants and the README table row.

The example synthesizes its own trajectories, a square pushed by the action, so it needs no simulator. Loss falls from 1.05 to 0.33 over 10 epochs. That shows the loop trains, not that it reproduces the paper. 12 CI checks green.

## Summary

- Adds LeWM latent world-model components for action-conditioned embedding prediction.
- Adds `LatentDynamicsPredictor`, `ActionEncoder`, `LeWMProjectionHead`, `LeWMLoss`, and `latent_distance`.
- Adds causal and conditional prediction, autoregressive rollout, SIGReg regularization, and PyTorch compatibility fallback.
- Adds documentation, examples, and comprehensive tests.


lightly/models/modules/world_model/predictor.py
```diff
@@ -0,0 +1,306 @@
...
+
+import torch
+from torch import Tensor, nn
+
+# torch added scaled_dot_product_attention in 2.0. The package still declares
+# torch>=1.10, so the fused call is optional and a manual path stands in.
+_FUSED_ATTENTION_AVAILABLE = hasattr(nn.functional, "scaled_dot_product_attention")
+
+
+def _manual_attention(
```
gabrielfruet: Let's not provide a manual attention, if fused attention is not available, let's just crash it.

examples/pytorch/lewm.py
```diff
@@ -0,0 +1,171 @@
...
+from torch.utils.data import Dataset
+
+from lightly.loss import LeWMLoss
+from lightly.models.modules import ActionEncoder, LatentDynamicsPredictor
+
+# Number of frames in one training clip. The predictor sees the first
+# num_frames - 1 frames and predicts the embedding of the following frame.
+num_frames = 4
+image_size = 64
+embed_dim = 192
```
gabrielfruet: Prefer either constants that are passed to the module, or are default params. DOn't just use them inside modules without being a param.

lightly/loss/lewm_loss.py
```diff
@@ -0,0 +1,143 @@
...
+        self.normalize = normalize
+        self.sigreg = SIGReg(
+            knots=sigreg_knots,
+            t_max=sigreg_t_max,
+            num_vectors=sigreg_num_vectors,
+            gather_distributed=gather_distributed,
+        )
+
+    def forward(
+        self, *, predicted: Tensor, target: Tensor, embeddings: Tensor
```
gabrielfruet: Maybe this could be a problem for ONNX export. 

Having kw only arguments. I'm not sure. Investigate into this and document it somewhere. 

I think it's not a problem here exactly, but may be in other places (modules)

examples/pytorch/lewm.py
```diff
@@ -0,0 +1,171 @@
...
+        # Any image encoder works here. LeWM trains it from scratch, but a
+        # frozen pretrained encoder can be used instead without changing the
+        # predictor or the loss.
+        self.backbone = vit_tiny_patch16_224(
+            pretrained=False,
+            img_size=image_size,
+            num_classes=0,
+            dynamic_img_size=True,
+        )
+        self.projection_head = Sequential(
```
gabrielfruet: Maybe this should be a head in our Head api?

lightly/loss/lewm_loss.py
```diff
@@ -109,6 +109,15 @@ def __init__(
             gather_distributed=gather_distributed,
         )
 
+    # forward takes keyword-only arguments so a caller cannot silently swap
+    # predicted and target. This is safe for a loss but is not a pattern to copy
```
gabrielfruet: This comment is not necessary.

examples/pytorch/lewm.py
gabrielfruet: UPPERCASE if module constants. only consider module constants if necessary, otherwise rely on the torch module defaults

examples/pytorch/lewm.py
```diff
@@ -0,0 +1,178 @@
...
+
+from lightly.loss import LeWMLoss
+from lightly.models.modules import (
+    ActionEncoder,
+    LatentDynamicsPredictor,
+    LeWMProjectionHead,
+)
+
+
+class MovingSquareTrajectories(Dataset):
```
gabrielfruet: Try Gym API here

PR #2038, opened 2026-08-16
(prototype) One SimCLR example, one benchmark, and the gate between them

gabrielfruet:
> **Prototype. Do not merge.** This is a design exploration for the LightlySSL 2.0
> refactor, opened to be read and argued with, not to land on `master`. The whole
> stack goes together or not at all, and the shape is still open.

6 of 7 in a stack. Base: #2037.

SimCLR exists four times in this tree and nothing compares the copies. The two shipped heads differ by 4,922,112 parameters, so `load_state_dict` raises. The split and fused forwards give gradients at cosine similarity 0.0837. Both examples take `NTXentLoss()`'s default 0.5 while the benchmark passes 0.1.

`examples/simclr.py` is plain torch with the loop in view, small enough to run: ResNet-18, CIFAR-10, batch 256. `benchmarks/simclr/` is the same method on Lightning, parameterised by one row of `datasets.py`, where the `imagenet` row is the paper's numbers. Every row states every field, so two rows side by side show every difference.

`tests/test_simclr_agrees.py` is what makes them one method. It compares the method, not the run: block classes, the view contract, the fused forward, the head's width against the backbone's, and that each file's criterion carries the temperature that file declares. Batch size, backbone, learning rate and temperature values are free to differ.

The old four are deleted in PR 7, not here.

Testing: 14 gate cases; mutating either side's forward to run one view at a time fails the gate and names the side.

### DIFF
--- a/benchmarks/simclr/benchmark.py
+++ b/benchmarks/simclr/benchmark.py
@@ -1,0 +1,248 @@
+"""SimCLR on Lightning, parameterised by one row of datasets.py.
+
+The same method as ``examples/simclr.py``, written for the other job. The example
+is the file to read; this one has to survive a week on a preemptible cluster, so
+it hands the loop, the checkpoint and the step counter to Lightning. What it adds
+on top of the example is the schedule, the probes, the logger, bf16, a checkpoint
+and the dataset axis.
+
+The two files restate each other on purpose, and ``tests/test_simclr_agrees.py``
+is what holds them to the same method. They do not share numbers: the example is
+one small configuration, the ``imagenet`` row is the paper's.
+
+Run it with::
+
+    torchrun --nproc_per_node=8 -m benchmarks.simclr.benchmark \\
+        --train-dir /datasets/imagenet/train --val-dir /datasets/imagenet/val
+
+    python -m benchmarks.simclr.benchmark --dataset cifar10 \\
+        --train-dir ./cifar/train --val-dir ./cifar/val
+"""
+
+from __future__ import annotations
+
+from argparse import ArgumentParser, Namespace
+from pathlib import Path
+from typing import Any, Dict, List, Tuple
+
+import torch
+from pytorch_lightning import LightningModule, Trainer, seed_everything
+from pytorch_lightning.callbacks import LearningRateMonitor, ModelCheckpoint
+from pytorch_lightning.loggers import TensorBoardLogger
+from torch import Tensor, no_grad
+from torch.optim import SGD, Optimizer
+from torchvision.models import resnet18, resnet50
+
+from benchmarks.bench import ImageDataModule
+from benchmarks.bench.probes import KNNProbe
+from benchmarks.simclr.datasets import SETTINGS, Setting
+from lightly.backbones import TorchvisionResNetBackbone, small_image_stem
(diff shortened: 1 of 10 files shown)


examples/simclr.py
```diff
@@ -0,0 +1,96 @@
...
+criterion = NTXentLoss(temperature=TEMPERATURE)
+transform = SimCLRTransform(
+    input_size=INPUT_SIZE,
+    cj_strength=COLOR_JITTER_STRENGTH,
+    gaussian_blur=GAUSSIAN_BLUR,
+    normalize=NORMALIZE,
+)
+
+
+def forward(sample: Sample) -> Tensor:
```
gabrielfruet: Use nn.Module instead here. For PyTorch users this will look very weird

examples/simclr.py
```diff
@@ -0,0 +1,96 @@
...
+from lightly.models.modules import SimCLRProjectionHead
+from lightly.optim import param_groups
+from lightly.transforms import SimCLRTransform
+from lightly.transforms.utils import CIFAR10_NORMALIZE
+
+# ResNet-18, CIFAR-10, 32 pixels. Small enough to run on one GPU.
+BATCH_SIZE, EPOCHS, LR, MOMENTUM = 256, 100, 0.06, 0.9
+TEMPERATURE, WEIGHT_DECAY = 0.5, 5e-4
+INPUT_SIZE, FEATURE_DIM, HIDDEN_DIM, OUT_DIM = 32, 512, 512, 128
+COLOR_JITTER_STRENGTH, GAUSSIAN_BLUR, NORMALIZE = 0.5, 0.0, CIFAR10_NORMALIZE
```
gabrielfruet: The example should be clearer. The thing is that the user should be able to **run it**, not exactly match the paper parameter.

For the benchmarks, it should be the "same as the paper"

benchmarks/simclr/benchmark.py
```diff
@@ -0,0 +1,248 @@
...
+    """Builds the row's transform. Restated from the example; the gate compares it."""
+    return SimCLRTransform(
+        input_size=setting.size,
+        cj_strength=setting.color_jitter_strength,
+        gaussian_blur=setting.blur,
+        normalize=setting.normalize,
+    )
+
+
+class SimCLR(LightningModule):
```
gabrielfruet: Benchmarks are a bit hidden now. They should have the source of truth for the parameters

"Anyone that is serious about running the benchmarks could easily look into github"

We need to think of a better way of saving the parameters (maybe exposing in the docs in someway, or making it easily accesible.

benchmarks/simclr/benchmark.py
```diff
@@ -0,0 +1,248 @@
...
+def transform(setting: Setting) -> SimCLRTransform:
+    """Builds the row's transform. Restated from the example; the gate compares it."""
+    return SimCLRTransform(
+        input_size=setting.size,
+        cj_strength=setting.color_jitter_strength,
+        gaussian_blur=setting.blur,
+        normalize=setting.normalize,
+    )
+
+
```
gabrielfruet: Maybe call the Examples -> Examples and Benchmarks and show the PyTorch and the Benchmark.

PR #2041, opened 2026-08-18
fix(models): handle zero masked patch count in random_inverse_block_mask (#2025)

### LINKED ISSUE #2025: random_inverse_block_mask raises RuntimeError when the masked count rounds to zero
## What happens

`random_inverse_block_mask` in `lightly/models/utils.py` crashes with a `RuntimeError` from `torch.empty(1).uniform_(...)` whenever `mask_ratio` is small enough that `int(num_patches * mask_ratio) == 0`. `mask_ratio=0.0` is the clearest case, but any value in `[0, 1/num_patches)` triggers it.

```
RuntimeError: uniform_ expects to return a [from, to) range, but found from=0 > to=-6.25e-07
```

## Where

https://github.com/lightly-ai/lightly/blob/f444cf36392621c0b75a9f2a6d90bf7f56caa4ea/lightly/models/utils.py#L789-L806

When `num_masked` rounds to 0, `num_visible == num_patches`, so on a square grid:

- `min_lar = max(log_aspect_ratio[0], math.log(num_visible / width**2))` becomes `math.log(1) = 0`,
- `max_lar = min(log_aspect_ratio[1], math.log(height**2 / (num_visible + 1e-5)))` becomes a tiny negative number (about `-6.25e-7` on a 4x4 grid).

`torch.empty(1).uniform_(min_lar, max_lar)` then receives `from > to` and raises. The `if num_visible > 0:` guard just above covers the opposite extreme (all patches masked, `mask_ratio=1.0`) but not the all-visible boundary.

## Reproduction

```python
from lightly.models.utils import random_inverse_block_mask

# size = (batch_size, sequence_length); sequence_length = num_prefix_tokens + num_patches
# here: 1 prefix token + 16 patches (a 4x4 grid)
random_inverse_block_mask(size=(2, 17), mask_ratio=0.0)
# RuntimeError: uniform_ expects to return a [from, to) range, but found from=0 > to=-6.25e-07
```

`mask_ratio=0.65` works. `mask_ratio=0.05` on the same grid fails the same way, because `int(16 * 0.05) == 0`.

## Why it matters

The sibling utility `random_grid_token_mask` explicitly supports and tests `mask_ratio=0.0` (its test comment reads "mask_ratio=0.0 keeps all patches and masks none"). A caller sweeping `mask_ratio` across a range that includes low values, or passing `0.0` to mean "mask nothing", gets a hard crash instead of an all-visible mask.

## Suggested fix

Handle the all-visible boundary (`num_visible == num_patches`) the way the `num_visible == 0` case is already handled: skip the aspect-ratio sampling and return an empty masked set with every patch visible. Alternatively clamp `max_lar` to be at least `min_lar` before calling `uniform_`.

_Automated report: this issue was produced and filed automatically, with no human review before posting. Two independent checks agreed it is a real bug, but if it misreads the code please say so and we will close it._

_Found while running [Ito](https://ito.ai) (AI code review that runs your application, free for open source) against recently merged PRs. [Full analysis](https://app.ito.ai/share/d9f35df6-e369-47ea-a76b-77cfedca2a70?tab=details&utm_source=oss-outreach&utm_medium=github-issue&utm_campaign=oss-autopilot)._

harmehak0173:
### Summary of Changes

Fixes #2025.

In `lightly/models/utils.py`, `random_inverse_block_mask` previously raised a `RuntimeError: uniform_ expects to return a [from, to) range, but found from=0 > to=-6.25e-07` when `mask_ratio` was `0.0` or small enough that `num_masked = int(num_patches * mask_ratio) == 0`.

### Root Cause
When `num_masked == 0`, `num_visible == num_patches`:
- `min_lar = max(log_aspect_ratio[0], math.log(num_visible / width**2))` evaluated to `0.0`.
- `max_lar = min(log_aspect_ratio[1], math.log(height**2 / (num_visible + 1e-5)))` evaluated to `-6.25e-7` due to the epsilon term.
- `torch.empty(1).uniform_(min_lar, max_lar)` failed because `min_lar > max_lar`.

### Solution
1. Explicitly handle the boundary case `if num_visible == num_patches:` by marking all patches visible (`visible.fill_(True)`), skipping the aspect-ratio sampling loop (consistent with sibling masking functions like `random_grid_token_mask`).
2. Added unit tests in `tests/models/test_ModelUtils.py` (`test_random_inverse_block_mask__mask_ratio_extremes`) verifying `mask_ratio=0.0`, `mask_ratio=0.05` (which rounds to 0 masked patches on a 16-patch grid), and `mask_ratio=1.0`.

### Testing
- `pytest tests/models/test_ModelUtils.py`: 88 passed, 8 skipped.
- `ruff check lightly/models/utils.py tests/models/test_ModelUtils.py`: Passed.
- `ruff format --check lightly/models/utils.py tests/models/test_ModelUtils.py`: Passed.

### DIFF
--- a/tests/models/test_ModelUtils.py
+++ b/tests/models/test_ModelUtils.py
@@ -1125,6 +1125,47 @@ def test_random_inverse_block_mask__not_square_raises() -> None:
         )
 
 
+def test_random_inverse_block_mask__mask_ratio_extremes() -> None:
+    num_prefix_tokens = 1
+    sequence_length = num_prefix_tokens + 16  # 4x4 patch grid
+
+    # mask_ratio=0.0 keeps all patches and masks none.
+    idx_keep, idx_mask = utils.random_inverse_block_mask(
+        size=(2, sequence_length),
+        mask_ratio=0.0,
+        num_prefix_tokens=num_prefix_tokens,
+    )
+    assert idx_mask.shape == (2, 0)
+    assert idx_keep.shape == (2, sequence_length)
+    assert torch.equal(
+        idx_keep, torch.arange(sequence_length).expand(2, sequence_length)
+    )
+
+    # mask_ratio=0.05 rounds to 0 masked patches (int(16 * 0.05) == 0).
+    idx_keep, idx_mask = utils.random_inverse_block_mask(
+        size=(2, sequence_length),
+        mask_ratio=0.05,
+        num_prefix_tokens=num_prefix_tokens,
+    )
+    assert idx_mask.shape == (2, 0)
+    assert idx_keep.shape == (2, sequence_length)
+    assert torch.equal(
+        idx_keep, torch.arange(sequence_length).expand(2, sequence_length)
+    )
+
+    # mask_ratio=1.0 keeps only the prefix tokens and masks all patches.
+    idx_keep, idx_mask = utils.random_inverse_block_mask(
+        size=(2, sequence_length),
+        mask_ratio=1.0,
+        num_prefix_tokens=num_prefix_tokens,
+    )
+    assert idx_mask.shape == (2, 16)
+    assert idx_keep.shape == (2, num_prefix_tokens)
+    assert torch.equal(
+        idx_keep, torch.arange(num_prefix_tokens).expand(2, num_prefix_tokens)
+    )
+
+
 def test_random_grid_token_mask__mask_ratio_extremes() -> None:
     num_prefix_tokens = 1
     sequence_length = num_prefix_tokens + 16  # 4 cells of 4 patches (grid=2)
--- a/tests/models/test_ModelUtils.py
+++ b/tests/models/test_ModelUtils.py
@@ -1125,68 +1125,60 @@ def test_random_inverse_block_mask__not_square_raises() -> None:
         )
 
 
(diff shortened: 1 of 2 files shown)


tests/models/test_ModelUtils.py
```diff
@@ -1125,6 +1125,47 @@ def test_random_inverse_block_mask__not_square_raises() -> None:
         )
 
 
+def test_random_inverse_block_mask__mask_ratio_extremes() -> None:
```
gabrielfruet: Let's use a parametrized test instead

gabrielfruet (changes requested):
Looking good! I just think we should use a parametrized test. A PR got merged back then that didn't used the parametrized test, you can also create a follow-up for that if you want (right below the test you added)

tests/models/test_ModelUtils.py
harmehak0173: Thanks @gabrielfruet! I have updated `test_random_inverse_block_mask__mask_ratio_extremes` with `@pytest.mark.parametrize` and also refactored the sibling `test_random_grid_token_mask__mask_ratio_extremes` right below it to use `@pytest.mark.parametrize` as well.

gabrielfruet (approved):
LGTM! Thank you for handling this :)

PR #2045, opened 2026-08-21
fix: correct torch.linalg.solve_triangular availability check in WMSELoss

### LINKED ISSUE #2044: WMSELoss / Whitening2d fail to build: torch.linalg.solve_triangular wrongly reported unavailable
## Bug description

`WMSELoss` and `Whitening2d` do not build. The constructor raises a `RuntimeError`. It reports that `torch.linalg.solve_triangular` is not available, but the function is available on PyTorch >=1.11.

The availability check at `lightly/loss/wmse_loss.py` line 10 uses `import torch.linalg.solve_triangular`. That statement imports a function as a module, so it always raises `ImportError`. The flag `_SOLVE_TRIANGULAR_AVAILABLE` is then always `False`, and the constructor always raises. This affects every supported PyTorch version.

The test module `tests/loss/test_wmse_loss.py` uses the same import at module level. So the test skips itself on every run, and CI does not catch the fault.

## To reproduce

```python
import torch
from lightly.loss import WMSELoss

print(torch.__version__)                          # e.g. 2.13.0
print(hasattr(torch.linalg, "solve_triangular"))  # True
WMSELoss()
```

```
RuntimeError: Whitening2d depends on torch.linalg.solve_triangular which is not available in your PyTorch installation. Please update to PyTorch 1.11 or newer.
```

## Environment

- lightly: `1.5.26` (current release) and `master`.
- torch: the fault does not depend on the torch version. Verified on `2.13.0`, where `torch.linalg.solve_triangular` is present.

## Cause

`import torch.linalg.solve_triangular` treats `solve_triangular` as a module. It is a function in `torch.linalg`. Python cannot import a function as a module, so the statement always raises `ImportError`. The check has been wrong since #1459.

## Suggested fix

Detect the function with `hasattr(torch.linalg, "solve_triangular")`, in the module and in the test. I have a fix ready and can open a PR.

lorinczszabolcs:
Closes #2044.

`WMSELoss` and `Whitening2d` gated construction on `_SOLVE_TRIANGULAR_AVAILABLE`, set by `import torch.linalg.solve_triangular`. That statement imports a function as a module, so it always raised `ImportError` and the flag was always `False`, both classes raised `RuntimeError` on construction on every supported torch, including PyTorch >=1.11 where the function exists. `tests/loss/test_wmse_loss.py` used the same import to skip itself, so the failure never ran in CI.

Detect the function with `hasattr(torch.linalg, "solve_triangular")`, in both the module and the test. The WMSE tests now run instead of skipping.

Verified `WMSELoss()` builds and `pytest tests/loss/test_wmse_loss.py` passes (6 tests, previously 1 skipped) on torch 2.13. `make format-check` and mypy pass.

### DIFF
--- a/lightly/loss/wmse_loss.py
+++ b/lightly/loss/wmse_loss.py
@@ -6,13 +6,8 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
-try:
-    import torch.linalg.solve_triangular
-except ImportError:
-    # Only available in PyTorch >=1.11.
-    _SOLVE_TRIANGULAR_AVAILABLE = False
-else:
-    _SOLVE_TRIANGULAR_AVAILABLE = True
+# torch.linalg.solve_triangular is only available in PyTorch >=1.11.
+_SOLVE_TRIANGULAR_AVAILABLE = hasattr(torch.linalg, "solve_triangular")
 
 
 def norm_mse_loss(x0: torch.Tensor, x1: torch.Tensor) -> torch.Tensor:
--- a/tests/loss/test_wmse_loss.py
+++ b/tests/loss/test_wmse_loss.py
@@ -3,9 +3,8 @@ import torch
 
 from lightly.loss.wmse_loss import WMSELoss
 
-try:
-    import torch.linalg.solve_triangular
-except ImportError:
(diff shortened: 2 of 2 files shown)


gabrielfruet (approved):
LGTM! thank you

PR #1809, opened 2025-03-03
LIG-6287 DINO ViT Benchmark Report

yutong-xiang-97:
- Fix default settings of `knn_t` in KNN eval
- Make `knn_k` configurable
- Update DINO ViT benchmark results on ImageNet1k on README and Docs

Updated Docs:
[Benchmarks — lightly 1.5.19 documentation.pdf](https://github.com/user-attachments/files/19052307/Benchmarks.lightly.1.5.19.documentation.pdf)

PR #2055, opened 2026-09-01
feat(benchmarks): add --resume flag and last.ckpt saving to imagenet benchmarks

### LINKED ISSUE #2053: ImageNet benchmark runs can't cleanly resume: no last.ckpt + timestamped run dirs
`--ckpt-path` already flows into `trainer.fit(ckpt_path=...)` (`benchmarks/imagenet/resnet50/main.py:323`, `benchmarks/imagenet/vitb16/main.py:269`), so Lightning restores epoch, optimizer and scheduler state. Two gaps stop it being a clean resume.

**1. No explicit `ModelCheckpoint`.** The `pretrain()` callback lists (`resnet50/main.py:306`, `vitb16/main.py:252`) rely on Lightning's implicit epoch-end checkpoint. No `last.ckpt`, no mid-epoch save, no retention control for 100-epoch runs.

**2. Every launch stamps a fresh run dir** with `datetime.now()` (`resnet50/main.py:122`, `vitb16/main.py:103`). A resume logs to a new folder and there is no discovery of the prior checkpoint, so you hand-locate the `.ckpt` path.

**Scope** (benchmark-only, both `main.py`):
- [ ] Add `ModelCheckpoint(save_last=True, ...)` to both `pretrain()` callback lists.
- [ ] Add `--resume` that resolves the newest run dir + its `last.ckpt` and reuses that dir instead of stamping a new one.
- [ ] Pass the resolved path to `ckpt_path`.
- [ ] Document the flag in `benchmarks/imagenet/README.md`.

7487:
closes #2053

## Description
- [ ] My change is breaking

Implements the scope checklist from the issue, for both `resnet50/main.py` and `vitb16/main.py`:

- `pretrain()` now includes an explicit `ModelCheckpoint(save_last=True)`, so every run writes a `last.ckpt` (plus one rolling epoch checkpoint) instead of relying on Lightning's implicit checkpointing.
- New `--resume` flag: `find_resume_run()` picks the newest timestamped run directory under `log_dir/<method>/` that contains a `last.ckpt`, reuses that directory for logging, and passes the checkpoint to `trainer.fit(ckpt_path=...)` (and to the `--epochs 0` weight-loading path). Since each resume logs to a new `version_N` subdirectory, the most recently modified `last.ckpt` within the run is chosen.
- If no checkpoint exists yet, `--resume` just starts a fresh timestamped run, so the flag can always be passed on preemptible clusters (e.g. in a SLURM requeue script).
- `--resume` and `--ckpt-path` are mutually exclusive and raise a `ValueError` when combined.

The helper is duplicated in both scripts on purpose, matching the self-contained benchmark script layout.

## Tests
- [x] I have manually tested the change. Verified `find_resume_run()` with an assertion script over a fabricated log tree: missing dir -> None, runs without checkpoints -> None, newest run with a checkpoint wins over both older runs and newer checkpoint-less runs, and the most recently modified `last.ckpt` is picked when a run contains multiple `version_N` subdirectories. Also checked that both scripts pass `ruff check` and `ruff format` at the locked version. There is no test infrastructure for `benchmarks/`, so no test files were added.

## Documentation
- [x] I have updated the documentation accordingly. (`benchmarks/imagenet/README.md` documents the new flag.)

- Mid-epoch time-based checkpointing (`train_time_interval`) was left out to keep the change minimal; easy follow-up if wanted.

Adds resumable ImageNet benchmark runs for ResNet-50 and ViT-B/16.

- Saves `last.ckpt` with explicit `ModelCheckpoint` callbacks.
- Adds `--resume` to find and reuse the newest run with a checkpoint.
- Rejects simultaneous `--resume` and `--ckpt-path`.
- Documents the new option in the ImageNet README.

### DIFF
--- a/benchmarks/imagenet/README.md
+++ b/benchmarks/imagenet/README.md
@@ -140,7 +140,11 @@ Common flags supported by both benchmark families:
 * `--devices`: Number of devices used by PyTorch Lightning.
 * `--accelerator`: Accelerator used by PyTorch Lightning. Defaults to `gpu`.
 * `--precision`: PyTorch Lightning precision. Defaults to `16-mixed`.
-* `--ckpt-path`: Load or resume from a checkpoint.
+* `--ckpt-path`: Load or resume from a specific checkpoint file.
+* `--resume`: Resume the newest run of each method. Reuses its run directory
+  and continues pretraining from its `last.ckpt`. Starts a new run if no
+  checkpoint exists yet, so the flag is safe to always pass on preemptible
+  clusters. Cannot be combined with `--ckpt-path`.
 * `--num-classes`: Number of dataset classes. Defaults to `1000`.
 * `--skip-knn-eval`: Skip kNN evaluation.
 * `--skip-linear-eval`: Skip linear evaluation.
--- a/benchmarks/imagenet/resnet50/main.py
+++ b/benchmarks/imagenet/resnet50/main.py
@@ -15,7 +15,7 @@ import inspect
 from argparse import ArgumentParser
 from datetime import datetime
 from pathlib import Path
-from typing import Dict, Sequence, Union
+from typing import Dict, Sequence, Tuple, Union
 
 import barlowtwins
 import byol
@@ -36,6 +36,7 @@ from pytorch_lightning.callbacks import (
     DeviceStatsMonitor,
     EarlyStopping,
     LearningRateMonitor,
+    ModelCheckpoint,
 )
 from pytorch_lightning.loggers import TensorBoardLogger
 from torch.utils.data import DataLoader
@@ -57,6 +58,7 @@ parser.add_argument("--accelerator", type=str, default="gpu")
 parser.add_argument("--devices", type=int, default=1)
 parser.add_argument("--precision", type=str, default="16-mixed")
 parser.add_argument("--ckpt-path", type=Path, default=None)
+parser.add_argument("--resume", action="store_true")
 parser.add_argument("--compile-model", action="store_true")
 parser.add_argument("--methods", type=str, nargs="+")
 parser.add_argument("--num-classes", type=int, default=1000)
@@ -105,11 +107,14 @@ def main(
     skip_linear_eval: bool,
     skip_finetune_eval: bool,
     ckpt_path: Union[Path, None],
+    resume: bool,
     float32_matmul_precision: str,
     strategy: str,
     seed: int | None = None,
 ) -> None:
     print_rank_zero(f"Args: {locals()}")
+    if resume and ckpt_path is not None:
+        raise ValueError("--resume and --ckpt-path cannot be used together.")
     if seed is not None:
         seed_everything(seed, workers=True, verbose=True)
 
@@ -118,9 +123,21 @@ def main(
     method_names = methods or METHODS.keys()
 
     for method in method_names:
-        method_dir = (
-            log_dir / method / datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
-        ).resolve()
+        method_ckpt_path = ckpt_path
+        method_dir = None
+        if resume:
+            resume_run = find_resume_run(log_dir / method)
+            if resume_run is None:
+                print_rank_zero(
+                    f"No checkpoint found in {log_dir / method}, starting a new run."
+                )
+            else:
+                method_dir, method_ckpt_path = resume_run
+                print_rank_zero(f"Resuming from checkpoint {method_ckpt_path}")
+        if method_dir is None:
+            method_dir = (
+                log_dir / method / datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+            ).resolve()
         print_rank_zero(f"Logging to {method_dir}")
         # Initialise the base arguments for all models
         model_kwargs = {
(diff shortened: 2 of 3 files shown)


benchmarks/imagenet/vitb16/main.py
```diff
@@ -114,9 +131,9 @@ def main(
 
         if epochs <= 0:
             print_rank_zero("Epochs <= 0, skipping pretraining.")
-            if ckpt_path is not None:
+            if method_ckpt_path is not None:
                 model.load_state_dict(
-                    torch.load(ckpt_path, weights_only=False)["state_dict"]
+                    torch.load(method_ckpt_path, weights_only=False)["state_dict"]
```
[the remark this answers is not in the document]
7487: Thanks for checking, but I believe this one is a false positive — keeping `weights_only=False` as is:

1. **Not introduced by this PR.** `weights_only=False` predates this change (added in #1809, March 2025); this diff only renames `ckpt_path` to `method_ckpt_path` on that line.
2. **The torch 1.10 floor doesn't apply to the benchmarks.** `torch>=1.10.0` is the library's `minimal` lowest-resolution test group in `pyproject.toml`, and that CI surface explicitly excludes these scripts (`.github/workflows/test_minimal_deps.yml` filters with `!benchmarks/**`). This script can't run on torch 1.10 anyway: it unconditionally calls `torch.set_float32_matmul_precision` (torch >= 1.12) and `seed_everything(seed, workers=True, verbose=True)` — the `verbose` parameter exists only in pytorch_lightning >= 2.4, which itself requires torch >= 2.1.
3. **Removing it would be the actual regression.** Since torch 2.6, `torch.load` defaults to `weights_only=True`, and full Lightning trainer checkpoints contain objects outside the weights-only allowlist, so dropping the argument would break this exact code path on current PyTorch — which is presumably why #1809 added it.

benchmarks/imagenet/resnet50/main.py
```diff
@@ -224,6 +241,30 @@ def main(
...
+
+    Returns:
+        Tuple with the run directory and its most recent last.ckpt checkpoint,
+        or None if no run with a checkpoint exists.
+    """
+    if not method_log_dir.is_dir():
+        return None
+    # Run directories are named after their start time, so the lexicographic
+    # order matches the chronological order.
+    for run_dir in sorted(
```
gabrielfruet, line 255: This skips newer runs that have no `last.ckpt` yet. Same in vitb16/main.py.
A run killed before its first checkpoint gets passed over, and `--resume` continues an older run instead, even a finished one.

```python
run_dirs = sorted(path for path in method_log_dir.iterdir() if path.is_dir())
if not run_dirs:
    return None
checkpoints = sorted(
    run_dirs[-1].rglob("last.ckpt"), key=lambda path: path.stat().st_mtime
)
if not checkpoints:
    return None
return run_dirs[-1].resolve(), checkpoints[-1].resolve()
```

benchmarks/imagenet/README.md
```diff
@@ -140,7 +140,11 @@ Common flags supported by both benchmark families:
 * `--devices`: Number of devices used by PyTorch Lightning.
 * `--accelerator`: Accelerator used by PyTorch Lightning. Defaults to `gpu`.
 * `--precision`: PyTorch Lightning precision. Defaults to `16-mixed`.
-* `--ckpt-path`: Load or resume from a checkpoint.
+* `--ckpt-path`: Load or resume from a specific checkpoint file.
+* `--resume`: Resume the newest run of each method. Reuses its run directory
+  and continues pretraining from its `last.ckpt`. Starts a new run if no
+  checkpoint exists yet, so the flag is safe to always pass on preemptible
+  clusters. Cannot be combined with `--ckpt-path`.
```
gabrielfruet, line 147: `--resume` isn't safe to always pass when experiments share a `--log-dir`.
A new experiment launched with it continues the previous run's `last.ckpt`, and nothing checks that against the new flags.

```suggestion
* `--resume`: Continue the newest run of each method from its `last.ckpt` and
  reuse its run directory. Starts a new run if the newest run has no
  checkpoint yet. The checkpoint is not checked against the current flags, so
  use a separate `--log-dir` per experiment. Cannot be combined with
  `--ckpt-path`.
```

benchmarks/imagenet/resnet50/main.py
```diff
@@ -305,6 +346,8 @@ def pretrain(
         devices=devices,
         callbacks=[
             LearningRateMonitor(),
+            # Save last.ckpt so interrupted runs can be resumed with --resume.
+            ModelCheckpoint(save_last=True),
```
gabrielfruet, line 350: Each resume opens a new pretrain/version_N, so one run's curves are split across versions. Same in vitb16/main.py.
Passing the checkpoint's version folder as `version=` writes back into it. I tried it and got one continuous curve in version_0.

```python
# main(), passed to pretrain()
logger_version = (
    method_ckpt_path.parent.parent.name
    if resume and method_ckpt_path is not None
    else None
)

# pretrain()
logger=TensorBoardLogger(
    save_dir=str(log_dir), name="pretrain", version=logger_version
),
```

gabrielfruet (changes requested):
Really nice work overall! Can you address my comments?

gabrielfruet:
Hi, are you still interested in working on this? If not, I’ll offer it to someone else.

PR #2062, opened 2026-09-14
fix(loss): allow deferring the DINO/iBOT center update

liopeer:
## Problem

`DINOLoss.forward()` applies an EMA momentum update to the `center` buffer on **every** call, and so do `IBOTPatchLoss` / `IBOTPlusPlusPatchLoss` via the shared `Center` module. Under **gradient accumulation** that means K momentum updates per optimizer step instead of one, so the effective decay rate of the center silently scales with the accumulation factor.

Only the caller knows when a micro-batch is the last of an optimizer step, and only the loss can skip the update — so the fix needs a hook on both sides.

## Approach

Separate **accumulating the batch statistic** from **applying the momentum update**. `forward()` always accumulates (cheap, `no_grad`, two small buffers); only the momentum step is deferrable, via a new keyword-only `update_center` flag:

```python
for micro_batch in micro_batches:          # K micro-batches
    loss = criterion(teacher_out, student_out, update_center=False)
    (loss / K).backward()
criterion.update_center()                  # one EMA step over the full batch
optimizer.step()
```

The center then sees the **full effective batch** rather than the last micro-batch alone. Simply skipping the update on all but the last micro-batch would drift ~`1.7e-2` from a single full-batch pass; today's behaviour drifts ~`3.9e-2`.

For iBOT the manual path is `criterion.center.update()`.

### Micro-batches are weighted by their element count

Accumulating one *mean* per micro-batch and dividing by K only reproduces a single full-batch pass when every micro-batch contributes the same number of elements. So accumulation is element-weighted: `mean × num_elements` is summed and divided by the total count.

This is not a corner case for iBOT. `IBOTPatchLoss` masks a **variable number of tokens per batch by construction**, so its micro-batches differ in size on every step — unweighted averaging drifts `3.8e-3`–`6.6e-3`. DINO is unaffected in practice (multi-crop needs fixed-size batches and `drop_last=True`), but it is weighted too for consistency.

`center_mean` is untouched: multiplying its result by the element count recovers the sum, so its contract does not change.

The element count is reduced across processes **inside `accumulate`**, mirroring `center_mean`. Collecting rank-local sums and reducing once in the apply step would be cheaper but is not safe — DDP broadcasts buffers, including non-persistent ones, from rank zero before every forward and would clobber the rank-local partial sums.

Verified end-to-end: K=2 accumulation over a small DINO training loop reproduces the single-full-batch center to `7.5e-9`, and the new unequal-size regression tests (unequal sample counts for DINO, unequal masked-token counts for iBOT, unequal patch counts for iBOT++) all fail on an unweighted implementation.

**Deliberately unchanged:** `center_mean` averages the per-rank means without weighting them by their element counts, which for iBOT already differs from a token-weighted global mean today. This PR preserves that exactly. Fixing it would be more correct but would shift existing DINOv2/iBOT training numerics, so it belongs in its own PR.

## Consistency with the existing opt-out APIs

Two components already solve this problem, and the flag follows their shape:

| | `MemoryBankModule` | `TiCoLoss` | this PR |
|---|---|---|---|
| flag name | `update` | `update_covariance_matrix` | `update_center` — same `update_<state>` shape |
| default | `False` | `True` | `True` — preserves current behaviour |
| documented in `Args:` | yes | yes | yes, marked `Experimental:` |
| extra public methods on the loss | none | none | none (`DINOLoss.update_center()` was already public) |
| deferred-apply semantics | no | no | **yes** — deliberate divergence |

The one deliberate divergence: both precedents are *skip-only*, which here would center on the last micro-batch alone. The accumulate-then-apply variant is exact instead. `TiCoLoss` has the same last-micro-batch weakness and could get the same treatment in a follow-up.

The flag is **keyword-only** because `IBOTPlusPlusPatchLoss.forward` inserts `visible_loss_weight` before it — a positional flag would put a `bool` and a `float` in the same slot across the override and trip mypy.

## ⚠️ Behaviour change to review as a unit

This also adds the missing **`if self.training` guard** — the center previously updated during validation/eval passes too. Nothing in this repo is affected (no test, example, or benchmark calls an SSL criterion outside a training path), but it *is* a change for anyone computing the SSL loss in eval mode. Happy to drop it if you'd rather ship it separately.

The guard lives in `forward` only, so the manual `update_center()` / `Center.update()` escape hatch still works on an eval-mode module.

Note that `self.training` now gates a distributed collective (the `all_reduce` in `center_mean`), so `.train()`/`.eval()` must be called on all ranks in lockstep — the same contract as `SyncBatchNorm`. Documented next to the guard.

## Backwards compatibility

- The accumulator buffers are registered `persistent=False`, so `state_dict()` keys are unchanged (`{"center"}` / `{"center.center"}`) and `load_state_dict(strict=True)` succeeds in **both** directions (old checkpoint → new code, new checkpoint → old code). Covered by regression tests.
- Default `update_center=True` reproduces today's behaviour — the existing golden-reference tests against the FB DINO implementation pass untouched.
- `DINOLoss.update_center(teacher_out)` keeps working, and now also accepts the list of per-view tensors that `forward` takes.

## Other notes

- Warn-once safeguard: `update_center=False` with no follow-up call leaves the center frozen at zeros — i.e. no centering, which is exactly the DINO collapse mode, silently. A `UserWarning` fires after 1000 accumulations (~15x beyond any realistic accumulation factor).
- Drive-by fixes: the wrong `self.center: Parameter` annotation in `DINOLoss` (it is a buffer), the dead `Center._center_mean` method, and `tests/loss/test_ibot_loss.py` which parametrized `device=["cpu","cuda"]` but never moved anything — exactly where a device-placement bug in the new accumulator would surface.
- Docs/examples/notebooks untouched, so `make generate-example-notebooks` is not needed.

## Follow-ups worth issues

Same class of bug, not touched here: `NTXentLoss` hardwires `memory_bank.forward(update=out0.requires_grad)`; `CO2Regularizer` hardwires `update=True`; `Whitening2d` updates running stats per forward; `SwaVPrototypes._freeze_prototypes_if_required` mutates `requires_grad` from `forward()`; `TiCoLoss.C` is a plain attribute rather than a buffer.

Plus the cross-process weighting in `center_mean` noted above.

## Testing

`make static-checks` clean. `make test`: 1587 passed; the 3 failures (2 pyav video-dataset, 1 `test_loss_dcl`) reproduce identically on clean `master` and are unrelated.

New tests cover accumulation equivalence for equal **and** unequal micro-batch sizes, the eval guard, the eval-mode manual escape hatch, `state_dict` key stability, and the warn-once safeguard.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016c2X21JXjgQRVp3NNLNTUS

Adds deferred, element-weighted center updates for DINO and iBOT to support gradient accumulation across unequal micro-batches and masked-token counts.

Evaluation no longer updates centers automatically. Accumulators remain outside `state_dict`, and missing updates trigger a warning.

Adds tests for accumulation equivalence, evaluation behavior, unequal counts, warnings, and checkpoint compatibility.


lightly/models/modules/center.py
```diff
@@ -61,23 +82,66 @@ def value(self) -> Tensor:
...
                 dimensions as self.size.
         """
+        # NOTE(Lionel, 09/26): The distributed all-reduce happens here, inside
+        # accumulate, which makes the accumulated sum identical on all ranks. DDP
+        # broadcasts buffers (including non-persistent ones) from rank zero before
+        # every forward pass, so that broadcast is a no-op for the accumulator.
+        # Moving the all-reduce to apply_update would break this.
         batch_center = self._center_fn(x=x, dim=self.dim)
+        self._batch_center_sum += batch_center
+        self._num_accumulated += 1
```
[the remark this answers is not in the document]
liopeer: Confirmed and fixed in 3ed0137f.

You are right, and the iBOT case is worse than I claimed in the original description: `IBOTPatchLoss` masks a variable number of tokens per batch *by construction*, so its micro-batches differ in size on every step, not occasionally. Measured drift from a single full-batch pass with unweighted averaging: `3.8e-3` to `6.6e-3` depending on how uneven the masked-token counts are. My earlier argument that unweighted accumulation preserved the reference semantics for iBOT conflated the across-step EMA weighting (which does weight each step equally) with the within-step token weighting (which the reference computes as one token-weighted mean). Those are different things.

Accumulation is now element-weighted, but without changing `center_mean`'s contract as suggested: multiplying its result by the element count recovers the sum, so `center_mean` stays as-is and only a small `center_num_elements` helper is added.

One deviation from the suggestion worth flagging: the element count is reduced across processes **inside `accumulate`**, not deferred to the apply step. Collecting rank-local sums and reducing once would be cheaper, but DDP broadcasts buffers — including non-persistent ones — from rank zero before every forward pass, which would clobber the rank-local partial sums. Keeping both reductions in `accumulate` makes the accumulators rank-identical, so the broadcast is a no-op.

Cross-*process* weighting is deliberately left alone: `center_mean` averages per-rank means without weighting by their element counts, which already differs from a token-weighted global mean for iBOT today. Changing that would shift existing DINOv2/iBOT training results, so it belongs in its own PR rather than riding along in a gradient-accumulation fix.

tests/loss/test_dino_loss.py
```diff
@@ -155,6 +156,96 @@ def test_center__equivalence() -> None:
...
+        """Deferring the update matches a single forward over the full batch."""
+        batch_size, output_dim = 8, 4
+        criterion = DINOLoss(output_dim=output_dim, center_momentum=0.9)
+        single = DINOLoss(output_dim=output_dim, center_momentum=0.9)
+
+        torch.manual_seed(0)
+        micro_batches = [
+            [torch.rand(batch_size, output_dim) for _ in range(2)]
+            for _ in range(num_micro_batches)
+        ]
```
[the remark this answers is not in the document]
liopeer: Added in 3ed0137f, for all three losses:

- `test_gradient_accumulation__unequal_batch_sizes` (DINO, sample counts `[2, 6]` and `[1, 8, 3]`)
- `test_gradient_accumulation__unequal_masked_tokens` (iBOT, masked tokens per image `[2, 6]` and `[1, 8, 3]`)
- `test_gradient_accumulation__unequal_patch_counts` (iBOT++, patch tokens `[4, 10]` and `[2, 12, 6]`)

Each asserts the deferred center matches a single pass over the concatenated inputs. I verified all six fail against the previous unweighted implementation and pass against the weighted one, so they do catch the case they are meant to catch rather than just passing by construction.

lightly/models/modules/center.py
```diff
@@ -100,6 +175,30 @@ def center_mean(x: Tensor, dim: Tuple[int, ...]) -> Tensor:
...
+            Dimensions along which the center is calculated.
+
+    Returns:
+        A scalar tensor with the number of elements.
+    """
+    num_elements = torch.tensor(
+        float(math.prod(x.shape[d] for d in dim)), device=x.device
+    )
+    if dist.is_available() and dist.is_initialized():
+        dist.all_reduce(num_elements)
```
gabrielfruet, line 198: `center_num_elements` here sum all batches sizes and `center_mean` compute the mean per rank. that means that a rank with 100 tokens and a rank with 10 tokens would have the same weight.

i'm not 100% sure if this can happen (a rank with 100 tokens and a rank with 10 tokens) in any standard pytorch training loop. but I think it's worth it to fix.

tests/loss/test_dino_loss.py
```diff
@@ -155,6 +156,123 @@ def test_center__equivalence() -> None:
     assert torch.allclose(criterion.center, center.value)
 
 
+def test_center__equivalence_accumulate() -> None:
+    """Check that deferred center updates are equivalent to Center's.
+
+    TODO(Guarin, 08/24): Remove this test once DINOLoss uses Center internally.
```
gabrielfruet, line 162: TODO Lionel I think instead? and date

lightly/models/modules/center.py
gabrielfruet, line 209: Nit: I thinkt his should be no top of the file maybe?

tests/loss/test_dino_loss.py
```diff
@@ -155,6 +156,123 @@ def test_center__equivalence() -> None:
...
+    """
+    criterion = DINOLoss(output_dim=32, center_momentum=0.9)
+    center = Center(size=(1, 1, 32), momentum=0.9)
+    for _ in range(3):
+        x = torch.rand(2, 32)
```
gabrielfruet, line 167: Shouldn't this be 2,1,32 for example? because it is going to be reduced

gabrielfruet (changes requested):
General shape is looking good, just small stuff

tests/loss/test_dino_loss.py
liopeer: Would leave as is. It's a small modification on top of what Guarin mentioned as the TODO.

